Issue #6, opened 2021-05-04
Add networking logic to perform HTTP calls to API


levibostian:
Technical requirement in order for the SDK to perform requests to the CIO API. 

# Why are we doing this for our customers? 

Today, our customers need to write custom code in their apps to perform requests to the CIO API to add data to their workspace. This task will setup the technical code in order for us to perform the API calls on your behalf. 

# How do we test that this change is successful? 

The end goal is to perform a successful HTTP request to the CIO API in an Android app via the SDK. This includes:

* [Authentication header](https://customer.io/docs/api/#section/Authentication)
* [Path is dynamic](https://customer.io/docs/api/#section/Authentication)

# How are we building this? 

Not 100% sure. The end goal is keeping the SDK slim so not using retrofit or volley is a great start. 

My current idea is to use kotlin coroutines. Google gave a nice code snippet in it's docs:

```kt
sealed class Result<out R> {
    data class Success<out T>(val data: T) : Result<T>()
    data class Error(val exception: Exception) : Result<Nothing>()
}

class LoginRepository(private val responseParser: LoginResponseParser) {
    private const val loginUrl = "https://example.com/login"

    // Function that makes the network request, blocking the current thread
    fun makeLoginRequest(
        jsonBody: String
    ): Result<LoginResponse> {
        val url = URL(loginUrl)
        (url.openConnection() as? HttpURLConnection)?.run {
            requestMethod = "POST"
            setRequestProperty("Content-Type", "application/json; utf-8")
            setRequestProperty("Accept", "application/json")
            doOutput = true
            outputStream.write(jsonBody.toByteArray())
            return Result.Success(responseParser.parse(inputStream))
        }
        return Result.Error(Exception("Cannot open HttpURLConnection"))
    }
}
```
- [x] Decide how we will do HTTP networking calls. Implement that code in the SDK. 

- [x] Add `User-Agent` header for all requests. Value set to `Customer.io Android SDK vX.X.X, Y Android app vX.X.X`  where `Y` is the name of the customer's Android app. Thinking you can specify an optional app identifier in the configuration? 

- [x] Create a common module to be shared among different modules and possible extension methods for networking, as well as common request/response models

levibostian:
From the iOS SDK, here is some help for you on this task:
* [the headers to set](https://github.com/customerio/customerio-ios/blob/5d77bb6e528f144f6a8d34a477563ce134851184/Sources/Tracking/Service/HttpClient.swift#L116-L118)
[...]

PR #23, opened 2021-09-30
chore: Add network layer and common module 

Shahroz16:
This PR tackles #6

A common module is added that can be shared among different modules and the barebone structure of the Network layer using Retrofit and OKHTTP is also added, along with utility methods to synchronously and asynchronously make network requests.


RussellCIO (approved):
Approved.

Issue #27, opened 2021-10-06
Track an Event


Shahroz16:
We need the ability to track a customer event.

# Why are we doing this for our customers? 

CIO gives you the ability today to [track an event](https://customer.io/docs/api/#operation/track), we need to cater that in the Android SDK. 

# How do we test that this change is successful? 

Upon tracking any event through SDK it should be reflected on Portal as well, also integration test. 

# How are we building this? 

After the customer is identified, store the identifier in preferences. Provide a method in the `CustomerIo` class to track an event, also able to stop identifying. 

- [ ] Store identifier in the prefrences.
- [ ] Create method to track an event `CustomerIO.track()`
- [ ] Create method to stop identifying events `CustomerIO. identifyStop()`

PR #28, opened 2021-10-08
feat: track event

Shahroz16:
Tackles #27 

- Event tracking
- Storage/Removal of `identity`
- Support of `map<String, Any>` for events tracking and identify your customer

### DIFF
--- a/sdk/src/main/java/io/customer/sdk/repository/PreferenceRepository.kt
+++ b/sdk/src/main/java/io/customer/sdk/repository/PreferenceRepository.kt
@@ -12,14 +12,17 @@ internal interface PreferenceRepository {
 
 internal class PreferenceRepositoryImpl(private val context: Context) : PreferenceRepository {
 
+    private val prefsName by lazy {
+        "io.customer.sdk.${context.packageName}.${CustomerIO.instance().config.siteId}"
+    }
+
     companion object {
-        private const val PREFS_NAME = "customer_io_store"
         private const val KEY_IDENTIFIER = "identifier"
     }
 
     private val prefs: SharedPreferences by lazy {
         context.applicationContext.getSharedPreferences(
-            CustomerIO.instance().config.siteId,
+            prefsName,
             Context.MODE_PRIVATE
         )
     }


app/src/main/java/io/customer/example/MainActivity.kt
```diff
@@ -23,24 +24,70 @@ class MainActivity : AppCompatActivity() {
...
+                attributes = mapOf("email" to "testemail@email.com")
+            ).enqueue(outputCallback)
     }
 
     private fun makeSynchronousRequest() {
         CoroutineScope(Dispatchers.IO).launch {
-            when (val result = CustomerIo.instance().identify("testcoo@email.com").execute()) {
+            when (
+                val result =
+                    CustomerIo.instance()
```
levibostian, line 87: I believe `CustomerIO` (all caps IO) is the more proper way for the name of the company.

app/src/main/java/io/customer/example/MainActivity.kt
```diff
@@ -23,24 +24,70 @@ class MainActivity : AppCompatActivity() {
...
+                        .identify("sample@email.com@email.com", mapOf("speed" to "slow"))
+                        .execute()
```
levibostian, line 89: I am guessing the idea behind using `Action<>.execute()` is for supporting coroutines?

Also, if I am someone using the SDK and I use `execute()`, how do I perform error handling since it is synchronous? 

Lastly, is there a scenario when a customer would ever want to cancel an `Action`? I am wondering why that functionality exists?

sdk/src/main/java/io/customer/sdk/CustomerIoClient.kt
```diff
@@ -1,19 +1,58 @@
...
+
+            override fun cancel() {
+                action.cancel()
+            }
+        }
+    }
+
+    override fun track(name: String, attributes: Map<String, Any>): Action<Unit> {
+        val identifier = preferenceRepository.getIdentifier()
+        return trackingRepository.track(identifier, name, attributes)
```
levibostian, line 49: What happens when this code runs and a profile is not identified? I would expect `preferenceRepository.getIdentifier()` to return `null`, yeah?

sdk/src/main/java/io/customer/sdk/repository/AttributesRepository.kt
```diff
@@ -0,0 +1,34 @@
...
+        map.entries.forEach {
+            validMap[it.key] = getValidValue(it.value)
+        }
+        return validMap.toMap()
+    }
+
+    // Unix timestamp date are only acceptable
+    private fun getValidValue(any: Any): Any {
+        return when (any) {
+            is Date -> any.time
```
levibostian, line 30: I just ran into this issue today on iOS: make sure that `any.time` does not have decimal values. The API does not accept dates with a decimal. Decimal values may exist if the Date object has milliseconds.

sdk/src/main/java/io/customer/sdk/repository/PreferenceRepository.kt
```diff
@@ -0,0 +1,41 @@
...
+internal class PreferenceRepositoryImpl(private val context: Context) : PreferenceRepository {
+
+    companion object {
+        private const val PREFS_NAME = "customer_io_store"
+        private const val KEY_IDENTIFIER = "identifier"
+    }
+
+    private val prefs: SharedPreferences by lazy {
+        context.applicationContext.getSharedPreferences(
+            PREFS_NAME,
```
levibostian, line 21: On iOS, we use the site id as the ID for the shared preferences store that we retrieve. This avoids the scenario when someone wants to change Customer.io workspaces in their app or they want to use multiple workspaces per app. 

Would that be better to do then having all data stored in 1 shared preferences store?

app/src/main/java/io/customer/example/MainActivity.kt
Shahroz16: `Action` provides you three ways to interact 
- `Execute being synchronous and it results back with a `Result<T>` object which is a sealed class, containing either Sucess or Error object.
- `enqueue(callback: Callback<T>)`, Where `Callback`has `fun onResult(result: Result<T>)` method.

So the client is provided with a `Result` object in all cases, except you when you want to ignore `callback`

- For `cancel` probably doesn't make much sense for `tracking` messages but for `transaction` requests that our system supports it's going to be very handy. Plus, in modern tooling you have ability to cancel `coroutine` `disposable` from RX world, and the cancel feature would compliment if anyone wants to clean up resources.

sdk/src/main/java/io/customer/sdk/CustomerIoClient.kt
Shahroz16: This case is handled in `TrackingRepositoryImpl`

if `identifier` is `null`, we send back 

`return ErrorResult(
                        error = ErrorDetail(
                            statusCode = StatusCode.UnIdentifiedUser
                        )
                    )`

sdk/src/main/java/io/customer/sdk/repository/AttributesRepository.kt
Shahroz16: Ack

sdk/src/main/java/io/customer/sdk/repository/PreferenceRepository.kt
Shahroz16: Absolutely, I missed the case for multiple workspaces. Would update as `site ID`being key.

sdk/src/main/java/io/customer/sdk/repository/AttributesRepository.kt
Shahroz16: Just confirmed, `any.time` covers into `Long` and doesn't have any decimal values.

app/src/main/java/io/customer/example/MainActivity.kt
levibostian: Thanks for helping me better understand the `Action` type. It's trying to offer a flexible way for customers to interact with the SDK. 

Since it's already built, I say we can leave it. Otherwise I would suggest keeping `Action` simple with the intention of making *all* SDK functions synchronous in the future like the iOS SDK is moving towards at the moment.

sdk/src/main/java/io/customer/sdk/CustomerIO.kt
```diff
@@ -0,0 +1,128 @@
...
+Create your own instance using
+`CustomerIo.Builder(siteId: "XXX", apiKey: "XXX", region: Region.US, appContext: Application context)`
+It is recommended to initialize the client in the `Application::onCreate()` method.
+
+After the instance is created you can access it via singleton instance: `CustomerIO.instance()` anywhere,
+ */
+
+class CustomerIO internal constructor(
+    val config: CustomerIOConfig,
+    private val api: CustomerIoApi,
```
levibostian, line 24: `CustomerIO.Companioninstance` creates a singleton instance of the `CustomerIO` class. 

Looking at all of the properties of the `CustomerIO` class, `CustomerIOConfig` at this time is safe to keep a strong reference of as it's just a collection of Strings and such. 

I am, however, nervous about keeping a strong reference to `CustomerIoApi` in this class as that class has dependencies which those have dependencies, etc, etc. It seems prone to a memory leak bug in the future. 

Any suggestions for how we can prevent in the code base? 

...I am glad I am seeing this now because [I did something similar on iOS](https://github.com/customerio/customerio-ios/blob/c0eb2737bc2727e7aa64ce1e43b33447766e943a/Sources/Tracking/CustomerIOImplementation.swift#L26) that should be addressed!

levibostian: I am not necessarily suggesting we change the way we do dependency injection. You using constructor injection is a good thing. 

I am curious if there is a safer way to manage the `CustomerIO` singleton class. It's the fact that the singleton class has a strong reference to a dependency (in this case a Repository) and the Repository has many nested dependencies inside of it is what I mean by this being error-prone. If we were able to not make `CustomerIO` class a singleton instance (not likely?) or try to limit the strong references in the class to instances that are light-weight (not having a lot of nested dependencies could fix this issue. 

This is more of a conversation. Not something to fix in this code review.

PR #31, opened 2021-10-31
feat: Push notification 

Shahroz16:
- Add/Remove device
- Track push notification metrics
- Push notification service and reciever


messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingService.kt
```diff
@@ -0,0 +1,159 @@
...
+            }
+
+            return true
+        }
+
+        /**
+         * Handles new or refreshed token
+         * Call this from [FirebaseMessagingService] to register the new device token
+         */
+        fun handleNewToken(token: String) {
```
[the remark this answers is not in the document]
levibostian: Oh, I see this now further down the class. From the function documentation, it didn't imply to me that this function is optional and that you can instead call `onNewToken()`.

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOPushActionReceiver.kt
```diff
@@ -0,0 +1,36 @@
...
+import io.customer.messagingpush.CustomerIOFirebaseMessagingService.Companion.DELIVERY_ID
+import io.customer.messagingpush.CustomerIOFirebaseMessagingService.Companion.DELIVERY_TOKEN
+import io.customer.messagingpush.CustomerIOFirebaseMessagingService.Companion.REQUEST_CODE
+import io.customer.sdk.CustomerIO
+import io.customer.sdk.data.request.MetricEvent
+
+class CustomerIOPushActionReceiver : BroadcastReceiver() {
+
+    companion object {
+        const val ACTION = "com.customer.messagingpush.PUSH_ACTION"
```
levibostian, line 16:
```suggestion
        const val ACTION = "io.customer.messagingpush.PUSH_ACTION"
```

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOPushActionReceiver.kt
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    override fun onReceive(context: Context?, intent: Intent?) {
+        // Dismiss the notification
+        val requestCode = intent!!.getIntExtra(REQUEST_CODE, 0)
+        val mNotificationManager =
+            context!!.getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
```
levibostian, line 23: Can we avoid the use of `!!`? It can use NPE at runtime and crash the SDK.

levibostian (approved):

PR #32, opened 2021-11-03
docs: add documentation for existing features

Shahroz16:


README.md
```diff
@@ -1,17 +1,36 @@
-![min Android SDK version is 16](https://img.shields.io/badge/min%20Android%20SDK-16-green)
-[![Maven Central](https://maven-badges.herokuapp.com/maven-central/io.customer.android/sdk/badge.svg)](https://maven-badges.herokuapp.com/maven-central/io.customer.android/sdk)
+![min Android SDK version is 16](https://img.shields.io/badge/min%20Android%20SDK-21-green)
```
levibostian, line 1:
```suggestion
![min Android SDK version is 21](https://img.shields.io/badge/min%20Android%20SDK-21-green)
```

README.md
```diff
@@ -22,26 +41,183 @@ allprojects {
...
 
-...Not yet ready. Under development. 
+Before you can use the Customer.io SDK, you need to initialize it. `CustomerIO` is a singleton: you'll create it once and re-use it across your application.
 
-# Contributing
+```kotlin
+class App : Application() {
+   override fun onCreate() {
+       super.onCreate()
+      val customerIO = CustomerIO.Builder(
```
levibostian, line 63: Is there a reason for holding a reference to the `.build()` result? (`val customerIO`)

If so, it may be a good idea to explain that with a code comment in this code sample.

README.md
```diff
@@ -22,26 +41,183 @@ allprojects {
 }
 ```
 
-2. Add your dependency to your app's `build.gradle` file:
+### Available Artifiacts
+
+2. Add relevant dependencies to your app's `build.gradle` file:
```
levibostian, line 46: I suggest we re-word this to say something instead like, "here are all of the available SDKs that you can install. We suggest you start only with `tracking` and install others later if you need them". Right now with the way the docs are written, customers don't know what SDK does what and what they need.

README.md
```diff
@@ -22,26 +41,183 @@ allprojects {
...
+            apiKey = "your-api-key",
+            appContext = this
+        ).build()
+   }
+}
+```
+The Builder for CustomerIO exposes configuration options for features such `region`,
+`timeout`. 
+
+> A best practice is to initialize CustomerIO in the Application class, this way you will be able to access that instance from any part of your application using the instance() method:
```
levibostian, line 74: > using the instance() method:

I see a `:` suggesting that maybe there is supposed to be more written after this?

README.md
levibostian: [The docs look like they do a good job of doing this already](https://github.com/customerio/customerio-android/pull/32/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R79) so maybe we don't even need this paragraph?

README.md
```diff
@@ -22,26 +41,183 @@ allprojects {
...
+## Identify a customer
+
+When you identify a customer, you:
+1. Add or update the customer's profile in your workspace.
+2. Save the customer's information on the device. Future calls to the SDK are linked to the last-identified customer. For example, after you identify a person, any events that you track are automatically associated with that person.
+
+You can only identify one customer at a time. The SDK "remembers" the most recently-identified customer.
+If you identify customer "A", and then call the identify function for customer "B", the SDK "forgets" customer "A" and assumes that customer "B" is the current app user. 
+
+[Learn more about the `Tracking` class][s-tracking].
```
levibostian, line 92: Don't forget to include [the markdown links](https://github.com/customerio/customerio-ios/blob/alpha/README.md?plain=1#L1-L5) to make this link work.

README.md
```diff
@@ -22,26 +41,183 @@ allprojects {
...
+            ).enqueue { result ->
+                when (result) {
+                    is ErrorResult -> {
+                        // Handle error
+                    }
+                    is Success -> {
+                        // Handle success
+                    }
+                }
+            }
```
levibostian, line 108: Formatting this would make this look a little cleaner.

README.md
```diff
@@ -22,26 +41,183 @@ allprojects {
...
+If you identify customer "A", and then call the identify function for customer "B", the SDK "forgets" customer "A" and assumes that customer "B" is the current app user. 
+
+[Learn more about the `Tracking` class][s-tracking].
+
+```kotlin
+CustomerIO.instance()
+            .identify(
+                identifier = "989388339",
+                attributes = mapOf("first_name" to "firstName")
+            ).enqueue { result ->
```
levibostian, line 99: I think it would be a good idea to introduce the design of the SDK in using `enqueue` and other functions of the `Action` class so customers can properly use the public functions.

README.md
```diff
@@ -22,26 +41,183 @@ allprojects {
...
+                when (result) {
+                    is ErrorResult -> {
+                        // Handle error
```
levibostian, line 102: The iOS docs have a section on [recommended error handling](https://github.com/customerio/customerio-ios#Error-handling). I know the Android SDK has some suggestions, too. It would be helpful to mention this here to be more helpful.

README.md
```diff
@@ -22,26 +41,183 @@ allprojects {
...
+                    }
+                }
+            }
+```
+
+`attributes` accepts the following data types:
+- Primitives (int, float, char...) and their boxed counterparts (Integer, Float, Character...).
+- Arrays, Collections, Lists, Sets, and Maps
+- Strings
+- Enums
```
levibostian, line 135: how are enums converted to JSON strings in the SDK?

README.md
```diff
@@ -22,26 +41,183 @@ allprojects {
...
+
+ override fun onMessageReceived(message: RemoteMessage) {
+        val handled = CustomerIOFirebaseMessagingService.onMessageReceived(this, message)
+        if (handled) {
+            logger.breadcrumb(this, "Push notification has been handled", null)
+        }
+ }
+ 
+ override fun onNewToken(token: String) {
+     CustomerIOFirebaseMessagingService.onNewToken(token)
```
levibostian, line 175: I suggest a note in the doc mentioning that at this time, the SDK [does not have error handling implemented](https://github.com/customerio/customerio-android/blob/c447a38b6d47d28b5200b23b855909f52eb0fb43/messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingService.kt#L185) so if there is an error adding the device token to the CIO profile, the device token will be lost.

README.md
```diff
@@ -22,26 +41,192 @@ allprojects {
...
+                     }
+                     is Success -> {
+                         // Handle success
+                     }
+                 }
+             }
+     }
+}
+```
+
```
levibostian, line 188: Somewhere in the push notifications section of the docs we should mention our default Android Notification channel and our plans to allow you to customize it in the future.

README.md
levibostian: 🤔 So, `customerIO` is an instance of `CustomerIO` where you can use it to call public SDK functions of the SDK instead of using the singleton `CustomerIO`? 

If that is the purpose of `customerIO`, I am confused on the use case for that since the README document says "CustomerIO is a singleton: you'll create it once and re-use it across your application.". This sentence makes me think that using the singleton is the only way of using the SDK. 

I like the idea of being able to use the SDK with a singleton or not a singleton (the iOS SDK does this). If this works, I think the readme needs to be updated to not convey an assumption that singleton is the only way of using the SDK.

README.md
levibostian: Cool! I do see these sections. I suggest adding a link to the callback/error handling section of the document. 

Here is an example of how the iOS docs does this:
2021-11-05 at 2 23 PM

README.md
levibostian: Oh, sorry. I meant what does the resulting JSON string look for an enum? If I was a customer reading this document, I would ask that question. 

If I was using this code:
```kotlin
CustomerIO.instance().trackEvent("edit profile", mapOf("section" to ProfileSection.PHOTO))
```

I would wonder, what does the resulting JSON string for attributes look like?

```json
{"section": ???}
```

README.md
levibostian: The code sample in the README has been updated to use `onNewToken()` with an error callback so this comment has been addressed 👍

README.md
Shahroz16: Okay so let me explain it a bit more, `CustomerIO` is indeed a `singleton ` class. And to initialize it we are using `Builder Pattern`.  And `build` method of the `Builder pattern` always returns the class object and the next time `CustomerIO` instance is accessed it's always the one you created before. 

``` Koltin
class CustomerIO internal constructor(
    val config: CustomerIOConfig,
    private val api: CustomerIoApi,
) {
    companion object {
        private var instance: CustomerIO? = null

        @JvmStatic
        fun instance(): CustomerIO {
            return instance
                ?: throw IllegalStateException("CustomerIo.Builder::build() must be called before obtaining CustomerIo instance")
        }
}
```

``` Koltin
// build method of the Builder class inside 
 fun build(): CustomerIO {

            if (apiKey.isEmpty()) {
                throw IllegalStateException("apiKey is not defined in " + this::class.java.simpleName)
            }

            if (siteId.isEmpty()) {
                throw IllegalStateException("siteId is not defined in " + this::class.java.simpleName)
            }

            val config = CustomerIOConfig(
                siteId = siteId,
                apiKey = apiKey,
                region = region,
                timeout = timeout
            )

            val customerIoComponent =
                CustomerIOComponent(customerIOConfig = config, context = appContext)

            val client = CustomerIO(config, customerIoComponent.buildApi())
            instance = client
            return client
        }
```

So as you can see `instance = client`, once the instance is created it will get reused when accessed via `CustomerIO.instance()`

README.md
Shahroz16: It would be exactly hor the enum is set unless the adapter is customized 

so in the mentioned case, it would look like this 

`{"section": "PHOTO"}`

README.md
levibostian: I understand how the code `CustomerIO.Builder(...).build()` works. Sorry, I believe that my comments were not asking about how it works but more about the docs and use cases around it. Mind reading my comments again?

README.md
levibostian: Ok. I suggest a comment in the docs mentioning that.

README.md
Shahroz16: hmm, I am not sure, what would be another way to phrase it because we wouldn't want the client to make a new instance each time with `Builder` and then use it. 

Maybe something like this,
`Before you can use the Customer.io SDK, you need to initialize it. `CustomerIO` is a singleton: once created it will be re-used until you decide to reconfigure and recreate the instance.`

README.md
Shahroz16: Mind suggesting it again, because I feel like it got lost among other comments and updates to README.?

levibostian (approved):

PR #34, opened 2021-11-10
feat: Rich push support 

Shahroz16:
- Support for links (universal and deep link)
- Support to override deep link handling in the host application
- Support for images and image style of notifications 
- Support for both Notification payload and data payload in notifications


messagingpush/src/main/java/io/customer/messagingpush/CustomerIOPushNotificationHandler.kt
```diff
@@ -0,0 +1,192 @@
...
+            .bigLargeIcon(null)
+            .setSummaryText(body)
+        val url = URL(imageUrl)
+        withContext(Dispatchers.IO) {
+            try {
+                val input = url.openStream()
+                BitmapFactory.decodeStream(input)
+            } catch (e: IOException) {
+                null
+            }
```
levibostian, line 183: Are there more exceptions that can be thrown besides `IOException` from this `try{}` block? I agree that this is a good approach to try to create a Bitmap and if for any reason it fails, just continue on with the push notification and display it. My concern is that there could be a different exception thrown and the SDK code does not catch it resulting in a runtime crash.

sdk/src/main/java/io/customer/sdk/CustomerIO.kt
```diff
@@ -52,6 +54,16 @@ class CustomerIO internal constructor(
...
+        /**
+         * Override url/deep link handling
+         *
+         * @param urlHandler callback called when deeplink push action is performed.
+         */
+        fun setCustomerIOUrlHandler(urlHandler: CustomerIOUrlHandler): Builder {
+            this.urlHandler = urlHandler
+            return this
+        }
+
```
levibostian, line 66: What use case would having a url handler help with? I have wondered about a url handler as well, but wasn't sure if it provided enough value because [apps can setup deep link handling](https://developer.android.com/training/app-links/deep-linking) in their app. My thought is that [setting up deep link handling](https://developer.android.com/training/app-links/deep-linking) is a universal way for an app to handle deep links not dependent on an SDK or tool.

messagingpush/src/main/AndroidManifest.xml
```diff
@@ -16,12 +16,25 @@
...
+            <action android:name="android.intent.action.VIEW" />
+            <category android:name="android.intent.category.BROWSABLE" />
+            <data android:scheme="https" />
+        </intent>
+        <intent>
+            <action android:name="android.intent.action.VIEW" />
+            <category android:name="android.intent.category.BROWSABLE" />
+            <data android:scheme="http" />
+        </intent>
+    </queries>
```
levibostian, line 38: What is the need of this code?

levibostian (approved):

messagingpush/src/main/AndroidManifest.xml
Shahroz16: We look for possible applications to open a deep link including browser links and for (API level 30) or higher this is needed to do that.

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOPushNotificationHandler.kt
Shahroz16: Good point, it would be safer to go with the `Exception` class and not just `IOException`

sdk/src/main/java/io/customer/sdk/CustomerIO.kt
Shahroz16: This is there for the applications that do the internal Deep linking managements via some Deep linking manager. For [example](https://proandroiddev.com/an-effective-deep-link-navigation-approach-for-multi-modular-android-applications-63860a3f45b3), applications that use multi modules approach and activities in each module. Or an activity, that has complex navigation graphs and supports fragment navigations of deep link internally, or any application that uses 3rd party deep linking [library](https://github.com/airbnb/DeepLinkDispatch). 

So we support both cases, android recommended `Manifest` way or URL handler.

PR #39, opened 2021-11-17
fix: pro-guarding removal of enums 

Shahroz16:
https://github.com/square/moshi#r8--proguard


sdk/src/main/java/io/customer/sdk/data/request/Metric.kt
```diff
@@ -3,6 +3,7 @@ package io.customer.sdk.data.request
 import com.squareup.moshi.Json
 import com.squareup.moshi.JsonClass
 
+@JsonClass(generateAdapter = false)
```
levibostian, line 6: 🤔 I have only used `generateAdapter = true`. Why are we using `false` here?

Shahroz16: R8/Pro-guard removes them, hence this is [recommended](https://github.com/square/moshi#r8--proguard) by Moshi. 

Moshi don't need adapters for Enums, it caters them internally.

levibostian: Ah, 🤦‍♂️. I saw your link (thank you) in the description but didn't see the section titled "Enum". Got it.

levibostian (approved):
I just confirmed that [we are QA testing r8 compiled builds of the sample app](https://github.com/customerio/RemoteHabits-Android/blob/f396fb57a7cb39b86bb6ec37b713a99a51c4aaaa/fastlane/Fastfile#L103) so we can confirm we are testing our SDK with proguard enabled. Good to double check. With that in mind, LGTM.

PR #47, opened 2021-12-01
chore: add unit test cases 

Shahroz16:


sdk/src/test/java/io/customer/sdk/CustomerIOClientTest.kt
```diff
@@ -0,0 +1,198 @@
...
+        verify(preferenceRepository, times(1)).saveIdentifier(any())
+    }
+
+    @Test
+    fun `verify when customer identify is cleared its removed in prefs repo`() {
+        Mockito.`when`(preferenceRepository.getIdentifier()).thenReturn("identify")
+
+        customerIOClient.clearIdentify()
+
+        verify(preferenceRepository, times(1)).removeIdentifier(any())
```
levibostian, line 85:
```suggestion
        verify(preferenceRepository, times(1)).removeIdentifier("identify")
```

To verify `clearIdentify` passes the expected value.

sdk/src/test/java/io/customer/sdk/CustomerIOClientTest.kt
```diff
@@ -0,0 +1,198 @@
...
+        Mockito.`when`(
+            preferenceRepository.getIdentifier()
+        ).thenReturn("identify")
+
+        Mockito.`when`(pushNotificationRepository.registerDeviceToken(any(), any()))
+            .thenReturn(ActionUtils.getEmptyAction())
+
+        customerIOClient.registerDeviceToken("token").execute()
+
+        verify(preferenceRepository, times(1)).saveDeviceToken(any())
```
levibostian, line 113:
```suggestion
        verify(preferenceRepository, times(1)).saveDeviceToken("token")
```

sdk/src/test/java/io/customer/sdk/CustomerIOClientTest.kt
```diff
@@ -0,0 +1,198 @@
...
+
+        Mockito.`when`(preferenceRepository.getDeviceToken()).thenReturn("token")
+
+        Mockito.`when`(
+            pushNotificationRepository.deleteDeviceToken(any(), any())
+        ).thenReturn(ActionUtils.getEmptyAction())
+
+        customerIOClient.deleteDeviceToken().execute()
+
+        verify(preferenceRepository, times(1)).removeDeviceToken(any())
```
levibostian, line 130:
```suggestion
        verify(preferenceRepository, times(1)).removeDeviceToken("token")
```

sdk/src/test/java/io/customer/sdk/CustomerIOClientTest.kt
```diff
@@ -0,0 +1,198 @@
...
+            identityRepository = identityRepository,
+            trackingRepository = trackingRepository,
+            pushNotificationRepository = pushNotificationRepository,
+            preferenceRepository = preferenceRepository
+        )
+    }
+
+    @Test
+    fun `verify client sends error when identify repo fails identifying`() {
+        Mockito.`when`(identityRepository.identify(any(), any()))
```
levibostian, line 43: I personally prefer when writing Android tests with mockito to use a static import for `when()`. 

```kt
Mockito.`when`()

to

when()
```

Do you have a preference? It's boilerplate to me to continuously see `Mockito.XXX` when writing tests.

sdk/src/test/java/io/customer/sdk/CustomerIOTest.kt
```diff
@@ -0,0 +1,182 @@
...
+        customerIO = mockCustomerIO.build()
+    }
+
+    @Test
+    fun `verify SDK configuration is correct`() {
+        customerIO.config.siteId `should be equal to` "mock-site"
+        customerIO.config.apiKey `should be equal to` "mock-key"
+        customerIO.config.timeout `should be equal to` 6000
+        customerIO.config.region `should be equal to` Region.US
+        customerIO.config.urlHandler `should be equal to` null
```
levibostian, line 35: I prefer to avoid hard-coded values in tests because if, for example, `MockCustomerIOBuilder.siteId` changes to `"mock-site-id"` by someone in the future, this test breaks. 

What about something like this? 

```kt
    fun `verify SDK configuration is set to default values`() {
        customerIO.config.siteId `should be equal to` MockCustomerIOBuild.siteId
        ...
```

sdk/src/main/java/io/customer/sdk/api/retrofit/CustomerIoCallAdapter.kt
```diff
@@ -1,5 +1,6 @@
 package io.customer.sdk.api.retrofit
```
levibostian, line 1: I can't seem to wrap my head around this whole file + it's unit test. I am not understanding what it's used for and what it's doing. Perhaps more descriptive names of classes, functions, variables + docs would help?

sdk/src/test/java/io/customer/sdk/repositories/AttributesRepositoryTest.kt
```diff
@@ -0,0 +1,108 @@
...
+        attributesRepository = MoshiAttributesRepositoryImp(parser)
+    }
+
+    @Test
+    fun `Verify string attributes are mapped correctly`() {
+        val expected = mapOf("key" to "pair", "key2" to "pair2")
+
+        val result = attributesRepository.mapToJson(mapOf("key" to "pair", "key2" to "pair2"))
+
+        (result == expected).shouldBeTrue()
```
levibostian, line 29: Can we write this as `result shouldBeEqual expected`? Using *should equal* over *should true*?

sdk/src/test/java/io/customer/sdk/repositories/AttributesRepositoryTest.kt
```diff
@@ -0,0 +1,108 @@
...
+    @Before
+    fun setup() {
+        attributesRepository = MoshiAttributesRepositoryImp(parser)
+    }
+
+    @Test
+    fun `Verify string attributes are mapped correctly`() {
+        val expected = mapOf("key" to "pair", "key2" to "pair2")
+
+        val result = attributesRepository.mapToJson(mapOf("key" to "pair", "key2" to "pair2"))
```
levibostian, line 27: I see the `mapOf()` code copy/pasted:

```kt
val expected = mapOf("key" to "pair", "key2" to "pair2")
val result = ... mapOf("key" to "pair", "key2" to "pair2")
```

Doesn't seem valuable. What are the test functions in this class testing?

levibostian (approved):

levibostian:
Any plans to write integration tests?

sdk/src/test/java/io/customer/sdk/repositories/AttributesRepositoryTest.kt
Shahroz16: `shouldBeEqual` in case of map, matches the order as well. But we are not using ordered maps, so it would fail unless we change all the maps to ordered maps.

sdk/src/main/java/io/customer/sdk/api/retrofit/CustomerIoCallAdapter.kt
Shahroz16: We have a `CustomerIOCallAdapter` that extends, Retrofit, `CallAdapter` class. 
These tests are around that adapter, to ensure, the conversion, responses, return types are correct.

PR #52, opened 2021-12-14
feat: automatic screen tracking

Shahroz16:


sdk/src/main/java/io/customer/sdk/CustomerIO.kt
```diff
@@ -56,6 +64,11 @@ class CustomerIO internal constructor(
             return this
         }
 
+        fun shouldAutoRecordScreenViews(shouldRecordScreenViews: Boolean): Builder {
```
levibostian, line 67: If you want to keep things consistent, in the iOS SDK we [`autoTrackScreenViews` as the function name](https://github.com/customerio/customerio-ios/blob/4db3dc097e3cf127264ef2d0830499bd10966b4d/Sources/Tracking/Store/SdkConfig.swift#L25) for configuration.

sdk/src/main/java/io/customer/sdk/CustomerIO.kt
```diff
@@ -178,4 +208,19 @@ class CustomerIO internal constructor(
...
+                activity.componentName, PackageManager.GET_META_DATA
+            )
+            val activityLabel = info.loadLabel(packageManager)
+            screen(activityLabel.toString(), attributes)
+        } catch (e: PackageManager.NameNotFoundException) {
+            ActionUtils.getErrorAction(ErrorResult(error = ErrorDetail(message = "Activity Not Found: $e")))
+        } catch (e: Exception) {
+            ActionUtils.getErrorAction(ErrorResult(error = ErrorDetail(message = "Unable to track, $activity")))
+        }
+    }
```
levibostian, line 225: Instead of this code, are we able to instead get the class name from the given Activity parameter? If an Activity class is named `LoginActivity` then we can use that class name?

buildSrc/src/main/kotlin/io.customer/android/Dependencies.kt
```diff
@@ -29,6 +29,7 @@ object Dependencies {
     const val junit4 = "junit:junit:${Versions.JUNIT4}"
     const val kluent = "org.amshove.kluent:kluent:${Versions.KLUENT}"
     const val kotlinGradlePlugin = "org.jetbrains.kotlin:kotlin-gradle-plugin:${Versions.KOTLIN}"
+    const val lifecycleCommon = "androidx.lifecycle:lifecycle-common:${Versions.LIFECYCLE_COMMON}"
```
levibostian, line 32: Do we need this dependency? Can we use [this](https://developer.android.com/reference/android/app/Application.ActivityLifecycleCallbacks) instead?

sdk/src/main/java/io/customer/sdk/CustomerIO.kt
Shahroz16: The benefit of having this is, instead of getting `class name` which doesn't necessarily depict a screen, we fetch the label given to that activity in the `Manifest`. 
A couple of clients we talked to wanted this approach as its quite standard.

buildSrc/src/main/kotlin/io.customer/android/Dependencies.kt
Shahroz16: `DefaultLifecycleObserver` needs this, but it's focused more on the latest API. We can ignore for now.

levibostian (approved):

PR #53, opened 2022-01-14
fix: updated host endpoints

Shahroz16:


hownowstephen (approved):

sdk/src/main/java/io/customer/sdk/data/model/Region.kt
```diff
@@ -5,6 +5,6 @@ package io.customer.sdk.data.model
  * The SDK will route traffic to the correct data center location depending on the `Region` that you use.
  */
 sealed class Region(val code: String, val baseUrl: String) {
-    object US : Region(code = "us", baseUrl = "https://track.customer.io/")
-    object EU : Region(code = "eu", baseUrl = "https://track-eu.customer.io/")
+    object US : Region(code = "us", baseUrl = "https://track-sdk.customer.io/")
+    object EU : Region(code = "eu", baseUrl = "https://track-sdk-eu.customer.io/")
```
levibostian, line 9: On the iOS SDK, we [added a comment](https://github.com/customerio/customerio-ios/blob/ac66abdd022f3f560aab8d595b56e4602f0ba344/Sources/Tracking/Store/SdkCredentials.swift#L16-L18) about these special SDK endpoints in case a customer reads the SDK code. We redirect readers to the official docs.

levibostian (approved):

PR #65, opened 2022-02-15
feat: create tasks in background queue

levibostian:
Task 1 or 4 for the background queue. Creating tasks by saving them to the file system. 

Closes: https://github.com/customerio/issues/issues/6672

- [x] Wait until the pull request has been reviewed *and approved* by a teammate
- [x] After pull request is approved, and you determine it's ready **add the label "Ready to merge"** to the pull request and it will automatically be merged.


sdk/src/main/java/io/customer/sdk/data/store/FileStorage.kt
```diff
@@ -0,0 +1,88 @@
...
+     queue/
+       inventory.json
+       tasks/
+         <task-id>.json
+ ```
+
+ Notice that we are using the <site id> as a way to isolate files from each other.
+ The file tree remains the same for all site ids.
+ */
+class FileStorage internal constructor(
```
Shahroz16, line 28: How do you feel about using `SQLite` instead of a file-based storage system? We will still not be adding any additional dependency, more command on how we interact with it. If we have to sort the tasks, add in priority it would be alot easier. What do you think?

sdk/src/main/java/io/customer/sdk/di/CustomerIOComponent.kt
```diff
@@ -90,11 +113,14 @@ internal class CustomerIOComponent(
         }
     }
 
+    fun buildMoshi(): Moshi = Moshi.Builder()
```
Shahroz16, line 116: should this be `lazy` or would we require to build a new instance?

sdk/src/main/java/io/customer/sdk/extensions/RandomExtensions.kt
```diff
@@ -0,0 +1,24 @@
+package io.customer.sdk.utils
+
+import java.util.*
+
+/**
+ * Set of extensions for primitive types to get a random value. Used for tests.
+ */
+
+fun String.Companion.random(length: Int): String {
```
Shahroz16, line 9: Any specific reason for using `String.Companion.random` instead of `String.random`, apart from dealing with `char` and `char arrays`?

sdk/src/main/java/io/customer/sdk/queue/Queue.kt
```diff
@@ -0,0 +1,39 @@
...
+import io.customer.sdk.util.Logger
+
+class Queue internal constructor(
+    private val storage: QueueStorage,
+    private val jsonAdapter: JsonAdapter,
+    private val sdkConfig: CustomerIOConfig,
+    private val logger: Logger
+) {
+
+    fun <TaskData : Any> addTask(type: String, data: TaskData): QueueModifyResult {
```
Shahroz16, line 16: Should the `type` of a task be an `enum`? will it give us any benifit?

sdk/src/main/java/io/customer/sdk/queue/QueueStorage.kt
```diff
@@ -0,0 +1,102 @@
...
+import io.customer.sdk.queue.type.QueueTaskRunResults
+import io.customer.sdk.util.JsonAdapter
+import java.util.*
+
+interface QueueStorage {
+    fun getInventory(): QueueInventory
+    fun saveInventory(inventory: QueueInventory): Boolean
+    fun create(type: String, data: String): QueueModifyResult
+    fun get(taskStorageId: String): QueueTask?
+    fun delete(taskStorageId: String): QueueModifyResult
```
Shahroz16, line 19: What happens if we switch `siteID` are we going to loop and delete all tasks?

sdk/src/main/java/io/customer/sdk/util/JsonAdapter.kt
```diff
@@ -0,0 +1,74 @@
...
+
+        val jsonAdapter = moshi.adapter(T::class.java)
+
+        return jsonAdapter.fromJson(json) as T
+    }
+
+    inline fun <reified T : Any> fromJsonList(json: String): List<T> {
+        val json = json.trim()
+
+        if (json.isNotEmpty() && json[0] != '[') throw IllegalArgumentException("String is not a list. Use `fromJson` instead.")
```
Shahroz16, line 54: do we need to trim it, if we are to check the 1st character in a string?

sdk/src/sharedTest/java/io/customer/sdk/data/store/DeviceStoreTest.kt
```diff
@@ -36,22 +32,22 @@ internal class DeviceStoreTest {
     }
 
     @Test
-    fun `verify build attributes in device`() {
-        deviceStore.deviceBrand `should be equal to` "Google"
-        deviceStore.deviceModel `should be equal to` "Pixel 6"
-        deviceStore.deviceManufacturer `should be equal to` "Google"
-        deviceStore.deviceOSVersion `should be equal to` 30
+    fun verifyBuildAttributesInDevice() {
+        deviceStore.deviceBrand shouldBeEqualTo "Google"
```
Shahroz16, line 36: Out of curiosity, why were these changes made? Was the previous version not that reader friendly?

Shahroz16 (approved):
Very excited to see whats next in like for BQ

sdk/src/main/java/io/customer/sdk/data/store/FileStorage.kt
levibostian: In the past, I have [built a background queue that uses SQLite](https://github.com/levibostian/wendy-android). I encountered some pains with releasing a 3rd party library that uses SQLite. I will admit I encountered more pains with [the iOS version](https://github.com/levibostian/wendy-ios). 

When building the background queue for our iOS SDK, I tried the file system approach instead of a database as an experiment. It went very well. I didn't encounter the same issues that using a database gave me. With that in mind, I suggest that we stick with using the file system knowing that if we ever encounter problems with this approach, we can migrate to a database in the future.

sdk/src/main/java/io/customer/sdk/di/CustomerIOComponent.kt
levibostian: As long as we don't make it a singleton (since it doesn't need to be), I am fine with using lazy or not. If we make the `moshi` property a `lazy`, that makes me want to ask the question, "Why not make everything `lazy`?" and to that question, I don't know if making lots of properties `lazy` is a good thing?

sdk/src/main/java/io/customer/sdk/extensions/RandomExtensions.kt
levibostian: Using `String.Companion.random` means I can do: `String.random` to get a random string. 

Using `String.random` means I have to call: `"".random` to get a random string. 

Having a static property (via `String.Companion`) makes more sense to me from the code above.

sdk/src/main/java/io/customer/sdk/queue/Queue.kt
levibostian: When I do [this task](https://github.com/customerio/issues/issues/6674) you will understand why we have to use a `String`. In iOS, I used an enum until I did that task and I had to change it to using a String. To describe it briefly, you can't reference an `enum` that exists in the `messagingpush` module from the `tracking` module where the queue code exists. 

My plan is the same as how the iOS implementation is:

```kt
enum QueueTask {
  FOO

  val string: String {
    when (this) {
      FOO -> return "foo"
    }
  }
}

queue.addTask(QueueTask.FOO.string....)
```

So, the queue requires a String, but we still use an `enum` when calling `addTask()`.

sdk/src/main/java/io/customer/sdk/queue/QueueStorage.kt
levibostian: At this time, no. Once a task gets added to the background queue, it will get saved to the file system in it's own "siteid sandbox" (a directory of files in the file system only belonging to that siteid). We only delete tasks from the background queue if a task runs successfully. 

We have [had discussions](https://github.com/customerio/issues/issues/6793) about doing something like this in the future but we have not made any decisions on moving forward with doing that.

sdk/src/sharedTest/java/io/customer/sdk/data/store/DeviceStoreTest.kt
levibostian: I encountered [this issue](https://github.com/MarkusAmshove/Kluent/issues/53) which required me to change our dependency from using `kluent` to `kluent-android`. Running tests on the Android OS via instrumentation testing throws a compile time error saying that the old syntax couldn't compile. 

It's the same compile time error if you create a test function like:
```kt
    @Test
    fun `given X, expect Y`() {}   
```

compared to:
```kt
    @Test
    fun givenX_expectY() {}   
```

sdk/src/main/java/io/customer/sdk/util/JsonAdapter.kt
levibostian: Are you asking a question about `val json = json.trim()` or `if (json.isNotEmpty() && json[0] != '[') throw IllegalArgumentException("String is not a list. Use `fromJson` instead.")`? The comment is made on line 54 but I think you're asking a question about line 52?

base/src/main/java/io/customer/base/extenstions/DateExtensions.kt
```diff
@@ -6,3 +6,9 @@ import java.util.concurrent.TimeUnit
 fun Date.getUnixTimestamp(): Long {
     return TimeUnit.MILLISECONDS.toSeconds(this.time)
 }
+
+fun Long.unixTimeToDate(): Date {
+    val seconds = this
+    val milliseconds = seconds * 1000L
```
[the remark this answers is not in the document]
levibostian: I can make this change. I'll wait until I determine what other changes to make based on other discussions in the PR.

sdk/src/main/java/io/customer/sdk/data/store/FileStorage.kt
Shahroz16: Sounds good we can start off with a file system, but if we want it to be flexible, should we then probably have a `QueueStorage` interface, with `FileStorage` as its implementation? 

```kotlin
class QueueStorageImpl internal constructor(
    private val siteId: String,
    private val fileStorage: QueueStorage,
    private val jsonAdapter: JsonAdapter
) ```

sdk/src/main/java/io/customer/sdk/di/CustomerIOComponent.kt
Shahroz16: > "Why not make everything lazy?"

Lazy initialization was designed to prevent the unnecessary initialization of objects and it comes down to how expensive is the `initialization`. 

Moshi does a maybe few milliseconds of reflection each time it creates a JSON adapter for a type it's never before seen. If you create many Moshi instances you potentially duplicate this work. This is what we can saved by having it `lazy`. 

The memory issue is not what I am not concerned with much, as its scope is bounded by `CustomerIOComponent`. But if you are comfortable with doing multiple initializations until we notice this performance hit, we can do that as well.

sdk/src/main/java/io/customer/sdk/util/JsonAdapter.kt
Shahroz16: Sorry, missed the `trim` on line 52, that's why I was suggesting to have it in like 54.

PR #71, opened 2022-03-08
feat: added support for device attributes

Shahroz16:

### DIFF
--- a/sdk/src/main/java/io/customer/sdk/data/store/ApplicationStore.kt
+++ b/sdk/src/main/java/io/customer/sdk/data/store/ApplicationStore.kt
@@ -1,11 +1,14 @@
 package io.customer.sdk.data.store
 
 import android.content.Context
+import androidx.core.app.NotificationManagerCompat
 
 interface ApplicationStore {
     // Customer App information
     val customerAppName: String
     val customerAppVersion: String
+
+    val isPushSubscribed: Boolean
 }
 
 internal class ApplicationStoreImp(val context: Context) : ApplicationStore {
@@ -16,6 +19,8 @@ internal class ApplicationStoreImp(val context: Context) : ApplicationStore {
         get() = appInfo.first
     override val customerAppVersion: String
         get() = appInfo.second
+    override val isPushSubscribed: Boolean
+        get() = NotificationManagerCompat.from(context).areNotificationsEnabled()
 
     private fun getAppInformation(): Pair<String, String> {
         val appName: String = try {
--- a/sdk/src/main/java/io/customer/sdk/data/store/DeviceStore.kt
+++ b/sdk/src/main/java/io/customer/sdk/data/store/DeviceStore.kt
@@ -16,6 +16,7 @@ interface DeviceStore : BuildStore, ApplicationStore {
      * `Customer.io Android Client/1.0.0-alpha.6`
      */
     fun buildUserAgent(): String
+    fun buildDeviceAttributes(): Map<String, Any>
 }
 
 internal class DeviceStoreImp(
@@ -32,6 +33,8 @@ internal class DeviceStoreImp(
         get() = buildStore.deviceManufacturer
     override val deviceOSVersion: Int
         get() = buildStore.deviceOSVersion
+    override val deviceLocale: String
+        get() = buildStore.deviceLocale
     override val customerAppName: String
         get() = applicationStore.customerAppName
     override val customerAppVersion: String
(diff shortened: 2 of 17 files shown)


sdk/src/main/java/io/customer/sdk/data/store/DeviceStore.kt
```diff
@@ -47,4 +50,15 @@ internal class DeviceStoreImp(
...
     }
+
+    override fun buildDeviceAttributes(): Map<String, Any> {
+        return mapOf(
+            "device_os" to deviceOSVersion,
+            "device_model" to deviceModel,
+            "app_version" to customerAppVersion,
+            "cio_sdk_version" to customerIOVersion,
+            "device_locale" to deviceLocale,
+            "push_subscribed" to true
```
levibostian, line 61: Is there a reason that this is always `true`? 

[I remember talking about what this attribute means as a squad](https://www.notion.so/custio/504e69222e0e43f3982f51e80aaaf7b3?v=bcaa50a92c5b47d6920bc9ee8ff5049a). We decided (for now) that on Android, this attribute would be true if at least 1 Android push channel was enabled. Is that possible for us to query?

sdk/src/test/java/io/customer/sdk/store/DeviceStoreTest.kt
```diff
@@ -54,4 +27,29 @@ internal class DeviceStoreTest {
...
+        deviceStore.deviceModel `should be equal to` "Pixel 6"
+        deviceStore.deviceManufacturer `should be equal to` "Google"
+        deviceStore.deviceOSVersion `should be equal to` 30
+        deviceStore.deviceLocale `should be equal to` "en"
+
+        resultDeviceAttributes.keys shouldBeEqualTo expectedDeviceAttributes.keys
+
+        expectedDeviceAttributes.keys.forEach { key ->
+            expectedDeviceAttributes[key] `should be equal to` resultDeviceAttributes[key]
+        }
```
levibostian, line 53:
```suggestion
        resultDeviceAttributes shouldBeEqualTo expectedDeviceAttributes
```

Should this work? I like this approach because it makes the test function future-proof. Each time that you update `DeviceStore.buildDeviceAttributes()`, you need to modify this test. If you use `shouldBeEqualTo`, you wouldn't need to. 

If it doesn't work, I am assuming it's because `shouldBeEqualTo` doesn't support `Map<String, Any>`? Could we create own our `shouldBeEqualToMap` to provide the same logic as:

```kotlin
resultDeviceAttributes.keys shouldBeEqualTo expectedDeviceAttributes.keys

        expectedDeviceAttributes.keys.forEach { key ->
            expectedDeviceAttributes[key] `should be equal to` resultDeviceAttributes[key]
        }
```

levibostian (commented):
In this PR, I don't see support for adding custom attributes? Is that planned for a future PR? On iOS, it got implemented in it's PR for adding device attributes. I assume Android should support it as well?

sdk/src/test/java/io/customer/sdk/store/DeviceStoreTest.kt
```diff
@@ -1,46 +1,19 @@
...
-            },
-            applicationStore = object : ApplicationStore {
-                override val customerAppName: String
-                    get() = "User App"
-                override val customerAppVersion: String
-                    get() = "1.0"
-            },
-            version = "1.0.0-alpha.6"
-        )
-    }
```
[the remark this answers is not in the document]
levibostian: That should work. Create a stub now and we can move it later 👍

sdk/src/main/java/io/customer/sdk/data/store/ApplicationStore.kt
```diff
@@ -16,6 +19,8 @@ internal class ApplicationStoreImp(val context: Context) : ApplicationStore {
         get() = appInfo.first
     override val customerAppVersion: String
         get() = appInfo.second
+    override val isPushSubscribed: Boolean
+        get() = NotificationManagerCompat.from(context).areNotificationsEnabled()
```
levibostian, line 23: This seems like a good idea to me. However, it doesn't match the plan that the squad thought up in the notion doc. So, I approve of this implementation however, I would send a message to the squad saying that this is how `push_subscribed` device attribute is going to work + update the notion doc table so LJ knows about this implementation.

levibostian:
In my last PR review, I left this comment
> In this PR, I don't see support for adding custom attributes? Is that planned for a future PR? On iOS, it got implemented in it's PR for adding device attributes. I assume Android should support it as well?

I don't see support for adding custom attributes added so this question still stands. Besides this question, the PR looks good to me.

sdk/src/main/java/io/customer/sdk/data/store/ApplicationStore.kt
Shahroz16: > In my last PR review, I left this comment
> 
> > In this PR, I don't see support for adding custom attributes? Is that planned for a future PR? On iOS, it got implemented in it's PR for adding device attributes. I assume Android should support it as well?
> 
> I don't see support for adding custom attributes added so this question still stands. Besides this question, the PR looks good to me.

Awesome thanks @levibostian ! `custome attributes` is part of future releases when BE/FE supports it.

Shahroz16:
> In my last PR review, I left this comment
> 
> > In this PR, I don't see support for adding custom attributes? Is that planned for a future PR? On iOS, it got implemented in it's PR for adding device attributes. I assume Android should support it as well?
> 
> I don't see support for adding custom attributes added so this question still stands. Besides this question, the PR looks good to me.

Awesome thanks @levibostian ! custome attributes is part of future releases when BE/FE supports it.

PR #74, opened 2022-03-14
fix: allow http requests in host app

levibostian:
Potential bug fix for a customer. 

This PR does include a modification to the snapshot deployment script as the previous script was not working (semantic-release was not running a dry run) so I changed the snapshot to use the git commit hash to always be unique. We can leave this script change in this PR, or remove it and fix that in a future fix if there is a better idea to fix it. I wanted to get a snapshot build made for the customer to test for us.

### DIFF
--- a/.github/workflows/snapshot-release.yml
+++ b/.github/workflows/snapshot-release.yml
@@ -17,8 +17,31 @@ jobs:
           distribution: adopt
           java-version: 11
 
+      # Get just version number for semantic snapshot release. It would just dry run and not publish, release
+      - name: Run semantic release to get version
+        uses: cycjimmy/semantic-release-action@v2
+        id: semantic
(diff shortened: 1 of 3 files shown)


levibostian:
PR blocked while customer tests out snapshot build for us. Once we hear back, we will proceed with deploying this fix if they report it fixes their issue.

levibostian:
Customer replied back with this fixing their issue. Ready for review + merge + deploy.

Shahroz16 (approved):
The fix looks good, regarding the snapshot release process. I don't think thats the best use of snapshot. The biggest benifit of snapshot releases is that they can be updated unlike staging/prod releases.
 
 And I believe industry practise is to name snapshot releases like 
`next-release-version-SNAPSHOT`

PR #75, opened 2022-03-21
fix: crash using invalid characters in HTTP header

levibostian:
Fixes: https://github.com/customerio/issues/issues/6980


base/src/main/java/io/customer/base/extenstions/OkHttpExtensions.kt
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * A runtime exception "Unexpected char..." will be thrown by OkHttp if certain characters exist in the Header name or value of the HTTP request.
+ *
+ * See function for the release/tag that we have installed for OkHttp:
+ * https://github.com/square/okhttp/blob/parent-4.9.1/okhttp/src/main/kotlin/okhttp3/Headers.kt#L431
+ *
+ * Here, we are removing all of the invalid characters to prevent OkHttp throwing an exception.
+ */
+fun String.filterHeaderValue(): String {
```
Shahroz16, line 11: I was browsing through Okhttp documentation, and it looks like they provide a method to?

something like
```
val header = Headers.Builder()
header.addUnsafeNonAscii("User-Agent", store.deviceStore.buildUserAgent())
```

https://square.github.io/okhttp/4.x/okhttp/okhttp3/-headers/-builder/add-unsafe-non-ascii/

jatinn: that function makes it sound like it would pass along the non-ascii characters in the request so if we decide to use it would need to make sure that wont cause any other issues.

levibostian: I asked team-engineering if non-ascii characters are going to cause issues. If it won't, we can go with this suggestion. Else, I suggest we stick with the proposed PR solution which strips out non-ascii characters.

levibostian: From the slack thread, I propose to fix this issue, we use the app's package name instead of app name for the `User-Agent` value. 

I have implemented this idea in this PR. Depending on what we decide as a team, I can do some git cherry-picking/reverting to change the state of this PR to a different solution.

PR #77, opened 2022-03-22
feat: support for custom device attributes and config

Shahroz16:

### DIFF
--- a/sdk/src/main/java/io/customer/sdk/data/store/BuildStore.kt
+++ b/sdk/src/main/java/io/customer/sdk/data/store/BuildStore.kt
@@ -17,7 +17,7 @@ interface BuildStore {
     // Android SDK Version: 21
     val deviceOSVersion: Int
 
-    // Device locale: en
+    // Device locale: en-US
     val deviceLocale: String
 }
 
@@ -32,5 +32,5 @@ internal class BuildStoreImp : BuildStore {
     override val deviceOSVersion: Int
         get() = Build.VERSION.SDK_INT
     override val deviceLocale: String
-        get() = Locale.getDefault().language
+        get() = Locale.getDefault().toLanguageTag()
 }
--- a/sdk/src/test/java/io/customer/sdk/store/DeviceStoreTest.kt
+++ b/sdk/src/test/java/io/customer/sdk/store/DeviceStoreTest.kt
@@ -13,7 +13,7 @@ internal class DeviceStoreTest : BaseTest() {
         deviceStore.deviceModel `should be equal to` "Pixel 6"
         deviceStore.deviceManufacturer `should be equal to` "Google"
         deviceStore.deviceOSVersion `should be equal to` 30
-        deviceStore.deviceLocale `should be equal to` "en"
+        deviceStore.deviceLocale `should be equal to` "en-US"
     }
 
     @Test
@@ -36,7 +36,7 @@ internal class DeviceStoreTest : BaseTest() {
             "device_model" to "Pixel 6",
             "app_version" to "1.0",
             "cio_sdk_version" to "1.0.0-alpha.6",
-            "device_locale" to "en",
+            "device_locale" to "en-US",
             "push_enabled" to true
         )
 
--- a/sdk/src/test/java/io/customer/sdk/utils/DeviceStoreStub.kt
+++ b/sdk/src/test/java/io/customer/sdk/utils/DeviceStoreStub.kt
@@ -20,7 +20,7 @@ class DeviceStoreStub {
                 override val deviceOSVersion: Int
                     get() = 30
                 override val deviceLocale: String
-                    get() = Locale.US.language
+                    get() = Locale.US.toLanguageTag()
             },
             applicationStore = object : ApplicationStore {
                 override val customerAppName: String


app/src/main/java/io/customer/example/MainActivity.kt
```diff
@@ -99,7 +106,7 @@ class MainActivity : AppCompatActivity() {
             when (
                 val result =
                     CustomerIO.instance().identify(
-                        identifier = "support-ticket-test",
+                        identifier = "device-attri",
```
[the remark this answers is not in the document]
Shahroz16: That was pushed by mistake, I was testing both cases.

sdk/src/main/java/io/customer/sdk/CustomerIO.kt
```diff
@@ -198,6 +221,14 @@ class CustomerIO internal constructor(
         deviceToken = deviceToken
     )
 
+    /**
+     * Use to provide additional and custom device attributes
+     * apart from the ones the SDK is programmed to send to customer workspace.
+     */
+    fun addCustomDeviceAttributes(attributes: Map<String, Any>) {
```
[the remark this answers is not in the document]
Shahroz16: Majorly because of in `Java` it will not look like this, it would look something like this

```
 CustomerIO.instance().getDeviceAttributes().put("", "");
```

Also, `deviceAttributes` would be `final`. But I can try parity with iOS.

sdk/src/main/java/io/customer/sdk/CustomerIO.kt
```diff
@@ -178,7 +193,15 @@ class CustomerIO internal constructor(
...
-        api.registerDeviceToken(deviceToken, store.deviceStore.buildDeviceAttributes())
+        api.registerDeviceToken(deviceToken, collectDeviceAttributes())
+
+    private fun collectDeviceAttributes(): Map<String, Any> {
+        return if (config.autoTrackDeviceAttributes) {
+            config.deviceAttributes + store.deviceStore.buildDeviceAttributes()
+        } else {
+            config.deviceAttributes
+        }
+    }
```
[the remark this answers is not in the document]
Shahroz16: added

sdk/src/main/java/io/customer/sdk/CustomerIOConfig.kt
```diff
@@ -9,5 +9,7 @@ class CustomerIOConfig(
     val region: Region,
     val timeout: Long,
     val urlHandler: CustomerIOUrlHandler?,
-    val autoTrackScreenViews: Boolean
+    val autoTrackScreenViews: Boolean,
+    val autoTrackDeviceAttributes: Boolean,
+    val deviceAttributes: MutableMap<String, Any>
```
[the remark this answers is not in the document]
Shahroz16: changed

sdk/src/main/java/io/customer/sdk/CustomerIO.kt
```diff
@@ -74,11 +72,6 @@ class CustomerIO internal constructor(
             return this
```
levibostian, line 72: 👏 Thanks, Android lint 👏. Without the lint warning here, I wouldn't have noticed this. 

This line of code is a bug. Should be:
```kt
this.shouldAutoRecordScreenViews = shouldTrackDeviceAttributes
```

Phew. Wouldn't have noticed that if it wasn't for the lint. 

Can we add to our `Builder` unit tests to test that this property gets set properly?

sdk/src/test/java/io/customer/sdk/CustomerIOTest.kt
```diff
@@ -156,6 +158,55 @@ internal class CustomerIOTest : BaseTest() {
...
+                any(),
+                any(),
+            )
+        ).thenReturn(getEmptyAction())
+
+        customerIO.deviceAttributes.putAll(mapOf("test" to "value"))
+
+        val expectedToken = "token"
+
+        val expectedAttributes = customerIO.deviceAttributes + deviceStore.buildDeviceAttributes()
```
levibostian, line 175: I have an idea to improve the quality of this test. 

For the expected value, have as little logic involved in generating the expected value is a good idea. What I mean by logic is:
```kt
// logic is involved to create the expected value
val expected = 2 + 2 // <-- if this had a bug in it, the test could give a false positive

// compared to no logic
val expected = 4 // <-- no logic == less chance of encountering a false positive. 
```

In my past experience, I have encountered times where my test functions gave false positives because what I truly expected to be the expected value was incorrect. That was because it had too much logic involved where the logic actually had a bug in it. 

So, I would suggest that this line of code instead says:
```kt
// instead of having logic generate the expected value, hard-code the expected value in. 
val expectedAttributes = mapOf("test" to "value, "push_enabled" to "true)
```

yes, this would require modifying the returned value from `deviceStore.buildDeviceAttriutes()` to something hard-coded/predictable but stubs should give the ability to modify returned values so it shouldn't be a problem.

sdk/src/test/java/io/customer/sdk/CustomerIOTest.kt
```diff
@@ -156,6 +158,55 @@ internal class CustomerIOTest : BaseTest() {
...
+
+        val response = customerIO.registerDeviceToken(expectedToken).execute()
+
+        verify(mockCustomerIO.api).registerDeviceToken(expectedToken, expectedAttributes)
+
+        verifySuccess(response, Unit)
+    }
+
+    @Test
+    fun `verify SDK skips default attributes on basis of config`() {
```
levibostian, line 185: In the background queue PR, I moved tests from `test/` into `sharedTest/` so that they can be run from JVM or Android OS. Since our SDK at that point does more operations on the OS, it's a good idea to test on Android OS as well as JVM. With this move, test function naming does not allow the syntax of using the ` character in it. If you could, please, rename test function names to snakeCase that would save us time to refactor in the future. 

No need to rename *all* test function names. Just for the test functions you have created. 

Same with `should be equal to` needs to be snakeCase, too.

sdk/src/test/java/io/customer/sdk/MockCustomerIOBuilder.kt
```diff
@@ -18,17 +18,20 @@ internal class MockCustomerIOBuilder {
         const val timeout = 6000
         val urlHandler = null
         const val shouldAutoRecordScreenViews = false
-    }
+        const val autoTrackDeviceAttributes = true
```
levibostian, line 21: As stated in a previous comment, there is a bug in the `CustomerIO.Builder.autoTrackDeviceAttributes`. Since `true` is the default value of `autoTrackDeviceAttributes` in the SDK, anytime that our automated tests run, it's not really testing the Builder. I suggest modifying the test function to test both true and false work or at least changing this to `false` to be opposite of SDK config's default value.

sdk/src/test/java/io/customer/sdk/CustomerIOTest.kt
```diff
@@ -28,13 +29,14 @@ internal class CustomerIOTest : BaseTest() {
...
     @Test
-    fun `verify SDK configuration is correct`() {
+    fun `verify SDK default configuration is correct`() {
         customerIO.config.siteId `should be equal to` MockCustomerIOBuilder.siteId
         customerIO.config.apiKey `should be equal to` MockCustomerIOBuilder.apiKey
         customerIO.config.timeout `should be equal to` MockCustomerIOBuilder.timeout.toLong()
         customerIO.config.region `should be equal to` MockCustomerIOBuilder.region
         customerIO.config.urlHandler `should be equal to` MockCustomerIOBuilder.urlHandler
         customerIO.config.autoTrackScreenViews `should be equal to` MockCustomerIOBuilder.shouldAutoRecordScreenViews
+        customerIO.config.autoTrackDeviceAttributes `should be equal to` MockCustomerIOBuilder.autoTrackDeviceAttributes
```
levibostian, line 39: Here is the false positive test. I believe that if you were to change `sdk/src/test/java/io/customer/sdk/MockCustomerIOBuilder.kt` default value of `autoTrackDeviceAttributes` to `false` this test function would fail.

sdk/src/main/java/io/customer/sdk/CustomerIO.kt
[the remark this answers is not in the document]
levibostian: I had a feeling it was for Java compatability. 

I would love to see our SDK be Kotlin-first with java support added on top. Make the SDK's API feel Kotlin-y. Right now it feels like our SDK is designed to be a Java SDK that happens to work with Kotlin with some nice Kotlin features built-in such as coroutines. 

This is my opinion and something we can address more long-term if we choose. Highest priority today is our SDK has Java support and you have done a great job making sure that is the case.

sdk/src/test/java/io/customer/sdk/MockCustomerIOBuilder.kt
Shahroz16: Added test for changing values should reflect accordingly,

levibostian (approved):
There are 3 unresolved conversations in the PR. I am unsure how you plan to address those? Make a change now, make a ticket to change later, we decide to disagree with the suggestion and we don't make a change ever. 

However, they are minor conversations so I am OK with moving this PR forward and into `alpha`!

sdk/src/main/java/io/customer/sdk/data/store/BuildStore.kt
```diff
@@ -17,7 +17,7 @@ interface BuildStore {
     // Android SDK Version: 21
     val deviceOSVersion: Int
 
-    // Device locale: en
+    // Device locale: en-US
```
levibostian, line 20: From conversations with LJ, I believe the backend is looking for `en_US` (notice underscore).

PR #78, opened 2022-03-22
feat: implement grouping tasks to run together background queue

levibostian:


sdk/src/sharedTest/java/io/customer/sdk/CustomerIOClientTest.kt
```diff
@@ -73,13 +81,26 @@ class CustomerIOClientTest : BaseTest() {
...
 
-        verify(backgroundQueueMock).addTask(QueueTaskType.IdentifyProfile, IdentifyProfileQueueTaskData(newIdentifier, givenAttributes))
-        verify(backgroundQueueMock, never()).addTask(eq(QueueTaskType.DeletePushToken), any())
-        verify(backgroundQueueMock).addTask(QueueTaskType.RegisterDeviceToken, RegisterPushNotificationQueueTaskData(newIdentifier, givenDeviceToken, dateUtilStub.givenDate))
+        inOrder(backgroundQueueMock).apply {
+            verify(backgroundQueueMock).addTask(
+                QueueTaskType.IdentifyProfile,
+                IdentifyProfileQueueTaskData(newIdentifier, givenAttributes),
+                groupStart = QueueTaskGroup.IdentifyProfile(newIdentifier),
+                blockingGroups = null
```
Shahroz16, line 93: `blockingGroups` can be `null` by default?

sdk/src/main/java/io/customer/sdk/CustomerIOClient.kt
```diff
@@ -47,7 +48,18 @@ internal class CustomerIOClient(
             }
         }
 
-        val queueStatus = backgroundQueue.addTask(QueueTaskType.IdentifyProfile, IdentifyProfileQueueTaskData(identifier, attributes))
+        // If SDK previously identified profile X and X is being identified again, no use blocking the queue with a queue group.
+        val queueGroupStart = if (isFirstTimeIdentifying || isChangingIdentifiedProfile) QueueTaskGroup.IdentifyProfile(identifier) else null
+        // If there was a previously identified profile, or, we are just adding attributes to an existing profile, we need to block
```
Shahroz16, line 53: Thank you for adding very useful comments, helps a lot in understanding the approach. We talked about this before as well but since these tasks were going to be smaller we said, it might not be worth moving tasks as separate classes/entities. 

Coming back to that, I feel like this logic should not belong to `CustomerIOClient` since it's `task-specific. How can we move it from her? if sub-tasks are bit overwork, maybe repositories/use-cases ?

sdk/src/main/java/io/customer/sdk/queue/type/QueueTaskGroup.kt
```diff
@@ -0,0 +1,19 @@
...
+    /**
+     * Be sure that the string is unique to each subclass of [QueueTaskGroup].
+     * The returned string will identify each group in the background queue. Therefore, it's important to use properties from the subclass constructor in the returned string to differentiate groups of the same type.
+     */
+    data class IdentifyProfile(val identifier: String) : QueueTaskGroup() {
+        override fun toString(): String = "identified_profile_$identifier"
+    }
+    data class RegisterPushToken(val token: String) : QueueTaskGroup() {
+        override fun toString(): String = "registered_push_token_$token"
+    }
```
Shahroz16, line 18: I really like the grouping feature, there are a lot of use cases for it, not just pre-defined but maybe runtime as well. 

For example, there is a particular event we want to track but we want some other events to get tracked before it, a real-life example could be, I create an event `Onboarding Completed` but then I block it by the group, `Onboarding`. And when a user is going through each onboarding page, I am adding the event with the group `Onboarding"` in it. Once the user goes through all onboarding pages, and the `Onboarding` group is completed, the previously blocked event `Onboarding Completed` gets sent as well. 

 but if we only accept `QueueTaskGroup` at QueueRunner, we then need another subclass that lets us create `QueueTaskGroup` subclasses. 
 
Thoughts?

sdk/src/sharedTest/java/io/customer/sdk/queue/QueueStorageIntegrationTest.kt
```diff
@@ -46,52 +47,57 @@ class QueueStorageIntegrationTest : BaseTest() {
...
+        newlyCreatedInventoryItem.groupStart shouldBeEqualTo "identified_profile_148"
+        newlyCreatedInventoryItem.groupMember shouldBeEqualTo listOf("registered_push_token_ABC")
     }
 
     @Test
     fun givenCreateNewTask_expectInventoryUpdated() {
         queueStorage.getInventory().isEmpty().shouldBeTrue()
 
-        queueStorage.create(String.random, String.random)
+        queueStorage.create(String.random, String.random, null, null)
```
Shahroz16, line 78: A bit general question, related to the `blocking group` concept. 

Are we doing anything to avoid the cyclic dependency? both groups being blocked by each other?

sdk/src/sharedTest/java/io/customer/sdk/CustomerIOClientTest.kt
levibostian: Are you asking why this code is:
```kt
            verify(backgroundQueueMock).addTask(
                QueueTaskType.IdentifyProfile,
                IdentifyProfileQueueTaskData(newIdentifier, givenAttributes),
                groupStart = QueueTaskGroup.IdentifyProfile(newIdentifier),
                blockingGroups = null
            )
```
instead of
```kt
            verify(backgroundQueueMock).addTask(
                QueueTaskType.IdentifyProfile,
                IdentifyProfileQueueTaskData(newIdentifier, givenAttributes),
                groupStart = QueueTaskGroup.IdentifyProfile(newIdentifier)
            )
```
(removing `blockingGroups = null`) because `blockingGroups` is null by default?

docs/dev-notes/BACKGROUND_QUEUE.md
```diff
@@ -0,0 +1,3 @@
+# Background queue
+
+See [the iOS SDK doc](https://github.com/customerio/customerio-ios/blob/develop/docs/dev-notes/BACKGROUND-QUEUE.md) to maintain a single source of truth of this knowledge.
```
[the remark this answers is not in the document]
levibostian: The .md docs in the Android and iOS SDK are currently *not* hosted on github pages. 

Are you suggesting instead of having a URL that is hard-coded to a git branch like this is:
```
https://github.com/customerio/customerio-ios/blob/develop/docs/dev-notes/BACKGROUND-QUEUE.md
                                             ^^^^^^^^^^^^
```

We have a URL for a webpage that has a higher chance of not being a 404?

sdk/src/main/java/io/customer/sdk/CustomerIOClient.kt
levibostian: I am having a hard time understanding your question. What logic should not being in `CustomerIOClient` and how is that logic task specific? Being more specific about the code would be very helpful.

sdk/src/main/java/io/customer/sdk/queue/type/QueueTaskGroup.kt
levibostian: 🤔 I have never thought of using groups at runtime before. I have only thought about the specific use case it's being used in today for helping ensure that HTTP events will run successfully in the queue. 

With the real-world example that you gave, I believe that you are suggesting that *events* are sent to CIO in a specific order, correct? 

If you don't mind, I want to ask product to see if this is a use case that is valid for customers. I asked in our slack channel. 

Good question!

sdk/src/sharedTest/java/io/customer/sdk/queue/QueueStorageIntegrationTest.kt
levibostian: Ah! That would indeed be a big problem if that happened. 

With all of the apps that I have built using a background queue, and our specific use cases of our SDK, I have not encountered a scenario where cyclic dependencies happened. However, a bug could easily cause this to happen when we didn't want it to happen. 

On the SDK level, we are currently not doing anything to avoid this from happening. Off the top of my head, some ideas for how we could address this concern is:
* We could write a series of automated tests that check if cyclic dependencies are possible. 
* We create and maintain a flow chart graph (like the one below) for all groups that are in the SDK to help us visualize our dependencies to try and catch cyclic dependencies. 

CleanShot 2022-04-13 at 15 58 33@2x

* We could write some code to catch cyclic dependencies in the SDK. But that solution has some drawbacks. For one, that code could also have a bug in it. Also, that code exists as more of a band-aid to cover up the bug after it already exists. I think that if we have preventative measures in place during development to avoid introducing cyclic dependencies in the first place, we wouldn't need this logic. 

I agree this is something important to think about. Cyclic dependencies could cause problems that prevent a lot of data from being sent to CIO. Any ideas on how to prevent it?

sdk/src/main/java/io/customer/sdk/queue/type/QueueTaskGroup.kt
Shahroz16: Added a bit more detail on it in the thread, just a *note* it's not a blocker, so this suggestion can be ignored and can be taken on later.

sdk/src/sharedTest/java/io/customer/sdk/CustomerIOClientTest.kt
Shahroz16: oh never mind, I thought it's not `null` by default, but just checked the code, it indeed is `null by default`. You just added it out of practice.

sdk/src/sharedTest/java/io/customer/sdk/queue/QueueStorageIntegrationTest.kt
Shahroz16: On the top of my head, we could use `LinkedList` instead of `List` and find a loop in it. 
This can be done while adding a task to the queue.

docs/dev-notes/BACKGROUND_QUEUE.md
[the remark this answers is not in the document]
levibostian: > also, redirecting to ios repo doesn't give the best experience. As it might suggest that information might be spread among different repos.

I do agree. I felt that this solution of linking to the iOS docs was a better idea then copying and pasting the identical document into this repo. 

I did think about making a new github repo just for storing SDK docs but felt that was overkill. I did think of an internal doc in notion but I felt having the doc live in the repo (1) helps us keep it more up-to-date and (2) it's a document that could be helpful for customers so it's good to live here. 

The solution I have here in the PR is what I found to be the best idea that I could come up with. If you have a better idea, I would love to hear it.

sdk/src/main/java/io/customer/sdk/queue/type/QueueTaskGroup.kt
levibostian: Let's continue the conversation outside GitHub. I'll close this.

sdk/src/sharedTest/java/io/customer/sdk/queue/QueueStorageIntegrationTest.kt
levibostian: Could you elaborate on "find a loop in it"? I can see how a `LinkedList` would be a good data structure for this scenario but I know the data structure by itself would not solve the problem. 

Want to create a ticket and you and I share ideas together there? I find this idea to be a good improvement to the background queue. I also do not believe this to be a blocker for this PR so a new ticket will help us keep the conversation going.

sdk/src/main/java/io/customer/sdk/CustomerIOClient.kt
levibostian: Besides file size, do you see any problems with the `CustomerIOClient` as it is in this PR? 

For me, when tests are difficult to write, that's an indicator that a function or class is getting difficult to maintain. I only needed to write [5 test functions](https://github.com/customerio/customerio-android/blob/32096d119fd63dc17ca9d76ea5620452fe7d8d3f/sdk/src/sharedTest/java/io/customer/sdk/CustomerIOClientTest.kt#L59-L181=) for `identify` and I did not feel they were difficult to create. 

I like to avoid putting code into separate files to simply keep file sizes down. I do agree that large files are difficult to read and maintain, but I also find that lots of small files are also difficult to read and maintain. 

What do you think about breaking up `CustomerIOClient` into separate files based on their purpose? Some functions have to deal with profiles (identify), some have to do with tracking (screens, events, metrics), etc. Should we have a `IdentifiedProfileRepository`, `TrackingRepository` to break this file up?

docs/dev-notes/BACKGROUND_QUEUE.md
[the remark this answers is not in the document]
Shahroz16: >I did think about making a new github repo just for storing SDK docs but felt that was overkill. I did think of an internal doc in notion but I felt having the doc live in the repo (1) helps us keep it more up-to-date and (2) it's a document that could be helpful for customers so it's good to live here.

Maybe not in the PR but we might have to move to one of these approaches, we already have a lot of information regarding the development environment, CI/CD pipelines, BQ and this list is going to grow. So either a separate repo or a notion would do. 

This can be closed and maybe a ticket can be created out of this.

sdk/src/main/java/io/customer/sdk/CustomerIOClient.kt
Shahroz16: File size and somewhat a `chain-of-responsibility`, right now all of the logic is present in `CustomerIOClient` hence its responsible for storing/fetching identifiers, checking the blocking groups logic, etc. 

We should have a breakdown of this responsibility at some level, and I feel like repositories like you also mentioned could be a good first candidate for it.

sdk/src/sharedTest/java/io/customer/sdk/queue/QueueStorageIntegrationTest.kt
Shahroz16: Yup, not a blocker, I'll create a ticket and propose a `LinkedList` based solution, which upon addition of any task would traverse itself to find a possible `loop/circular depdency`.  👍 

This conversation can be closed.

Shahroz16: Created a ticker for it, In there I'll propose how I believe `LinkedList` can be utilized, that every time a new task is added, it's traversed to find a possible `loop/circular dependency` in it. 

This conversation can be closed as well.

sdk/src/main/java/io/customer/sdk/CustomerIOClient.kt
levibostian: Since we want to get this feature released in 1-2 days, I [created a ticket](https://github.com/customerio/issues/issues/7160) to break up the `CustomerIOClient` file after this ships.

PR #85, opened 2022-04-14
feat: add http pauses prevent hitting API too hard

levibostian:
Pauses HTTP request in certain scenarios. [Learn more here](https://github.com/customerio/customerio-android/blob/1ce3e0eda49b43d69e30c8c2f401bc3f446e072b/sdk/src/main/java/io/customer/sdk/api/TrackingHttpClient.kt#L92-L96=)

- [x] Wait until the pull request has been reviewed *and approved* by a teammate
- [x] After pull request is approved, and you determine it's ready **add the label "Ready to merge"** to the pull request and it will automatically be merged.


sdk/src/main/java/io/customer/sdk/api/HttpRequestRunner.kt
```diff
@@ -0,0 +1,117 @@
...
+                    }
+                } else {
+                    pauseHttpRequests()
+                    prepareForNextRequest() // after retry policy is finished, reset to prepare for the next HTTP request
+
+                    Result.failure(CustomerIOError.ServerDown())
+                }
+            }
+            401 -> {
+                pauseHttpRequests()
```
Shahroz16, line 69: Why would we `pause` the requests if it's `401` unauthenticated user? 

My use case is if a client has entered wrong workspace credentials, initially they wouldn't get to know about it until we have hit `X seconds` after that we will pause the `http` requests for another `Y` seconds. 

Its going to take atleast X+Y seconds to find out if the requests are going successfully now?

sdk/src/main/java/io/customer/sdk/api/HttpRequestRunner.kt
```diff
@@ -0,0 +1,117 @@
...
+        retryPolicy.reset() // in case retry policy was used, reset it so the next request can use it.
+        // do not edit the HTTP pausing. Let that get modified somewhere else and get reset by timing out.
+    }
+
+    // In certain scenarios, it makes sense for us to pause making any HTTP requests to the
+    // Customer.io API. Because HTTP requests are performed by the background queue, there is
+    // a chance that the background queue could make a lot of HTTP requests in
+    // a short amount of time from lots of devices. This makes a performance impact on our API.
+    // By pausing HTTP requests, we mitigate the chance of customer devices causing harm to our API.
+    internal fun pauseHttpRequests() {
```
Shahroz16, line 109: I am having a bit of a hard time understanding why would we need to `pause` making any HTTP requests?

Shouldn't be either rate-limiting from the start so this issue never arise? if we throttle our requests when making from BQ, we wont get to the point that it starts impacting our APIs performance? Thoughts? 

Maybe we can take this to the squad as well for their input on this. ?

sdk/src/main/java/io/customer/sdk/util/SimpleTimer.kt
```diff
@@ -19,29 +26,58 @@ class AndroidSimpleTimer(
...
     private val instanceIdentifier = String.random
 
+    override suspend fun <T> scheduleAndCancelPreviousSuspend(
+        seconds: Seconds,
+        block: suspend () -> T
+    ): T {
+        return coroutineScope {
+            // using deferred as it allows us to cancel our timer.
+            val newTimer: Deferred<T> = synchronized(this@AndroidSimpleTimer) {
+                timerAlreadyScheduled = true
```
Shahroz16, line 40: This is a bit confusing here, `deferred` coroutines can run in parallel and all of them will be using  `timerAlreadyScheduled` value. 

On top of that, we are using a `synchronized` block as well. 

So if we are not planning on running the coroutines in parallel, shouldn't we just run them synchronously?

levibostian: I removed the `suspend` function from the `SimpleTimer` so this code no longer exists. I replaced the logic with a simple `delay()` in the coroutine that performs HTTP requests.

sdk/src/main/java/io/customer/sdk/api/HttpRequestRunner.kt
levibostian: This has been discussed outside of GitHub. We decided to move forward with this implementation but create a ticket to see if we can improve it in the future.

sdk/src/main/java/io/customer/sdk/api/HttpRequestRunner.kt
levibostian: This has been discussed outside of GitHub. We decided to move forward with this implementation but create a ticket to see if we can improve it in the future.

sdk/src/main/java/io/customer/sdk/CustomerIO.kt
```diff
@@ -141,7 +141,8 @@ class CustomerIO internal constructor(
                 urlHandler = urlHandler,
                 autoTrackScreenViews = shouldAutoRecordScreenViews,
                 backgroundQueueMinNumberOfTasks = 10,
-                backgroundQueueSecondsDelay = 30.0
+                // TODO put this value back to 30
+                backgroundQueueSecondsDelay = 10.0
```
Shahroz16, line 145: was this meant to be pushed? commenting for your visibility

Shahroz16 (approved):
Looks good, just added one comment for visibility

sdk/src/main/java/io/customer/sdk/CustomerIO.kt
levibostian: fixed. thanks for the reminder.

PR #100, opened 2022-05-04
feat: added support for custom track url

Shahroz16:

### DIFF
--- a/sdk/src/main/java/io/customer/sdk/data/model/Region.kt
+++ b/sdk/src/main/java/io/customer/sdk/data/model/Region.kt
@@ -11,5 +11,7 @@ sealed class Region(val code: String, val baseUrl: String) {
     // to find the correct hostname for what you're trying to do.
     object US : Region(code = "us", baseUrl = "https://track-sdk.customer.io/")
     object EU : Region(code = "eu", baseUrl = "https://track-sdk-eu.customer.io/")
+
+    // Utilise custom `baseUrl` provided and ignore `code`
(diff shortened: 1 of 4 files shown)


sdk/src/main/java/io/customer/sdk/data/model/Region.kt
```diff
@@ -11,4 +11,7 @@ sealed class Region(val code: String, val baseUrl: String) {
     // to find the correct hostname for what you're trying to do.
     object US : Region(code = "us", baseUrl = "https://track-sdk.customer.io/")
     object EU : Region(code = "eu", baseUrl = "https://track-sdk-eu.customer.io/")
+
+    // Utilise custom `baseUrl` provided and ignore `code`
+    class Custom(baseUrl: String) : Region(code = "", baseUrl = baseUrl)
```
levibostian, line 16: I am hesitant on an implementation like this because

1. The `Region` is a data type. `Region`'s job is to keep track of all of the regions that are compatible with the CustomerIO service. CustomerIO has a finite number of regions that we provide our service to. By adding a *Custom* region, that is altering the data type to have a different meaning. 
2. If we ever decide that we need to use the `Region.code` for any logic, having this empty string will be an edge case to now handle. Instead of having a finite set of edge cases to handle (because there are is a finite set of Regions that CustomerIO's service is offered), this is making an infinite set of edge cases. 

Although I do like this implementation from a simplicity standpoint, I do not currently like the idea of us modifying the use of `Region` as a data type.

PR #101, opened 2022-05-04
feat!: register device attributes when set

levibostian:

### DIFF
--- a/sdk/src/main/java/io/customer/sdk/CustomerIO.kt
+++ b/sdk/src/main/java/io/customer/sdk/CustomerIO.kt
@@ -302,7 +302,12 @@ class CustomerIO internal constructor(
      * Use to provide additional and custom device attributes
      * apart from the ones the SDK is programmed to send to customer workspace.
      */
-    val deviceAttributes: MutableMap<String, Any> = mutableMapOf()
+    var deviceAttributes: CustomAttributes = emptyMap()
+        set(value) {
+            field = value
+
+            api.addCustomDeviceAttributes(value)
+        }
 
     private fun recordScreenViews(activity: Activity, attributes: CustomAttributes) {
         val packageManager = activity.packageManager
--- a/sdk/src/sharedTest/java/io/customer/sdk/CustomerIOClientTest.kt
+++ b/sdk/src/sharedTest/java/io/customer/sdk/CustomerIOClientTest.kt
@@ -253,6 +253,44 @@ class CustomerIOClientTest : BaseTest() {
         prefRepository.getDeviceToken() shouldBeEqualTo givenDeviceToken
     }
 
+    // addCustomDeviceAttributes
+
+    @Test
+    fun addCustomDeviceAttributes_givenNoPushToken_expectDoNotRegisterPushToken() {
+        val givenAttributes = mapOf(String.random to String.random)
+
+        customerIOClient.addCustomDeviceAttributes(givenAttributes)
+
+        // no token registered
+        verifyNoInteractions(backgroundQueueMock)
+    }
+
+    @Test
+    fun addCustomDeviceAttributes_givenExistingPushToken_expectRegisterPushTokenAndAttributes() {
+        val givenAttributes = mapOf(String.random to String.random)
+        val givenDeviceToken = String.random
+        val givenIdentifier = String.random
+        prefRepository.saveDeviceToken(givenDeviceToken)
+        prefRepository.saveIdentifier(givenIdentifier)
+
+        customerIOClient.addCustomDeviceAttributes(givenAttributes)
+
+        // a token got registered
+        verify(backgroundQueueMock).addTask(
+            QueueTaskType.RegisterDeviceToken,
+            RegisterPushNotificationQueueTaskData(
+                givenIdentifier,
+                Device(
+                    token = givenDeviceToken,
+                    lastUsed = dateUtilStub.givenDate,
+                    attributes = deviceStore.buildDeviceAttributes() + givenAttributes
+                )
+            ),
+            groupStart = QueueTaskGroup.RegisterPushToken(givenDeviceToken),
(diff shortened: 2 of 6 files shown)


sdk/src/sharedTest/java/io/customer/sdk/CustomerIOClientTest.kt
```diff
@@ -253,6 +253,44 @@ class CustomerIOClientTest : BaseTest() {
...
+                givenIdentifier,
+                Device(
+                    token = givenDeviceToken,
+                    lastUsed = dateUtilStub.givenDate,
+                    attributes = deviceStore.buildDeviceAttributes() + givenAttributes
+                )
+            ),
+            groupStart = QueueTaskGroup.RegisterPushToken(givenDeviceToken),
+            blockingGroups = listOf(QueueTaskGroup.IdentifyProfile(givenIdentifier))
+        )
```
levibostian, line 291: I do not like the code used here to test if `identify()` was called. This will be addressed with [a refactor splitting up the `CustomerIOClient` class](https://github.com/customerio/issues/issues/7160). These lines of code could then be much easier to understand and easier to maintain. I imagine the code will be changed to something similar to this:

```kotlin
verify(deviceRepository).registerDeviceToken(givenDeviceToken)
```

sdk/src/sharedTest/java/io/customer/sdk/CustomerIOClientTest.kt
```diff
@@ -253,6 +253,44 @@ class CustomerIOClientTest : BaseTest() {
...
+    // addCustomDeviceAttributes
+
+    @Test
+    fun addCustomDeviceAttributes_givenNoPushToken_expectDoNotRegisterPushToken() {
+        val givenAttributes = mapOf(String.random to String.random)
+
+        customerIOClient.addCustomDeviceAttributes(givenAttributes)
+
+        // no token registered
+        verifyNoInteractions(backgroundQueueMock)
```
levibostian, line 265: I do not like the code used here to test if `registerDeviceToken()` was called. This will be addressed with [a refactor splitting up the `CustomerIOClient` class](https://github.com/customerio/issues/issues/7160). These lines of code could then be much easier to understand and easier to maintain. I imagine the code will be changed to something similar to this:

```kotlin
verify(deviceRepository, never()).registerDeviceToken(givenDeviceToken)
```

sdk/src/main/java/io/customer/sdk/CustomerIO.kt
```diff
@@ -302,7 +302,12 @@ class CustomerIO internal constructor(
      * Use to provide additional and custom device attributes
      * apart from the ones the SDK is programmed to send to customer workspace.
      */
-    val deviceAttributes: MutableMap<String, Any> = mutableMapOf()
+    var deviceAttributes: CustomAttributes = emptyMap()
```
Shahroz16, line 305: >This PR changes the API to syntax that is more familiar to Kotlin developers

This isn't just a syntax change, this changes the behavior of how `CustomAttributes` will be set. 

I intentionally kept it as a `MutableMap`, because I didn't want users to be restricted to provide all the information at once or come up with a mechanism to store all information in one place and then send it to us. The proposed change removes that capability. 

```
// On Sign-up page 1,
CustomerIO.instance().deviceAttributes.putAll(mapOf("language" to "english"))
// On Sign-up page 2           
CustomerIO.instance().deviceAttributes.putAll(mapOf("ads" to false))
           
// Now when you send a device token all of these stored attributes can go along with it.    

Whereas,  in the proposed change, user will have to do it in one shot
CustomerIO.instance().deviceAttributes = mapOf()      
```

levibostian: The use case mentioned in the comment above is still covered by this PR. 

```
// On Sign-up page 1,
CustomerIO.instance().deviceAttributes = mapOf("language" to "english")
// Device attributes "language" get registered to the device on the API

// On Sign-up page 2           
CustomerIO.instance().deviceAttributes = mapOf("ads" to false)
// Device attributes "ads" get registered to the device on the API
```

I don't see a problem with the above use case? Customers still have the convenience of not having to add all of the device attributes in one shot. Plus they get the added benefits outlined in the description of this PR in the "Why this change is being proposed" section.

levibostian: You are correct. The use case of custom device attributes being ignored in the SDK when a device token has not been registered [has been documented and the squad has had conversations together about](https://github.com/customerio/issues/issues/6982). 

We as a squad decided that we want to ignore custom device attributes when a push token is not yet registered. We will make a note of this in the SDK docs. Then, we will collect customer feedback and make changes to the SDK behavior in the future if we choose to do so.

sdk/src/sharedTest/java/io/customer/sdk/CustomerIOClientTest.kt
levibostian: implemented https://github.com/customerio/customerio-android/pull/103

sdk/src/sharedTest/java/io/customer/sdk/CustomerIOClientTest.kt
levibostian: implemented https://github.com/customerio/customerio-android/pull/103

sdk/src/main/java/io/customer/sdk/CustomerIO.kt
Shahroz16: Alright, I think I missed this conversation as I must not be working that day. But by just making `deviceAttributes` a `MutableMap` and then emptying it when the task is added to BQ can easily solve the mentioned use-case. 

>The use case of custom device attributes being ignored in the SDK when a device token has not been registered

I'll unblock this PR but I do believe we should move to solve this as it's a very quick fix and doesn't `Why this change is being proposed` part as well.

Shahroz16 (approved):

PR #106, opened 2022-05-23
feat: inapp messaging module

Shahroz16:
## *Test for hooks to follow in next PR*


messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
```diff
@@ -0,0 +1,74 @@
...
+import io.customer.sdk.hooks.ModuleHook
+import io.customer.sdk.hooks.ModuleHookProvider
+import io.customer.sdk.repository.TrackRepository
+
+class ModuleMessagingInApp internal constructor(
+    private val overrideDiGraph: CustomerIOComponent?,
+    private val organizationId: String
+) : CustomerIOModule {
+
+    constructor(organizationId: String) : this(
```
levibostian, line 19:
```suggestion
    constructor(gistOrganizationId: String) : this(
```

This might be a question for product - because CIO does not use the term "organization id" anywhere in the product, I think that just saying "organization id" will confuse customers where they don't understand what we are asking of them. Instead, if we label it as `gistOrganizationId` customers should understand better what we are asking.

messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
```diff
@@ -0,0 +1,74 @@
...
+                event = MetricEvent.opened
+            )
+        }
+    }
+
+    private fun setupHooks() {
+        hooksManager.add(object : ModuleHookProvider() {
+            override fun profileIdentifiedHook(hook: ModuleHook.ProfileIdentifiedHook) {
+                gistProvider.setUserToken(hook.identifier)
+            }
```
levibostian, line 56: 🤔 will this cause memory issues in the future? 

`object : ModuleHookProvider()` creates a instance that is stored in `HooksManager` > `gistProvider` inside of the `ModuleHookProvider` is holding a strong reference to the `gistProvider` instance inside of the `ModuleMessagingInApp` class. This means that the `ModuleHookProvider` instance holds a strong reference to `GistProvider` and to `ModuleMessagingInApp`. 

I was expecting to see less strong references to instances. 

With the current implementation of `ModuleMessagingInApp`, this might not be a problem today. However, having these strong references here might cause issues in the future if we are not careful. 

Would it work if we had `ModuleHookProvider` create *new* instances of `GistProvider` on each function call? 

```kotlin
hooksManager.add(object : ModuleHookProvider() {
  private val diGraph: CustomerIOComponent 
    get() = CustomerIO.instance().diGraph

  val profileIdentifiedHook: ModuleHook.ProfileIdentifiedHook {
    return diGraph.gistProvider
  }
```

I like how an implementation like this doesn't have any strong references to anything. It's lightweight and less error-prone.

messagingpush/src/main/java/io/customer/messagingpush/provider/FCMTokenProvider.kt
```diff
@@ -18,17 +18,22 @@ class FCMTokenProviderImpl(private val logger: Logger) : FCMTokenProvider {
     override fun getCurrentToken(onComplete: (String?) -> Unit) {
         logger.debug("getting current FCM device token...")
 
-        FirebaseMessaging.getInstance().token.addOnCompleteListener { task ->
-            if (task.isSuccessful) {
-                val existingDeviceToken = task.result
-                logger.debug("got current FCM token: $existingDeviceToken")
+        try {
```
levibostian, line 21: Why is a `try/catch` needed?

sdk/src/main/java/io/customer/sdk/di/CustomerIOComponent.kt
```diff
@@ -59,27 +41,48 @@ class CustomerIOComponent(
...
+            queueRunner,
+            queueStorage,
+            logger,
+            queueQueryRunner
+        )
 
     val logger: Logger
         get() = override() ?: LogcatLogger(sdkConfig)
 
+    val hooksManager: HooksManager by lazy { override() ?: CioHooksManager() }
```
levibostian, line 74: For consistency, instead of `lazy`, should we use `getInstanceOrCreate()` like we do with the `Queue` class?

sdk/src/main/java/io/customer/sdk/hooks/HooksManager.kt
```diff
@@ -0,0 +1,29 @@
+package io.customer.sdk.hooks
+
+interface HooksManager {
+    fun add(subscriber: ModuleHookProvider)
+    fun onHookUpdate(hook: ModuleHook)
+}
+
+internal class CioHooksManager : HooksManager {
+
+    val list: MutableList<ModuleHookProvider> = mutableListOf()
```
levibostian, line 10: Instead of a List, I recommend using a Map<String, ModuleHookProvider> to avoid having 2+ hooks per module registered. We don't want to send 2+ events to the same module in the SDK.

sdk/src/main/java/io/customer/sdk/queue/QueueRunner.kt
```diff
@@ -40,13 +47,27 @@ internal class QueueRunnerImpl(
...
         val taskData: IdentifyProfileQueueTaskData = jsonAdapter.fromJson(task.data)
 
-        return cioHttpClient.identifyProfile(taskData.identifier, taskData.attributes)
+        return cioHttpClient.identifyProfile(taskData.identifier, taskData.attributes).apply {
+            if (this.isSuccess) {
+                hooksManager.onHookUpdate(
+                    hook = ModuleHook.ProfileIdentifiedHook(taskData.identifier)
+                )
+            }
+        }
```
levibostian, line 56: Is there a reason why module hook events are being called here in the background queue execution instead of when `CustomerIO.identify()` is called?

levibostian (changes requested):
Couple questions about the implementation of module hooks before I feel comfortable approving PR. Looks great otherwise. 

I see quite a bit of code formatting being done in the PR. Made a lot of noise in the PR for review. I am assuming that it's Android Studio making these code format changes for you? If so, it would probably be a good idea for everyone in the Android dev team to share the same code format rules for Android Studio to prevent generating this noise every time that someone modifies code.

messagingpush/src/main/java/io/customer/messagingpush/provider/FCMTokenProvider.kt
Shahroz16: Added the context here, 
https://github.com/customerio/issues/issues/7425

but i believe we might encounter some other issue as well so its safer to wrap it under the `try/catch` block.

messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
Shahroz16: While the concern is valid, I thought about the implementation where we are creating new instances as well, but I, later on, reached the conclusion that it shouldn't cause memory leakages. 

Since the SDK is `singleton` and the hooksManager is `singleton` too and all of its scopes is tied to the Application, it would all remain in memory anyways. 

```
modules: MutableMap<String, CustomerIOModule> = mutableMapOf()
```
inside `CustomerIO` is also going to be holding the same strong references for `ModuleMessagingInApp` and `ModuleMessagingPush`.

>this might not be a problem today However, having these strong references here might cause issues in the future if we are not careful.

I agree we will need to be careful on how we add on top of it, but I think with SDK being singleton we do have a certain blank of safety here and when we later work on, we can even look into keeping `WeakReference` both in `CustomerIO` class and `HooksManager`

sdk/src/main/java/io/customer/sdk/queue/QueueRunner.kt
Shahroz16: Good point, two clarifications needed on it, 

- Where do we notify hook, I agree, I would rather have it repositories rather than Queue, it would make more sense and be more consistant. 
-  When do we notify? I am currently going for when the queue has run the task to be consistent, otherwise, behaviour can be a bit different, because imagine on fly we dont see a person being notified for the next 30 secs, but they are getting in-app messages on phone since we already set the `user-token` on Gist.

messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
levibostian: I asked product about this to see if they have a preference. This is an easy fix to change later.

messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
levibostian: I do understand where you're coming from. I appreciate the thought you put into the implementation as you were writing it. 

It looks like you and I both agree that we do not see a problem with the code as it is today and we both agree that we need to be careful on how we add on top of this code in the future. 

With that being said, I want to ask:
1. Can we come up with a solution right now that removes the error prone code and so we no longer *need to be careful*? Is this new solution better then the current implementation in this PR? Can the new solution be implemented now before merging if it doesn't take too much time? 
2. If we can't do ^^^, create a ticket for this and work on it after we get the SDK deployed. This SDK is important to get out the door but I do think that the code in this implementation is important to improve if we come up with something better. Along with creating a ticket to improve the code, I think it would also be a good idea to add a comment to the code saying that it's code to be careful with to warn future developers touching that code until we come up with something better. 

I like code that we don't need to be careful using. We are only human and if we have code we know we need to be careful with, it's easy to make mistakes with it. I believe it's important to document careful code and try to improve it as a responsible way to address careful code.

sdk/src/main/java/io/customer/sdk/queue/QueueRunner.kt
levibostian: From your reply ^^^, it seems to me that you are planning on moving the notify hooks code from here in the `QueueRunner` to the repositories? I can't tell from your comment what your action item is. 

I agree that the repositories would be a better place for this code.

levibostian:
@Shahroz16, we will need to add [to the CI configuration file](https://github.com/customerio/customerio-android/blob/a6c87c22d2b1b758147cefa4cd40b09800184967/.github/workflows/test.yml#L10=) the new in-app Gradle module. I see that the tests are not being executed on the CI server yet.

messagingpush/src/main/java/io/customer/messagingpush/provider/FCMTokenProvider.kt
Shahroz16: Yup, definitely a wrapper around `Firebase` may be in a similar way we have around Gist, to encapsulate its implementation and other possible crashes since FCM is something we don't have complete hold of, its always to make the whole wrapped under `try/catch`

messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
Shahroz16: Thank you for bringing this up, I agree this needs a thorough discussion so it doesn't remain a repetitive topic. 

So, I'll try to target each comment. 

>Can we come up with a solution right now that removes the error-prone code so we no longer need to be careful? 

What is error-prone code over here, it's the fact that our `Modules`, i.e,  ModuleMessagingInApp and ModuleMessagingPush are going to be referenced by the `tracking` module, either in the form of `modules` in `CustomerIO` class or by `HooksManager`. 

>we do not see a problem with the code as it is today

We both agree, for now, it's not much of a problem because we have managed to keep `Modules` lightweight, and because of the fact SDK is a singleton, so only one instance is created and it's scoped by the Application lifecycle and even the modules are sharing the same `digraph`. 


>Can we come up with a solution right now that removes the error-prone code so we no longer need to be careful?

From what I can think of right now, whatever we have right now is the best way possible and it will stay that way until we don't bloat our `Modules`. For example, for all eventing in the modules, we just utilize the same `repositories`. If we were to start creating new classes and references that will then go on to be referenced by the tracking module, that's where the problems will start arising. 

If you have other ideas to improve, please share them here, keeping in mind the current implementation of iOS and Android is different in a sense, iOS does create and hold multiple instances of SDK, whereas Android doesn't. 

> think it would also be a good idea to add a comment to the code saying that it's code to be careful with to warn future developers touching that code until we come up with something better.

I would suggest, mentioning it on Modules to keep them light and to rely on repositories and mutually shared digraph 
for common code. Let me know, if this much sounds good or what would you add.

Shahroz16:
> @Shahroz16, we will need to add [to the CI configuration file](https://github.com/customerio/customerio-android/blob/a6c87c22d2b1b758147cefa4cd40b09800184967/.github/workflows/test.yml#L10=) the new in-app Gradle module. I see that the tests are not being executed on the CI server yet.

thanks for raising this, updated

messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
levibostian: Product decided to keep the naming as it is in this PR.

sdk/src/main/java/io/customer/sdk/data/request/DeliveryEvent.kt
```diff
@@ -0,0 +1,23 @@
...
+
+@JsonClass(generateAdapter = false)
+internal enum class DeliveryType {
+    in_app
+}
+
+@JsonClass(generateAdapter = true)
+internal data class DeliveryPayload(
+    @field:Json(name = "delivery_id") val deliveryID: String,
+    val event: MetricEvent,
```
levibostian, line 15: For iOS, I had to create a new `Metric` object specifically for in-app because in-app has `clicked` and push doesn't.

messaginginapp/src/main/java/io/customer/messaginginapp/provider/GistInAppMessagesProvider.kt
```diff
@@ -0,0 +1,61 @@
...
+    override fun clearUserToken() {
+        GistSdk.clearUserToken()
+    }
+
+    override fun subscribeToEvents(onMessageShown: (String) -> Unit) {
+        GistSdk.addListener(object : GistListener {
+            override fun embedMessage(message: Message, elementId: String) {
+            }
+
+            override fun onAction(message: Message, currentRoute: String, action: String) {
```
levibostian, line 46: I think this is where we track the `clicked` event. Filter out `action == "gist://close"` but track the rest as a click.

messaginginapp/src/main/java/io/customer/messaginginapp/provider/GistInAppMessagesProvider.kt
```diff
@@ -0,0 +1,61 @@
...
+            }
+
+            override fun onError(message: Message) {
```
levibostian, line 49: Log to the `Logger` an error. In iOS I did something like this:

```
logger.error("in-app message error occurred $message")
```

Since there isn't an `Throwable` parameter, we don't know what the error was. This was the most detailed error message I could come up with.

sdk/src/main/java/io/customer/sdk/hooks/HooksManager.kt
```diff
@@ -0,0 +1,37 @@
+package io.customer.sdk.hooks
+
+import io.customer.sdk.util.Singleton
+
+interface HooksManager {
+    fun add(modules: HookModules, subscriber: ModuleHookProvider)
+    fun onHookUpdate(hook: ModuleHook)
+}
+
+enum class HookModules {
```
levibostian, line 10:
```suggestion
enum class HookModule {
```

Since each enum case is a singular object, this name is less confusing. When I read `HookModules`, I expect to see something like a List and not a single enum case.

sdk/src/main/java/io/customer/sdk/repository/TrackRepository.kt
```diff
@@ -41,7 +46,13 @@ class TrackRepositoryImpl(
         }
 
         // if task doesn't successfully get added to the queue, it does not break the SDK's state. So, we can ignore the result of adding task to queue.
```
levibostian, line 48: This comment is no longer correct because the next line of code uses the queue status.

messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
levibostian: In my previous comment where I say, "error-prone code" and "code to be careful using", I am not referring to 1 specific line of code. I am referring to the whole implementation of module hooks in this pull request. Because we have identified that we need to be careful when using the current implementation of module hooks, I am calling the entire implementation error-prone. 

I am suggesting that we either change the implementation of module hooks in this pull request to code that we no longer need to be careful using it, or, we create a ticket to come up with a solution in the future that we no longer need to be careful using. 

When I see code that we need to be careful with, I don't like to merge it without doing something about it first. If we were to merge the code as it is today without improving the implementation or at a minimum create a ticket for it, we are setting our team up for creating bugs in the future. I am trying to avoid future bugs from happening now before merging the code in. 

---

Looking at the implementation of the module hooks again in this class:
```kotlin
hooksManager.add(object : ModuleHookProvider() {
            override fun profileIdentifiedHook(hook: ModuleHook.ProfileIdentifiedHook) {
                gistProvider.setUserToken(hook.identifier)
            }
```

Besides talking about memory resources, I don't find this implementation scalable. At this time, each of the `ModuleHookProvider` functions implemented only call 1 dependency: `gistProvider`. But in the future, this could increase to many more dependencies making this file more complex. 

I would be OK if we move `object : ModuleHookProvider()` out of this class and into it's own class. Something like:
```kotlin
class ModuleInAppHookProvider: ModuleHookProvider() {

  private var diGraph: CustomerIOComponent {
    CustomerIO.instance().diGraph
  }

  private var gistProvider: GistProvider {
    diGraph.gistProvider
  }

  override fun profileIdentifiedHook(hook: ModuleHook.ProfileIdentifiedHook) {
                gistProvider.setUserToken(hook.identifier)
            }
```

I like how an implementation like this is simple but scales better then having everything in 1 file. Also, it removes the parts of the code that we keep saying "we need to be careful using" as we are getting dependencies directly out of the di graph here and keeping strong references to code to a minimum.

messagingpush/src/main/java/io/customer/messagingpush/provider/FCMTokenProvider.kt
levibostian: I added a comment to the ticket.

sdk/src/main/java/io/customer/sdk/queue/QueueRunner.kt
levibostian: Looks like the change was made. Looks good!

messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
Shahroz16: >I am referring to the whole implementation of module hooks in this pull request. Because we have identified that we need to be careful when using the current implementation of module hooks, I am calling the entire implementation error-prone.

Even if we remove the `implementation of module hooks`, we will still need to be careful. Because in the whole `module implementation` initialization is done for `FCM token`, the `CustomerIO` class holds a strong reference of each `module` class through it. Now that would become error-prone as well?. So in any case, we need always make sure our modules are lighter. 

So regardless of any implementation, whenever multi-module need to communicate with each other, the base module would need to hold a reference. Having said that, as mentioned earlier, since our SDK is a singleton, we have the luxury of being in memory and all references are pointed to the same instance. 

> Besides talking about memory resources, I don't find this implementation scalable. At this time, each of the ModuleHookProvider functions implemented only call 1 dependency: gistProvider. But in the future, this could increase to many more dependencies making this file more complex.

Not sure what you mean by that, why would there be more dependencies in `ModuleMessagingInApp` other than gist or when we replace it with ours? and even if there are let's say even 2, not sure how that would make it complex, cause it would be just another line of code added here, and the provider is going to handle its own implementation in its own class. 

>I would be OK if we move object : ModuleHookProvider() out of this class and into it's own class.

Sure, I can do that, but as mentioned earlier, "we `still` need to be careful using" modules because its still being strongly referenced by `tracking` module.

messaginginapp/src/main/java/io/customer/messaginginapp/provider/GistInAppMessagesProvider.kt
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Wrapper around Gist SDK
+ */
+internal interface InAppMessagesProvider {
+    fun initProvider(application: Application, organizationId: String)
+    fun setUserToken(userToken: String)
+    fun setCurrentRoute(route: String)
+    fun clearUserToken()
+    fun subscribeToEvents(onMessageShown: (String) -> Unit)
```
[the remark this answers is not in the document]
Shahroz16: `Clicked` wasn't done by BE, so they are part of the next PR.

sdk/src/main/java/io/customer/sdk/data/request/DeliveryEvent.kt
Shahroz16: Same as the previous comment.

messaginginapp/src/main/java/io/customer/messaginginapp/provider/GistInAppMessagesProvider.kt
Shahroz16: yup, next PR, since this was decided i call after

sdk/src/main/java/io/customer/sdk/hooks/HooksManager.kt
Shahroz16: Agreed, but did it to keep naming consistent with [iOS SDK](https://github.com/customerio/customerio-ios/blob/develop/Sources/Common/hooks/HooksManager.swift#L10)

levibostian: Ah! Good catch. I am just now realizing that the iOS SDK does that. I don't think that it should and I just changed that on the iOS codebase to match `HookModule`.

messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
levibostian: > Sure, I can do that, but as mentioned earlier, "we still need to be careful using" modules because its still being strongly referenced by tracking module.

I agree that in regards to modules, we do still need to be careful. I think it would be a good idea to add a comment to [the module interface](https://github.com/customerio/customerio-android/blob/213165333b947eedb8506f3bf4c3020dfb88305f/sdk/src/main/java/io/customer/sdk/CustomerIOModule.kt#L3-L7=) reminding the reader to keep modules lightweight since they are strongly referenced by the `tracking` module. Maybe there is a better place for a comment but I think a reminder would be helpful to avoid issues in the future. 

---

I see the [commit](https://github.com/customerio/customerio-android/pull/106/commits/213165333b947eedb8506f3bf4c3020dfb88305f) made to this PR. Looks great! I'll approve of this PR as it looks like feedback as all been resolved.

levibostian (approved):

messaginginapp/build.gradle
```diff
@@ -43,7 +43,7 @@ dependencies {
     implementation Dependencies.coroutinesCore
     implementation Dependencies.coroutinesAndroid
     implementation Dependencies.retrofit
-    implementation Dependencies.gist
+    api Dependencies.gist
```
levibostian, line 46: Why was this changed from `implementation`?

Shahroz16:
Updated branch name and created a new PR

https://github.com/customerio/customerio-android/pull/112

Shahroz16:
Closing this

https://github.com/customerio/customerio-android/pull/106#issuecomment-1161330272

PR #107, opened 2022-05-30
refactor: add missing test and metric events

Shahroz16:


messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
```diff
@@ -34,19 +34,31 @@ class ModuleMessagingInApp internal constructor(
...
     private fun setupGistCallbacks() {
-        gistProvider.subscribeToEvents { deliveryID ->
-            trackRepository.trackInAppMetric(
-                deliveryID = deliveryID,
-                event = MetricEvent.opened
-            )
-        }
+        gistProvider.subscribeToEvents(
+            onMessageShown = { deliveryID ->
+                trackRepository.trackInAppMetric(
```
levibostian, line 48: I recommend adding a `logger.debug` statement for onMessageShown, onAction for easier debugging of the SDK. 

```suggestion
               logger.debug("in-app message shown $deliveryID")

                trackRepository.trackInAppMetric(
```

messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
```diff
@@ -34,19 +34,31 @@ class ModuleMessagingInApp internal constructor(
...
+                    event = MetricEvent.opened
+                )
+            },
+            onAction = { deliveryID: String, _: String, _: String ->
+                trackRepository.trackInAppMetric(
+                    deliveryID = deliveryID,
+                    event = MetricEvent.clicked
+                )
+            }, onError = { error ->
+                logger.error("in-app message error occurred $error")
```
levibostian, line 59: I wish that we could include some reference to the failed message in the error message. That way when there is an error, we know what message the error happened for. 

Do you think there is a way? 

```suggestion
                logger.error("in-app message error occurred $error. message: $message")
```

messaginginapp/src/sharedTest/java/io/customer/messaginginapp/ModuleMessagingInAppTest.kt
```diff
@@ -31,6 +34,13 @@ internal class ModuleMessagingInAppTest : BaseTest() {
 
         module.initialize()
 
+        // verify gist is initialized
         verify(gistInAppMessagesProvider).initProvider(any(), eq("test"))
+
+        // verify hook was added
+        verify(hooksManager).add(any(), any())
```
levibostian, line 41: I think it would be good to verify the correct module was registered with the hooks manager. 

```suggestion
        verify(hooksManager).add(equal(.messagingInApp), any())
```

messaginginapp/src/sharedTest/java/io/customer/messaginginapp/ModuleMessagingInAppTest.kt
```diff
@@ -31,6 +34,13 @@ internal class ModuleMessagingInAppTest : BaseTest() {
...
+
+        // verify events
+        verify(gistInAppMessagesProvider).subscribeToEvents(any(), any(), any())
```
levibostian, line 44: I am ok with this test function testing for multiple expected behaviors. 
1. The gist SDK is initialized
2. Module hooks initialized. 
3. Module listening for Gist events. 

I think it would be good to either (1) update the name of the test function to reflect all 3 things being tested against in the test function or (2) split up this test function into 3 separate test functions where each test function tests 1 of the 3 things above.

sdk/src/main/java/io/customer/sdk/hooks/HooksManager.kt
```diff
@@ -4,7 +4,7 @@ import io.customer.sdk.util.Singleton
 
 interface HooksManager {
     fun add(module: HookModule, subscriber: ModuleHookProvider)
-    fun onHookUpdate(hook: ModuleHook)
+    fun onHookUpdate(hook: ModuleHook?)
```
levibostian, line 7: Why did this get changed to optional?

sdk/src/sharedTest/java/io/customer/sdk/hook/HookManagerTest.kt
```diff
@@ -0,0 +1,62 @@
...
+                override fun screenTrackedHook(hook: ModuleHook.ScreenTrackedHook) {
+                    assertEquals(hook, screenTrackedHook)
+                    assertEquals(hook.screen, screen)
+                }
+            }
+        )
+
+        cioHooksManager.onHookUpdate(profileIdentifiedHook)
+        cioHooksManager.onHookUpdate(beforeProfileStoppedBeingIdentifiedHook)
+        cioHooksManager.onHookUpdate(screenTrackedHook)
```
levibostian, line 60: What is the reason for calling 3 separate hook functions in this test function?

sdk/src/sharedTest/java/io/customer/sdk/repository/ProfileRepositoryTest.kt
```diff
@@ -92,6 +94,29 @@ class ProfileRepositoryTest : BaseTest() {
         verifyNoMoreInteractions(backgroundQueueMock)
     }
 
+    @Test
+    fun identify_givenFirstTimeIdentify_givenNoDeviceTokenRegistered_expectProfileIdentifiedHookUpdateWithCorrectIdentifier() {
```
levibostian, line 98: What do you think about combining this test function with `identify_givenFirstTimeIdentify_givenDeviceTokenExists_expectIdentifyBackgroundQueue_expectDoNotDeleteToken_expectRegisterDeviceToken`? 

I am making the suggestion of combining the 2 test functions before the test environment is identical between the two. Both test functions are testing what happens when you call `identify()` when it's a first time identifying the profile *and* there is no device token registered. It would be more difficult to manage multiple tests that all have the same test environment. 

I see that you modified the test function `clearIdentify_givenNoPreviouslyIdentifiedProfile_expectIgnoreRequest_expectDontDeleteDeviceToken` to add `_expectDontUpdateHook`. This is what I am suggesting here, too. Add hooks logic to existing test function.

levibostian: Also, it would be good to add logic to the other `identify` test functions. If hooks should be called when `identify()` gets called in certain scenarios, we should also be testing that hooks *do not* get called in certain scenarios.

sdk/src/sharedTest/java/io/customer/sdk/repository/ProfileRepositoryTest.kt
```diff
@@ -92,6 +94,29 @@ class ProfileRepositoryTest : BaseTest() {
...
+            )
+        ).thenReturn(QueueModifyResult(true, QueueStatus(siteId, 1)))
+
+        repository.identify(newIdentifier, givenAttributes)
+
+        val argumentCaptor = nullableArgumentCaptor<ModuleHook.ProfileIdentifiedHook>()
+
+        verify(hooksManager, times(1)).onHookUpdate(argumentCaptor.capture())
+
+        assertEquals(newIdentifier, argumentCaptor.firstValue?.identifier)
```
levibostian, line 117: I have found that `.apply{}` works really nicely with argument captor. I find it's easier to write and read. I am OK with using it or not. 

```suggestion
        nullableArgumentCaptor<ModuleHook.ProfileIdentifiedHook>().apply {
            verify(hooksManager, times(1)).onHookUpdate(capture())
            assertEquals(newIdentifier, firstValue?.identifier)
        }
```

sdk/src/sharedTest/java/io/customer/sdk/repository/ProfileRepositoryTest.kt
```diff
@@ -189,11 +214,33 @@ class ProfileRepositoryTest : BaseTest() {
...
+
+        repository.clearIdentify()
+
+        prefRepository.getIdentifier().shouldBeNull()
+
+        val argumentCaptor =
+            nullableArgumentCaptor<ModuleHook.BeforeProfileStoppedBeingIdentified>()
+
+        verify(hooksManager, times(1)).onHookUpdate(
+            argumentCaptor.capture() ?: ModuleHook.BeforeProfileStoppedBeingIdentified(
```
levibostian, line 239: I don't understand the reason for `?: ModuleHook.BeforeProfileStoppedBeingIdentified(`? 

Are we sure it's a good idea to give a default value here? Shouldn't we test when `onHookUpdate()` gets called with `null`?

Shahroz16: updated it, we don't need it anymore, as `ModuleHook` doesn't need to be nullable.

messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
levibostian: Yes, a log statement would more then likely exist when adding a task to the background queue. 

Why I still suggest to add a log statement here is that the point of logs is to help walk you through the code step by step. What if there is a bug in the code between `onMessageShown()` and the background queue code? Having a log statement in `onMessageShown()` would help to help us pinpoint where the bug is. 

Having too many logs can indeed get noisy. From my experience, noisy is much better then not enough logs and not being able to debug the bug as quickly.

messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
levibostian: Ah, I see. I think the variable `error` was what got me confused. I thought that it was an object. Perhaps a variable name such as `errorMessage` or `errorDescription` would be better?

sdk/src/sharedTest/java/io/customer/sdk/hook/HookManagerTest.kt
levibostian: Ah, ok. The test function's job is to call each of the module hook event types and make sure that the callbacks are called for each type. 

I agree that this test function provides value. My main concern is the maintainability of it. If we were to add a new hook event type to the project, we would need to remember to go into this test function and add another `cioHooksManager.onHookUpdate()` call with the new hook event type. 

To avoid this issue, I wonder if there is a way to use an `enum` or something similar? Something that allows us to write code like this:

```kotlin
HookEventTypes.forEach { hookEventType -> 
  cioHooksManager.onHookUpdate(hookEventType)
} 
```

The code could be very different then what's above, but the idea is that we write a test function once and then anytime that a new module hook event type is added to the project, the test function will automatically test that new event type?

sdk/src/sharedTest/java/io/customer/sdk/repository/ProfileRepositoryTest.kt
levibostian: > the test case doesn't get bloated but we can separate them if that happens.

I agree the test function could continue to get bloated and hard to maintain. 

If it gets too bloated to handle, I suggest we consider refactoring the `identify` function. In my experience, a bloated test function is a red flag that your code's logic is too complex.

sdk/src/sharedTest/java/io/customer/sdk/hook/HookManagerTest.kt
Shahroz16: I don't think that would be quite possible, since hook types are expecting parameters, `identify` would require an `identifier` similarly, and others would require their respective parameters and those parameters are validated as well so, can't be sending `any`

> we would need to remember to go into this test function

So what I can do to resolve the remembering part, is make it an enum so, it forces you to handle the newly added hook type.

messaginginapp/src/main/java/io/customer/messaginginapp/di/DIGraphMessaginIApp.kt
```diff
@@ -1,8 +1,12 @@
 package io.customer.messaginginapp.di
 
+import io.customer.messaginginapp.provider.GistApiProvider
 import io.customer.messaginginapp.provider.GistInAppMessagesProvider
 import io.customer.messaginginapp.provider.InAppMessagesProvider
 import io.customer.sdk.di.CustomerIOComponent
 
+internal val CustomerIOComponent.gistApiProvider: GistApiProvider
+    get() = override() ?: GistApiProvider()
```
levibostian, line 9:
```suggestion
internal val CustomerIOComponent.gistApiProvider: GistApi
    get() = override() ?: GistApiProvider()
```

sdk/src/sharedTest/java/io/customer/sdk/hook/HookManagerTest.kt
levibostian: It looks like this has been implemented [here](https://github.com/customerio/customerio-android/blob/ccca81d528cbeeb458290debedaecc27c2b6eff3/sdk/src/sharedTest/java/io/customer/sdk/hook/HookManagerTest.kt#L74-L83=). 

Because the test function is a bit long and *could* be simplified by someone in the future, I suggest adding a comment to the test function explaining why the test function is written this way. A comment saying that the test function is designed to remind the developer to update the test function when a new hook event type is added. 

Besides that, this looks to be resolved.

levibostian (approved):

levibostian:
@Shahroz16 I have accepted the PR because all previous conversations have been resolved. There are 2 not resolved comments left but they are minor so I'll let you view these and do what you wish with them. 

* https://github.com/customerio/customerio-android/pull/107#discussion_r890495921=
* https://github.com/customerio/customerio-android/pull/107#discussion_r890504856=

PR #108, opened 2022-05-31
feat: delete expired background queue tasks

levibostian:


sdk/src/main/java/io/customer/sdk/di/DiGraph.kt
```diff
@@ -27,4 +27,35 @@ abstract class DiGraph {
      * Or you may not be able to successfully mock in test functions.
      */
     inline fun <reified DEP> override(): DEP? = overrides[DEP::class.java.simpleName] as? DEP
+
+    /**
+     * We prefer to have all of the SDK's singleton instances held in the dependency injection graph. This makes it easier for automated tests to be able to delete all
+     * singletons between each test function and prevent test flakiness.
+     */
+    val singletons: MutableMap<String, Any> = mutableMapOf()
```
levibostian, line 35: I changed the way that we create and store singles in the SDK. I had some failing automated tests and it was because singleton instances got created when the first unit test got executed and that same instance was used in all tests. I needed to delete all singleton instances after each test. 

So, to help automate this, I thought of this idea where all singletons are stored in the dependency graph so we can easily delete them all after each test. Fixed the issue I was having with tests not working.

sdk/src/main/java/io/customer/sdk/util/Seconds.kt
```diff
@@ -9,8 +9,11 @@ data class Seconds(
     val value: Double
 ) {
     companion object {
-        // used for automated tests. Value that is small to make tests run fast, but long enough to test code's logic.
-        fun testValue(): Seconds = Seconds(0.01)
+        fun fromDays(numDays: Int): Seconds {
```
Shahroz16, line 12: We could utilize `TimeUnit` methods for these? 
`TimeUnit.DAYS.toSeconds()` and similarly 
`TimeUnit.SECONDS.toMillis()` for `toMilliseconds`

sdk/src/main/java/io/customer/sdk/di/DiGraph.kt
Shahroz16: My only concern here would be what we have been discussing before, storing references in the `singletons` map. So even if that instance isn't being used it would still be referenced by `Digraph`?

levibostian: If in instance is saved in `Digraph.singletons` is not being used by another class then, yes, it would be referenced by `Digraph`. I currently can't think of any problems that could occur with `Digraph` containing all of these references. Can you think of any problems?

sdk/src/main/java/io/customer/sdk/util/Seconds.kt
levibostian: Great suggestion. I forget that `TimeUnit` exists. 

I made the change:
```kotlin
data class Seconds(
    val value: Double
) {
    companion object {
        fun fromDays(numDays: Int): Seconds {
            return Seconds(TimeUnit.DAYS.toSeconds(numDays.toLong()).toDouble())
        }
    }

    val toMilliseconds: Milliseconds
        get() = Milliseconds(TimeUnit.SECONDS.toMillis(value.toLong()))

    override fun toString(): String = "$value seconds"
}
```

But the unit tests all failed. I think it's because we are converting `Double` (for seconds) into `Long` for `TimeUnit.ZZZ.toZZZ()` functions as they all require `Long`. `Double.toLong()` will remove the fractional part from `Double` so we are losing accuracy. 

So, we could need to manually convert `Double` to `Long` (probably by manually converting seconds into milliseconds) and then use `TimeUnit`. Which, at that point, we might as well skip using `TimeUnit`.

sdk/src/main/java/io/customer/sdk/queue/QueueStorage.kt
```diff
@@ -119,6 +126,44 @@ class QueueStorageImpl internal constructor(
         }
     }
 
+    override fun deleteExpired(): List<QueueTaskMetadata> {
+        // important to lock the queue storage until entire process is complete to avoid race conditions with querying and deleting tasks.
+        synchronized(this) {
```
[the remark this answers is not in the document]
levibostian: Great suggestion. I forget to use the annotation. For this class, the annotation would work well. I switched over to using it.

Shahroz16 (approved):

sdk/src/main/java/io/customer/sdk/di/DiGraph.kt
Shahroz16: I think it should be okay for now.

PR #111, opened 2022-06-13
feat: save sdk logs to file

levibostian:
Closes: https://github.com/customerio/issues/issues/6620
Closes: https://github.com/customerio/issues/issues/6635


base/src/main/java/io/customer/base/extenstions/DateExtensions.kt
```diff
@@ -1,8 +1,26 @@
 package io.customer.base.extenstions
 
+import java.text.SimpleDateFormat
 import java.util.*
 import java.util.concurrent.TimeUnit
 
+enum class DateFormat {
```
Shahroz16, line 7: maybe `DateFormat` should be in `utils` folder rather than an extension class? to follow the conventions also, If someone was looking for something similar. it would be hard to guess it would be in extension folder

sdk/src/main/java/io/customer/sdk/CustomerIO.kt
```diff
@@ -111,6 +111,7 @@ class CustomerIO internal constructor(
         private var logLevel = CioLogLevel.ERROR
         internal var overrideDiGraph: CustomerIOComponent? = null // set for automated tests
         private var trackingApiUrl: String? = null
+        private var developerMode: Boolean = false
```
Shahroz16, line 114: what do you think about `debugMode` ?  

on another thought, `debugMode` might be a bit confusing with `debug` log mode maybe?

sdk/src/main/java/io/customer/sdk/util/Logger.kt
```diff
@@ -54,4 +60,46 @@ class LogcatLogger(
...
+    private fun logMessageToFile(level: Int, message: String) {
+        // Writing to external storage on Android has changed a lot over the many versions of the OS. Therefore, there are
+        // many use cases where an exception can be thrown and crash the app. This code is not critical to the functionality
+        // of the SDK so we wrap all of it's logic in a try/catch to prevent crashing the app.
+        try {
+            // Attempt to write to Downloads folder on the Android device. This allows customers/QA to send us logs really easily
+            // by navigating to their Downloads folder on their Android device and sharing the file. This is unlike saving the log
+            // file to internal storage of the mobile app where it's difficult to access.
+
+            // `getExternalStoragePublicDirectory` is deprecated but from my testing, it is still working.
```
Shahroz16, line 81: I won't suggest using this, as this is `deprecated` for security concerns and OS will abide by it. API 30+ it's not likely to work. 

Why don't we use `Context.getExternalFilesDir()`? the only downside to it, is you will be able to find files, using `File explorer` but `getExternalFilesDir()` does not get indexed by the MediaStore on Android 10+ so you won't have the gallery view.

But I think the tradeoff is worth it.

sdk/src/main/java/io/customer/sdk/CustomerIO.kt
levibostian: I do think a better naming should be done. I would ask the product team to see what they think. 

I think a better name should be chosen because I think this mode should be enabled for every build of a customer's app except for production. Because when they are testing staging or they are doing QA on their mobile app, we should be doing things like saving logs for them. 

Some name that better illustrates that this mode is for helping debugging but can be used by more then just developers developing the app.

sdk/src/main/java/io/customer/sdk/util/Logger.kt
levibostian: I do understand that `getExternalStoragePublicDirectory` is deprecated. It's funny because at Google I/O 2022, they actually showed the use of `getExternalStoragePublicDirectory` as a code sample in one of the video sessions. So, that made me wonder how serious of a deprecation it really is? 

Anyway, the reason I went with `getExternalStoragePublicDirectory` was because (1) it's a simple API and (2) it allows you to save to the Downloads folder without a UI popup or permission. We want to save the logs file in Downloads because it allows much easier sharing capabilities. 

Why downloads folder? Imagine you're a QA engineer testing your app. You encounter a bug with the CIO SDK. Your test device is not connected to your computer to be able to run `adb` or dump logs. You just want to email/slack/etc the logs file to us to take a look at. It's easy to navigate to your Downloads folder and send the file. 

Why it is nice to not have to ask permissions or show a UI popup for Downloads folder? Android is changing a lot of how files are saved. I am seeing a trend where the user of an app has to explicitly grant your app permission to save files to the Downloads folder via a popup window. This is fine, but if we can avoid having to show a UI popup saying that we need the user to grant us permission to the Downloads folder, that would be great. Because our SDK is installed on customer's apps, our customers might be confused that while they are QA testing their app they are asked to grant permission to Downloads folder. They might ask, "We don't use the Downloads folder in our mobile app. Why is it asking for it?". 

I figured that we could use `getExternalStoragePublicDirectory` until it no longer works. At that point, switch to the next best alternative. 

--- 

When I get a moment, I was going to try `Context.getExternalFilesDir()` to see if it gives us the same benefit of saving to Downloads without permission. If so, let's use it.

levibostian:
Closing PR as it's stale. We can implement again in the future if we decide to.

Issue #83, opened 2022-04-11
Ability to set notification icon/color


atreat:
The notification icon is currently set to the application icon inside [CustomerIOPushNotificationHandler](https://github.com/customerio/customerio-android/blob/c61a43210c8fb0aed028323768968fb0c967f2e7/messagingpush/src/main/java/io/customer/messagingpush/CustomerIOPushNotificationHandler.kt#L109). I'd like a way to set a different notification icon as for the majority of cases, the application icon will have a shape background which will be masked on Lollipop+ devices (Display a white circle/square). 

See screenshot:
![Screen Shot 2022-04-10 at 9 44 42 PM](https://user-images.githubusercontent.com/1031555/162656555-0c092355-92ad-454c-8d89-b86d3502c529.png)

# Why are we doing this for our customers? 

--

# How do we test that this change is successful? 

Send test notification when a custom notification icon has been set.

# How are we building this? 

Currently setting the application icon when building the notification:

https://github.com/customerio/customerio-android/blob/1.1.0-alpha.2/messagingpush/src/main/java/io/customer/messagingpush/CustomerIOPushNotificationHandler.kt#L118

FCM has some documentation that sets defaults when the icon does not exists. Unfortunately because the call to `setSmallIcon` these defaults do not seem to take.

```
<!-- Set custom default icon. This is used when no icon is set for incoming notification messages.
     See README(https://goo.gl/l4GJaQ) for more. -->
<meta-data
    android:name="com.google.firebase.messaging.default_notification_icon"
    android:resource="@drawable/ic_stat_ic_notification" />
<!-- Set color used with incoming notification messages. This is used when no color is set for the incoming
     notification message. See README(https://goo.gl/6BKBk7) for more. -->
<meta-data
    android:name="com.google.firebase.messaging.default_notification_color"
    android:resource="@color/colorAccent" />
```

I'm not sure if I'd prefer to set ApplicationManifest metadata to set the notification, or if a programmatic method similar to `CustomerIOFirebaseMessagingService.onMessageReceived` in order to achieve this.

Shahroz16:
Thank you for bringing this to our attention @atreat

We can surely look into it and add support for notification icon, you can expect this in the upcoming releases.

PR #118, opened 2022-07-07
feat: customize push notifications from app

mrehan27:
Closes: [83](https://github.com/customerio/customerio-android/issues/83)

### Changes

After discussing the changes internally, I have made following changes so that client apps can modify notification attributes without affecting SDK metrics.

- Added `PushNotificationListener` that notifies app side whenever we are about to show the push notification on device
- Added `CustomerIOPushNotificationService` to hold client app listeners
- Updated `CustomerIOPushNotificationHandler` to make sure `PushNotificationListener` gets correct callbacks
- Added `PushNotificationPayload` so that app side can read any data received for the push
- Added docs

### Sample Usage

```
CustomerIOPushNotificationService.setPushNotificationListener(listener = object : PushNotificationListener {
        override fun onNotificationComposed(
            payload: PushNotificationPayload,
            builder: NotificationCompat.Builder
        ) {
            builder.setSmallIcon(R.drawable.ic_small_push_notification)
        }
    }
)
```


levibostian (changes requested):
I have a question about the API of the SDK. See discussion with squad. 

https://customerio.slack.com/archives/C01QYDKD0SU/p1657207989846349

mrehan27:
Updated as discussed with squad, let me know if there are any questions or suggestions

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOPushNotificationService.kt
```diff
@@ -0,0 +1,20 @@
+package io.customer.messagingpush
+
+import io.customer.messagingpush.notification.PushNotificationListener
+
+/**
+ * Helper class to hold client listeners for push notifications
+ */
+object CustomerIOPushNotificationService {
+    internal var pushNotificationListener: PushNotificationListener? = null
```
levibostian, line 9: Does the SDK need to hold a reference to the listener? 

We should avoid holding any sort of strong references like this whenever possible. I think that we can avoid this reference if we design the API of this feature differently. 

If we are able to pass in a `PushNotificationListener` instance into a SDK function to be used in the function instead of having to first *set* a listener object and *then* call a function that uses the listener. 

Or, maybe we can make `PushNotificationListener.onNotificationComposed`  a function that you can override in the `CustomerIOFirebaseMessagingService`?

levibostian (changes requested):

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOPushNotificationService.kt
mrehan27: Are you suggesting to handle this in similar fashion like we did for `CustomerIOUrlHandler` in `CustomerIOConfig`? I don't think that's possible since we are in `messagingpush` module in this case and the notification call hierarchy is triggered by FCM so this might be tricky.

For the other suggestion, I think that might not be a very good idea as `CustomerIOFirebaseMessagingService` uses `CustomerIOPushNotificationHandler`, and updating `CustomerIOPushNotificationHandler` to reference back `CustomerIOFirebaseMessagingService` will create a cyclic dependency which isn't a very good idea in my opinion.

However, if we can think of doing this by adding some bridge class in between, that might be a good idea than holding listeners in an object. I'll think about handling it this way, meanwhile if you have something to suggest, please feel to share your thoughts.

mrehan27:
Closing this as we can now have a simpler solution with changes for Android 12 merged recently. Opened PR #129 for updated solution.

PR #125, opened 2022-07-25
fix!: android 12 deep link fix

mrehan27:
Fixes: [7801](https://github.com/customerio/issues/issues/7801)

## Changes

To fix deep link navigation from notifications, this PR has following changes

- **(BREAKING CHANGE)** - Replaced `CustomerIOUrlHandler` with `CustomerIOPushNotificationCallback` to return `TaskStackBuilder` instead of `boolean` for wider options
- **(BREAKING FOR INTERNAL MODULES ONLY)** - Modified base module classes to have their own configurations without changing anything in the SDK
- Fixed all modules to support config options, made sure the initialization isn't affected with it
- Added `MessagingPushModuleConfig` to make redirecting apps outside the host app from SDK optional
- Moved a few irrelevant classes from sdk to `messagingpush` module to reduce unnecessary coupling from the SDK
- Added `DeepLinkUtil` responsible for finding the right activity navigation
- Made `BigTextStyle` default for custom notifications
- Added kdocs
- Fixed test cases

### Apps targeting Android 12 or greater

- For apps targeting Android 12 or greater, updated `CustomerIOPushNotificationHandler` to create `PendingIntent` for correct navigation instead of starting `CustomerIOPushReceiver`
- For metrics like Opened, we'll now be completely relying on `PushTrackingUtil` triggered from `ActivityLifecycleCallbacks`
- Links supportive with other apps will be opened with the host app being opened in the background

### Apps targeting Android 11 or lower

- For apps targeting Android 11 or lower, added backward support to minimise behavior changes
- Metrics will be tracked the same way as they were tracked before
- Links supportive with other apps will be opened without opening the host app

### Sample Usage

```
setCustomerIOUrlHandler(object : CustomerIOUrlHandler {
    override fun createTaskStackFromPayload(
        context: Context,
        payload: CustomerIOParsedPushPayload
    ): TaskStackBuilder? {
        return if (url?.contains("home") == true) TaskStackBuilder.create(this).run {
          addNextIntentWithParentStack(Intent(context, MainActivity::class.java))
          getPendingIntent(
              0,
              PendingIntent.FLAG_UPDATE_CURRENT or PendingIntent.FLAG_IMMUTABLE
          )
        } else null
    }
})
```

------------------------------


sdk/src/main/java/io/customer/sdk/util/DeepLinkUtil.kt
```diff
@@ -0,0 +1,61 @@
...
+
+class DeepLinkUtilImpl : DeepLinkUtil {
+    override fun createIntentsForLink(context: Context, deepLink: String?): List<Intent> {
+        val intent = deepLink?.let { link -> resolveDeepLinkIntent(context, link) }
+        return if (intent == null) {
+            Log.v(
+                TAG,
+                "No supporting application for this deepLink $deepLink," +
+                    " launching default activity"
+            )
```
levibostian, line 31:
```suggestion
            logger.verbose(
                "No supporting application for this deepLink $deepLink," +
                    " launching default activity"
            )
```

The SDK has a `Logger` class that you inject into the constructor to do all logging with. We centralize it so we can log to files, log to logcat, etc.

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOPushNotificationHandler.kt
```diff
@@ -139,22 +142,44 @@ internal class CustomerIOPushNotificationHandler(private val remoteMessage: Remo
...
+    ): PendingIntent? {
+        // In Android 12, you must specify the mutability of each PendingIntent
+        val flags = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+            PendingIntent.FLAG_IMMUTABLE or PendingIntent.FLAG_UPDATE_CURRENT
+        } else {
+            PendingIntent.FLAG_UPDATE_CURRENT
+        }
+
+        val pushContentIntents = sdkConfig.urlHandler?.createIntentsForLink(deepLink)
+            ?: deepLinkUtil.createIntentsForLink(context, deepLink)
```
levibostian, line 170:
```suggestion
        val pushContentIntents = sdkConfig.urlHandler?.createIntentsForLink(deepLink)
            ?: deepLinkUtil.createDefaultDeepLinkHandlerIntent(context, deepLink) ?: createDefaultOpenAppIntent(context)
```

I think something like this reads better. The code right now reads like we only support deep links but if a push doesn't have a deep link in it, the SDK can just launch an Activity.

sdk/src/main/java/io/customer/sdk/data/communication/CustomerIOUrlHandler.kt
```diff
@@ -1,13 +1,27 @@
...
     /**
      * Callback called when deeplink action is performed.
-     * @param uri Deeplink URL
-     * @return Boolean return TRUE, if the URI was handled otherwise false.
+     *
+     * @param url link received in notification data
+     * @return intent to launch on notification click; null to let the SDK handle this
      */
-    fun handleCustomerIOUrl(uri: Uri): Boolean
+    fun createIntentForLink(url: String?): Intent?
```
levibostian, line 14: This function's job is for customers to implement in their app to be able to return back to the SDK an `Intent` to start when a push is clicked. 

However, the function name and parameters `create...ForLink()` and `url: String?` make it seem like customers should only implement this function if they want to handle a deep link when a push is pressed. 

I would like to suggest that we instead change this to: 

```suggestion
    fun createIntentForPush(parsedPush: CustomerIOParsedPushPayload): Intent?
```

where `CustomerIOParsedPushPayload` is an object that contains properties inside for the parsed push like this:
```kotlin
val push: CustomerIOParsedPushPayload = ...
push.deepLink: String? 
push.cioDeliveryId: String 
push.title: String 
push.body: String 
```

This is how to do "deep link handling" on the iOS SDK now. Instead of allowing customers to handle deep links in a push, we give them a parsed push object and allow them to handle a push when it's clicked, not just when it's a deep link that's pushed.

sdk/src/main/java/io/customer/sdk/util/DeepLinkUtil.kt
mrehan27: Thanks for sharing this, I'll update it to use `logger` instead

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOPushNotificationHandler.kt
```diff
@@ -62,7 +65,7 @@ internal class CustomerIOPushNotificationHandler(private val remoteMessage: Remo
     ): Boolean {
         // Check if message contains a data payload.
         if (bundle.isEmpty) {
-            Log.d(TAG, "Message data payload: $bundle")
+            logger.debug("Message data payload: $bundle")
```
levibostian, line 68: Currently, the `Logger` doesn't differentiate logs from 1 file to another. So, I recommend log statements to be more descriptive then `message data payload` because maybe some other file in the SDK also uses `message data payload`. 

```suggestion
            logger.debug("Push message received is empty.")            
```

Then outside of this `if` statement where the bundle is not empty:
```kotlin
logger.debug("Handling push message. Bundle: $bundle")
```

sdk/src/main/java/io/customer/sdk/data/model/CustomerIOParsedPushPayload.kt
```diff
@@ -0,0 +1,16 @@
...
+ * @property deepLink url received in notification data
+ * @property cioDeliveryId Customer.io message delivery id
+ * @property title notification content title text
+ * @property body notification content body text
+ */
+data class CustomerIOParsedPushPayload(
+    val deepLink: String?,
+    val cioDeliveryId: String,
+    val title: String,
+    val body: String
```
levibostian, line 15: I recommend adding the raw push payload's `Bundle` to this object since a customer might be adding custom `data` properties to the JSON push payload sent through CIO.

messagingpush/src/main/java/io/customer/messagingpush/data/model/CustomerIOParsedPushPayload.kt
```diff
@@ -0,0 +1,52 @@
...
+data class CustomerIOParsedPushPayload(
+    val extras: Bundle,
+    val deepLink: String?,
+    val cioDeliveryId: String,
+    val title: String,
+    val body: String
+) : Parcelable {
+    constructor(parcel: Parcel) : this(
+        extras = parcel.readBundle(Bundle::class.java.classLoader) ?: Bundle(),
+        deepLink = parcel.readString(),
```
jatinn, line 25: this can also be empty, can it not?

messagingpush/src/main/java/io/customer/messagingpush/data/model/CustomerIOParsedPushPayload.kt
```diff
@@ -0,0 +1,52 @@
...
+        cioDeliveryId = parcel.readString().orEmpty(),
+        title = parcel.readString().orEmpty(),
+        body = parcel.readString().orEmpty()
+    )
+
+    override fun writeToParcel(parcel: Parcel, flags: Int) {
```
jatinn, line 31: are flags needed here?

jatinn (approved):

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOPushReceiver.kt
```diff
@@ -39,49 +46,50 @@ internal class CustomerIOPushReceiver : BroadcastReceiver() {
             context.getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
         mNotificationManager.cancel(requestCode)
 
-        val bundle = intent.extras
-        val deliveryId = bundle?.getString(DELIVERY_ID_KEY)
-        val deliveryToken = bundle?.getString(DELIVERY_TOKEN_KEY)
+        val payload: CustomerIOParsedPushPayload? = intent.extras?.getParcelable(PUSH_PAYLOAD_KEY)
+        val deliveryId = payload?.cioDeliveryId
+        val deliveryToken = payload?.extras?.getString(DELIVERY_TOKEN_KEY)
```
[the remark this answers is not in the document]
mrehan27: **Adding notes from our discussion on this earlier**

Yes, we can extract delivery the same way. But since we already stored it in `cioDeliveryId`, we decided not to extract again and use one from data class.

messagingpush/src/main/java/io/customer/messagingpush/data/model/CustomerIOParsedPushPayload.kt
mrehan27: This can, since it is `nullable`, we don't need to add fallback empty value to this

messagingpush/src/main/java/io/customer/messagingpush/data/model/CustomerIOParsedPushPayload.kt
mrehan27: Not in our case though. But since they are added to base method, we need to keep them when overriding.

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOPushReceiver.kt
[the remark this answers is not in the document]
mrehan27: Yes, updated ✅

PR #134, opened 2022-09-30
feat: added shared instance for independent components

mrehan27:
### Changes

- Added a top level graph named `CustomerIOSharedComponent` to hold instances of classes that can be initialized without any dependency. Currently, the graph contains only following classes and can hold more where needed.
  - `Logger`
  - `DispatchersProvider`
- Added `CustomerIOShared` class to initialize our SDK on first access without any dependency and can be used to minimise challenged for delayed initializations, particularly helpful when building wrapper SDKs for non native frameworks
- Added test cases to verify log modifications and and shared instance behavior

### DIFF
--- a/sdk/src/main/java/io/customer/sdk/di/CustomerIOComponent.kt
+++ b/sdk/src/main/java/io/customer/sdk/di/CustomerIOComponent.kt
@@ -6,12 +6,7 @@ import io.customer.sdk.BuildConfig
 import io.customer.sdk.CustomerIOActivityLifecycleCallbacks
 import io.customer.sdk.CustomerIOConfig
 import io.customer.sdk.Version
-import io.customer.sdk.api.CustomerIOApiRetryPolicy
-import io.customer.sdk.api.HttpRequestRunner
-import io.customer.sdk.api.HttpRequestRunnerImpl
-import io.customer.sdk.api.HttpRetryPolicy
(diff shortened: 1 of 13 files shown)


sdk/src/main/java/io/customer/sdk/di/CustomerIOComponent.kt
```diff
@@ -71,7 +66,7 @@ class CustomerIOComponent(
         )
 
     val logger: Logger
-        get() = override() ?: LogcatLogger(sdkConfig)
+        get() = sharedComponent.logger
```
Shahroz16, line 69: a little confused here, should we keep the ability to override the logger? would there be any benifit? for example in test?

Shahroz16 (approved):

sdk/src/main/java/io/customer/sdk/di/CustomerIOComponent.kt
mrehan27: That's a good catch 😅 

Although, the only benefit I can think of doing this is; any tests using shared dependencies would be unchanged and will not having to override in shared instance separately, as doing it in current instance will work for them. Which makes sense to me. As long as we are using these dependencies from `CustomerIOComponent` instead of `CustomerIOSharedComponent` directly. I'll update this soon.

PR #139, opened 2022-10-05
chore: cleaned up code styles in sdk classes

mrehan27:
### Changes

- Fixed some typos, formatting and warnings, and deleted some unused code blocks
- Removed graph dependency from `CustomerIOFirebaseMessagingService` since we were only using graph for context. Added similar method to pass context and deprecated the current one to remove graph dependency from service
- Added constants to keep default configuration values at a single place. Categorized constants into two parts to keep core SDK and module level constants separate (Let me know if there are some concerns on this 👀)


Shahroz16 (commented):
Looks good, just need to resync changes with my branch because of overlap

mrehan27:
@Shahroz16 Yeah, agreed. I'll update it once your branch is merged to avoid any possible conflicts.

Shahroz16 (approved):
LGTM

PR #143, opened 2022-10-18
fix: user agent client support in preferences repo

mrehan27:
### Changes

- Added support for client support in preferences repository, so that user agent doesn't break when initializing from context
- Updated `CustomerIOStoredValues` and `SharedPreferenceRepository` visibility to `internal`
- Updated `Client` class to add support for initializing from raw value (helpful when initializing from preferences and react/expo sdk wrappers)
- Added and updated relevant tests

### DIFF
--- a/sdk/src/main/java/io/customer/sdk/di/CustomerIOSharedComponent.kt
+++ b/sdk/src/main/java/io/customer/sdk/di/CustomerIOSharedComponent.kt
@@ -6,7 +6,7 @@ import io.customer.sdk.repository.preference.SharedPreferenceRepositoryImp
 
 class CustomerIOSharedComponent(context: Context) : DiGraph() {
 
-    val sharedPreferenceRepository: SharedPreferenceRepository by lazy {
+    internal val sharedPreferenceRepository: SharedPreferenceRepository by lazy {
         override() ?: SharedPreferenceRepositoryImp(
             context = context
         )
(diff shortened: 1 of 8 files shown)


sdk/src/main/java/io/customer/sdk/di/CustomerIOSharedComponent.kt
```diff
@@ -6,7 +6,7 @@ import io.customer.sdk.repository.preference.SharedPreferenceRepositoryImp
 
 class CustomerIOSharedComponent(context: Context) : DiGraph() {
 
-    val sharedPreferenceRepository: SharedPreferenceRepository by lazy {
+    internal val sharedPreferenceRepository: SharedPreferenceRepository by lazy {
```
Shahroz16, line 9: 👍

sdk/src/main/java/io/customer/sdk/repository/preference/SharedPreferenceRepository.kt
```diff
@@ -41,6 +44,10 @@ internal class SharedPreferenceRepositoryImp(context: Context) : SharedPreferenc
                 siteId = getString(SITE_ID, null).orEmpty(),
                 apiKey = getString(API_KEY, null).orEmpty(),
                 region = getString(REGION, null).toRegion(),
+                client = Client.fromRawValue(
+                    source = getString(CLIENT_SOURCE, null) ?: "Unknown",
```
Shahroz16, line 48: In the defaults of the class `CustomerIOStoredValues`, we use
```kotlin
Client.Android(sdkVersion = Version.version)
```
 but here we set `Unknown` if the value doesn't exist. 
 
 should we probably have both defaults the same?

Shahroz16 (approved):

PR #148, opened 2022-11-15
fix: cio sdk version attribute using client value

mrehan27:
Helps: [8515](https://github.com/customerio/issues/issues/8515)

### Changes

- Uses wrapper SDK version from client value attributes passed in `DeviceStore`
- Added support for Flutter client


Shahroz16 (approved):
it would be great to have a test for it aswell

mrehan27:
@Shahroz16 Agreed, pushed test cases as well

PR #156, opened 2023-01-06
test: assert BQ can handle lots of tasks inside of it

levibostian:
Created a test to make sure that the SDK handles the edge case of lots of items in the BQ. 

This idea came from the iOS SDK [having a bug](https://github.com/customerio/customerio-ios/pull/245) in it when there are lots of items in the BQ. The Android SDK doesn't have this bug 😄.


sdk/src/main/java/io/customer/sdk/CustomerIO.kt
```diff
@@ -175,7 +175,7 @@ class CustomerIO internal constructor(
         private val modules: MutableMap<String, CustomerIOModule<out CustomerIOModuleConfig>> =
             mutableMapOf()
         private var logLevel: CioLogLevel = SDKConstants.LOG_LEVEL_DEFAULT
-        internal var overrideDiGraph: CustomerIOComponent? = null // set for automated tests
+        public var overrideDiGraph: CustomerIOComponent? = null // set for automated tests
```
mrehan27, line 178: Even though it is in builder and I don't expect customers playing around with this a lot, but I would still recommend to annotate it with `InternalCustomerIOApi` and add slightly more descriptive comment.

mrehan27 (approved):
Just one minor comment, rest looks good 👍🏻

Shahroz16 (approved):

PR #159, opened 2023-01-18
fix: remove currentRoute parameter in in-app event listener

levibostian:
Removing extra parameter from in-app event listener that [the team did not mention is needed](https://github.com/customerio/customerio-ios/pull/211#issuecomment-1332281565).

The parameter has also been removed in iOS SDK.


Shahroz16 (approved):
Looks good! just [asked the team about this](https://customerio.slack.com/archives/C036FTE6UJ0/p1674076544784709?thread_ts=1669742853.819709&cid=C036FTE6UJ0),  as I think this parameter might be helpful.

Shahroz16:
@levibostian its good to go, Colby just confirmed web SDK doesn't has it. 👍

PR #170, opened 2023-02-15
set gist dependency to use latest 3.X.Y version

jatinn:

### DIFF
--- a/buildSrc/src/main/kotlin/io.customer/android/Versions.kt
+++ b/buildSrc/src/main/kotlin/io.customer/android/Versions.kt
@@ -15,7 +15,7 @@ object Versions {
     internal const val FIREBASE_MESSAGING = "23.1.0"
     internal const val GRADLE_NEXUS_PUBLISH_PLUGIN = "1.1.0"
     internal const val GRADLE_VERSIONS_PLUGIN = "0.39.0"
-    internal const val GIST = "3.0.3"
+    internal const val GIST = "3.+"
     internal const val GOOGLE_PLAY_SERVICES_BASE = "17.6.0"
     internal const val KLUENT = "1.72"
     internal const val KOTLIN_BINARY_VALIDATOR = "0.10.1"


buildSrc/src/main/kotlin/io.customer/android/Versions.kt
```diff
@@ -15,7 +15,7 @@ object Versions {
     internal const val FIREBASE_MESSAGING = "23.1.0"
     internal const val GRADLE_NEXUS_PUBLISH_PLUGIN = "1.1.0"
     internal const val GRADLE_VERSIONS_PLUGIN = "0.39.0"
-    internal const val GIST = "3.0.3"
+    internal const val GIST = "3.+"
```
Shahroz16, line 18: This would fetch major updates as well, we need to get only minor as per [slack](https://customerio.slack.com/archives/C01QYDKD0SU/p1676471996614339?thread_ts=1675937862.413369&cid=C01QYDKD0SU)

```suggestion
    internal const val GIST = "3.0.+"
```

jatinn: hmm... i may have misunderstood the version doc then i thought
`3.+` meant 3.0 <= version < 4.0 so
`3.1.+` would mean 3.1 <= version < 3.2

guess with the other notation we could have `[3.1,4.0[` which should be 3.1 <= version < 4 is that more clear?

Shahroz16: > 3.1.+ would mean 3.1 <= version < 3.2

This is exactly what we want, no? 

```
3.X.Y => X major update, Y minor updates
```

levibostian:
I modified the PR title to be a `fix:` PR title. 

It was `chore:` and those types of commits do not create a new deployment. We want this PR to trigger a new deployment to customers.

levibostian (approved):

buildSrc/src/main/kotlin/io.customer/android/Versions.kt
levibostian: Keep in mind that this syntax of using `+` will install pre-production versions of the Gist SDK, if any are ever made. I don't anticipate any be made so I am OK with the syntax here. 

If we want to make sure we only install production releases, we could change to: `[3.0.0,4.0)`

lj-cio: I think Jatin is right here.
* MAJOR version when you make incompatible API changes
* MINOR version when you add functionality in a backwards compatible manner
* PATCH version when you make backwards compatible bug fixes

`MAJOR.MINOR.PATCH`

PR #174, opened 2023-02-22
fix: update CustomerIOFirebaseMessagingService to open

mrehan27:
Helps: https://github.com/customerio/issues/issues/9477

### Changes

- Updated `CustomerIOFirebaseMessagingService` access to open so we can inherit it in wrapper SDKs


Shahroz16 (approved):
Do we ned to make methods `open` too? are do you plan on making them open when the need arise?

mrehan27:
@Shahroz16 Not really I think, as the only two methods outside companion object are already overriden.

PR #177, opened 2023-03-01
fix: remove tasks from queue with 400 http response

Shahroz16:
closes: https://github.com/customerio/issues/issues/9443


sdk/src/sharedTest/java/io/customer/sdk/queue/QueueStorageIntegrationTest.kt
```diff
@@ -109,6 +109,19 @@ class QueueStorageIntegrationTest : BaseTest() {
...
+    fun givenDeleteGroupTask_expectUpdateInventory() {
+        val givenStartOfTheGroup = QueueTaskGroup.IdentifyProfile(String.random)
+
+        queueStorage.create(String.random, String.random, givenStartOfTheGroup, null)
+        queueStorage.create(String.random, String.random, null, listOf(givenStartOfTheGroup))
+        queueStorage.create(String.random, String.random, null, listOf(givenStartOfTheGroup))
+
+        val itemsDeleted = queueStorage.deleteGroup(givenStartOfTheGroup.toString())
+        itemsDeleted.count() shouldBeEqualTo 3
+        queueStorage.getInventory().count() shouldBeEqualTo 0
```
mrehan27, line 122: Shouldn't we also need to add test for case when there are other tasks in the queue as well to make sure only the desired tasks have been deleted?

mrehan27 (changes requested):

sdk/src/sharedTest/java/io/customer/sdk/queue/QueueStorageIntegrationTest.kt
Shahroz16: I added those tests, but forgot to push them 😆 
Screenshot 2023-03-02 at 9 31 32 PM

sdk/src/main/java/io/customer/sdk/queue/QueueStorage.kt
```diff
@@ -104,6 +105,20 @@ internal class QueueStorageImpl internal constructor(
...
+
+        val groupStart = inventory.filter { it.groupStart == groupStartTask }
+        val groupMember = inventory.filter { task ->
+            task.groupMember != null && groupStartTask in task.groupMember
+        }
+
+        val tasksToBeDeleted = groupStart + groupMember
+        tasksToBeDeleted.forEach { delete(it.taskPersistedId) }
+
+        return tasksToBeDeleted
```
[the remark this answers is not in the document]
Shahroz16: I could add some comments but this [doc](https://github.com/customerio/customerio-ios/blob/main/docs/dev-notes/BACKGROUND-QUEUE.md) could help add more clarity on the `groups` concept.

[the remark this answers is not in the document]
mrehan27: Thanks for sharing the docs. Also, the comments you now added are enough to read the code 😄

mrehan27 (approved):
I don't have a very good understanding of queue implementation yet 😅 But tried going through everything and looked good to me 👍🏻

sdk/src/main/java/io/customer/sdk/queue/QueueRunRequest.kt
```diff
@@ -60,25 +64,47 @@ internal class QueueRunRequestImpl internal constructor(
...
+                        logger.debug("queue deleting task $nextTaskStorageId because it will always fail")
+
+                        // check if its a groupStart task, if so delete the group otherwise just delete the task
+                        if (nextTaskToRunInventoryItem.groupStart == null) {
+                            queueStorage.delete(nextTaskStorageId)
+                        } else {
+                            queueStorage.deleteGroup(groupStartTask = nextTaskToRunInventoryItem.groupStart)
+                        }
+
+                        // since failed task isn't being saved, and the group is also deleted no need to update `lastFailedTask`
```
levibostian, line 86: 🤔 I understand that there is no need to updating `lastFailedTask`, but is there a problem with updating `lastFailedTask`? 

I am worried about a side-effect happening from this code decision in the future. What if the logic of `goToNextTask` changes and passing `null` here as `lastFailedTask` causes a bug? 

I am wondering if maybe it's a better idea for us to use a non-null value here even if it doesn't make any difference to the code today. Maybe it will make a difference in the future? The task did indeed fail, so makes sense to use a non-null value?

sdk/src/main/java/io/customer/sdk/api/HttpRequestRunner.kt
```diff
@@ -79,19 +82,24 @@ internal class HttpRequestRunnerImpl(
...
+                        errorBodyString // if we can't parse the error body json, the raw response string is good to capture.
 
                     parseCustomerIOErrorBody(errorBodyString)?.message?.let { parsedErrorMessage ->
                         errorMessage = parsedErrorMessage
                     }
                 }
 
-                val customerIOError = CustomerIOError.UnsuccessfulStatusCode(statusCode, errorMessage)
+                val customerIOError =
+                    CustomerIOError.UnsuccessfulStatusCode(statusCode, errorMessage)
```
levibostian, line 102: Not a blocker to this PR but I am curious. 

What is causing these format changes? ktlint? A Intellij plugin? 

It would be ideal if the formatting tools that were used in the project were the same across all engineers on the team. So if, for example, this change was caused by an intellij plugin, I think that plugin should be disabled or that rule gets applied to ktlint (or another tool that the team uses).

sdk/src/main/java/io/customer/sdk/queue/QueueStorage.kt
```diff
@@ -104,6 +105,23 @@ internal class QueueStorageImpl internal constructor(
         return jsonAdapter.fromJsonOrNull(fileContents)
     }
 
+    override fun deleteGroup(groupStartTask: String): List<QueueTaskMetadata> {
+        val inventory = getInventory()
+
+        // find the start of the group.
+        val groupStart = inventory.filter { it.groupStart == groupStartTask }
```
levibostian, line 112: I might have found an edge case that could cause a bug. 

If the caller of `deleteGroup` ever calls `delete(taskStorageId:)` before calling `deleteGroup`, this might cause a bug since the task `groupStart` will no longer exist. 

I think it would be good for this function to handle that edge case and this could be another reason for encapsulating the delete task + delete group members into this class could be a good idea.

sdk/src/sharedTest/java/io/customer/sdk/api/HttpRequestRunnerTest.kt
```diff
@@ -192,6 +192,18 @@ class HttpRequestRunnerTest : BaseTest() {
         actual shouldBeEqualTo expected
     }
 
+    @Test
+    fun performHttpRequest_given400_expectNotPausedHttpRequestsAndReturnFailure(): Unit = runBlocking {
```
levibostian, line 196: Good test to add. 

Does this test exist for the iOS SDK PR? I think it's a valuable test.

sdk/src/sharedTest/java/io/customer/sdk/queue/QueueRunRequestTest.kt
```diff
@@ -67,41 +74,109 @@ class QueueRunRequestTest : BaseTest() {
     }
 
     @Test
-    fun test_start_givenHttpRequestsPaused_expectDontDeleteTask_expectDontUpdateTask(): Unit = runBlocking {
+    fun test_run_givenRunTask400Failure_expectDeleteTask(): Unit = runBlocking {
```
levibostian, line 77: Do we have integration tests setup for the Queue? Aka: tests that do not require so much mocking such as `storageMock`? 

I think the tests you added are testing good edge cases but they could be higher quality and less annoying to maintain (since we don't use as many mocks) if they were integration tests.

sdk/src/sharedTest/java/io/customer/sdk/queue/QueueRunRequestTest.kt
```diff
@@ -67,41 +74,109 @@ class QueueRunRequestTest : BaseTest() {
...
+            QueueModifyResult(
+                true,
+                QueueStatus(siteId, 0)
+            )
+        )
         runRequest.run()
 
-        verify(storageMock, never()).delete(anyOrNull())
-        verify(storageMock, never()).update(anyOrNull(), anyOrNull())
+        verify(storageMock).delete(givenTaskId)
```
levibostian, line 98: Would be good to check that `storage.deleteGroup` was *not* called.

sdk/src/sharedTest/java/io/customer/sdk/queue/QueueRunRequestTest.kt
```diff
@@ -67,41 +74,109 @@ class QueueRunRequestTest : BaseTest() {
...
-        whenever(storageMock.delete(any())).thenReturn(QueueModifyResult(true, QueueStatus(siteId, 0)))
+        whenever(storageMock.deleteGroup(eq(givenTaskId))).thenReturn(givenListOfQueueTaskMetadata)
 
         runRequest.run()
 
-        verify(storageMock).delete(givenTaskId)
-        verify(storageMock).delete(givenTaskId2)
-        verify(runnerMock).runTask(givenQueueTask)
-        verify(runnerMock).runTask(givenQueueTask2)
+        verify(storageMock).deleteGroup(givenGroupOfTasks.toString())
```
levibostian, line 124: Would be good to check that `storage.delete` was *not* called.

sdk/src/sharedTest/java/io/customer/sdk/queue/QueueStorageIntegrationTest.kt
```diff
@@ -109,6 +109,49 @@ class QueueStorageIntegrationTest : BaseTest() {
         queueStorage.getInventory().count() shouldBeEqualTo 0
     }
 
+    @Test
+    fun givenDeleteGroupTask_expectAllTasksInGroupToDelete() {
+        val givenStartOfTheGroup = QueueTaskGroup.IdentifyProfile(String.random)
+
+        queueStorage.create(String.random, String.random, givenStartOfTheGroup, null)
+        queueStorage.create(String.random, String.random, null, listOf(givenStartOfTheGroup))
+        queueStorage.create(String.random, String.random, null, listOf(givenStartOfTheGroup))
```
levibostian, line 118: What about this edge case: 

What if we had this:
```
```
```suggestion
        queueStorage.create(String.random, String.random, groupStartOfAnotherGroup, listOf(givenStartOfTheGroup))
        queueStorage.create(String.random, String.random, null, listOf(givenStartOfTheGroup))
```

So a task was the member of a group *and* was also the start of a separate group. 

If *that* task gets deleted (because it's a member of a group), should we also delete all of the tasks that *it* starts? 

I think that register push token is one of these edge cases. It belongs to the identify profile group but it also starts it's own group. Delete device token is a member of that group that register starts.

sdk/src/sharedTest/java/io/customer/sdk/queue/QueueStorageIntegrationTest.kt
```diff
@@ -109,6 +109,49 @@ class QueueStorageIntegrationTest : BaseTest() {
...
+        val givenStartOfAnotherGroup = QueueTaskGroup.RegisterPushToken(String.random)
+
+        queueStorage.create(String.random, String.random, givenStartOfTheGroup, null)
+        queueStorage.create(String.random, String.random, null, listOf(givenStartOfTheGroup))
+        queueStorage.create(String.random, String.random, null, null)
+        queueStorage.create(String.random, String.random, null, listOf(givenStartOfAnotherGroup))
+
+        val itemsDeleted = queueStorage.deleteGroup(givenStartOfTheGroup.toString())
+        itemsDeleted.count() shouldBeEqualTo 2
+        queueStorage.getInventory().count() shouldBeEqualTo 2
```
levibostian, line 137: Be good to check that the 2 tasks in inventory are tasks you expected to be in the inventory.

sdk/src/sharedTest/java/io/customer/sdk/queue/QueueIntegrationTests.kt
```diff
@@ -27,24 +29,85 @@ class QueueIntegrationTests : BaseTest() {
...
+        mockWebServer.enqueue(MockResponse().setResponseCode(400))
+
+        queue.queueTrack(givenIdentifier, String.random, EventType.event, emptyMap())
+        mockWebServer.enqueue(MockResponse().setResponseCode(404))
+
         queue.run()
 
-        // expect all of tasks to run and run successfully
-        queueStorage.getInventory().count() shouldBeEqualTo 0
+        // expect tasks with 400 still present
```
levibostian, line 84:
```suggestion
        // expect tasks that did not receive a 400 response still present in queue
```

sdk/src/sharedTest/java/io/customer/sdk/queue/QueueIntegrationTests.kt
```diff
@@ -27,24 +29,85 @@ class QueueIntegrationTests : BaseTest() {
...
+        queueStorage.getInventory().count() shouldBeEqualTo 1
```
levibostian, line 85: I think it would be good to check that the 1 item remaining in the queue is the item we expect to see (the 404 response). 

You have done a lot of work on this PR. I'll help with this task.

sdk/src/sharedTest/java/io/customer/sdk/queue/QueueIntegrationTests.kt
```diff
@@ -27,24 +29,85 @@ class QueueIntegrationTests : BaseTest() {
...
         mockWebServer.requestCount shouldBeEqualTo 3
     }
+
+    @Test
+    fun givenRunQueueAndMultipleTaskGroupsFailWith400Tasks_expectQueueToBeEmpty(): Unit =
```
levibostian, line 90: Let's also make sure to test the edge case of a task being deleted because it's a member of a group but it's also the start of a group. 

I'll help add this test to keep the PR moving along.

sdk/src/sharedTest/java/io/customer/sdk/queue/QueueStorageIntegrationTest.kt
```diff
@@ -132,9 +134,74 @@ class QueueStorageIntegrationTest : BaseTest() {
...
+        queueStorage.getInventory().count() shouldBeEqualTo 0
+    }
+
+    @Test
+    fun givenDeleteGroupTask_givenMembersTasksBelongToDifferentGroups_givenDifferentOrder_expectAllStartTasksAndTheirMembersToBeDeleted() {
+        val givenStartOfTheGroup = QueueTaskGroup.IdentifyProfile(String.random)
+        val givenStartOfAnotherGroup = QueueTaskGroup.RegisterPushToken(String.random)
+
+        queueStorage.create(String.random, String.random, null, listOf(givenStartOfTheGroup))
+        queueStorage.create(String.random, String.random, givenStartOfTheGroup, null)
```
levibostian, line 176: This scenario *should* never occur where a BQ task that is the start of a group exists *after* a BQ task that belongs to that group. 

So, I don't see this test function adding value. Do you see a reason to keep this test function?

sdk/src/sharedTest/java/io/customer/sdk/queue/QueueStorageIntegrationTest.kt
```diff
@@ -132,9 +134,74 @@ class QueueStorageIntegrationTest : BaseTest() {
...
+            listOf(givenStartOfTheGroup)
+        )
+
+        val itemsDeleted = queueStorage.deleteGroup(givenStartOfTheGroup.toString())
+        itemsDeleted.count() shouldBeEqualTo 4
+        queueStorage.getInventory().count() shouldBeEqualTo 1
+    }
+
+    @Test
+    fun givenDeleteGroupTask_givenIncorrectQueue_expectNotToGetInfiniteLoop() {
```
levibostian, line 192:
```suggestion
    fun givenDeleteGroupTask_givenTaskStartsAndBelongsToSameGroup_expectNotToGetInfiniteLoop() {
```

I think defining "incorrect queue" would document what this test function is ultimately trying to test.

sdk/src/sharedTest/java/io/customer/sdk/queue/QueueStorageIntegrationTest.kt
Shahroz16: It's still an edge case that I like to keep, to make sure we safe guard against it.

levibostian (approved):

PR #178, opened 2023-03-03
fix: prevent crash for file not found exception

mrehan27:
closes: https://github.com/customerio/issues/issues/9530

### Changes

- Moved reading file text code inside `try-catch` to prevent crashing the app
- Added null safety operator for recently nullable file list method

> NOTE: This method doesn't fix the root cause for files being not found. We're not yet able to reproduce this, but the code change helps preventing customer apps from crashing

### DIFF
--- a/sdk/src/main/java/io/customer/sdk/data/store/FileStorage.kt
+++ b/sdk/src/main/java/io/customer/sdk/data/store/FileStorage.kt
@@ -59,8 +59,13 @@ class FileStorage internal constructor(
 
         if (!filePath.exists()) return null
 
-        val fileContents = filePath.readText()
-        if (fileContents.isBlank()) return null
+        val fileContents = try {
+            filePath.readText()
+        } catch (ex: Exception) {
+            logger.error("error while reading file $type. path ${filePath.absolutePath}. message: ${ex.message}")
+            null
+        }
+        if (fileContents.isNullOrBlank()) return null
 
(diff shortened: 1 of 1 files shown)


sdk/src/main/java/io/customer/sdk/data/store/FileStorage.kt
```diff
@@ -59,8 +59,13 @@ class FileStorage internal constructor(
 
         if (!filePath.exists()) return null
 
-        val fileContents = filePath.readText()
-        if (fileContents.isBlank()) return null
+        val fileContents = try {
+            filePath.readText()
+        } catch (ex: Exception) {
```
levibostian, line 64: Mind explaining why a try/catch is the solution to the issue?

levibostian (changes requested):

sdk/src/main/java/io/customer/sdk/data/store/FileStorage.kt
levibostian: I now see the note in the description 😄

mrehan27: 😅 Added inline comment as well

levibostian (approved):

Shahroz16 (approved):
Thanks for this, we should get it out asap. 🚀

PR #184, opened 2023-04-04
fix: push opened metrics tracked on Android 12

mrehan27:
fixes: https://github.com/customerio/issues/issues/9483

### Changes

- Updated `instanceOrNull` method to support modules initialization
- Updated `CustomerIOFirebaseMessagingService` to initialize push module when we initialize SDK using `context`

### DIFF
--- a/messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingService.kt
+++ b/messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingService.kt
@@ -86,8 +86,19 @@ open class CustomerIOFirebaseMessagingService : FirebaseMessagingService() {
             handleNewToken(context = context, token = token)
         }
 
+        /**
+         * Gets initialized instance of SDK. If the SDK is not initialized, it
+         * attempts to initialize with messaging push module so the module classes
+         * are initialized as well.
+         */
+        private fun getSDKInstanceOrNull(context: Context): CustomerIO? = try {
+            CustomerIO.instanceOrNull(context, listOf(ModuleMessagingPushFCM()))
+        } catch (e: Exception) {
+            null
+        }
+
         private fun handleNewToken(context: Context, token: String) {
-            CustomerIO.instanceOrNull(context)?.registerDeviceToken(deviceToken = token)
+            getSDKInstanceOrNull(context = context)?.registerDeviceToken(deviceToken = token)
         }
 
         private fun handleMessageReceived(
@@ -96,9 +107,8 @@ open class CustomerIOFirebaseMessagingService : FirebaseMessagingService() {
             handleNotificationTrigger: Boolean = true
         ): Boolean {
             // if CustomerIO instance isn't initialized, we can't handle the notification
-            if (CustomerIO.instanceOrNull(context, listOf(ModuleMessagingPushFCM())) == null) {
-                return false
-            }
+            getSDKInstanceOrNull(context = context) ?: return false
+
             val handler = CustomerIOPushNotificationHandler(remoteMessage = remoteMessage)
             return handler.handleMessage(context, handleNotificationTrigger)
         }
--- a/messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingService.kt
+++ b/messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingService.kt
@@ -87,9 +87,14 @@ open class CustomerIOFirebaseMessagingService : FirebaseMessagingService() {
         }
 
         /**
-         * Gets initialized instance of SDK. If the SDK is not initialized, it
-         * attempts to initialize with messaging push module so the module classes
-         * are initialized as well.
+         * Gets initialized instance of SDK. If the SDK is not initialized, we
+         * try to initialize the SDK and messaging push module earlier than requested
+         * by wrapper SDKs using stored values if available.
+         *
+         * By initializing the module early, we can register activity lifecycle callback
+         * required by messaging push module whenever we initialize the SDK using context.
+         * This helps us tracking metrics in wrapper SDKs for notifications received when
+         * app was in terminated state.
          */
         private fun getSDKInstanceOrNull(context: Context): CustomerIO? {
             return CustomerIO.instanceOrNull(context, listOf(ModuleMessagingPushFCM()))
--- a/messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingService.kt
+++ b/messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingService.kt
@@ -91,11 +91,9 @@ open class CustomerIOFirebaseMessagingService : FirebaseMessagingService() {
          * attempts to initialize with messaging push module so the module classes
          * are initialized as well.
          */
-        private fun getSDKInstanceOrNull(context: Context): CustomerIO? = try {
-            CustomerIO.instanceOrNull(context, listOf(ModuleMessagingPushFCM()))
-        } catch (e: Exception) {
-            null
-        }
+private fun getSDKInstanceOrNull(context: Context): CustomerIO? {
+    return CustomerIO.instanceOrNull(context, listOf(ModuleMessagingPushFCM()))
+}
 
         private fun handleNewToken(context: Context, token: String) {
             getSDKInstanceOrNull(context = context)?.registerDeviceToken(deviceToken = token)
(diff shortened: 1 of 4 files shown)


Shahroz16 (approved):

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingService.kt
```diff
@@ -86,8 +86,19 @@ open class CustomerIOFirebaseMessagingService : FirebaseMessagingService() {
...
+        /**
+         * Gets initialized instance of SDK. If the SDK is not initialized, it
+         * attempts to initialize with messaging push module so the module classes
+         * are initialized as well.
+         */
+        private fun getSDKInstanceOrNull(context: Context): CustomerIO? = try {
+            CustomerIO.instanceOrNull(context, listOf(ModuleMessagingPushFCM()))
+        } catch (e: Exception) {
+            null
+        }
```
Shahroz16, line 98: We can probably remove the try-catch block, as the `instanceOrNull` method inside it already handles the exceptions? 

```suggestion
fun getSDKInstanceOrNull(context: Context): CustomerIO? {
    return CustomerIO.instanceOrNull(context, listOf(ModuleMessagingPushFCM()))
}
```

Shahroz16 (commented):
found a small improvement 😄

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingService.kt
mrehan27: Oops 🤦🏻 Yeah 😅

sdk/src/main/java/io/customer/sdk/CustomerIO.kt
```diff
@@ -337,6 +344,14 @@ class CustomerIO internal constructor(
             val client = CustomerIO(diGraph)
             val logger = diGraph.logger
 
+            // cleanup of old reference if it exists
```
levibostian, line 347: I appreciate the comment here. It is missing explanation for why we need to cleanup the old reference. It would be great to have this comment modified for that.

sdk/src/main/java/io/customer/sdk/CustomerIO.kt
```diff
@@ -337,6 +344,14 @@ class CustomerIO internal constructor(
...
+            instance?.let {
+                // we need to unregister the previous activity lifecycle because it doesn't get garbage collected
+                // and will continue to hold a reference to the previous app context and return callbacks
+                appContext.unregisterActivityLifecycleCallbacks(it.diGraph.activityLifecycleCallbacks)
+                instance = null
```
levibostian, line 352: Why set instance to null? Can we remove this?

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingService.kt
```diff
@@ -86,8 +86,17 @@ open class CustomerIOFirebaseMessagingService : FirebaseMessagingService() {
             handleNewToken(context = context, token = token)
         }
 
+        /**
+         * Gets initialized instance of SDK. If the SDK is not initialized, it
+         * attempts to initialize with messaging push module so the module classes
+         * are initialized as well.
+         */
+        private fun getSDKInstanceOrNull(context: Context): CustomerIO? {
```
levibostian, line 94: Why was this function added?

levibostian:
I modified the title of the PR to explain why this commit exists instead of what it is doing in the code.

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingService.kt
mrehan27: Whenever we initialize SDK instance using `context` from `CustomerIOFirebaseMessagingService`, we should ideally initialize push-messaging module as well. For all instances in service requesting SDK initialization, they can call this method instead of copying the code again.

sdk/src/main/java/io/customer/sdk/CustomerIO.kt
Shahroz16: What's wrong with setting the instance to `null`?  it will just make it available earlier for GC collection. Whats the downside of having it?

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingService.kt
levibostian: From looking at the pull request diff, it looks like the SDK is currently calling `CustomerIO.instanceOrNull(context)`. For example, in `handleNewToken` in this file, the code today is `CustomerIO.instanceOrNull(context)` but this PR is changing it to using this function. 

What problem does `CustomerIO.instanceOrNull(context, listOf(ModuleMessagingPushFCM()))` solve instead of using what the code has today: `CustomerIO.instanceOrNull(context)`? I am assuming that the SDK works as intended today, so why change this code?

sdk/src/main/java/io/customer/sdk/CustomerIO.kt
levibostian: It's a nit-picky thing that shouldn't block the PR. 

I didn't expect to see this line of code and didn't understand the reason for adding it (especially because the next line of code is `instance = client`), so I thought I would ask the purpose it serves. 

Since I don't see a reason to add it and it isn't really hurting anything, I don't have much else to add here?

levibostian (approved):

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingService.kt
mrehan27: By initializing the module early, we can register activity lifecycle callback required by messaging push module whenever we initialize SDK using context from the service. So this is actually what fixes the problem for notifications received when app was in terminated state. Also, since this file is part of messaging push module, so requesting initialization of module with SDK from these files is actually the correct behavior in my opinion.

levibostian: I think that explanation you gave is really good! Can you add this explanation as a comment to the `getSDKInstanceOrNull` function explaining what problem that function solves in it's implementation?

mrehan27: Sure, updated docs on `getSDKInstanceOrNull`.

levibostian: Great, thanks. 

I think we should still ship this PR if it works, but I did have an idea for a refactor. 

The purpose of `getSDKInstanceOrNull` is to make sure that Activity lifecycle callbacks get registered with the SDK on initialization. The complexity exists because Activity lifecycle callbacks are in the messaging push module. 

If we were to make the tracking module register empty Activity lifecycle callbacks during SDK initialization and then using some sort of communication mechanism (event bus, callback functions) we can make the messaging push module get code executed when the methods in the tracking module's Activity lifecycle callback get called. 

If we were to make the SDK always register Activity lifecycle callbacks during SDK initialization, then that makes me think that this could replace something such as `getSDKInstanceOrNull` and make the SDK more stable?

sdk/src/main/java/io/customer/sdk/CustomerIO.kt
```diff
@@ -350,6 +350,10 @@ class CustomerIO internal constructor(
                 // we need to unregister the previous activity lifecycle because it doesn't get garbage collected
                 // and will continue to hold a reference to the previous app context and return callbacks
                 appContext.unregisterActivityLifecycleCallbacks(it.diGraph.activityLifecycleCallbacks)
+                // Cancelling queue timer as the new queue timer will take care of running queue tasks.
+                // If we do not cancel old timer, it results in multiple timers being run and accessing
+                // the same tasks.
+                it.diGraph.queue.cancelTimer()
```
levibostian, line 356: Thanks for adding this along with a comment explaining why this is here. 

In a call together, we talked about a long-term solution where `CustomerIO.Builder.build()` can be called 2+ times and the SDK config is the only thing that gets modified but the SDK instance or any dependencies do not get modified. 

Do we have a ticket to capture that refactor? I want to make sure before we introduce code like this that we need to maintain in the meantime before we do the refactor.

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingService.kt
mrehan27: Sure Levi. Since this apparently has worked well for us and fixes the bug, we'll ship this as soon as testing is complete. If you have ideas for refactoring, please feel free to create a ticket and suggest the changes in a separate PR.

sdk/src/main/java/io/customer/sdk/CustomerIO.kt
levibostian: Created a ticket for this work https://github.com/customerio/issues/issues/9915

PR #187, opened 2023-04-12
feat: in app click tracking

Shahroz16:
closes: https://github.com/customerio/issues/issues/8657

### DIFF
--- a/sdk/src/main/java/io/customer/sdk/data/request/DeliveryEvent.kt
+++ b/sdk/src/main/java/io/customer/sdk/data/request/DeliveryEvent.kt
@@ -13,7 +13,8 @@ internal enum class DeliveryType {
 internal data class DeliveryPayload(
     @field:Json(name = "delivery_id") val deliveryID: String,
     val event: MetricEvent,
-    val timestamp: Date
+    val timestamp: Date,
+    @field:Json(name = "metadata") val metaData: Map<String, String>
 )
 
 @JsonClass(generateAdapter = true)
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
@@ -84,11 +84,12 @@ internal constructor(
                 deliveryID = deliveryID,
                 event = MetricEvent.opened
             )
-        }, onAction = { deliveryID: String, _: String, _: String, _: String ->
+        }, onAction = { deliveryID: String, _: String, action: String, name: String ->
             logger.debug("in-app message clicked $deliveryID")
             trackRepository.trackInAppMetric(
                 deliveryID = deliveryID,
-                event = MetricEvent.clicked
+                event = MetricEvent.clicked,
+                metadata = mapOf("action_name" to name, "action_value" to action)
             )
         }, onError = { errorMessage ->
             logger.error("in-app message error occurred $errorMessage")
--- a/sdk/api/sdk.api
+++ b/sdk/api/sdk.api
@@ -351,8 +351,12 @@ public abstract interface class io/customer/sdk/repository/ProfileRepository {
 public abstract interface class io/customer/sdk/repository/TrackRepository {
 	public abstract fun screen (Ljava/lang/String;Ljava/util/Map;)V
 	public abstract fun track (Ljava/lang/String;Ljava/util/Map;)V
-	public abstract fun trackInAppMetric (Ljava/lang/String;Lio/customer/sdk/data/request/MetricEvent;)V
+	public abstract fun trackInAppMetric (Ljava/lang/String;Lio/customer/sdk/data/request/MetricEvent;Ljava/util/Map;)V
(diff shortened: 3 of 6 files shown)


sdk/src/main/java/io/customer/sdk/data/request/DeliveryEvent.kt
```diff
@@ -13,7 +13,8 @@ internal enum class DeliveryType {
 internal data class DeliveryPayload(
     @field:Json(name = "delivery_id") val deliveryID: String,
     val event: MetricEvent,
-    val timestamp: Date
+    val timestamp: Date,
+    @field:Json(name = "metadata") val metaData: Map<String, String>
```
levibostian, line 17:
```suggestion
    val metadata: Map<String, String>
```

I think metadata is spelled as one word. having it as one word could make this code more simple, too.

levibostian: Separately, [from the ticket](https://github.com/customerio/issues/issues/8657), the `metadata` payload is defined currently. I would prefer instead of using `Map<String, String>`, we have a data class with `action_name` and `action_value` inside of it.

mrehan27 (approved):

sdk/src/main/java/io/customer/sdk/data/request/DeliveryEvent.kt
levibostian: > I think metadata is spelled as one word
Huh. That is the first time I have seen `metaData`. In this PR, I see `metaData` and `metadata` both being used and I would prefer consistency being used in the codebase. Especially with a `@Json` annotation of `metadata`. I am not going to block this PR based on this, however. 

> The metadata payload is defined currently in ticket
If the SDK sends a pre-determined data set like this PR currently does (we send 2 properties, and the keys are hard-coded), then I would strongly prefer we stick with a strongly typed data set. However, such as the feature of profile/device attributes where we allow customers to send an undetermined data set with undetermined keys, that's where I see a `Map` being helpful.

Shahroz16: >If the SDK sends a pre-determined data set

I think that's our assumption that BE would always require these 2 values because the `meta-data` could be more than that. We only need these two variables, but in the future, we might need more. I don't think it's fixed it's always going to be only these 2 variables forever. 

So that is why I am more comfortable, with `Map`, I think the product shared the same idea as this is also a `Map` in the web SDK as mentioned earlier. So I would like to stick with the `Map`, if you feel more strongly about it, we can check with the broader team for you.

levibostian: I agree that `Map` is a great solution when the backend accepts a not pre-determined data set. I also agree that the backend today accepts a pre-determined data set and could change that requirement. 

The backend may never make a requirement change. It might make a requirement change to a different endpoint, not this one.

In my code review, I am suggesting to the team that when a payload is pre-determined in the API today, we use a data class. When the payload is not pre-determiend in the API today, we use a `Map`. When the backend makes a change to it's requirements, we make a modification to the SDKs to conform to that change. 

Since we cannot predict the future and the backend may never make a requirements change, I would rather use strongly typed data classes when possible to make our code easier to maintain.

Shahroz16: Thanks for the insights, I agree on the backend `may or may not` require changes, I don't want to add another class that we need to keep and maintain for this. It's the same thing as using `Pair`, why would you use it and not a static data class. I think it comes down to personal preference, and I see Rehan has approved it already, web is also using `Map`. So I would like to stick with it as well, I don't feel like its a blocker because it's not customer-facing nor is it performance related.

levibostian: To try and make communication easier on this thread, I played with the code [and made a refactor for this branch](https://github.com/customerio/customerio-android/pull/188) illustrating what I was referring to in my comments above. 

I do have a personal preference to use strongly-typed data classes when possible. However, I am open to not use a strong data type for this use case if the rest of the team prefers not to. 

To prevent this PR from being a blocker, I'll approve this PR and leave it up to the rest of the team to decide if this PR gets deployed as-is or include the refactor. 

Thanks for the work on the PR and the PR review conversations 👍

levibostian (approved):

sdk/src/main/java/io/customer/sdk/data/request/DeliveryEvent.kt
Shahroz16: Appreciate the PR Levi, it helps unblock and make sure that even if the work refactoring doesn't get merged in this round it can get merged in the second one. Refactor looks good, but I might not be able to test out if the `json` structuring is coming out correctly and test it manually with RH to see if all cases are working. 

If you can do that and confirm, I'll happily merge otherwise, If I manage to test it out myself I'll merge it otherwise I'll go with the current implementation which was tested by me. 👍

PR #189, opened 2023-04-20
chore: java layout sample app

mrehan27:

### Changes

- Java+Layouts sample app project in SDK
- App icon, base theme and colors
- `SampleApplication` and `BaseActivity` to help reduce similar code


samples/java_layout/src/main/java/io/customer/android/sample/java_layout/ui/core/BaseActivity.java
```diff
@@ -0,0 +1,28 @@
+package io.customer.android.sample.java_layout.ui.core;
+
+import android.os.Bundle;
+
+import androidx.annotation.EmptySuper;
+import androidx.annotation.Nullable;
+import androidx.appcompat.app.AppCompatActivity;
+import androidx.viewbinding.ViewBinding;
+
+public abstract class BaseActivity<VB extends ViewBinding> extends AppCompatActivity {
```
Shahroz16, line 10:
```suggestion
public abstract class BaseActivity<VB extends ViewBinding & Serializable> extends AppCompatActivity {
```

This ensures that the view binding classes used with this activity can be serialized, which is important for Android's state saving mechanism.

samples/java_layout/build.gradle
```diff
@@ -0,0 +1,49 @@
...
+        targetCompatibility JavaVersion.VERSION_1_8
+    }
+    buildFeatures {
+        viewBinding true
+    }
+}
+
+dependencies {
+    // Updating lifecycle version causes duplicate classes error because a large
+    // number of Lifecycle classes have been converted to Kotlin since 2.6
```
Shahroz16, line 34: we can exclude those? using something like this ?

```
    packagingOptions {
        exclude 'META-INF/lifecycle_runtime.kotlin_module'
        exclude 'META-INF/rxjava.properties'
        exclude 'META-INF/proguard/androidx-annotations.pro'
        exclude 'META-INF/proguard/androidx-core.pro'
    }
```

mrehan27: As discussed on slack, [this answer](https://stackoverflow.com/a/75298544) has helped solving this problem. Thanks for sharing 👍🏻

Shahroz16 (approved):

levibostian (approved):

levibostian:
Why is the path to this sample app `samples/java_layout/`? I was expecting it to be `samples/java/`.

samples/java_layout/build.gradle
```diff
@@ -0,0 +1,55 @@
...
+    id 'com.android.application'
+}
+
+android {
+    namespace 'io.customer.android.sample.java_layout'
+    compileSdk 33
+
+    defaultConfig {
+        applicationId "io.customer.android.sample.java_layout"
+        minSdk 21
```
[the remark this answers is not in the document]
mrehan27: We can, but I don't think it is a very good idea. We don't generally reuse constants in test cases as well so the values are updated manually to make sure everything works as expected. This is similar IMO, the current value matches the one supported by SDK right now and updates in future should be done manually so we can assure we know the behavior with apps using lower min sdks and can be added to our docs accordingly.

samples/java_layout/build.gradle
```diff
@@ -0,0 +1,55 @@
...
+    // causes duplicate classes error because a large number of Lifecycle classes have been
+    // converted to Kotlin since 2.6
+    // see: https://developer.android.com/jetpack/androidx/releases/lifecycle#2.6.0
+    constraints {
+        implementation("org.jetbrains.kotlin:kotlin-stdlib-jdk7:1.8.0") {
+            because("kotlin-stdlib-jdk7 is now a part of kotlin-stdlib")
+        }
+        implementation("org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.8.0") {
+            because("kotlin-stdlib-jdk8 is now a part of kotlin-stdlib")
+        }
```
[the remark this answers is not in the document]
mrehan27: We haven't added them directly. The [discussion above](https://github.com/customerio/customerio-android/pull/189#discussion_r1180474985) and [comment on the block](https://github.com/customerio/customerio-android/pull/189/files/f3273769ca54272d07596de5a92baa7b5e4c7fbf#diff-82bfd67dff87b442354e11655db6249d10cb394b1981a4e4de5b2f3d0198a3dbR43-R46) already explains why we need this.

mrehan27:
@levibostian `java` is a keyword used by Android project e.g. `src/main/java` files, it can sometimes conflict in refactoring and can be confusing for IDE. So I avoided using common keyword to be on safer side.

Also, it adds symmetry to sample apps name by following a pattern, e.g.

- `kotlin_compose` => Kotlin + Compose
- `java_layout` => Java + Layout files

PR #191, opened 2023-04-27
chore: added base project and icons

Shahroz16:
Part of https://github.com/customerio/issues/issues/9844


build.gradle
```diff
@@ -18,6 +18,7 @@ buildscript {
         classpath Dependencies.gradleNexusPublishPlugin
         classpath Dependencies.dokka
         classpath Dependencies.kotlinGradlePlugin
+        classpath 'org.jetbrains.kotlin:kotlin-gradle-plugin:1.7.20'
```
mrehan27, line 21: Why do we need this plugin here? Isn't `Dependencies.kotlinGradlePlugin` the same? 🤔

settings.gradle
```diff
@@ -5,3 +5,4 @@ include ':base'
 include ':messagingpush'
 include ':common-test'
 include ':messaginginapp'
+include ':kotlin_compose'
```
mrehan27, line 8: Does this show up in Android project view if you open the SDK? Asking because it is placed in `samples` folder, and without specifying path I think it might not show sample code when opening the SDK.

I suggest updating as follows so we can group all sample apps inside sample directory in Android project view when opening the SDK in Android Studio

```suggestion
include ':samples:kotlin_compose'
```

kotlin_compose/src/main/ic_launcher-playstore.png
mrehan27, line 1: Can you please double check why all these icons are not in `sample` folder?

mrehan27 (approved):

samples/kotlin_compose/build.gradle.kts
```diff
@@ -0,0 +1,71 @@
...
+    id("org.jetbrains.kotlin.android")
+}
+
+android {
+    namespace = "io.customer.android.sample.kotlin_compose"
+    compileSdk = 33
+
+    defaultConfig {
+        applicationId = "io.customer.android.sample.kotlin_compose"
+        minSdk = 24
```
levibostian, line 12: Are we able to set the min to the min version of our SDK? I think this will improve the quality of our testing with these apps being able to test the min SDK version of our SDK is appropriate.

samples/kotlin_compose/src/main/res/values/themes.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+    <style name="Theme.CustomerioSDK" parent="android:Theme.Material.Light.NoActionBar" />
```
levibostian, line 4: I thought that Jetpack Compose ignores `<style>` themes? I thought that themes were defined via Compose? 

So, is this file needed?

samples/kotlin_compose/build.gradle.kts
Shahroz16: I think we can use it, but I think that would be more of a restriction rather than flexibility. Because we would want our SDK to be compiled against different SDK versions, if all our sample compiles against one then we lose the change to test it.

samples/kotlin_compose/.gitignore
```diff
@@ -0,0 +1 @@
+/build
```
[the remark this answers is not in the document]
Shahroz16: Good idea.

samples/kotlin_compose/src/main/res/values/themes.xml
Shahroz16: I plan to use [Splash screen](https://developer.android.com/develop/ui/views/launch/splash-screen), which requires `styles` for lower version APIs. 

Using this saves us from adding a custom new splash screen compasable and logic of how many X seconds it should stay etc.

samples/kotlin_compose/build.gradle.kts
levibostian: I do think it's important that at least 1 of our sample apps uses the min SDK version. 

Maybe we make the Java sample app do that? We designate our Kotlin sample app to be more modern and use modern tools and SDK versions?

samples/kotlin_compose/src/main/res/values/themes.xml
levibostian: Do you believe that we need Splash screen? I want to avoid our sample apps having code that's irrelevant to QA or to customers using our SDK. Do you see a need for Splash screen and therefore this themes file?

samples/kotlin_compose/build.gradle.kts
mrehan27: Java app already uses [21 for `minSdk`](https://github.com/customerio/customerio-android/blob/main/samples/java_layout/build.gradle#L19) which is same as our [`minSdk` in SDK](https://github.com/customerio/customerio-android/blob/main/buildSrc/src/main/kotlin/io.customer/android/Configurations.kt#L6).

Are you trying to say we should reference `Configurations.minSdk` from Java sample app as well?

levibostian: :facepalm: I apologize, I made a mistake that caused this confusion. 

For some reason, I was thinking the min sdk of our sdk was lower then 21. After checking it again, it is indeed 21 which matches the Java sample app as expected. 

Sorry for causing confusion here. We are good with both apps 👍

samples/kotlin_compose/src/main/res/values/themes.xml
Shahroz16: Splash is for the UX, these sample apps are not just going to be used by QA's but I think like before marketing team might use it to demo and test too. So having a black screen while the app loads seems like a bad experience to me and this is a very small tradeoff for it. 

And generally regarding 
> I want to avoid our sample apps having code that's irrelevant to QA or to customers

I think it varies, and I don't want all the sample apps to be same, because customers have different implementations themselves too. So its good if there are different approaches taken in different sample app and customers can relate to the one it finds closest to their implementation. 

We have had customers asking how they can use CIO using DI, so they won't have to replicate stuff etc, so the different variations don't hurt.

PR #193, opened 2023-05-02
chore: added db and di layer for user related features

Shahroz16:

### DIFF
--- a/samples/kotlin_compose/build.gradle.kts
+++ b/samples/kotlin_compose/build.gradle.kts
@@ -1,6 +1,17 @@
+import java.util.Properties
+import java.io.FileInputStream
+
 plugins {
     id("com.android.application")
     id("org.jetbrains.kotlin.android")
+    id("org.jetbrains.kotlin.kapt")
+}
+
+val localProperties = Properties()
+try {
+    localProperties.load(FileInputStream(rootProject.file("local.properties")))
+} catch (ex: Exception) {
+    println("$ex")
 }
 
 android {
@@ -18,6 +29,9 @@ android {
         vectorDrawables {
             useSupportLibrary = true
         }
+
+        buildConfigField("String", "API_KEY", "\"" + localProperties["apiKey"] + "\"")
+        buildConfigField("String", "SITE_ID", "\"" + localProperties["siteId"] + "\"")
     }
 
     buildTypes {
@@ -49,6 +63,11 @@ android {
     }
 }
 
+val hiltVersion = "2.44.2"
+val composeVersion = "1.4.1"
+val coroutinesVersion = "1.5.2"
+val roomVersion = "2.4.2"
+
 dependencies {
 
     // compose compiler requires an updated version of kotlin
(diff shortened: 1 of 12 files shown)


samples/kotlin_compose/build.gradle.kts
```diff
@@ -18,6 +29,9 @@ android {
         vectorDrawables {
             useSupportLibrary = true
         }
+
+        buildConfigField("String", "API_KEY", "\"" + localProperties["apiKey"] + "\"")
+        buildConfigField("String", "SITE_ID", "\"" + localProperties["siteId"] + "\"")
```
mrehan27, line 34: Nothing to take action in this PR, but just wanted to add that in future, we might be able to move out similar settings in a common gradle file and use `apply from` to add config fields and customer.io dependencies 💭

samples/kotlin_compose/build.gradle.kts
```diff
@@ -49,18 +63,45 @@ android {
...
     implementation("androidx.compose.ui:ui-graphics")
     implementation("androidx.compose.ui:ui-tooling-preview")
     implementation("androidx.compose.material3:material3")
+
+    implementation(platform("androidx.compose:compose-bom:2022.10.00"))
+
+    // SDK
+    implementation(project(":sdk"))
+    implementation(project(":messagingpush"))
+    implementation(project(":messaginginapp"))
```
mrehan27, line 88: This should be easier to test specific version of SDK. May be use `IS_DEVELOPMENT` flag for this? Open to other options as well, just want to keep it easier to use remote/local versions.

samples/kotlin_compose/src/main/java/io/customer/android/sample/kotlin_compose/MainApplication.kt
```diff
@@ -0,0 +1,47 @@
...
+    lateinit var dataStore: DataStore<Preferences>
+    override fun onCreate() {
+        super.onCreate()
+        val (trackingApiUrl, siteId, apiKey) = runBlocking {
+            return@runBlocking try {
+                listOf(
+                    dataStore.data.first()[TRACK_API_URL_KEY],
+                    dataStore.data.first()[SITE_ID],
+                    dataStore.data.first()[API_KEY]
+                )
```
mrehan27, line 29: We can avoid repetition by using variable, something like follows:

```suggestion
                val data = dataStore.data.first()
                listOf(
                    data[TRACK_API_URL_KEY],
                    data[SITE_ID],
                    data[API_KEY]
                )
```

mrehan27 (approved):
Just some minor suggestions

samples/kotlin_compose/src/main/java/io/customer/android/sample/kotlin_compose/MainApplication.kt
Shahroz16: ah ! very good suggestion. It would not just avoid repetition but `data.first()` call is synchronous, so it won't block multiple times either.

samples/kotlin_compose/build.gradle.kts
levibostian: This comment might come off a little bias, haha, but [here is a tool for us to consider](https://github.com/levibostian/dotenv-android) for providing the site id and api key to the app at compile time. 

I like the gradle plugin because we can create environment variables on the CI server and then all of our sample apps compile based on those environment variables. No need to worry about generating build config properties or `gradle.properties` files which are popular in Android projects.

Shahroz16: Thanks for the suggestions Levi, currently I am not sure how this solution is making our lives easier. Because the gradle and build config seems easier to me. 

All we need in the CI pipeline is to create a `local.properties` file in the root project with the keys as we do it in RH and a gradle script

```
buildConfigField("String", "API_KEY", "\"" + localProperties["apiKey"] + "\"")
buildConfigField("String", "SITE_ID", "\"" + localProperties["siteId"] + "\"")
``` 

that all sample apps `apply` from. And these keys would be available at compile time. 

Whereas the plugin you mentions, looks great but also needs a `file` to be used, and some other config and would provide the keys in compile time.

mrehan27: Agreed with Shahroz here. Also, the changes I [pushed in java sample PR](https://github.com/customerio/customerio-android/pull/198/files#diff-82bfd67dff87b442354e11655db6249d10cb394b1981a4e4de5b2f3d0198a3db) makes it more easier. Once they are merged, I can pull the changes to separate gradle file use `apply from` for all our sample apps in the SDK. That will not only avoid repetition in gradle files, but will also help us setup CIO properties using the same idea that we have followed in RH and also making it simple for CI server.

PR #194, opened 2023-05-02
chore: java sample di and preferences

mrehan27:

### Changes

- Added SDK dependencies
- Loaded SDK credentials from `local.properties`
- Base DI graph
- Preferences store
- User object
- Factory for ViewModels
- `AuthViewModel` for user authentication
- Basic logic for authentication using splash screen
- Helper classes and functions

### DIFF
--- a/samples/java_layout/build.gradle
+++ b/samples/java_layout/build.gradle
@@ -2,6 +2,14 @@ plugins {
     id 'com.android.application'
 }
 
+def localProperties = new Properties()
+def localPropertiesFile = rootProject.file('local.properties')
+if (localPropertiesFile.exists()) {
+    localPropertiesFile.withReader('UTF-8') { reader ->
+        localProperties.load(reader)
+    }
+}
+
 android {
     namespace 'io.customer.android.sample.java_layout'
     compileSdk 33
@@ -12,6 +20,9 @@ android {
         targetSdk 33
         versionCode 1
         versionName "1.0"
+
+        buildConfigField "String", "API_KEY", "\"${localProperties["apiKey"]}\""
+        buildConfigField "String", "SITE_ID", "\"${localProperties["siteId"]}\""
     }
 
     buildTypes {
@@ -30,6 +41,18 @@ android {
 }
 
 dependencies {
+    // TODO Replace this with generic implementation for all sample SDKs to referencing remote versions is easier
+    def cioSDKVersion = localProperties["sdkVersion"]
+    if (cioSDKVersion == null || cioSDKVersion.toString().isBlank()) {
+        implementation(project(":sdk"))
+        implementation(project(":messagingpush"))
+        implementation(project(":messaginginapp"))
+    } else {
+        implementation "io.customer.android:tracking:$cioSDKVersion"
+        implementation "io.customer.android:messaging-push-fcm:$cioSDKVersion"
+        implementation "io.customer.android:messaging-in-app:$cioSDKVersion"
+    }
+
     def datastore_version = "1.0.0"
     def lifecycle_version = "2.6.1"
     implementation "androidx.annotation:annotation:1.6.0"
--- a/samples/java_layout/build.gradle
+++ b/samples/java_layout/build.gradle
@@ -30,13 +30,17 @@ android {
 }
 
 dependencies {
+    def datastore_version = "1.0.0"
     def lifecycle_version = "2.6.1"
     implementation "androidx.annotation:annotation:1.6.0"
     implementation "androidx.appcompat:appcompat:1.6.1"
     implementation "androidx.constraintlayout:constraintlayout:2.1.4"
     implementation "androidx.core:core-splashscreen:1.0.1"
+    implementation "androidx.datastore:datastore-preferences:$datastore_version"
+    implementation "androidx.datastore:datastore-preferences-rxjava3:$datastore_version"
     implementation "androidx.lifecycle:lifecycle-common-java8:$lifecycle_version"
     implementation "androidx.lifecycle:lifecycle-livedata:$lifecycle_version"
+    implementation "androidx.lifecycle:lifecycle-reactivestreams:$lifecycle_version"
     implementation "androidx.lifecycle:lifecycle-viewmodel:$lifecycle_version"
     implementation "com.google.android.material:material:1.8.0"
 
--- a/samples/java_layout/src/main/java/io/customer/android/sample/java_layout/data/PreferencesDataStore.java
+++ b/samples/java_layout/src/main/java/io/customer/android/sample/java_layout/data/PreferencesDataStore.java
@@ -1,0 +1,57 @@
+package io.customer.android.sample.java_layout.data;
+
+
+import androidx.annotation.MainThread;
+import androidx.datastore.preferences.core.MutablePreferences;
+import androidx.datastore.preferences.core.Preferences;
+import androidx.datastore.preferences.core.PreferencesKeys;
+import androidx.datastore.preferences.rxjava3.RxPreferenceDataStoreBuilder;
+import androidx.datastore.rxjava3.RxDataStore;
+
+import java.util.HashMap;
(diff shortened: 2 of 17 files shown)


samples/java_layout/src/main/java/io/customer/android/sample/java_layout/data/PreferencesDataStore.java
```diff
@@ -0,0 +1,55 @@
...
+import androidx.datastore.preferences.core.Preferences;
+import androidx.datastore.preferences.rxjava3.RxPreferenceDataStoreBuilder;
+import androidx.datastore.rxjava3.RxDataStore;
+
+import java.util.Map;
+
+import io.customer.android.sample.java_layout.SampleApplication;
+import io.reactivex.rxjava3.core.Flowable;
+
+public class PreferencesDataStore {
```
Shahroz16, line 14: It would be good to maybe create an `interface` for it, so in future if we are to add test cases we can mock do it easily.

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/data/PreferencesStoreUtils.java
```diff
@@ -0,0 +1,51 @@
...
+import androidx.datastore.preferences.core.PreferencesKeys;
+import androidx.datastore.rxjava3.RxDataStore;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import io.reactivex.rxjava3.core.Flowable;
+import io.reactivex.rxjava3.core.Single;
+
+class PreferencesStoreUtils {
```
Shahroz16, line 16: Maybe the methods in here can be part of `PreferencesDataStoreImpl` instead of separate utils class? that is reliant on `RxDataStore<Preferences>`  as well ?

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/data/model/User.java
```diff
@@ -0,0 +1,57 @@
...
+
+import android.text.TextUtils;
+
+import androidx.annotation.Nullable;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class User {
+    private static class Keys {
```
Shahroz16, line 11: we can probably, remove the separate class and have constats?

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/data/model/User.java
```diff
@@ -0,0 +1,57 @@
...
+        static final String DISPLAY_NAME = "display_name";
+        static final String EMAIL = "email";
+        static final String IS_GUEST = "is_guest";
+    }
+
+    @Nullable
+    public static User fromMap(Map<String, String> bundle) {
```
Shahroz16, line 18: `Optional<User>` instead of `null` maybe?

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/ui/user/AuthViewModel.java
```diff
@@ -0,0 +1,49 @@
...
+                bundle -> Optional.ofNullable(User.fromMap(bundle)));
+        this.userDataObservable = LiveDataReactiveStreams.fromPublisher(
+                publisher.filter(Optional::isPresent).map(Optional::get));
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+            this.userLoggedInStateObservable = LiveDataReactiveStreams.fromPublisher(
+                    publisher.map(Optional::isPresent));
+        } else {
+            //noinspection Convert2MethodRef
+            this.userLoggedInStateObservable = LiveDataReactiveStreams.fromPublisher(publisher.map(optional -> optional.isPresent()));
+        }
```
Shahroz16, line 31:
```suggestion
 public AuthViewModel(PreferencesDataStore preferencesDataStore) {
        this.preferencesDataStore = preferencesDataStore;
        Flowable<Optional<User>> publisher = preferencesDataStore.userData()
                .map(User::fromMap);
        this.userDataObservable = LiveDataReactiveStreams.fromPublisher(
                publisher.filter(Optional::isPresent).map(Optional::get));
        this.userLoggedInStateObservable = Transformations.map(userDataObservable, user -> user != null);
    }
```

We can probably use `Transformations` ?

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/data/PreferencesDataStore.java
mrehan27: I don't think we ever need to setup unit tests for sample apps. This will only add complexity IMO. So I tried to keep things simple as the main goal for sample apps is to test integration with SDK.

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/data/PreferencesStoreUtils.java
mrehan27: I did this to shorten class length and avoid code repetition. Moving this back to `PreferencesDataStore` will add little too much code to the class.

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/data/model/User.java
mrehan27: It actually is not required by any other class than `User`. So I would suggest not to move it outside the `User` class. I can replace it with static constants if you like?

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/ui/user/AuthViewModel.java
mrehan27: Had to tweak the suggestion slightly, updated code

Shahroz16 (approved):

samples/java_layout/build.gradle
```diff
@@ -30,13 +41,29 @@ android {
...
+    def cioSDKVersion = localProperties["sdkVersion"]
+    if (cioSDKVersion == null || cioSDKVersion.toString().isBlank()) {
+        implementation(project(":sdk"))
+        implementation(project(":messagingpush"))
+        implementation(project(":messaginginapp"))
+    } else {
+        implementation "io.customer.android:tracking:$cioSDKVersion"
+        implementation "io.customer.android:messaging-push-fcm:$cioSDKVersion"
+        implementation "io.customer.android:messaging-in-app:$cioSDKVersion"
+    }
```
levibostian, line 54: For our sample apps, I don't think that we need to reference remote versions. 

The idea is: Do you want to test a specific version of the CIO SDK? Checkout that SDK tag/commit in git and then compile the sample apps to test that commit/tag. 

I don't see a reason we would need to test remote versions?

Shahroz16: I don't know what the ideal solution would be, but 
>The idea is: Do you want to test a specific version of the CIO SDK? Checkout that SDK tag/commit in git and then compile the sample apps to test that commit/tag.

this would always mean testing a local SDK, there should be a way to able to test code checked out from maven etc to be sure that works too. 

An example of it could be, what if while publishing artifacts, one fails? you would never realize that using local version all the time but could be caught when testing remote version

mrehan27: +1 to what Shahroz said. And since this allows us to test both local and remote versions conveniently, we don't need to change anything here IMO.

levibostian: If the gradle configuration for the sample apps is: default to compiling the app against local SDK version but allow developer during development to install a remote version of SDK, then I am OK with this change. 

As part of the [QA CI workflow](https://github.com/customerio/issues/issues/9028), having sample apps compile on CI against local SDK for pull requests is important. Therefore, it should be the default behavior. Which, from looking at this gradle config file, that seems to be the case so I am OK with the configuration that exists here right now.

PR #195, opened 2023-05-03
chore: java sample login screen

mrehan27:

### Changes

- Login Screen UI Complete
- Dummy Settings Screen
- Random value generator class
- Theme fixes and updates


samples/java_layout/src/main/res/values/themes.xml
```diff
@@ -1,6 +1,6 @@
 <resources>
     <!-- Base application theme. -->
-    <style name="Theme.SampleApp" parent="Theme.MaterialComponents.DayNight.NoActionBar">
+    <style name="Theme.SampleApp.Base" parent="Theme.Material3.DayNight.NoActionBar">
```
Shahroz16, line 3: fancy 😮

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/ui/login/LoginActivity.java
```diff
@@ -29,10 +34,8 @@ protected void setupContent() {
     }
 
     private void setupObservers() {
-        authViewModel.getUserLoggedInStateObservable().observe(this, isLoggedIn -> {
-        });
-        authViewModel.getUserDataObservable().observe(this, user -> {
-        });
+        // TODO: Fetch and set user agent here
+        binding.userAgentTextView.setText("User agent will be shown here");
```
levibostian, line 38: I am assuming this comment is suggesting that we refactor our SDKs to expose the user-agent to public so we can display it in our sample apps. 

If that assumption is true, then I have a suggestion because I am currently not a fan of that solution of directly using the user-agent. 

For the iOS sample apps, in the sample apps, I [display 3 pieces of information:](https://github.com/customerio/customerio-ios/blob/ffe002ec0c131ecd103d89613c87088d764cf13b/Apps/CocoaPods-FCM/src/View/Widget/EnvironmentText.swift#L7-L8) 
1. the SDK version
2. the sample app version number
3. the sample app build number 

To accomplish this, I did need refactor the iOS SDK to [expose the SDK version publicly. ](https://github.com/customerio/customerio-ios/pull/299). But, I find that as a feature that customers may want to use. Which is why I like the solution - we are adding a feature to the SDK that customers might use, but our sample apps will definitely use. User-agent is a hack, IMO.

mrehan27: This was just added as reminder so we don't ignore anything from original designs. I'm not sure about exact idea behind adding the user agent, but if it was just added to figure out version numbers, the information can help us achieve the goal. I'll discuss with the squad to see what have we done for other sample apps and will make updates accordingly.

PR #200, opened 2023-05-10
chore: java sample events tracking

mrehan27:

### Changes

- Random event tracking
- Custom event tracking
- Device attributes tracking
- Profile attributes tracking
- Activity labels for auto screen tracking
- Updates in random generator class


samples/java_layout/src/main/java/io/customer/android/sample/java_layout/ui/core/BaseFragment.java
```diff
@@ -0,0 +1,50 @@
...
+import androidx.lifecycle.ViewModelProvider;
+import androidx.viewbinding.ViewBinding;
+
+import io.customer.android.sample.java_layout.SampleApplication;
+import io.customer.android.sample.java_layout.di.ApplicationGraph;
+
+public abstract class BaseFragment<VB extends ViewBinding> extends Fragment {
+
+    protected VB binding;
+    protected ApplicationGraph applicationGraph;
```
levibostian, line 21: Do you believe that holding a strong reference to the dependency injection graph in a Fragment will increase the risk of a memory leak from occurring?

levibostian (approved):
Wow, Android app code has come a *long* way since Java. I forgot about so much of the design patterns that exist in this PR.

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/ui/core/BaseFragment.java
mrehan27: Oh yeah, thanks for pointing out. Looking at [viewbinding docs](https://developer.android.com/topic/libraries/view-binding#fragments), we need to clear binding reference in `onDestroy`. I've [pushed the changes](https://github.com/customerio/customerio-android/pull/200/commits/289ea62f7051f5312c40636e4aba558ea117a66f) for it.

PR #201, opened 2023-05-11
chore: added settings screen and persistence of config

Shahroz16:


samples/kotlin_compose/src/main/java/io/customer/android/sample/kotlin_compose/MainApplication.kt
```diff
@@ -19,31 +15,35 @@ import kotlinx.coroutines.runBlocking
...
+            }
+            autoTrackDeviceAttributes(configuration.trackDeviceAttributes)
+            autoTrackScreenViews(configuration.trackScreen)
+
             addCustomerIOModule(ModuleMessagingInApp())
             addCustomerIOModule(ModuleMessagingPushFCM())
-            setLogLevel(CioLogLevel.DEBUG)
+
             build()
         }
```
levibostian, line 48: I like the idea of trying to make the sample apps source code helpful for customers to read and learn how to use our SDK in their app. 

The `configuration` object is an object that is unique to our sample apps and not relevant to customers. Therefore, I think that it would be ideal if the code that configures the SDK via the `configuration` object is abstracted away from this code block. 

Here is some code to explain the idea:

```kt
        // Initialize the Customer.io SDK
        CustomerIO.Builder(siteId: siteId, apiKey: apiKey, appContext: this).let { builder -> 
            // Configure the Customer.io SDK through the builder: 
			builder.setLogLevel(CioLogLevel.DEBUG) // For all of our sample apps, we prefer to set debug logs to make our internal testing easier. You may not need to do this.

            // This line of code is internal to Customer.io for testing purposes. Do not add this code to your app.
            configureSdkFromAppSettings(configuration, builder)

			// Finally, build the builder to finish initialization 
			builder.build()
        }
```

The idea is that we abstract away the `configuration` object modifications into a separate function `configureSdkFromAppSettings`. 

This code is similar to [what is in the FCM iOS sample app](https://github.com/customerio/customerio-ios/blob/a0224143b12b9bcdb1548965f35fbd38c8b6d61a/Apps/CocoaPods-FCM/src/AppDelegate.swift#L18-L24) and [the java sample app](https://github.com/customerio/customerio-android/blob/3aa5addd752a748eba53be61baae57e619d11540/samples/java_layout/src/main/java/io/customer/android/sample/java_layout/SampleApplication.java#L35-L52). 

For more details, you can see [a similar comment I left in java sample app](https://github.com/customerio/customerio-android/pull/198#discussion_r1188786852).

samples/kotlin_compose/src/main/java/io/customer/android/sample/kotlin_compose/ui/login/Login.kt
```diff
@@ -1,14 +1,184 @@
...
+    Column(
+        modifier = Modifier.padding(horizontal = 16.dp, vertical = 16.dp),
+        horizontalAlignment = Alignment.CenterHorizontally
+    ) {
+        OutlinedTextField(
+            value = name,
+            onValueChange = { newValue -> name = newValue },
+            label = { Text(stringResource(R.string.name)) },
+            modifier = Modifier.fillMaxWidth().padding(start = 0.dp, end = 0.dp),
+            isError = nameErrorState.isNotEmpty(),
```
levibostian, line 129: I think QA should be able to test what happens to the SDK if an empty string is used as an identifier in the SDK. Not because our SDK should handle that scenario, but simply allow QA the ability to test this scenario. 

Therefore, I think that textfields in the app should not error if the textfield is empty.

samples/kotlin_compose/src/main/java/io/customer/android/sample/kotlin_compose/ui/login/LoginViewModel.kt
```diff
@@ -1,5 +1,113 @@
...
+        }
+    }
+
+    private fun login(
+        email: String,
+        name: String,
+        isGuest: Boolean = false,
+        onLoginSuccess: () -> Unit
+    ) {
+        viewModelScope.launch(Dispatchers.IO) {
```
levibostian, line 71: Is this needed? Can we do everything on the main thread since our sample apps are not doing anything intensive?

samples/kotlin_compose/src/main/java/io/customer/android/sample/kotlin_compose/data/models/Configuration.kt
```diff
@@ -0,0 +1,20 @@
...
+    var apiKey: String
+) {
+    var trackUrl: String? = null
+
+    var backgroundQueueSecondsDelay: Double = BACKGROUND_QUEUE_SECONDS_DELAY
+    var backgroundQueueMinNumTasks: Int = BACKGROUND_QUEUE_MIN_NUMBER_OF_TASKS
+
+    var trackScreen: Boolean = SHOULD_AUTO_RECORD_SCREEN_VIEWS
+    var trackDeviceAttributes: Boolean = AUTO_TRACK_DEVICE_ATTRIBUTES
+    var debugMode: Boolean = false
```
levibostian, line 19:
```suggestion
    var trackUrl: String? = null

    var backgroundQueueSecondsDelay: Double? 
    var backgroundQueueMinNumTasks: Int? 

    var trackScreen: Boolean?
    var trackDeviceAttributes: Boolean?
    var debugMode: Boolean?
```

I am assuming that the code `var backgroundQueueSecondsDelay: Double = BACKGROUND_QUEUE_SECONDS_DELAY` is giving a default CIO SDK config value in the app? Which means that the sample app is providing a default value that overrides the default value that is provided by the CIO SDK? 

If this assumption is true, I suggest that the sample app does not provide it's own set of default values. The sample app either configures the CIO SDK with a value that has been set in the settings screen (so the user of the app decided to override the SDK configuration) or the CIO SDK default value is used. 

From this suggestion, I would expect all of the `Configuration` values in this data class to be optional unless the user goes into the settings screen and changes something.

Shahroz16: >I am assuming that the code var backgroundQueueSecondsDelay: Double = BACKGROUND_QUEUE_SECONDS_DELAY is giving a default CIO SDK config value in the app? Which means that the sample app is providing a default value that overrides the default value that is provided by the CIO SDK?

We are referencing the same default values that the CIO SDK is using, hence we are not overriding it but using the same values. if the CIO SDK defaults change so would the app's default. And once user change the apps, app would then used the persisted ones.

samples/kotlin_compose/src/main/java/io/customer/android/sample/kotlin_compose/di/RepositoryModule.kt
```diff
@@ -1,21 +1,27 @@
...
+import dagger.hilt.components.SingletonComponent
 import io.customer.android.sample.kotlin_compose.data.persistance.UserDao
+import io.customer.android.sample.kotlin_compose.data.repositories.PreferenceRepository
+import io.customer.android.sample.kotlin_compose.data.repositories.PreferenceRepositoryImp
 import io.customer.android.sample.kotlin_compose.data.repositories.UserRepository
 import io.customer.android.sample.kotlin_compose.data.repositories.UserRepositoryImpl
 
 @Module
-@InstallIn(ViewModelComponent::class)
+@InstallIn(SingletonComponent::class)
```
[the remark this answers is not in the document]
Shahroz16: We want to use the values in `Application` launch. The alternative would be to have another class that just accesses `datastore` and only expose that. 

But the additional work is not needed for the sample apps.

samples/kotlin_compose/src/main/java/io/customer/android/sample/kotlin_compose/ui/login/LoginViewModel.kt
Shahroz16:
```
userRepository.login(email = email, name = name, isGuest = isGuest)
```

We are inserting things in DB and even though data isn't much i don't feel comfortable inserting stuff in DB in main thread

samples/kotlin_compose/src/main/java/io/customer/android/sample/kotlin_compose/ui/login/Login.kt
Shahroz16: I don't particularly agree, it just unnecessarily breaks the user flow. If QA asks for this scenario we can create one for them. But we can't just guess and add it ourselves. 

I think others sample apps are doing the same except the FCM-COCOPODS one.

samples/kotlin_compose/src/main/java/io/customer/android/sample/kotlin_compose/MainApplication.kt
Shahroz16: Let me see what can be done about this.

samples/kotlin_compose/src/main/java/io/customer/android/sample/kotlin_compose/ui/settings/Settings.kt
```diff
@@ -1,9 +1,305 @@
...
+
+    Column(verticalArrangement = Arrangement.spacedBy(4.dp), modifier = Modifier.fillMaxWidth()) {
+        OutlinedTextField(
+            modifier = Modifier.fillMaxWidth(),
+            value = secondsDelays.toString(),
+            onValueChange = {
+                if (it.isNotEmpty()) {
+                    secondsDelays = it.toDouble()
+                    onBackgroundQueueSecondsDelayChange(it.toDouble())
+                }
```
mrehan27, line 206: Shouldn't we use default value if it is empty? e.g.

1. Change the value to `10`
2. Save settings
3. Go to settings screen again
4. Remove the value completely
5. Save settings

Will the value saved be `10` (last saved) or `30` (default value) in this case? 🤓

samples/kotlin_compose/src/main/java/io/customer/android/sample/kotlin_compose/MainApplication.kt
```diff
@@ -19,31 +15,35 @@ import kotlinx.coroutines.runBlocking
...
-                listOf(
-                    data[TRACK_API_URL_KEY],
-                    data[SITE_ID],
-                    data[API_KEY]
-                )
-            } catch (e: Exception) {
-                listOf(null, BuildConfig.SITE_ID, BuildConfig.API_KEY)
-            }
+        val configuration = runBlocking {
+            preferences.getConfiguration().first()
```
mrehan27, line 23: Does it guarantee it will return some value? With fresh install as well? 🤔

samples/kotlin_compose/src/main/java/io/customer/android/sample/kotlin_compose/MainApplication.kt
mrehan27: In case it helps, I have moved all SDK calls to a repository [like this](https://github.com/customerio/customerio-android/pull/203/files#diff-d69d87bddf0123142bdd66da3e46acad1ded0fa900f1877c5617d23bfca5276b) in Java sample.

mrehan27 (approved):

samples/kotlin_compose/src/main/java/io/customer/android/sample/kotlin_compose/MainApplication.kt
Shahroz16: Yes it initializes it with `defaults`

samples/kotlin_compose/src/main/java/io/customer/android/sample/kotlin_compose/ui/settings/Settings.kt
Shahroz16: It makes it 0 if you remove it, so when you save it its saved as 0.

PR #204, opened 2023-05-12
fix: improve delivered metrics

mrehan27:
closes: https://github.com/customerio/issues/issues/10066

### Changes

- `CustomerIOFirebaseMessagingReceiver` for tracking `delivered` using C2DM callbacks from GCM
- `JvmStatic` to methods from `CustomerIOFirebaseMessagingService`
- Extension function on `context` to create and retrieve SDK instance in push module
- Refactored `PushTrackingUtil` to remove redundant code

### DIFF
--- a/messagingpush/src/main/AndroidManifest.xml
+++ b/messagingpush/src/main/AndroidManifest.xml
@@ -22,6 +22,15 @@
                 <action android:name="io.customer.messagingpush.PUSH_ACTION" />
             </intent-filter>
         </receiver>
+        <!-- Broadcast receiver for listening to push events from GCM -->
+        <receiver
+            android:name=".CustomerIOFirebaseMessagingReceiver"
+            android:exported="true"
+            android:permission="com.google.android.c2dm.permission.SEND">
+            <intent-filter>
+                <action android:name="com.google.android.c2dm.intent.RECEIVE" />
+            </intent-filter>
+        </receiver>
     </application>
 
     <queries>
--- a/messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingReceiver.kt
+++ b/messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingReceiver.kt
@@ -3,9 +3,25 @@ package io.customer.messagingpush
 import android.content.BroadcastReceiver
 import android.content.Context
 import android.content.Intent
+import io.customer.messagingpush.di.moduleConfig
+import io.customer.messagingpush.di.pushTrackingUtil
+import io.customer.sdk.CustomerIO
 
 class CustomerIOFirebaseMessagingReceiver : BroadcastReceiver() {
+    private fun getSDKInstanceOrNull(context: Context): CustomerIO? {
+        return CustomerIO.instanceOrNull(context, listOf(ModuleMessagingPushFCM()))
+    }
+
     override fun onReceive(context: Context, intent: Intent) {
-        // Handle receive event here
+        val extras = intent.extras ?: return
+        val sdkInstance = getSDKInstanceOrNull(context) ?: return
+
+        val moduleConfig = sdkInstance.diGraph.moduleConfig
+        val pushTrackingUtil = sdkInstance.diGraph.pushTrackingUtil
+
+        // Track delivered event only if auto-tracking is enabled
+        if (moduleConfig.autoTrackPushEvents) {
+            pushTrackingUtil.parseIntentExtrasForTrackingDelivered(extras)
+        }
     }
 }
--- a/messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingReceiver.kt
+++ b/messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingReceiver.kt
@@ -1,0 +1,11 @@
+package io.customer.messagingpush
+
+import android.content.BroadcastReceiver
+import android.content.Context
+import android.content.Intent
+
+class CustomerIOFirebaseMessagingReceiver : BroadcastReceiver() {
+    override fun onReceive(context: Context, intent: Intent) {
+        // Handle receive event here
+    }
+}
--- a/messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingReceiver.kt
+++ b/messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingReceiver.kt
@@ -5,16 +5,14 @@ import android.content.Context
 import android.content.Intent
 import io.customer.messagingpush.di.moduleConfig
 import io.customer.messagingpush.di.pushTrackingUtil
-import io.customer.sdk.CustomerIO
+import io.customer.messagingpush.extensions.getSDKInstanceOrNull
 
 class CustomerIOFirebaseMessagingReceiver : BroadcastReceiver() {
-    private fun getSDKInstanceOrNull(context: Context): CustomerIO? {
-        return CustomerIO.instanceOrNull(context, listOf(ModuleMessagingPushFCM()))
-    }
 
     override fun onReceive(context: Context, intent: Intent) {
         val extras = intent.extras ?: return
-        val sdkInstance = getSDKInstanceOrNull(context) ?: return
+        // if CustomerIO instance isn't initialized, we cannot process the notification
+        val sdkInstance = context.getSDKInstanceOrNull() ?: return
 
         val moduleConfig = sdkInstance.diGraph.moduleConfig
         val pushTrackingUtil = sdkInstance.diGraph.pushTrackingUtil
--- a/messagingpush/src/main/java/io/customer/messagingpush/util/PushTrackingUtil.kt
+++ b/messagingpush/src/main/java/io/customer/messagingpush/util/PushTrackingUtil.kt
@@ -6,6 +6,7 @@ import io.customer.sdk.repository.TrackRepository
 
 interface PushTrackingUtil {
     fun parseLaunchedActivityForTracking(bundle: Bundle): Boolean
+    fun parseIntentExtrasForTrackingDelivered(bundle: Bundle): Boolean
 
     companion object {
         const val DELIVERY_ID_KEY = "CIO-Delivery-ID"
@@ -16,8 +17,7 @@ interface PushTrackingUtil {
 class PushTrackingUtilImpl(
     private val trackRepository: TrackRepository
 ) : PushTrackingUtil {
-
-    override fun parseLaunchedActivityForTracking(bundle: Bundle): Boolean {
+    private fun trackMetricEvent(bundle: Bundle, event: MetricEvent): Boolean {
         val deliveryId = bundle.getString(PushTrackingUtil.DELIVERY_ID_KEY)
         val deliveryToken = bundle.getString(PushTrackingUtil.DELIVERY_TOKEN_KEY)
 
@@ -26,9 +26,16 @@ class PushTrackingUtilImpl(
         trackRepository.trackMetric(
             deliveryID = deliveryId,
             deviceToken = deliveryToken,
-            event = MetricEvent.opened
+            event = event
         )
-
         return true
+    }
+
+    override fun parseLaunchedActivityForTracking(bundle: Bundle): Boolean {
+        return trackMetricEvent(bundle, MetricEvent.opened)
+    }
+
+    override fun parseIntentExtrasForTrackingDelivered(bundle: Bundle): Boolean {
+        return trackMetricEvent(bundle, MetricEvent.delivered)
     }
 }
--- a/messagingpush/api/messagingpush.api
+++ b/messagingpush/api/messagingpush.api
@@ -6,7 +6,10 @@ public final class io/customer/messagingpush/CustomerIOFirebaseMessagingReceiver
 public class io/customer/messagingpush/CustomerIOFirebaseMessagingService : com/google/firebase/messaging/FirebaseMessagingService {
 	public static final field Companion Lio/customer/messagingpush/CustomerIOFirebaseMessagingService$Companion;
 	public fun <init> ()V
+	public static final fun onMessageReceived (Landroid/content/Context;Lcom/google/firebase/messaging/RemoteMessage;)Z
+	public static final fun onMessageReceived (Landroid/content/Context;Lcom/google/firebase/messaging/RemoteMessage;Z)Z
 	public fun onMessageReceived (Lcom/google/firebase/messaging/RemoteMessage;)V
+	public static final fun onNewToken (Landroid/content/Context;Ljava/lang/String;)V
 	public fun onNewToken (Ljava/lang/String;)V
 }
 
--- a/messagingpush/api/messagingpush.api
+++ b/messagingpush/api/messagingpush.api
@@ -120,6 +120,7 @@ public abstract interface class io/customer/messagingpush/util/PushTrackingUtil 
 	public static final field Companion Lio/customer/messagingpush/util/PushTrackingUtil$Companion;
 	public static final field DELIVERY_ID_KEY Ljava/lang/String;
 	public static final field DELIVERY_TOKEN_KEY Ljava/lang/String;
+	public abstract fun parseIntentExtrasForTrackingDelivered (Landroid/os/Bundle;)Z
 	public abstract fun parseLaunchedActivityForTracking (Landroid/os/Bundle;)Z
 }
 
@@ -130,6 +131,7 @@ public final class io/customer/messagingpush/util/PushTrackingUtil$Companion {
 
 public final class io/customer/messagingpush/util/PushTrackingUtilImpl : io/customer/messagingpush/util/PushTrackingUtil {
 	public fun <init> (Lio/customer/sdk/repository/TrackRepository;)V
+	public fun parseIntentExtrasForTrackingDelivered (Landroid/os/Bundle;)Z
 	public fun parseLaunchedActivityForTracking (Landroid/os/Bundle;)Z
 }
 
--- a/messagingpush/api/messagingpush.api
+++ b/messagingpush/api/messagingpush.api
@@ -1,3 +1,8 @@
+public final class io/customer/messagingpush/CustomerIOFirebaseMessagingReceiver : android/content/BroadcastReceiver {
+	public fun <init> ()V
+	public fun onReceive (Landroid/content/Context;Landroid/content/Intent;)V
+}
+
 public class io/customer/messagingpush/CustomerIOFirebaseMessagingService : com/google/firebase/messaging/FirebaseMessagingService {
 	public static final field Companion Lio/customer/messagingpush/CustomerIOFirebaseMessagingService$Companion;
 	public fun <init> ()V
--- a/messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingService.kt
+++ b/messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingService.kt
@@ -3,6 +3,7 @@ package io.customer.messagingpush
 import android.content.Context
 import com.google.firebase.messaging.FirebaseMessagingService
 import com.google.firebase.messaging.RemoteMessage
+import io.customer.messagingpush.extensions.getSDKInstanceOrNull
 import io.customer.sdk.CustomerIO
 
 open class CustomerIOFirebaseMessagingService : FirebaseMessagingService() {
@@ -20,6 +21,7 @@ open class CustomerIOFirebaseMessagingService : FirebaseMessagingService() {
          * @return Boolean indicating whether this will be handled by CustomerIO
          */
         @JvmOverloads
+        @JvmStatic
         fun onMessageReceived(
             context: Context,
             remoteMessage: RemoteMessage,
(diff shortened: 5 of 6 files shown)


messagingpush/src/main/java/io/customer/messagingpush/util/PushTrackingUtil.kt
```diff
@@ -26,9 +26,16 @@ class PushTrackingUtilImpl(
...
         return true
     }
+
+    override fun parseLaunchedActivityForTracking(bundle: Bundle): Boolean {
+        return trackMetricEvent(bundle, MetricEvent.opened)
+    }
+
+    override fun parseIntentExtrasForTrackingDelivered(bundle: Bundle): Boolean {
+        return trackMetricEvent(bundle, MetricEvent.delivered)
+    }
```
Shahroz16, line 40: This looks like an overkill...

Shahroz16, line 40: This looks like an overkill...

messagingpush/src/main/java/io/customer/messagingpush/util/PushTrackingUtil.kt
```diff
@@ -6,6 +6,7 @@ import io.customer.sdk.repository.TrackRepository
 
 interface PushTrackingUtil {
     fun parseLaunchedActivityForTracking(bundle: Bundle): Boolean
+    fun parseIntentExtrasForTrackingDelivered(bundle: Bundle): Boolean
```
Shahroz16, line 9:
```suggestion
    fun trackMetricEvent(bundle: Bundle, event: MetricEvent): Boolean
```

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingReceiver.kt
```diff
@@ -0,0 +1,25 @@
...
+        // if CustomerIO instance isn't initialized, we cannot process the notification
+        val sdkInstance = context.getSDKInstanceOrNull() ?: return
+
+        val moduleConfig = sdkInstance.diGraph.moduleConfig
+        val pushTrackingUtil = sdkInstance.diGraph.pushTrackingUtil
+
+        // Track delivered event only if auto-tracking is enabled
+        if (moduleConfig.autoTrackPushEvents) {
+            pushTrackingUtil.parseIntentExtrasForTrackingDelivered(extras)
+        }
```
Shahroz16, line 23: trackMetricEvent(extras, MetricEvent.delivered)

messagingpush/src/main/java/io/customer/messagingpush/util/PushTrackingUtil.kt
Shahroz16: Yes, `trackMetricEvent ` already takes in argument for `MetricEvent `, making 2 method that just gives different values doesn't provide much value for me

mrehan27: Agreed, But `parseLaunchedActivityForTracking` is a public method and was the only reason for keeping it there. I can replace all calls with `trackMetricEvent`, but not going to delete `parseLaunchedActivityForTracking` in this PR. I'm going to mark it deprecated only, so we can remove it in future releases.

messagingpush/src/main/java/io/customer/messagingpush/util/PushTrackingUtil.kt
mrehan27: Resolving this as [discussed above](https://github.com/customerio/customerio-android/pull/204#discussion_r1193710660)

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingReceiver.kt
mrehan27: Resolving this as [discussed above](https://github.com/customerio/customerio-android/pull/204#discussion_r1193710660)

messagingpush/src/main/java/io/customer/messagingpush/util/PushTrackingUtil.kt
mrehan27: I updated it to `parseAndTrackMetricEvent` for clarity. Also, I marked it as internal api because I think customer's probably don't really need to use this directly. But if you think differently, let me know so I can remove the annotation from it.

messagingpush/src/main/java/io/customer/messagingpush/util/PushTrackingUtil.kt
```diff
@@ -1,10 +1,18 @@
...
 import io.customer.sdk.repository.TrackRepository
 
 interface PushTrackingUtil {
+    @InternalCustomerIOApi
+    fun parseAndTrackMetricEvent(bundle: Bundle, event: MetricEvent): Boolean
+
+    @Deprecated(
+        "This method is deprecated and will be removed in future releases. Use parseAndTrackMetricEvent instead",
+        ReplaceWith("parseAndTrackMetricEvent(bundle = bundle, event = MetricEvent.opened)")
+    )
```
levibostian, line 15: `PushTrackingUtil` is an internal class, right? Assuming that this is true and the code is not a public facing function, do we need to deprecate the function instead of removing it?

messagingpush/src/main/AndroidManifest.xml
```diff
@@ -22,6 +22,15 @@
                 <action android:name="io.customer.messagingpush.PUSH_ACTION" />
             </intent-filter>
         </receiver>
+        <!-- Broadcast receiver for listening to push events from GCM -->
```
levibostian, line 25: I think we should expand these comments to explain why we decided to add this broadcast receiver to the code base. For future readers of the code wondering why there seems to be 2 implementations of push delivered metrics in the SDK.

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingReceiver.kt
```diff
@@ -0,0 +1,26 @@
...
+
+import android.content.BroadcastReceiver
+import android.content.Context
+import android.content.Intent
+import io.customer.messagingpush.di.moduleConfig
+import io.customer.messagingpush.di.pushTrackingUtil
+import io.customer.messagingpush.extensions.getSDKInstanceOrNull
+import io.customer.sdk.data.request.MetricEvent
+
+class CustomerIOFirebaseMessagingReceiver : BroadcastReceiver() {
```
levibostian, line 11: At this time, this class's only responsibility is delivered push metrics. I think we should add a comment to this class explaining this limited responsibility and why for future readers. 

I am comfortable with relying on this broadcast receiver more in the future once we test it in production and see it's reliable. But until then, keeping it limited adds trust to our code.

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingReceiver.kt
```diff
@@ -0,0 +1,26 @@
...
+        // if CustomerIO instance isn't initialized, we cannot process the notification
+        val sdkInstance = context.getSDKInstanceOrNull() ?: return
+
+        val moduleConfig = sdkInstance.diGraph.moduleConfig
+        val pushTrackingUtil = sdkInstance.diGraph.pushTrackingUtil
+
+        // Track delivered event only if auto-tracking is enabled
+        if (moduleConfig.autoTrackPushEvents) {
+            pushTrackingUtil.parseAndTrackMetricEvent(extras, MetricEvent.delivered)
+        }
```
levibostian, line 24: Can we add automated tests to this code here? Help us confirm that we will only track metrics if the customer has enabled it.

levibostian (approved):

messagingpush/src/main/java/io/customer/messagingpush/util/PushTrackingUtil.kt
mrehan27: I too don't expect customers to use `PushTrackingUtil` directly. But it is a public class currently and customers can use it. We should annotate it as internal probably?

Since the updated changes revert the class to its current stable state, I'm marking this comment as resolved.

messagingpush/src/main/AndroidManifest.xml
mrehan27: Added more comments to it

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingReceiver.kt
mrehan27: Added comments to it

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingReceiver.kt
mrehan27: We unfortunately don't currently support `autoTrackPushEvents` in case of auto-initialization. So this may not work well with wrapper SDKs if the app was killed. I have created a [bug](https://github.com/customerio/issues/issues/10254) for this so this doesn't get ignored. Meanwhile, I'm writing tests to see how can we test both normal and auto-initialization, I'll push them soon.

mrehan27: Looking at the code changes, I think the [bug linked](https://github.com/customerio/issues/issues/10254) is the right ticket to write tests for testing this with auto-initialization. For now, I have added tests that covers metrics tracking for `PushMessageProcessor`.

messagingpush/src/main/java/io/customer/messagingpush/processor/PushMessageProcessorImpl.kt
```diff
@@ -0,0 +1,89 @@
...
+     * that the notification is processed already.
+     *
+     * @param deliveryId unique message id received in push payload.
+     * @return true if the message was processed previously; false otherwise.
+     * Callers should generally process notifications only if false was returned.
+     */
+    @Synchronized
+    @VisibleForTesting
+    internal fun getOrUpdateMessageAlreadyProcessed(deliveryId: String?): Boolean {
+        val logger = CustomerIOShared.instance().diStaticGraph.logger
```
Shahroz16, line 29: maybe initialize it as class level, so other methods can use it too if needed?

messagingpush/src/main/java/io/customer/messagingpush/processor/PushMessageProcessorImpl.kt
```diff
@@ -0,0 +1,89 @@
...
+     * Responsible for storing and updating recent messages in queue atomically.
+     * Once this method is called, the current implementation marks the notification
+     * as processed and future calls for same [deliveryId] will receive true indicating
+     * that the notification is processed already.
+     *
+     * @param deliveryId unique message id received in push payload.
+     * @return true if the message was processed previously; false otherwise.
+     * Callers should generally process notifications only if false was returned.
+     */
+    @Synchronized
```
Shahroz16, line 26: We talked about it before, `ArrayDeque<String>` is not thread-safe, hence we need to add this keyword. We can probably, go for `ConcurrentLinkedDeque` which is thread-safe and then wont even have to use this keyword.  

But any reason, we stuck to `ArrayDeque` ?

Shahroz16: Also, there is `LinkedBlockingDeque` which could be even better for our use case, thread-safe but allows you to set a maximum size. When you try to add an element while the queue is full, it will automatically remove elements from the other end.

mrehan27: I'm okay with even using a simple list that fulfills the purpose here, but `Synchronized` annotations should stay there IMO as we are checking and inserting elements at different lines. And without locking, we may run into race condition resulting in multiple `delivered` metrics.

`LinkedBlockingDeque` looks good to me, I'll take look at other options as well and update this.

Shahroz16: `Synchronized` and `ArrayDeque` does the same thing as `LinkedBlockingDeque` or `ConcurrentLinkedDeque` without `Synchronized` annotation. 

When multiple threads access a thread-safe collection in Java, the collection will be synchronized, which means that only one thread can access it at a time. This prevents race conditions, which are errors that can occur when multiple threads are trying to access the same data at the same time.

Shahroz16 (approved):

messagingpush/src/main/java/io/customer/messagingpush/processor/PushMessageProcessorImpl.kt
mrehan27: Updated

messagingpush/src/main/java/io/customer/messagingpush/processor/PushMessageProcessorImpl.kt
```diff
@@ -0,0 +1,88 @@
...
+
+    override fun processRemoteMessageDeliveredMetrics(deliveryId: String, deliveryToken: String) {
+        if (!getOrUpdateMessageAlreadyProcessed(deliveryId = deliveryId)) {
+            // We only track delivered metrics here for Firebase messages as the caller
+            // processes and displays the notification already and FCM guarantees to send
+            // callbacks only once for each notification, so duplication is not possible within
+            // Firebase message receivers
+            trackDeliveredMetrics(deliveryId, deliveryToken)
+        }
+    }
```
levibostian, line 76: The implementation of these 2 functions is almost identical except `processGCMMessageIntent` parses an `Intent` while the other doesn't. 

When you look at the caller of `processRemoteMessageDeliveredMetrics` in `CustomerIOPushNotificationHandler.kt`, an `Intent` is parsed before calling `processRemoteMessageDeliveredMetrics`. 

Can we refactor the code to avoid this code duplication?

messagingpush/src/main/java/io/customer/messagingpush/processor/PushMessageProcessorImpl.kt
levibostian: Unresolving this conversation because I would like to ask for a conclusion to this conversation. 

In the PR, it looks like `ArrayDequeu` and `@Sychronized` is implemented. I would like to ask the reason for this solution instead of using a different data type that was suggested in this conversation.

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingReceiver.kt
```diff
@@ -0,0 +1,34 @@
...
+class CustomerIOFirebaseMessagingReceiver : BroadcastReceiver() {
+
+    override fun onReceive(context: Context, intent: Intent) {
+        val extras = intent.extras
+        // Ignore event if no data was received in extras
+        if (extras == null || extras.isEmpty) return
+        // If CustomerIO instance isn't initialized, we cannot process the notification
+        val sdkInstance = context.getSDKInstanceOrNull() ?: return
+
+        sdkInstance.diGraph.pushMessageProcessor.processGCMMessageIntent(intent = intent)
```
levibostian, line 32: Why does the class name contain `Firebase` in the name but it's calling a function that uses `GCM` in the name?

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingReceiver.kt
levibostian: I have a separate comment from the one above. 

Why does the class name contain `Firebase` but the comment and the code indicates that this is a class only used by GCM? I am assuming this was a typo?

messagingpush/src/main/java/io/customer/messagingpush/processor/PushMessageProcessorImpl.kt
```diff
@@ -0,0 +1,88 @@
...
+
+import android.content.Intent
+import androidx.annotation.VisibleForTesting
+import io.customer.messagingpush.MessagingPushModuleConfig
+import io.customer.messagingpush.util.PushTrackingUtil
+import io.customer.sdk.data.request.MetricEvent
+import io.customer.sdk.repository.TrackRepository
+import io.customer.sdk.util.Logger
+
+internal class PushMessageProcessorImpl(
```
levibostian, line 11: I am not sure if I find this class very useful at the moment. It seems to contain mostly Queue logic for the `PushMessageProcessor.recentMessagesQueue` data structure and a small bit of code beyond that. 

I am leaning towards removing it. 

For the Queue data structure, if we were to [use a different data type for the `PushMessageProcessor.recentMessagesQueue` data structure](https://github.com/customerio/customerio-android/pull/204/files#r1204632825) or we create our own thread-safe, maximum size, Queue data structure class. This would remove a lot of the need for the `getOrUpdateMessageAlreadyProcessed` function. 

As for the rest of the code that is specific to FCM and GCM push processing, I would rather have the code contained inside of the existing FCM Service code and the GCM Broadcast Receiver. 

Having a separate class that only contains *mostly* this core code: 

```kt
if (!deliveryMetricsQueue.processedMessageAlready(deliveryId = deliveryId)) {
     trackRepository.trackMetric(...)
  }
}
```

Seems not very useful to me at the moment.

messagingpush/src/main/java/io/customer/messagingpush/processor/PushMessageProcessorImpl.kt
Shahroz16: Me and Rehan discussed this, and the usage of `@Synchronized` is valid in this case, because the critical section (the method in this case) has multiple operations that depend on each other and need to be performed atomically.

Let's dry-run it with multiple threads:

`Thread 1` enters the method and checks if the `deliveryId` is in the `recentMessagesQueue`. It finds that the `deliveryId` is not present. 

`Thread 1` moves on to the else block, and starts adding the `deliveryId` to the `queue`.

At this point, if `Thread 2` is allowed to enter the method, it could also check and find that the deliveryId is not in the recentMessagesQueue (because Thread 1 has not yet finished adding it). This could result in the deliveryId being added to the recentMessagesQueue twice. So, Thread 2 must wait until Thread 1 has completely finished the method.

With the @Synchronized keyword, this problem is avoided because only one thread can execute the method at a time. This is an example of where synchronization is necessary to ensure atomicity of operations.

Regarding the other part, `ArrayDeque` is fine as long as no other method accesses it, which is the case for now. Because the only access point to this array is from a method that is synchronized.

messagingpush/src/main/java/io/customer/messagingpush/processor/PushMessageProcessorImpl.kt
mrehan27: They look slightly identical at the moment but they both have different purpose. They should remain 2 separate methods because:

`processGCMMessageIntent` is the called directly from GCM receiver and is planned to display notification as well for wrapper SDKs in future

`processRemoteMessageDeliveredMetrics` was initially designed to receive `RemoteMessage` and parse delivery parameters by itself to make it easier for the caller. But since `deliveryId` and `deliveryToken` are being used by `CustomerIOPushNotificationHandler` to determine whether it is a CIO push notification or not, and also to generate push payload, moving them out of `CustomerIOPushNotificationHandler` was resulting in even more confusions.

So, out of all the solutions I tried before, the current structure looked the best one to me. I would stick to current changes for now. We can then gradually move more changes to processor and utilize it better.

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingReceiver.kt
mrehan27: I missed it renaming it 🤦🏻  Renamed it to `CustomerIOCloudMessagingReceiver`

messagingpush/src/main/java/io/customer/messagingpush/processor/PushMessageProcessorImpl.kt
mrehan27: Our current queue structure doesn't have any support for preventing duplications, the processor class not only takes the responsibility for avoiding duplication but also makes it easier for other classes. With the introduction of GCM receiver, we'll now be doing several push related operations at multiple places. So there has to be one class which holds all push related operations at a single place and avoid code repetition. The processor class takes this responsibility and will be used for adding more operations in future as well.

messagingpush/src/main/java/io/customer/messagingpush/processor/PushMessageProcessorImpl.kt
levibostian: Ok. Glad to hear you both talked about it. 

The code as-is in this PR is thread safe, but I do not believe that the code helps prevent miss-use of the `ArrayDequeue` in the future. It would be easy for someone in the future to write code that uses this object and forget to make it thread safe. 

One idea I have to help prevent this is making a thread safe version of `ArrayDequeue`? A class that is a wrapper around a private `ArrayDequeue` object where all access is via `@Synchronized` functions? 

```kt
class ThreadSafeArrayDequeue(maxSize: Int) {

  private dequeue: ArrayDequeue

  @Synchronized 
  get() {}

  @Synchronized
  set() {}
}
```

Sorry the code above is very short. I am writing this before heading out for the weekend. I wanted to try and simply illustrate the idea of a wrapper with `@Synchronized` functions.

mrehan27: Thanks Levi. I think using one of the structures suggested above by Shahroz should address your concern as well. I've replaced `ArrayDeque` with `LinkedBlockingDeque` which is thread safe and should work well for us.

PR #206, opened 2023-05-16
chore: java sample universal scheme deep links

mrehan27:

### Changes

- Deep links support for app scheme URLs
  - `java-sample://settings`
  - `java-sample://events/custom`
  - `java-sample://attributes/device`
  - `java-sample://attributes/profile`
- Deep links support for universal scheme URLs
  - https://www.java-sample.com/settings
  - https://www.java-sample.com/events/custom
  - https://www.java-sample.com/attributes/device
  - https://www.java-sample.com/attributes/profile
- Fixes and improvements in deep link handling
- Fixed user agent width and text alignment

### DIFF
--- a/samples/java_layout/build.gradle
+++ b/samples/java_layout/build.gradle
@@ -20,7 +20,7 @@ android {
         versionCode 1
         versionName "1.0"
 
-        String universalDeepLinkHost = "ciosample.page.link"
+        String universalDeepLinkHost = "www.java-sample.com"
 
         buildConfigField "String", "UNIVERSAL_DEEP_LINK_HOST", "\"${universalDeepLinkHost}\""
         buildConfigField "String", "API_KEY", "\"${localProperties["apiKey"]}\""
--- a/samples/java_layout/build.gradle
+++ b/samples/java_layout/build.gradle
@@ -20,8 +20,13 @@ android {
         versionCode 1
         versionName "1.0"
 
+        String universalDeepLinkHost = "ciosample.page.link"
+
+        buildConfigField "String", "UNIVERSAL_DEEP_LINK_HOST", "\"${universalDeepLinkHost}\""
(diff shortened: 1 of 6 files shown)


Shahroz16 (approved):

samples/java_layout/build.gradle
```diff
@@ -20,8 +20,13 @@ android {
         versionCode 1
         versionName "1.0"
 
+        String universalDeepLinkHost = "www.java-sample.com"
+
+        buildConfigField "String", "UNIVERSAL_DEEP_LINK_HOST", "\"${universalDeepLinkHost}\""
```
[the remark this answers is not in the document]
mrehan27: Only one benefit of this, and that's whenever we need to change host url, updating one variable will update it both for code files and the `manifest`.

samples/java_layout/build.gradle
```diff
@@ -20,8 +20,13 @@ android {
         versionCode 1
         versionName "1.0"
 
+        String universalDeepLinkHost = "www.java-sample.com"
```
levibostian, line 23: Since we do not own this domain name, Android will not be able to do auto verification for the app links and therefore, I expect that the web browser will open deep link URL or show a share dialog before opening our app? 

With the goal being this app should allow QA to test that our SDK works with Android app links, it's important that the URLs used by QA can be auto verifiable by Android and skip opening the browser and showing a dialog when opening a URL. 

Details on a domain name that is auto verifiable by our apps [is mentioned in slack](https://customerio.slack.com/archives/C01QYDKD0SU/p1683807088350899).

Shahroz16: The firebase dynamic link is a good solution, but this would work too. In Android, you can go to App settings and there you can find these links there once you allow them from settings, next time whenever that link is opened it will open your app rather than the browser, so QA should be able to test it.

PR #208, opened 2023-05-17
chore: extract common gradle props for sample apps

mrehan27:
Helps https://github.com/customerio/issues/issues/9844, https://github.com/customerio/issues/issues/9845

### Changes

- `sample-app.gradle` file to hold common gradle properties for sample apps
- gradle file updates in sample apps to apply from newly added gradle file

### DIFF
--- a/samples/sample-app.gradle
+++ b/samples/sample-app.gradle
@@ -1,0 +1,31 @@
+Properties localProperties = new Properties()
+File localPropertiesFile = rootProject.file('local.properties')
+if (localPropertiesFile.exists()) {
+    localProperties.load(new FileInputStream(localPropertiesFile))
+}
+
+android {
+    defaultConfig {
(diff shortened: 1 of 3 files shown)


samples/sample-app.gradle
```diff
@@ -0,0 +1,33 @@
+Properties localProperties = new Properties()
```
Shahroz16, line 1: can we add the common `plugin` as well?

Shahroz16 (approved):
Looks good, added one questions

samples/sample-app.gradle
mrehan27: I tried it too but received the following error. Looks like gradle doesn't allow `plugin` block in this file 😞 

```
Only Project and Settings build scripts can contain plugins {} blocks
```

PR #211, opened 2023-05-22
ci: build all sample apps on each pr and main branch merge

levibostian:
Part of ticket: https://github.com/customerio/issues/issues/9028

Sets up the CI to iterate through all of the sample apps in this repository, builds them, and [comments on PRs](https://github.com/customerio/customerio-android/pull/211#issuecomment-1557785673) with useful information related to the build.


.github/workflows/build-sample-apps.yml
```diff
@@ -0,0 +1,126 @@
+name: Build sample apps
```
levibostian, line 1: This file was taken from the iOS SDK and modified to build Android apps. 

There is a possibility that we can re-use the contents in this file with all of the SDK wrapper projects but I am not sure yet. That is my next task to determine with this ticket.

.github/workflows/build-sample-apps.yml
```diff
@@ -0,0 +1,126 @@
...
+        issue-number: ${{ github.event.pull_request.number }}
+        body: |
+          <!-- sample app builds --> 
+          # Sample app builds 📱
+          
+          Below you will find the list of the latest versions of the sample apps. It's recommended to always download the latest builds of the sample apps to accurately test the pull request. 
+
+          ---
+          ${{ steps.build.outputs.build-log }}
+        edit-mode: replace # replace the existing comment with new content since we are creating new builds
```
Shahroz16, line 45: Not sure what is the best way, but I feel like updating comments seems too big of a hassle considering we might be adding more information to it?.

What do you think about consolidating the build information into a single comment? This will require adding a step to create an artifact that stores the build information and then reading from that artifact to generate the PR comment at the end. 

so instead of incremental changes, we just do one comment at the end? 

In future, if we have to add something, we don't go about changing the comment again, but rather add something to that artifact. 

something like, 
```
const fs = require('fs');
          const buildInfos = fs.readdirSync('.').filter(name => name.startsWith('build-info'));
```

Shahroz16 (approved):

samples/fastlane/Fastfile
```diff
@@ -0,0 +1,58 @@
+# There are some re-usable functions in other Fastfile files in the org we can re-use. 
+import_from_git(
+  url: "https://github.com/customerio/customerio-ios.git", 
+  branch: "levi/fastlane-github-reusable",
```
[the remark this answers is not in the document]
levibostian: We can change it [after this PR gets merged](https://github.com/customerio/customerio-ios/pull/327)

.github/workflows/build-sample-apps.yml
levibostian: Sorry, I don't understand your question or concern? 

The way this workflow is setup today is: 
* When a new commit is pushed to a PR, delete the old PR comment and replace with a new blank comment saying that builds are starting. 
* After each build is created, they separately update the PR comment with a success or failed message for that 1 build. 
* This PR comment is only meant to be for sample apps compiling. If we want to automate more then building sample apps in the future, we can create separate PR comments for that purpose.

Shahroz16: It's not a blocker, just observation. My concern was about the part 

> After each build is created, they separately update the PR comment with a success or failed message for that 1 build.

I believe we can just add a comment once we have a success or failure, or we can keep storing its state in some artifacts and once both apps processing is done then make a comment. 

> we can create separate PR comments for that purpose.

I think if we keep on adding/updating comments like these, they might lose value. So I prefer fewer automated comments on PR. So we know what information each comment is going to have. I think that way it would be easier to maintain. 

Like I said earlier, these are just my thought no actions step need to come out of this, if more folks on the team doesnt feel the same.

PR #213, opened 2023-05-29
fix: unique name for di graph singletons

mrehan27:
closes: https://github.com/customerio/issues/issues/10311

### Changes

- Uses fully qualified name for `DiGraph` keys instead of simple name to avoid duplication


Shahroz16 (approved):
This a good short-term fix, but for the future maybe we should consider using unique identifiers that are not affected by obfuscation, such as string constants or enumerated types (enums). This would avoid issues with obfuscation and make your DI setup more robust and easier to use.

mrehan27:
@Shahroz16 Agreed. I too initially tried with similar options first, but eventually decided to go with short-term fix to prevent any potential bugs first. We can discuss this more internally and see how to improve this in future.

levibostian (approved):

PR #217, opened 2023-06-09
chore: java sample fixes and improvements

mrehan27:

### Changes

- Aligning app design/behavior with other sample apps and squad discussions
- Updated default config to match with SDK
- Show push prompt button to request notification permission
- Validation updates
- Updates in random value generation
- Build info updates
- App icon updates
- Material3 support
- Bug fixes and behavior improvements
- Deep links support for app scheme URLs
  - `java-sample://login`
  - `java-sample://dashboard`
  - `java-sample://settings`
  - `java-sample://settings?site_id=xxx&api_key=yyy`
- Deep links support for universal scheme URLs
  - https://www.java-sample.com/login
  - https://www.java-sample.com/dashboard
  - https://www.java-sample.com/settings
  - https://www.java-sample.com/settings?site_id=xxx&api_key=yyy


samples/java_layout/src/main/java/io/customer/android/sample/java_layout/data/model/CustomerIOSDKConfig.java
```diff
@@ -31,13 +30,12 @@ private static class Keys {
...
-                null,
-                null,
-                null,
-                null);
+                "https://track-sdk.customer.io/",
+                30.0,
+                10,
+                true,
+                true,
+                true);
```
levibostian, line 38: I am not the biggest fan of our sample apps providing it's own set of SDK default values. I was expecting to see our sample apps only override a SDK configuration when the settings screen UI was modified to indicate it wants to change 1 of the SDK values. 

We have other sample apps that also do this (app providing it's own SDK default values to override the SDK's default values) so I don't see this as a blocker. [I documented this in this ticket](https://github.com/customerio/issues/issues/10243) for the squad to discuss in the future if we wish.

levibostian (approved):

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/data/model/CustomerIOSDKConfig.java
mrehan27: > sample apps only override a SDK configuration when the settings screen UI was modified

I am also in favor of this, as this allows us to test default behavior of the SDK purely.

But based on [this discussion](https://customerio.slack.com/archives/C05C2HF5VB2/p1686159852015699?thread_ts=1686145893.748049&cid=C05C2HF5VB2), the decision was to provide default values.

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/ui/login/LoginActivity.java
```diff
@@ -68,7 +65,7 @@ private void setupViews() {
         });
         binding.randomLoginButton.setOnClickListener(view -> {
             Randoms randoms = new Randoms();
-            authViewModel.setLoggedInUser(new User(randoms.email(), randoms.displayName(), true));
+            authViewModel.setLoggedInUser(new User(randoms.email(), "", true));
```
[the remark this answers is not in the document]
mrehan27: From Figma

> First name is optional. Email is required to log in.

And the [discussion on slack](https://customerio.slack.com/archives/C05C2HF5VB2/p1686161223135229?thread_ts=1686148582.678599&cid=C05C2HF5VB2) suggests the name should be empty.

Shahroz16 (approved):

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/data/model/CustomerIOSDKConfig.java
```diff
@@ -167,4 +154,23 @@ public Boolean isDeviceAttributesTrackingEnabled() {
     public Boolean isDebugModeEnabled() {
         return debugModeEnabled;
     }
+
+    /**
+     * Features by default are nullable to help differentiate between default/null values and
+     * values set by user.
+     * Unwrapping nullable values here for ease of use by keeping single source of truth for whole
+     * sample app.
+     */
```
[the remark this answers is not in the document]
mrehan27: Yes we do, but they are still nullable, so values like `backgroundQueueMinNumberOfTasks` can be set to `null` to use SDK default values.

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/data/model/CustomerIOSDKConfig.java
levibostian: Thanks for linking the conversation. Very helpful. 

I interrupt that conversation in Slack as, "Populate the UI fields of the settings screen by reading the values from the SDK's config object" instead of, "Populate the UI fields of the settings screen by the app providing it's own SDK config object". 

Can this be done in the Java app? Reading values from the SDK's currently set config object? 

This is how I do it in the iOS Swift UI sample app. I think in order to do that, I had to add to the iOS SDK the ability to get the SDK config object to the public API: `CustomerIO.instance.config`.

mrehan27: Yeah, that would be easier to maintain. We do have `AnalyticsConstants` for default values but they are not currently accessible in Java files due to how they are structured. If that is not being used in wrappers directly, we can tweak it and make accessible to Java files. But to do the same in wrappers, we might need to make more changes.

For `CustomerIO.instance.config`, it looks like only accessible after the SDK has been initialized? If this is correct then it might be difficult to use it for initial default values?

I'm okay with whatever we decide for this, but I would like to make it consistent for wrappers as well. Let's move this discussion to slack and finalize it for wrappers as well.

PR #219, opened 2023-06-12
chore: sample apps release build signing

mrehan27:
### Background

- Our sample app release builds currently fail on installation because release build are currently unsigned

### Changes

- Add debug signing keystore file to sample apps directory
- Use newly add debug signing for sample app debug and release builds


samples/sample-app.gradle
```diff
@@ -9,6 +9,24 @@ android {
...
+            storeFile file("${rootDir}/samples/debug.keystore")
+            storePassword 'android'
+            keyAlias 'androiddebugkey'
+            keyPassword 'android'
+        }
+    }
+    buildTypes {
+        debug {
+            signingConfig signingConfigs.debug
+        }
```
levibostian, line 25: Is there a reason we want to sign debug builds of the sample apps? I was expecting to only see code signing used for release builds.

mrehan27: No, not really a specific reason. I just specified to avoid confusions as the default debug signingConfigs are from local machine, and new one we copied are also debug ones as we discussed on slack earlier. I'm okay with removing this for debug builds, but in that case, I'll suggest that we should rename the newly copied signing to release (or any other name than debug) to avoid any confusion.

levibostian: I agree with everything said here. I think it's OK to leave because debug code signing files are commonly known amongst Android devs and I don't see the configuration as it is in this PR causing confusion now or in the future.

PR #222, opened 2023-06-20
chore: java sample validations

mrehan27:

### Changes

- Settings screen updated to not allow any empty value
- Updated text field errors with generic error messages

### DIFF
--- a/samples/java_layout/src/main/java/io/customer/android/sample/java_layout/ui/settings/SettingsActivity.java
+++ b/samples/java_layout/src/main/java/io/customer/android/sample/java_layout/ui/settings/SettingsActivity.java
@@ -100,58 +100,7 @@ public class SettingsActivity extends BaseActivity<ActivitySettingsBinding> {
             ClipData clip = ClipData.newPlainText(getString(R.string.device_token), deviceToken);
             clipboard.setPrimaryClip(clip);
         });
-        binding.saveButton.setOnClickListener(view -> {
-            String siteId = ViewUtils.getTextTrimmed(binding.siteIdTextInput);
-            String apiKey = ViewUtils.getTextTrimmed(binding.apiKeyTextInput);
-            String trackingURL = ViewUtils.getTextTrimmed(binding.trackingUrlTextInput);
-
-            boolean isFormValid = true;
-            if (!isTrackingURLValid(trackingURL)) {
-                ViewUtils.setError(binding.trackingUrlInputLayout, getString(R.string.error_tracking_url));
-                isFormValid = false;
-            } else {
-                ViewUtils.setError(binding.trackingUrlInputLayout, null);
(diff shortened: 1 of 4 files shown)


samples/java_layout/src/main/java/io/customer/android/sample/java_layout/ui/settings/SettingsActivity.java
```diff
@@ -192,4 +141,75 @@ private void updateIOWithConfig(@NonNull CustomerIOSDKConfig config) {
...
+                    .subscribeOn(Schedulers.io())
+                    .observeOn(AndroidSchedulers.mainThread())
+                    .subscribe(preferences -> {
+                        binding.progressIndicator.hide();
+                        Toast.makeText(this, R.string.settings_save_msg, Toast.LENGTH_SHORT).show();
+                        OSUtils.restartApp();
+                    });
+            disposables.add(disposable);
+        }
+    }
```
Shahroz16, line 214: I think there is a lot of duplicated code, can we combine them and make it easier? 

something like this? 


```
private boolean isValidForm() {
    boolean isFormValid = true;

    isFormValid &= validateAndSetError(binding.trackingUrlInputLayout, this::isTrackingURLValid, R.string.error_tracking_url);
    isFormValid &= validateAndSetError(binding.siteIdInputLayout, TextUtils::isEmpty, R.string.error_text_input_field_blank);
    isFormValid &= validateAndSetError(binding.apiKeyInputLayout, TextUtils::isEmpty, R.string.error_text_input_field_blank);
    isFormValid &= validateAndSetError(binding.bqDelayInputLayout, TextUtils::isEmpty, R.string.error_text_input_field_blank);
    isFormValid &= validateAndSetError(binding.bqTasksInputLayout, TextUtils::isEmpty, R.string.error_text_input_field_blank);

    return isFormValid;
}

private boolean validateAndSetError(TextInputLayout inputLayout, Predicate<String> predicate, int errorStringId) {
    String input = ViewUtils.getTextTrimmed(inputLayout.getEditText());

    if (predicate.test(input)) {
        ViewUtils.setError(inputLayout, getString(errorStringId));
        return false;
    } else {
        ViewUtils.setError(inputLayout, null);
        return true;
    }
}
```

levibostian (approved):

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/ui/settings/SettingsActivity.java
mrehan27: I had to make [some tweaks](https://github.com/customerio/customerio-android/pull/222/commits/bcc20b1f7d861dee3cdd243becb9f6f817f2c974), pushed after following similar approach

PR #225, opened 2023-06-23
chore: java sample appium tags

mrehan27:
part of: https://github.com/customerio/issues/issues/9845
helps: https://github.com/customerio/issues/issues/9844

### Changes

- Sets content description for views which is common practice for automation tests, in case we need to changes this later because of some technical limitation, we can do it in a separate PR
- Added `appium.xml` containing tags for all views
- Added `appium.xml` as resource for all sample app so it can be reused

### DIFF
--- a/samples/sample-app.gradle
+++ b/samples/sample-app.gradle
@@ -33,6 +33,13 @@ android {
         // For snapshot builds
         maven { url 'https://s01.oss.sonatype.org/content/repositories/snapshots/' }
     }
+    sourceSets {
+        main {
+            res.srcDirs += [
+                    "${rootDir}/samples/resources",
+            ]
+        }
+    }
 }
 
 dependencies {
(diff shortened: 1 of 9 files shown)


samples/sample-app.gradle
```diff
@@ -33,6 +33,13 @@ android {
         // For snapshot builds
         maven { url 'https://s01.oss.sonatype.org/content/repositories/snapshots/' }
     }
+    sourceSets {
+        main {
+            res.srcDirs += [
+                    "${rootDir}/samples/resources",
+            ]
+        }
+    }
```
mrehan27, line 42: @Shahroz16 I added this to include `appium.xml` for all sample apps so this can be reused. In case you think it doesn't help, please let me know so I can remove it and include for Java sample only.

Shahroz16 (approved):
Looks good and thanks for the shared strings resources.

PR #230, opened 2023-06-27
chore: java sample fixes and improvements

mrehan27:
part of: https://github.com/customerio/issues/issues/9845

### Changes

- Save settings automatically after restore
- Background Queue settings validations
- Event screen attribute fix
- Validation updates
- Improved support for notification permission if disabled on Android 12 and lower

### DIFF
--- a/samples/java_layout/src/main/res/values/strings.xml
+++ b/samples/java_layout/src/main/res/values/strings.xml
@@ -58,4 +58,5 @@
     <string name="filter_view_universal_link">Java Sample Universal Scheme Links</string>
     <string name="error_text_input_field_blank">This field cannot be blank</string>
     <string name="error_text_input_field_empty">This field cannot be empty</string>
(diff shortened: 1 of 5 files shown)


samples/java_layout/src/main/res/values/strings.xml
```diff
@@ -58,4 +58,5 @@
     <string name="filter_view_universal_link">Java Sample Universal Scheme Links</string>
     <string name="error_text_input_field_blank">This field cannot be blank</string>
     <string name="error_text_input_field_empty">This field cannot be empty</string>
+    <string name="error_number_input_field_small">The value must be greater than %1$s</string>
```
Shahroz16, line 61: Shouldn't it be greater than 0? 1 should be allowed.?

mrehan27: Updated error message

levibostian (approved):

Shahroz16 (approved):

PR #233, opened 2023-07-03
ci: run android instrumentation tests on CI

Shahroz16:


.github/workflows/instrumentation-tests.yml
```diff
@@ -0,0 +1,50 @@
+name: Instrumentation tests
+
+on: [push, pull_request]
+
+jobs:
+  instrumentation-test:
```
levibostian, line 6: I suggest moving this job into the existing [test workflow](https://github.com/customerio/customerio-android/blob/main/.github/workflows/test.yml) to keep all tests together.

.github/workflows/instrumentation-tests.yml
```diff
@@ -0,0 +1,50 @@
...
+    - name: Setup Android SDK
+      uses: android-actions/setup-android@v2
+
+    - name: Gradle cache
+      uses: gradle/gradle-build-action@v2
+
+    - name: Start emulator
+      uses: reactivecircus/android-emulator-runner@v2
+      with:
+        # Configuration options for the emulator for the best optimization.
```
levibostian, line 32: Can you link to the place you found these configuration settings? So we can try and keep them up-to-date?

levibostian (approved):

.github/workflows/instrumentation-tests.yml
Shahroz16: I thought about this at first, but instrumentation tests take a lot longer than unit tests, and at times you have to run the test again because the AVD didn't run or was unknown other problems can happen with tests that involve an emulator running on the fly. 

If we tied both of them together, will they slow us down? we would need to re-run both tests on failure and also, will have to probably wait for long to know even if the problem was with unit tests etc.

.github/workflows/instrumentation-tests.yml
Shahroz16: Configuration can be found action document, do you prefer me adding the link to [reactivecircus/android-emulator-runner@v2](https://github.com/ReactiveCircus/android-emulator-runner#configurations) github?

.github/workflows/instrumentation-tests.yml
levibostian: Jobs run in parallel by default and that is what I am suggesting - moving the entire job into the test workflow file. 

This will unit tests and instrumentation tests run in parallel and will allow you to re-run failed tests, only. You do not need to re-run both.

.github/workflows/instrumentation-tests.yml
levibostian: I am trying to find where you found this configuration list *for the best optimization*. From [the docs](https://github.com/ReactiveCircus/android-emulator-runner#configurations), I see all the configuration options with default values but don't see where you came up with the configuration options you included. That's what I am looking for.

Shahroz16: Just read the description, took out the ones that aren't needed, and ran the workflow multiple times to test others.

mrehan27 (approved):

PR #236, opened 2023-07-07
ci: sample apps using separate workspaces

levibostian:
We need each sample app to work with different CIO workspaces. 

See [status of PR](https://github.com/customerio/customerio-android/pull/236#issuecomment-1626004989). 

Made this before my EOD to get early reviews and ideas on how to fix proxyman config.

### DIFF
--- a/.github/workflows/build-sample-apps.yml
+++ b/.github/workflows/build-sample-apps.yml
@@ -59,10 +59,12 @@ jobs:
         - "kotlin_compose"
         include: # Add additional variables to each sample app build: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrixinclude
         - sample-app: "java_layout"
-          cio-workspace: "java"
+          cio-siteid-secret-key: "CUSTOMERIO_JAVA_WORKSPACE_SITE_ID"
+          cio-apikey-secret-key: "CUSTOMERIO_JAVA_WORKSPACE_API_KEY"
         - sample-app: "kotlin_compose"
-          cio-workspace: "kotlin"
-
+          cio-siteid-secret-key: "CUSTOMERIO_KOTLIN_WORKSPACE_SITE_ID"
+          cio-apikey-secret-key: "CUSTOMERIO_KOTLIN_WORKSPACE_API_KEY"
+          
     runs-on: ubuntu-latest
     name: Building app...${{ matrix.sample-app }}
     steps:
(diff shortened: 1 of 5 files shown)


levibostian:
In this PR, I tried to get proxyman proxying setup for sample apps to make verifying the correct workspace faster. 

Debug builds work in proxyman but CI builds do not. 

I would prefer if proxyman works but debug and release builds made on the CI but it might not be possible? 

If it's a fast fix, I would like to have that put into this PR. Else, remove the proxyman config and let's ship only the CI build sample app GH Action workflow.

.github/workflows/build-sample-apps.yml
```diff
@@ -83,11 +88,19 @@ jobs:
...
-        echo "apiKey=${{ secrets.CIO_API_KEY }}" >> "samples/local.properties"          
+        echo "siteId=${{ secrets.CUSTOMERIO_JAVA_WORKSPACE_SITE_ID }}" >> "samples/local.properties"
+        echo "apiKey=${{ secrets.CUSTOMERIO_JAVA_WORKSPACE_API_KEY }}" >> "samples/local.properties" 
+
+    - name: Setup Customer.io workspace credentials for Kotlin
+      if: ${{ matrix.cio-workspace == 'kotlin' }}
+      run: |
+        touch "samples/local.properties"
+        echo "siteId=${{ secrets.CUSTOMERIO_KOTLIN_WORKSPACE_SITE_ID }}" >> "samples/local.properties"
+        echo "apiKey=${{ secrets.CUSTOMERIO_KOTLIN_WORKSPACE_API_KEY }}" >> "samples/local.properties"
```
Shahroz16, line 103: can we add `samples/local.properties` into an `env` variable?
It would make it much easier to change in all places and the chances of messing up the spelling on one of the occurrences gets lower as well.

Shahroz16:
```
env:
      LOCAL_PROP_FILE: "samples/local.properties"
```

Shahroz16 (approved):

.github/workflows/build-sample-apps.yml
levibostian: We did that strategy in RH and I strongly recommend we don't do that. That strategy required versioning the secret file and gets complex quickly. Also, there is an added risk of credentials getting leaked when you do not save the credential values as secrets. Another reason to do this approach instead. 

However, I do not like the copy/paste nature of the workflow as-is. 

I [found a way](https://github.com/customerio/customerio-android/pull/236/commits/c5a77546f69174ff12ae3b62d89ba063d2c9eaf4) to dynamically reference secrets which allows the workflow to be dynamic with it's generating of the local.properties file.

Issue #245, opened 2023-08-20
Crash inapp message 


arseneHuot:
**SDK version:** 3.6.3

**Environment:**Production

**Are logs available?** 
Fatal Exception: java.util.ConcurrentModificationException:
       at java.util.ArrayList$Itr.next(ArrayList.java:860)
       at io.customer.messaginginapp.gist.presentation.GistSdk.handleGistError$messaginginapp_release(GistSdk.kt:228)
       at io.customer.messaginginapp.gist.presentation.GistView.error(GistView.kt:138)
       at io.customer.messaginginapp.gist.presentation.engine.EngineWebView$setupTimeout$1.run(EngineWebView.kt:98)
       at java.util.TimerThread.mainLoop(Timer.java:562)
       at java.util.TimerThread.run(Timer.java:512)

**Describe the bug**
Crash but user without any inapp message (not deployed on our automations)

**To Reproduce**
no idea

**Expected behavior**
no cash

mrehan27:
Thank you for reaching out and sharing the details. We apologize for the inconvenience caused by the crash. Based on the stack-trace you've provided, we're actively working on a fix. You can expect this issue to be resolved in an upcoming release. We appreciate your patience and cooperation.

arseneHuot:
Great ! 
No problem that was only crashing for few users but we like inapp messages so we prefer to keep it and have 0 crash :)

mrehan27:
Thank you for your understanding and continued support! I'm excited to inform you that the issue has been fixed in version `3.6.5`. Please don't hesitate to reach out if there's anything else we can assist you with.

PR #246, opened 2023-08-21
fix: prevent concurrency issues in in-app listeners

mrehan27:
fixes: https://github.com/customerio/customerio-android/issues/245

### Changes

- Replaced for each loop with iterator to prevent concurrency issues

### DIFF
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistSdk.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistSdk.kt
@@ -170,7 +170,7 @@ object GistSdk : Application.ActivityLifecycleCallbacks {
     }
 
     fun clearListeners() {
-        iterateOnListeners { iterator ->
+        forEachListener { iterator ->
             val listener = iterator.next()
             val listenerPackageName = listener.javaClass.`package`?.name
             if (!listenerPackageName.toString().startsWith("build.gist.")) {
@@ -214,28 +214,28 @@ object GistSdk : Application.ActivityLifecycleCallbacks {
     }
 
     internal fun handleGistLoaded(message: Message) {
-        iterateOnListeners { iterator ->
+        forEachListener { iterator ->
             val listener = iterator.next()
             listener.onMessageShown(message)
         }
     }
 
     internal fun handleGistClosed(message: Message) {
-        iterateOnListeners { iterator ->
+        forEachListener { iterator ->
             val listener = iterator.next()
             listener.onMessageDismissed(message)
         }
     }
 
     internal fun handleGistError(message: Message) {
-        iterateOnListeners { iterator ->
+        forEachListener { iterator ->
             val listener = iterator.next()
             listener.onError(message)
         }
     }
 
     internal fun handleEmbedMessage(message: Message, elementId: String) {
-        iterateOnListeners { iterator ->
+        forEachListener { iterator ->
             val listener = iterator.next()
             listener.embedMessage(message, elementId)
         }
@@ -247,7 +247,7 @@ object GistSdk : Application.ActivityLifecycleCallbacks {
         action: String,
         name: String
     ) {
-        iterateOnListeners { iterator ->
+        forEachListener { iterator ->
             val listener = iterator.next()
             listener.onAction(message, currentRoute, action, name)
         }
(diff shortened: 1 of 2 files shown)


messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistSdk.kt
```diff
@@ -251,6 +257,14 @@ object GistSdk : Application.ActivityLifecycleCallbacks {
         return sharedPreferences.getString(SHARED_PREFERENCES_USER_TOKEN_KEY, null)
     }
 
+    private fun forEachListener(action: (MutableIterator<GistListener>) -> Unit) {
```
Shahroz16, line 260: lets add a unit test for this method?

Shahroz16 (commented):
We could probably move to another thread-collection to instead of `List` to avoid this but happy either way

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistSdk.kt
mrehan27: Yeah, I'm writing down some tests for this, will push once I'm able to fail it for the current code and pass for the new changes.

mrehan27:
I tried different approaches and Shahroz's [suggestion for trying thread safe collections](https://github.com/customerio/customerio-android/pull/246#pullrequestreview-1586735619) looked the best option. I have used [CopyOnWriteArrayList](https://developer.android.com/reference/java/util/concurrent/CopyOnWriteArrayList) which works well and added tests. The tests passes with current implementation and fail if we run them on previous implementation.

Shahroz16 (approved):
Thank you for adding tests 👍

messaginginapp/src/sharedTest/java/io/customer/messaginginapp/GistSdkListenersTest.kt
```diff
@@ -0,0 +1,124 @@
...
+import java.util.concurrent.TimeUnit
+import kotlin.concurrent.thread
+import org.amshove.kluent.internal.assertEquals
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+internal class GistSdkListenersTest : BaseTest() {
+    @Test
+    fun processAndRemoveListenersIndividually_givenConcurrentModification_expectSuccessfulCompletion() {
```
levibostian, line 18: Could you add a comment to this function explaining what edge case the test function is testing against (why this PR was created) and a small description about how this test function is designed? 

I don't expect anything long.

mrehan27: Added [docs](https://github.com/customerio/customerio-android/pull/246/commits/19fce997bd0bd4953128031ebf457ecb515c6578)

messaginginapp/src/sharedTest/java/io/customer/messaginginapp/GistSdkListenersTest.kt
```diff
@@ -0,0 +1,124 @@
...
+
+        assertEquals(
+            expected = 0L,
+            actual = threadsCompletionLatch.count,
+            message = "Threads did not complete within the timeout"
+        )
+    }
+
+    @Test
+    fun processAndClearListenersAllAtOnce_givenConcurrentModification_expectSuccessfulCompletion() {
```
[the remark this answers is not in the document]
mrehan27: Yes, because: 

- `removeListener` and `clearListeners` are two different functions in `GistSdk`
- They are independent and do not reuse each other
- They have different logics (one removes listener instantly while the latter loops over all listeners)

messaginginapp/src/sharedTest/java/io/customer/messaginginapp/GistSdkListenersTest.kt
levibostian: Great docs. Thanks!

messaginginapp/src/sharedTest/java/io/customer/messaginginapp/GistSdkListenersTest.kt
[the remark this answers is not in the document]
levibostian: Sorry, I didn't realize there were 2 different functions. Yes, having 2 test functions is good then.

PR #247, opened 2023-09-04
feat: improve push click behavior

mrehan27:

### Changes

- Add `TrackableScreen` interface to provide screen name for auto screen tracking
- Implement `TrackableScreen` in Java sample

### Pending Items

- [x] Update default values after squad discussion

### Note

This is base PR for [improving push click behavior](https://github.com/customerio/issues/issues/10830). Changes planned in this PR directly are final, but will be kept in draft till the feature is complete. Further changes for the feature will be merged through more PRs into this branch and then released after final testing.


sdk/src/main/java/io/customer/sdk/CustomerIO.kt
```diff
@@ -545,17 +546,25 @@ class CustomerIO internal constructor(
...
-            val activityLabel = info.loadLabel(packageManager)
-
-            val screenName = activityLabel.toString().ifEmpty {
-                activity::class.java.simpleName.getScreenNameFromActivity()
+        try {
+            val screenName: String? = if (activity is TrackableScreen) {
+                // TrackableScreen takes precedence over manifest label
+                activity.getScreenName()
+            } else {
+                val info = packageManager.getActivityInfo(
```
Shahroz16, line 554: Thinking out load, we could have just exposed an option in the `meta data` of the manifest to skip these Activities which we don't intend to track.

Shahroz16 (approved):
Looks good, just one alternative solution added.

sdk/src/main/java/io/customer/sdk/CustomerIO.kt
mrehan27: Interesting 🤔 I didn't think of it before. But having an interface gives more flexibility on determining the screen name, so let's stick with it.

levibostian (changes requested):
I am not quite understanding the reason for this pull request. I see "Update auto tracking to skip screen event for specific activity" as the task that was completed by this pull request in the ticket. 
 
That makes me think the reason for skipping screen events for specific activity is because you will be introducing a new Activity to the CIO SDK and you want to make sure that the new Activity does not get tracked by CIO auto screenview tracking. 

If that is all true, is there a simplified solution? Adding a new interface `TrackableScreen` feels like a new feature added to the CIO SDK for customers to use. I was not expecting a new feature, but a different solution such as a hard-coded filter in the SDK's screenview tracking logic that skipped Activities in the CIO SDK that should not be tracked.

mrehan27:
@levibostian Yes the change prevents new activity from being tracked by CIO auto-screen tracking. The PR does exactly what is described in the [task breakdown](https://github.com/customerio/issues/issues/10830#issuecomment-1698909361). Current approach is much better than solutions like hard-coding filters. And even if it adds a feature, it does without breaking existing functionality and requires minimal effort on our part. I don’t see any reason for not continuing with it. If you have any suggestions for improvement or see any issues with this approach, feel free to mention them. Otherwise, let’s continue the discussion in the parent ticket or on Slack.

levibostian:
I am open to a solution such as this one where an interface is used to determine if a screen should be tracked or not. 

My hesitations come up from: 

* The use of this new interface `TrackableScreen` in the sample apps. If we are to begin using `TrackableScreen` in the sample apps, that makes me think we are suggesting it as our recommended approach to doing automatic screenview tracking in the SDK. Especially with the use of `getScreenName()` where we are returning different values for certain fragments, `TrackableScreen` seems as though it's trying to fix an unrelated problem to push click behavior. I suggest we introduce a new feature such as this with a new ticket and new PR. 
* The naming of `TrackableScreen` makes it seem as though you need to use it in order for a screen to be tracked. A customer might ask, "If you do not inherit `TrackableScreen`, will my screen still be tracked?". We can discuss naming if this feature is turned into a new ticket with new PR. 

I have an idea to suggest. Would it work for a refactor to the PR where `TrackableScreen` is renamed to `DoNotTrackScreen` and the interface is `internal` or undocumented? If it's an interface that is currently only used internally (not used in sample apps), I would be OK with this. Then, we can expose this interface to the public and make it a documented feature in the future if we believe it's important to add.

mrehan27:
@levibostian I actually started implementing this with `DoNotTrackScreen` but ended up on `TrackableScreen`. Because I think the number of changes are almost same but `TrackableScreen` is more flexible. I'm okay with removing it from sample apps though if we have more votes on it.

For the interface, I don't think this can be confusing for customers because we are not yet going to add this in docs. If the customers find out the interface themselves, the doc on it already explains it is fully optional. We can discuss any improvements in docs or name on the file, if you have any ideas in my mind, please post it on [`TrackableScreen`](https://github.com/customerio/customerio-android/pull/247/files#diff-cd2a752d343c24cfd5a25dc5fb7451da738eb7c4dfb584de6e426ae543501cba) so it is easier to discuss in thread.

PR #249, opened 2023-09-06
chore: add activity to handle push click

mrehan27:

### Changes

- `NotificationClickReceiverActivity` for handling notification clicks
- `NotificationClickBehavior` for customizing notification click behavior
- `DeepLinkUtil` updates to avoid adding unwanted flags to intents

### Notes

- Intent flag changes in PR are not fully final and still being tested with all sample apps, which may force changes for flags in future PRs
- Options and default values for `NotificationClickBehavior` may be re-reviewed after feature changes are complete so we have clear picture of there implementation and behavior with different sample apps

### DIFF
--- a/messagingpush/src/main/java/io/customer/messagingpush/activity/NotificationClickReceiverActivity.kt
+++ b/messagingpush/src/main/java/io/customer/messagingpush/activity/NotificationClickReceiverActivity.kt
@@ -18,6 +18,12 @@ import io.customer.sdk.extensions.takeIfNotBlank
 import io.customer.sdk.tracking.TrackableScreen
 import io.customer.sdk.util.Logger
 
+/**
+ * Activity to handle notification click events.
+ *
+ * This activity is launched when a notification is clicked. It tracks opened
+ * metrics, handles the deep link and opens the desired activity in the host app.
+ */
 class NotificationClickReceiverActivity : Activity(), TrackableScreen {
     val logger: Logger by lazy { CustomerIOShared.instance().diStaticGraph.logger }
 
--- a/messagingpush/src/main/java/io/customer/messagingpush/activity/NotificationClickReceiverActivity.kt
+++ b/messagingpush/src/main/java/io/customer/messagingpush/activity/NotificationClickReceiverActivity.kt
@@ -64,17 +64,30 @@ class NotificationClickReceiverActivity : Activity(), TrackableScreen {
             return
         }
 
-        val diGraph = sdkInstance.diGraph
-        val deepLinkUtil: DeepLinkUtil = diGraph.deepLinkUtil
-        val moduleConfig: MessagingPushModuleConfig = diGraph.moduleConfig
+        val moduleConfig: MessagingPushModuleConfig = sdkInstance.diGraph.moduleConfig
+        trackMetrics(moduleConfig, payload)
+        handleDeepLink(moduleConfig, payload)
+    }
 
-        val deliveryId = payload.cioDeliveryId
-        val deliveryToken = payload.cioDeliveryToken
-        val deepLink = payload.deepLink?.takeIfNotBlank()
-
+    private fun trackMetrics(
+        moduleConfig: MessagingPushModuleConfig,
+        payload: CustomerIOParsedPushPayload
+    ) {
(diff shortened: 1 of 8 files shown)


levibostian (commented):
Sorry, github is giving me error messages while trying to leave inline comments. 

Before I leave those, I do want to point out that github annotations are showing up for many lint errors for deprecated code being used and automated tests not being written. 

Do you plan on adding automated tests for this PR and addressing lint warnings? 

Hopefully by the time you address these concerns, github will no longer give me errors can I can leave a more detailed review!

messagingpush/src/main/java/io/customer/messagingpush/activity/NotificationClickReceiverActivity.kt
```diff
@@ -0,0 +1,135 @@
...
+            } else {
+                processNotificationIntent(payload = payload)
+            }
+        }.onFailure { ex ->
+            logger.error("Failed to process notification intent: ${ex.message}")
+        }
+        finish()
+    }
+
+    private fun processNotificationIntent(payload: CustomerIOParsedPushPayload) {
```
Shahroz16, line 60: can we further break down the logic here so it gets a little less complicated and more maintable? maybe a separate method of `handling deep linking` ?

messagingpush/src/main/java/io/customer/messagingpush/config/NotificationClickBehavior.kt
```diff
@@ -0,0 +1,37 @@
...
+     * This is similar to Android's "Set up a special activity PendingIntent."
+     * For more info, see [Android Documentation](https://developer.android.com/develop/ui/views/notifications/navigation#ExtendedNotification).
+     */
+    ACTIVITY_PREVENT_RESTART,
+
+    /**
+     * Starts the deep-linked activity without adding any intent flags.
+     * - Example: Stack (A -> B) becomes (A -> B -> D) if D is the deep-linked target activity.
+     *
+     * This behavior relies on the launch mode or flags specified for the activity in the Android manifest.
```
Shahroz16, line 34: maybe we should add, that system default behaviors will take over if no flag is mentioned? so that we make it clear if you dont put in any flags and chose this option, our default option wont be applied but rather systems.

Shahroz16 (approved):

mrehan27:
@levibostian I think the only deprecated code we are using in this PR is what isn't available on older APIs. I fixed one small issue. Please feel free to point any are that is still using deprecated code.

For tests, I think a lot of changes we made in this PR rely on Android OS callbacks. So testing them may be quite hard with mocks, but I do plan to write some tests. And they will probably be coming in the last PR for this feature (after wrapper ones).

levibostian:
I would prefer that tests get added to this PR because it does make reviewing much easier. 

GitHub gives you annotations helping you find parts of the code missing tests. If a future PR was made adding tests, the annotations would not show up and therefore, would make reviewing the PR for missing tests much harder. 

CleanShot 2023-09-08 at 08 15 51@2x

--- 

As far as the deprecations, I also use github annotations to highlight them. There are lint warnings that are annotated (see screenshot above). I suggest enabling them if you have not already.

messagingpush/src/main/java/io/customer/messagingpush/activity/NotificationClickReceiverActivity.kt
mrehan27: Updated

messagingpush/src/main/java/io/customer/messagingpush/config/NotificationClickBehavior.kt
mrehan27: Updated docs

mrehan27:
The deprecations shown are for existing code, not the code added in this PR. Fixing the warnings is definitely a good idea, but this PR isn't the right place for those updates.

Holding up this PR for tests complicates testing and finalizing changes on wrapper SDKs. The PRs were split to keep them short and simplify reviews. I'll write tests after validating changes in the wrapper SDKs. Since this PR is merging into a feature PR, we can still use GitHub annotations to identify code that isn't covered with tests.

Issue #248, opened 2023-09-06
Crash on 1 user but he seems blocked on this crash


arseneHuot:
**SDK version:** 3.6.5

**Environment:** Production Android

**Are logs available?** 

**Describe the bug**
Crash

**Screenshots**
<img width="1792" alt="Capture d’écran 2023-09-06 à 12 41 33" src="https://github.com/customerio/customerio-android/assets/24872405/a58224ad-d4bc-406d-a9d3-ee9edec0d167">

[com.beebs.mobile_issue_01beba4b5431fe8dc3e200cb66500a7f_crash_session_64F8572302F900013EB935D48388E27F_DNE_0_v2_stacktrace.txt](https://github.com/customerio/customerio-android/files/12536472/com.beebs.mobile_issue_01beba4b5431fe8dc3e200cb66500a7f_crash_session_64F8572302F900013EB935D48388E27F_DNE_0_v2_stacktrace.txt)

Shahroz16:
Hey @arseneHuot thank you for reaching out and apologize that you are facing this issue, We are looking into it right away, can you add some more device context such as memory/storage, etc that might also show on Firebase?

PR #251, opened 2023-09-08
fix: stack-overflow caused by BQ recursion

Shahroz16:
closes: https://github.com/customerio/customerio-android/issues/248

### DIFF
--- a/sdk/src/main/java/io/customer/sdk/queue/QueueRunRequest.kt
+++ b/sdk/src/main/java/io/customer/sdk/queue/QueueRunRequest.kt
@@ -1,6 +1,7 @@
 package io.customer.sdk.queue
 
 import io.customer.sdk.error.CustomerIOError
+import io.customer.sdk.queue.type.QueueTaskMetadata
 import io.customer.sdk.util.Logger
 
 interface QueueRunRequest {
@@ -19,8 +20,18 @@ internal class QueueRunRequestImpl internal constructor(
         val inventory = queueStorage.getInventory()
         val tasksToRun = inventory.toMutableList()
 
+        var lastFailedTask: QueueTaskMetadata? = null
+
         while (tasksToRun.isNotEmpty()) {
-            val currentTaskMetadata = tasksToRun.removeFirst()
+            // get the next task to run using the query criteria
+            val currentTaskMetadata = queryRunner.getNextTask(tasksToRun, lastFailedTask)
+            if (currentTaskMetadata == null) {
+                logger.debug("queue out of tasks to run...")
+                break
+            }
+
+            tasksToRun.remove(currentTaskMetadata)
+
             val taskStorageId = currentTaskMetadata.taskPersistedId
             val taskToRun = queueStorage.get(taskStorageId)
 
@@ -63,6 +74,7 @@ internal class QueueRunRequestImpl internal constructor(
                             queueStorage.update(taskStorageId, newRunResults)
                         }
                     }
+                    lastFailedTask = currentTaskMetadata
                 }
             }
         }
--- a/sdk/src/main/java/io/customer/sdk/queue/QueueRunRequest.kt
+++ b/sdk/src/main/java/io/customer/sdk/queue/QueueRunRequest.kt
@@ -1,8 +1,6 @@
 package io.customer.sdk.queue
 
 import io.customer.sdk.error.CustomerIOError
-import io.customer.sdk.queue.type.QueueInventory
-import io.customer.sdk.queue.type.QueueTaskMetadata
 import io.customer.sdk.util.Logger
 
 interface QueueRunRequest {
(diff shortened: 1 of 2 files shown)


sdk/src/main/java/io/customer/sdk/queue/QueueRunRequest.kt
```diff
@@ -19,86 +18,67 @@ internal class QueueRunRequestImpl internal constructor(
...
+                    queueStorage.delete(taskStorageId)
+                }
 
-                        goToNextTask(inventory, totalNumberOfTasksToRun, lastFailedTask = nextTaskToRunInventoryItem)
+                result.isFailure -> {
+                    val error = result.exceptionOrNull()
+                    logger.debug("queue task $taskStorageId run failed $error")
+
+                    when (error as? CustomerIOError) {
+                        is CustomerIOError.HttpRequestsPaused, is CustomerIOError.NoHttpRequestMade -> {
```
Shahroz16, line 60: added `is CustomerIOError.NoHttpRequestMade` because if there is no connection, there is no point trying out every request and using resources, might as well just wait till next time.

levibostian (changes requested):
So far, this looks like a good change, but I am not confident that it works yet. 

GitHub annotations highlights a couple places where there is code that does not get executed by tests (meaning possible missing tests) and some of the code in the loop I think isn't correct. 

This is a critical part of the code base, execution of the BQ. I strongly suggest that more tests are written for this `run` function to cover all the possible edge cases.

Shahroz16:
@levibostian Thanks for looking into it,

> GitHub annotations highlights a couple places where there is code that does not get executed by tests (meaning possible missing tests)

The only highlight seems towards `log` statements it seems.  

> This is a critical part of the code base, execution of the BQ. I strongly suggest that more tests are written for this run function to cover all the possible edge cases.

Can you explain what specific tests are you looking for? asking because we didn't change any logic except move from recursion to iterative approach. So, what gave you confidence in the previous implementation with the current test suite, but now makes you feel less confident?

levibostian:
@Shahroz16 Great questions. 

The latest test that you added for when a task cannot be found in storage was 1 missing test case. That's a great test case to add, thank you. 

Github annotations is telling me that there is 1 more missing branch for [this conditional statement](https://github.com/customerio/customerio-android/pull/251/files#diff-84815b58149a230f1d5cd0ea839c2d756898864fed99d8442358848fb8339b7eR29-R30). Adding a test for that scenario would be great. 

You do have a good point in that this is a refactor so existing tests should be all that we need. In my original comment I could have explained better that even though this is a refactor, I don't want to ship this code until the test suite is improved upon. If there are missing tests, I wanted to make sure that we *add tests* to the existing suite to feel more confident in the code. 

I recently found a missing test case in this exact same code block in the iOS SDK. When I added the missing test case, I actually found a bug. This makes me want to double check our test suite around this code in the Android SDK to make sure we trust it. I believe this PR is a good opportunity to do that. 

I reviewed the current test suite and it looks good to me except: 
* [this conditional statement](https://github.com/customerio/customerio-android/pull/251/files#diff-84815b58149a230f1d5cd0ea839c2d756898864fed99d8442358848fb8339b7eR29-R30) missing a test 
* Non-blocker for this PR, `QueueIntegrationTests` - could we add a test to reproduce this stackoverflow error? Then make sure that this refactor fixes the issue?

Shahroz16:
@levibostian added more tests.

Regarding,
> could we add a test to reproduce this stackoverflow error? Then make sure that this refactor fixes the issue?

I did try, but wasn't able to do it because it depends on multiple factors like heap size and limiting it isn't that straight forward with unit tests.

levibostian:
Tests are great! Thanks!

Oh, I just realized there is [1 more conversation to have resolved](https://github.com/customerio/customerio-android/pull/251#discussion_r1319777532). I do consider this a blocker for merging this PR.

Shahroz16:
@levibostian just noticed my comment to it was never posted so did it again, it says `2 days ago` for me that's why I originally added it. Just making sure, you can see it now.

sdk/src/main/java/io/customer/sdk/queue/QueueRunRequest.kt
Shahroz16: I don't think we should hinder getting improvement out unless we have a strong reason not to do so, it's a very small change but with a big performance improvement. I am quite comfortable making this change because `NoHttpRequestMade` is **only** being set when the network responds back with `null`. 

If you can kindly go through the code and express what makes you worry, please mention that I can respond to that concern but otherwise, renaming data types and iOS mirroring the same behavior shouldn't be the reason to block an improvement going out. The risk involved if this error shows up is it will skip an iteration unless the next time the connection goes through. 

I can request the other squad members to check out the update as well if that makes you feel more confident.

Shahroz16: @mrehan27 could you also go through this change when you get chance for more 👁️

mrehan27: I don't see any other usage for `NoHttpRequestMade` at the moment.  So the change looks good, no reason to block unless a we find a use case that fails and can be verified by tests.

mrehan27 (approved):

sdk/src/main/java/io/customer/sdk/queue/QueueRunRequest.kt
Shahroz16: That's my point as well, this is a serious performance concern and i am comfortable with the check and the code backs it. But i would still add more tests for it to improve confidence in later releases.

PR #268, opened 2023-10-02
refactor: move push click handling to processor

mrehan27:

### Changes

- Moved code out of `NotificationClickReceiverActivity` to `PushMessageProcessor` for reusability and better testing
- Added tests for `CustomerIOPushNotificationHandler`, `ModuleMessagingConfig` and `PushMessageProcessor`

### DIFF
--- a/messagingpush/src/main/java/io/customer/messagingpush/activity/NotificationClickReceiverActivity.kt
+++ b/messagingpush/src/main/java/io/customer/messagingpush/activity/NotificationClickReceiverActivity.kt
@@ -3,18 +3,9 @@ package io.customer.messagingpush.activity
 import android.app.Activity
 import android.content.Intent
 import android.os.Bundle
-import androidx.core.app.TaskStackBuilder
-import io.customer.messagingpush.MessagingPushModuleConfig
-import io.customer.messagingpush.config.PushClickBehavior
-import io.customer.messagingpush.data.model.CustomerIOParsedPushPayload
-import io.customer.messagingpush.di.deepLinkUtil
-import io.customer.messagingpush.di.moduleConfig
-import io.customer.messagingpush.extensions.parcelable
-import io.customer.messagingpush.util.DeepLinkUtil
+import io.customer.messagingpush.di.pushMessageProcessor
 import io.customer.sdk.CustomerIO
 import io.customer.sdk.CustomerIOShared
-import io.customer.sdk.data.request.MetricEvent
-import io.customer.sdk.extensions.takeIfNotBlank
 import io.customer.sdk.tracking.TrackableScreen
 import io.customer.sdk.util.Logger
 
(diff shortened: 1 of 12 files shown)


Shahroz16 (commented):
`getApplicationInfo(String, PackageManager.ApplicationInfoFlags)` lets maybe use this, since the other method is deprecated.

messagingpush/src/main/java/io/customer/messagingpush/activity/NotificationClickReceiverActivity.kt
```diff
@@ -43,98 +34,23 @@ class NotificationClickReceiverActivity : Activity(), TrackableScreen {
...
     private fun handleIntent(data: Intent?) {
-        kotlin.runCatching {
-            val payload: CustomerIOParsedPushPayload? =
-                data?.extras?.parcelable(NOTIFICATION_PAYLOAD_EXTRA)
-            if (payload == null) {
-                logger.error("Payload is null, cannot handle notification intent")
+        if (data == null) {
+            // This should never happen ideally
+            logger.error("Intent is null, cannot process notification click")
+        } else {
```
levibostian, line 40:
```suggestion
        if (data == null) {
            // This should never happen ideally
            logger.error("Intent is null, cannot process notification click")
            return 
        }      
```

I think the guard design pattern would be nice here instead of using `if/else`. Not a blocker, just nit-picky.

messagingpush/src/main/java/io/customer/messagingpush/processor/PushMessageProcessorImpl.kt
```diff
@@ -85,4 +94,92 @@ internal class PushMessageProcessorImpl(
...
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    fun processNotificationIntent(activityContext: Context, payload: CustomerIOParsedPushPayload) {
+        trackMetrics(payload)
+        handleDeepLink(activityContext, payload)
+    }
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    fun trackMetrics(payload: CustomerIOParsedPushPayload) {
+        if (moduleConfig.autoTrackPushEvents) {
```
levibostian, line 120: Is there a chance that `trackMetrics()` would be called manually? If so, the `if (moduleConfig.autoTrackPushEvents)` should be moved into maybe `processNotificationIntent` function.

messagingpush/src/sharedTest/java/io/customer/messagingpush/CustomerIOPushNotificationHandlerTest.kt
```diff
@@ -0,0 +1,57 @@
...
+            deepLink = String.random,
+            cioDeliveryId = String.random,
+            cioDeliveryToken = String.random,
+            title = String.random,
+            body = String.random
+        )
+    }
+
+    @Test
+    fun createIntentForNotificationClick_validPayload_shouldStartDeepLinkedActivity() {
```
levibostian, line 41: Would it add value if there was another test function added where `given not valid payload, should not start Activity`?

levibostian (commented):
Looking good! I am OK with approving, but with a big PR like this, I would prefer multiple reviews and approvals on this one. So, I'll let Shahroz help me and approve this one.

messagingpush/src/main/java/io/customer/messagingpush/activity/NotificationClickReceiverActivity.kt
mrehan27: We unfortunately can't use guard here because it should finish activity in all cases before exiting from this method.

messagingpush/src/main/java/io/customer/messagingpush/activity/NotificationClickReceiverActivity.kt
```diff
@@ -43,98 +34,23 @@ class NotificationClickReceiverActivity : Activity(), TrackableScreen {
...
+            val sdkInstance = CustomerIO.instanceOrNull(context = this)
+            if (sdkInstance == null) {
+                logger.error("SDK is not initialized, cannot handle notification intent")
             } else {
-                processNotificationIntent(payload = payload)
+                sdkInstance.diGraph.pushMessageProcessor.processNotificationClick(
```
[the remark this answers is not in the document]
mrehan27: We can. But I avoided because I don't think this adds a lot of value. It mostly relies on Android OS lifecycle and couple of other methods that are already being tested. Also, it requires setting up Activity lifecycle for testing which I plan to improve later to gain more confidence in testing. However, if you think we should still cover this, I can write a test for it.

messagingpush/src/main/java/io/customer/messagingpush/processor/PushMessageProcessorImpl.kt
mrehan27: I don't want to pass this responsibility to `processNotificationIntent` actually, because I think `trackMetrics` should be responsible for checking this.

However, I rechecked and realized that latest versions of tests don't require these methods to be exposed, so I have updated the visibility. Also, looking at the code again, I feel like the method names should have been more descriptive after they were moved to `PushMessageProcessorImpl`, I have pushed changes after renaming them so they make more sense.

messagingpush/src/sharedTest/java/io/customer/messagingpush/CustomerIOPushNotificationHandlerTest.kt
mrehan27: No actually. The activity should be started in all cases because this is being triggered when notification is clicked. So even if the payload is broken, Android OS still need something to start when user taps the notification. Whether we open anything next or not will be decided by `NotificationClickReceiverActivity`.

But I see the test name is confusing, I have updated it.

mrehan27:
@Shahroz16 I can see many warnings on `getApplicationInfo`, but the part of code using it is not linked with this change and exist in current `main` branch. I agree we should fix this warning, but it should be done and tested in a separate PR.

mrehan27:
@levibostian Most of the changes are actually copied from `NotificationClickReceiverActivity` to `PushMessageProcessorImpl` to facilitate testing. I avoid splitting into multiple PRs because except tests, all other changes are already been approved in #249. But I understand the concern, and will wait for both of you to approve before merging.

messagingpush/src/main/java/io/customer/messagingpush/activity/NotificationClickReceiverActivity.kt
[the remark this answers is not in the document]
Shahroz16: I think it still alots value because the whole logic revolves around what flags to expect and what behaviour it would display.

Shahroz16 (approved):

levibostian (approved):

messagingpush/src/main/java/io/customer/messagingpush/activity/NotificationClickReceiverActivity.kt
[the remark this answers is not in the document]
mrehan27: I have pushed more tests, hope it helps.

PS: Github is having issues in PRs, so it might take some time to show new changes.

Issue #270, opened 2023-10-16
Crash with EngineWebView


Vnicius:
**SDK version:** 
3.6.6

**Environment:** Development or Production
Production

**Are logs available?** 

```
Fatal Exception: android.util.AndroidRuntimeException: android.util.AndroidRuntimeException: android.webkit.WebViewFactory$MissingWebViewPackageException: Failed to load WebView provider: No WebView installed
       at android.webkit.WebViewFactory.getProvider(WebViewFactory.java:382)
       at android.webkit.WebView.getFactory(WebView.java:2599)
       at android.webkit.WebView.ensureProviderCreated(WebView.java:2593)
       at android.webkit.WebView.setOverScrollMode(WebView.java:2661)
       at android.view.View.<init>(View.java:5648)
       at android.view.View.<init>(View.java:5823)
       at android.view.ViewGroup.<init>(ViewGroup.java:707)
       at android.widget.AbsoluteLayout.<init>(AbsoluteLayout.java:59)
       at android.webkit.WebView.<init>(WebView.java:426)
       at android.webkit.WebView.<init>(WebView.java:368)
       at android.webkit.WebView.<init>(WebView.java:350)
       at android.webkit.WebView.<init>(WebView.java:337)
       at android.webkit.WebView.<init>(WebView.java:327)
       at io.customer.messaginginapp.gist.presentation.engine.EngineWebView.<init>(EngineWebView.kt:28)
       at io.customer.messaginginapp.gist.presentation.engine.EngineWebView.<init>(EngineWebView.kt:20)
       at io.customer.messaginginapp.gist.presentation.GistView.<init>(GistView.kt:27)
       at io.customer.messaginginapp.gist.presentation.GistSdk.init$lambda$0(GistSdk.kt:106)
       at android.os.Handler.handleCallback(Handler.java:942)
       at android.os.Handler.dispatchMessage(Handler.java:99)
       at android.os.Looper.loopOnce(Looper.java:226)
       at android.os.Looper.loop(Looper.java:313)
       at android.app.ActivityThread.main(ActivityThread.java:8757)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:571)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1067)
```

```
[...]

PR #271, opened 2023-10-17
fix: added exception handling for when browser doesn't exist

Shahroz16:
fixes: https://github.com/customerio/customerio-android/issues/270

Testing:
- This is how i replicated, created a new emulator 
- Run the command 
```
adb shell pm uninstall --user 0 com.google.android.webview
```

Verified the app didn't crash when an in-app was recieved.

### DIFF
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
@@ -29,6 +29,7 @@ internal class EngineWebView @JvmOverloads constructor(
     private var elapsedTimer: ElapsedTimer = ElapsedTimer()
 
     init {
+        // exception handling is required for webview in-case webview is not supported in the device
         try {
             webView = WebView(context)
             this.addView(webView)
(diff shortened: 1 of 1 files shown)


messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
```diff
@@ -38,40 +43,60 @@ internal class EngineWebView @JvmOverloads constructor(
...
-                    listener?.error()
-                }
-
-                override fun onReceivedError(view: WebView?, request: WebResourceRequest?, error: WebResourceError?) {
-                    listener?.error()
-                }
-
-                override fun onReceivedSslError(view: WebView?, handler: SslErrorHandler?, error: SslError?) {
-                    listener?.error()
+            webView?.let {
```
mrehan27, line 49: Could be only me, but if the code below doesn't need to execute, I prefer early returns instead of putting inside a block as it looks cleaner to me 😅

mrehan27 (approved):
Thanks for sharing the details on testing 👍🏻

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
Shahroz16: I am bit hesitant to do an early return because we are logging some important things and running timers that i don't want to miss running.

PR #274, opened 2023-10-27
fix: update notification receiver activity launchMode

mrehan27:

### Changes

- Added `android:launchMode="singleTask"` to `NotificationClickReceiverActivity` because `launchMode="singleTop"` does not work well with `ACTIVITY_NO_FLAGS` if customer app also has `launchMode="singleTop"` (e.g. Flutter apps) and have multiple instances launched from notifications in one session

### DIFF
--- a/messagingpush/src/main/AndroidManifest.xml
+++ b/messagingpush/src/main/AndroidManifest.xml
@@ -9,14 +9,17 @@
         <!-- Transparent activity for push interactions 
 
         Notes: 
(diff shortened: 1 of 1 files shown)


levibostian (approved):
Reviewing the comments in the PR and PR description, it sounds like `singleTask` works like `singleTop` but fixes edge cases introduced by flutter apps. 

Assuming that this change does not add any negative consequences to native Android SDK customers, approved.

PR #275, opened 2023-10-31
fix: encrypt stored sdk config

mrehan27:
### Background

Following the [Slack discussion](https://customerio.slack.com/archives/C02580R06/p1698523559693929), we realized that we should encrypt the SDK config to prevent possible reverse engineering and unauthorized extraction of keys.

### Changes

- Utilized [EncryptedSharedPreferences](https://developer.android.com/reference/androidx/security/crypto/EncryptedSharedPreferences) for secure storage of the SDK config
- Renamed the existing preferences file for the SDK config to avoid corrupt values
- Disabled backup for encrypted preferences, as recommended in the [Android docs](https://developer.android.com/reference/androidx/security/crypto/EncryptedSharedPreferences).

### Testing

- Native SDK users will not experience any impact, as they initialize the SDK in the Application class, eliminating the need for auto-initialization
- For existing wrapper customers, the SDK won't be auto-initialized from the killed state until the app is launched after this update. As soon as the app is launched, stored config values will be updated, and everything will proceed as expected.
- To test auto backup changes, enable `allowBackup` in the app and run the following command:

```
adb shell bmgr backupnow <app-package-name>
```

### DIFF
--- a/buildSrc/src/main/kotlin/io.customer/android/Versions.kt
+++ b/buildSrc/src/main/kotlin/io.customer/android/Versions.kt
@@ -2,6 +2,9 @@ package io.customer.android
 
 object Versions {
     internal const val ANDROID_GRADLE_PLUGIN = "7.2.0"
+
+    // Using alpha version as current stable version (1.0.0) has minSdkVersion 23
+    internal const val ANDROIDX_SECURITY_CRYPTO = "1.1.0-alpha06"
     internal const val ANDROIDX_TEST_JUNIT = "1.1.4"
     internal const val ANDROIDX_TEST_RUNNER = "1.4.0"
     internal const val ANDROIDX_TEST_RULES = "1.4.0"
--- a/messagingpush/src/main/res/xml/backup_rules.xml
+++ b/messagingpush/src/main/res/xml/backup_rules.xml
@@ -1,7 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <full-backup-content>
-    <!-- Prevent EncryptedSharedPreferences file from being backed up -->
+    <!-- Encrypted preferences should not be backed up as the key used to encrypt it will no
+    longer be present when backup is restored -->
     <exclude
         domain="sharedpref"
-        path="io.customer.sdk.SECURE_PREFERENCE_FILE_KEY.xml" />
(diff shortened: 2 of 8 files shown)


levibostian (changes requested):
I have a couple questions about this PR before merging. I'll reach out with my questions shortly.

Shahroz16 (approved):
Since it passes the manual test I am good with the change.

buildSrc/src/main/kotlin/io.customer/android/Versions.kt
```diff
@@ -2,6 +2,9 @@ package io.customer.android
 
 object Versions {
     internal const val ANDROID_GRADLE_PLUGIN = "7.2.0"
+
+    // Using alpha version as current stable version (1.0.0) has minSdkVersion 23
+    internal const val ANDROIDX_SECURITY_CRYPTO = "1.1.0-alpha06"
```
levibostian, line 7: Because 1.1-alpha releases support minSdk 21 while 1.0 supports minSdk 23, do we anticipate that 1.1 stable will move to minSdk 21? 

I would like to avoid us having to bump our minSdk because of a dependency.

messagingpush/src/main/res/xml/backup_rules.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<full-backup-content>
+    <!-- Encrypted preferences should not be backed up as the key used to encrypt it will no
+    longer be present when backup is restored -->
+    <exclude
+        domain="sharedpref"
+        path="io.customer.sdk.EncryptedConfigCache.xml" />
```
levibostian, line 7: Does this hard-coded string need to match [this hard-coded string](https://github.com/customerio/customerio-android/pull/275/files#diff-c1a277cd6b2e000cf26e555a3c571509c7d81202c4e8f26d62015e2421308f4dR28)? 

If so, I suggest using a string resource value to store this string in 1 place.

mrehan27:
Closing this PR based on our [discussion on Slack](https://customerio.slack.com/archives/C01QYDKD0SU/p1698776833120709). We'll continue exploring improvements for this and revisit later.

PR #276, opened 2023-11-07
fix: add priority field to message and sorting by priority within local store

BernardGatt:
Fixes issue: https://github.com/customerio/issues/issues/11663

### DIFF
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/listeners/Queue.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/listeners/Queue.kt
@@ -24,7 +24,6 @@ import retrofit2.converter.gson.GsonConverterFactory
 class Queue : GistListener {
 
     private var localMessageStore: MutableList<Message> = mutableListOf()
-    private val cacheMap = mutableMapOf<String, Response>()
 
     init {
         GistSdk.addListener(this)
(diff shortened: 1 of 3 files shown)


levibostian (approved):

levibostian:
@BernardGatt, before you merge this PR, can you fix the CI checks? 

I think that you don't have git hooks installed on your development machine? I suggest this because I believe that the "API check" tasks on the CI only fail when local files on your development machine didn't get updated as expected. 

I suggest trying to install git hooks, make a new commit and push it up. If you get stuck, I am OK with you merging this PR as-is since it's not introducing any public facing changes to the SDK and then ask the squad in Slack for help fixing the the API checks on `main`.

mrehan27 (approved):

messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/listeners/Queue.kt
```diff
@@ -24,7 +24,6 @@ import retrofit2.converter.gson.GsonConverterFactory
 class Queue : GistListener {
 
     private var localMessageStore: MutableList<Message> = mutableListOf()
-    private val cacheMap = mutableMapOf<String, Response>()
```
BernardGatt, line 27: @Shahroz16 just to confirm, this is not needed right?

Shahroz16: @BernardGatt yes!! feel free to remove it.

Shahroz16 (approved):

PR #277, opened 2023-11-08
ci: fix installing sd tool needed for deployment

levibostian:
The iOS SDK experienced a deployment failure on `main` because of a GH Action that we depend on. 

I installed a workaround in the iOS SDK until the GH action gets fixed. This PR applies the same workaround here.

### DIFF
--- a/.github/workflows/build-sample-apps.yml
+++ b/.github/workflows/build-sample-apps.yml
@@ -74,7 +74,8 @@ jobs:
     # CLI to replace strings in files. The CLI recommends using `cargo install` which is slow. This Action is fast because it downloads pre-built binaries. 
     # If using sd on macos, "brew install" works great. for Linux, this is the recommended way. 
     - name: Install sd CLI to use later in the workflow 
-      uses: kenji-miyake/setup-sd@v1
+      # uses: kenji-miyake/setup-sd@v1
+      uses: levibostian/setup-sd@add-file-extension # Using fork until upstream Action has bug fixed in it. 
     
     - name: Install tools from Gemfile (ruby language) used for building our apps with 
       uses: ruby/setup-ruby@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/build-sample-apps.yml
```diff
@@ -74,7 +74,8 @@ jobs:
     # CLI to replace strings in files. The CLI recommends using `cargo install` which is slow. This Action is fast because it downloads pre-built binaries. 
     # If using sd on macos, "brew install" works great. for Linux, this is the recommended way. 
     - name: Install sd CLI to use later in the workflow 
-      uses: kenji-miyake/setup-sd@v1
+      # uses: kenji-miyake/setup-sd@v1
```
Shahroz16, line 77: `kenji-miyake/setup-sd@v1` doesn't seem to have many users/stars, can we look for an alternative so we don't bump into issues like these?

Shahroz16 (approved):
Approving to unblock PRs but can we make a follow up ticket to get a permanent solution for this?

.github/workflows/build-sample-apps.yml
levibostian: Good suggestion to bring up. 

The `sd` CLI is a more stable `sed` tool. None of us like having to use `sed` in any of our codebases so I would love to not find an alternative `setup-sd` action, but find an alternative solution to no longer need `sd` at all. 

I would personally like to move to [a semantic-release plugin](https://github.com/jpoehnelt/semantic-release-replace-plugin) since updating the version string in our code is part of the deployment process. Plus, by doing string replacement via code instead of CLI, the developer experience for us will be better.

levibostian: [Ticket made](https://github.com/customerio/issues/issues/11701).

PR #279, opened 2023-12-06
fix: fixes a thread safety issue with currentMessage being out of sync

BernardGatt:
Addresses: https://github.com/customerio/issues/issues/11854

### DIFF
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistModalManager.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistModalManager.kt
@@ -14,23 +14,25 @@ internal class GistModalManager : GistListener {
     }
 
     internal fun showModalMessage(message: Message, position: MessagePosition? = null): Boolean {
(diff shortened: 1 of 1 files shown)


messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistModalManager.kt
```diff
@@ -14,23 +14,25 @@ internal class GistModalManager : GistListener {
     }
 
     internal fun showModalMessage(message: Message, position: MessagePosition? = null): Boolean {
```
Shahroz16, line 16: can we not just make the method `synchronized`?

mrehan27 (approved):

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistModalManager.kt
BernardGatt: Updated and retested. Thanks for the suggestion!

Shahroz16 (approved):

PR #285, opened 2024-02-09
feat: use new header to set polling interval

BernardGatt:
Updates the polling interval remotely through the new `X-Gist-Queue-Polling-Interval` header.

Part of: [INAPP-12243](https://linear.app/customerio/issue/INAPP-12243/update-the-in-app-sdks-to-configure-polling-based-on-remote)

### DIFF
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/listeners/Queue.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/listeners/Queue.kt
@@ -146,13 +146,14 @@ class Queue : GistListener {
     }
 
     private fun updatePollingInterval(headers: Headers) {
-        headers["X-Gist-Queue-Polling-Interval"]?.let { pollingInterval ->
-            if (pollingInterval.toInt() > 0) {
-                val newPollingInterval = (pollingInterval.toInt() * 1000).toLong()
-                if (newPollingInterval != GistSdk.pollInterval) {
-                    GistSdk.pollInterval = newPollingInterval
+        headers["X-Gist-Queue-Polling-Interval"]?.toIntOrNull()?.let { pollingIntervalSeconds ->
+            if (pollingIntervalSeconds > 0) {
+                val newPollingIntervalMilliseconds = (pollingIntervalSeconds * 1000).toLong()
+                if (newPollingIntervalMilliseconds != GistSdk.pollInterval) {
+                    GistSdk.pollInterval = newPollingIntervalMilliseconds
+                    // Queue check fetches messages again and could result in infinite loop.
                     GistSdk.observeMessagesForUser(true)
-                    Log.i(GIST_TAG, "Polling interval changed to: ${pollingInterval.toInt()} seconds")
+                    Log.i(GIST_TAG, "Polling interval changed to: $pollingIntervalSeconds seconds")
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/listeners/Queue.kt
```diff
@@ -141,6 +145,19 @@ class Queue : GistListener {
...
+    private fun updatePollingInterval(headers: Headers) {
+        headers["X-Gist-Queue-Polling-Interval"]?.let { pollingInterval ->
+            if (pollingInterval.toInt() > 0) {
+                val newPollingInterval = (pollingInterval.toInt() * 1000).toLong()
+                if (newPollingInterval != GistSdk.pollInterval) {
+                    GistSdk.pollInterval = newPollingInterval
+                    GistSdk.observeMessagesForUser(true)
+                    Log.i(GIST_TAG, "Polling interval changed to: ${pollingInterval.toInt()} seconds")
+                }
+            }
```
levibostian, line 157:
```suggestion
        headers["X-Gist-Queue-Polling-Interval"]?.toIntOrNull()?.let { pollingIntervalSeconds ->
            if (pollingInterval.toInt() > 0) {
                val pollingIntervalMilliseconds = (pollingIntervalSeconds * 1000).toLong()
                if (pollingIntervalMilliseconds != GistSdk.pollInterval) {
                    GistSdk.pollInterval = pollingIntervalMilliseconds
                    
                    // Queue check fetches messages again and could result in infinite loop. 
                    GistSdk.observeMessagesForUser(skipQueueCheck = true)
                    Log.i(GIST_TAG, "Polling interval changed to: ${pollingIntervalSeconds} seconds")
                }
            }
```

I have a couple suggestions that makes the code easier for me to follow. Nothing major though. 

* Labeling the variables with "seconds" and "milliseconds" made code faster to read. 
* Added comment on why `GistSdk.observeMessagesForUser` is skipping queue check. Maybe my comment is incorrect? I left the comment based on my understanding of the code after reading. 
* Moved the `pollingInterval.toInt()` to the `let {}` statement, to prevent having to call it multiple times.

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistSdk.kt
```diff
@@ -25,12 +25,12 @@ const val GIST_TAG: String = "[CIO]"
 object GistSdk : Application.ActivityLifecycleCallbacks {
     private const val SHARED_PREFERENCES_NAME = "gist-sdk"
     private const val SHARED_PREFERENCES_USER_TOKEN_KEY = "userToken"
-    private const val POLL_INTERVAL = 10_000L
 
     private val sharedPreferences by lazy {
         application.getSharedPreferences(SHARED_PREFERENCES_NAME, MODE_PRIVATE)
     }
 
+    internal var pollInterval = 600_000L
```
levibostian, line 33: Why is the default poll value increased?

levibostian (approved):

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistSdk.kt
BernardGatt: This will be our new default, the server will then adjust it remotely, you can read more here: [INAPP-12248](https://linear.app/customerio/issue/INAPP-12248/update-the-queue-composer-to-automatically-set-polling-interval-based)

messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/listeners/Queue.kt
BernardGatt: Agreed, I'll update the PR and implement some other minor changes.

PR #287, opened 2024-03-21
fix: last_used format for device

Shahroz16:
closes: https://linear.app/customerio/issue/MBL-170/incorrect-last-used-value-being-sent-from-android-sdk

Device payload was using `lastUsed` instead of `last_used` while adding a device/device attributes which BE doesn't accept.

### DIFF
--- a/sdk/src/main/java/io/customer/sdk/data/request/Device.kt
+++ b/sdk/src/main/java/io/customer/sdk/data/request/Device.kt
@@ -9,7 +9,7 @@ import java.util.*
 data class Device(
     @field:Json(name = "id") val token: String,
     val platform: String = "android",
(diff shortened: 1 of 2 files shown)


sdk/src/main/java/io/customer/sdk/data/request/Device.kt
```diff
@@ -9,7 +9,7 @@ import java.util.*
 data class Device(
     @field:Json(name = "id") val token: String,
     val platform: String = "android",
-    val lastUsed: Date,
+    @field:Json(name = "last_used") val lastUsed: Date?,
```
mrehan27, line 12: Can you please add a comment to explain why we made it nullable for future reference?

mrehan27 (approved):
LGTM 👍🏻

PR #288, opened 2024-03-26
chore: track binary size of SDK

mrehan27:
closes: https://linear.app/customerio/issue/MBL-157/track-the-binary-size-of-our-android-sdk

### Changes

- Added `apkScale` dependency to evaluate SDK binary size
- Added gradle tasks to generate and compare reports for SDK binary size

### _Notes_

_Final reports will be generated only using Github actions whenever a PR is merged to `main`. Changes for it will be covered in a separate PR._


scripts/sdk-binary-size.gradle
levibostian, line 1: Is it a requirement of APK scale to create new gradle tasks? 

I was expecting that APK scale shipped with tasks that we could use after installing the APK scale plugin into the project?

mrehan27: No, it's not a requirement from APKScale. APKScale includes the `measureSize` task, which we are indeed using. The purpose of the added gradle tasks is to facilitate its execution from CI. The gradle task does the following for us:

- Generates AARs for APKScale to compare
- Executes APKScale tasks for all feature modules
- Collects reports generated by APKScale into a unified simple JSON file
- Utilizes another task to compare two such JSON files, helping in identifying changes in binary size

scripts/sdk-binary-size.gradle
```diff
@@ -0,0 +1,160 @@
...
+// This task can be run locally by executing `./gradlew generateSdkSizeReport`
+// Run `./gradlew generateSdkSizeReport -PwriteReportTo=<writeReportTo>` to save the report to
+// provided file path
+tasks.register("generateSdkSizeReport") {
+    // List all task dependencies for each library module
+    List<String> taskDependencies = new ArrayList<>()
+    // Add the assembleRelease task first for each library module as a dependency
+    // So that the apkscale can use the generated aar files to calculate the binary size
+    project.subprojects.findAll { Project subproject ->
+        subproject.plugins.hasPlugin('com.android.library')
```
Shahroz16, line 23: do we need to run for all library modules? or could we just run for `targetModules` ?

mrehan27: We can run it for `targetModules` only. However, running it for all library modules gives me more confidence that APKScale will be able to locate all the AARs it requires. I lean towards running it for all library modules for this reason, but I'm flexible and have no strong preference as long as it works. If we believe running it only for `targetModules` could save time without causing issues, I'm happy to make that change and test it.

Shahroz16: I think we can make a more informed decision once we see the time taken by both approaches. Happy to go ahead with this and if we see its taking more time than we anticipate we can, revisit it.

Shahroz16 (approved):

scripts/sdk-binary-size.gradle
levibostian: Ok, now I understand. 

When I have worked in the past on gradle tasks, I didn't have the most enjoyable development experience. Simple tasks were OK (example: I dont mind the part of `generateSdkSizeReport` that combines multiple reports in 1) but the more logic I would add to a gradle task, the higher the learning curve I experienced and I had to spend time every now and then keeping the tasks updated with new gradle releases. 

From these past experiences, I personally prefer to use other tools besides gradle, when possible. 

Shahroz and I encountered a similar experience when I [added Ruby logic to generate build notes for iOS app builds](https://github.com/customerio/customerio-ios/blob/96d12daa647d64ab0b328cb5cfc997eac3d0d8b3/Apps/fastlane/Fastfile#L173-L275). After talking about it, we both decided that we would like to instead move this logic out of Ruby code and use alternative tools to accomplish this on the CI. 

I am not strongly against these 2 tasks where I need to reject the PR. But I would like it to be considered to reduce the scope of these gradle tasks so we don't need to maintain them over the long-term.

mrehan27: I have really good experience with gradle tasks 😄 They're easy to use, powerful, and let us test easily without needing other tools like CI.

I don’t think we need to change the plan. It's simple: generate, combine, and compare. Unless other members strongly feel we should rethink this, I believe we should continue keeping these gradle tasks.

mrehan27:
**SDK Binary Size / Generate Comparison Report** is expected to fail as `generateSdkSizeReport` is not yet available in `main`. Merging this.

PR #290, opened 2024-03-27
refactor: rename sdk module to tracking

mrehan27:

### Changes

- `apkScale` was breaking for `tracking` module as the module name in `setting.gradle` was different from the artifact id, so renamed it to match the artifact id. This should not impact customers as the artifact id is unchanged.


levibostian (commented):
I am good with a module rename. 

I would prefer that we also rename the directory from `sdk` to `tracking` as well. I understand this will drastically increase the number of files changed, but as long as the files do not get modified, I approve of it. 

As someone coming into our code looking for the `tracking` module, it's confusing the the directory name for it is `sdk`.

mrehan27:
@levibostian I agree. I avoided initially because it might change many files, and we already have plans to rename it with the data pipeline changes anyway. But I agree this only affects file paths, not the content. Tagging @Shahroz16 for his thoughts. If we all agree to proceed, I'm happy to change the name of the directory.

Shahroz16 (commented):
Looks good to me, I would rather not rename the directory right now, even if it just means no content change for the same reason as Rehan also suggested, I,e this is going to be deleted after the addition of datapipeline module. 

So I would avoid unnecessary risk and testing that would then need to be done to make sure there are no ripples.

mrehan27:
Agreed. Given that our next immediate task involves working on the data pipelines for Android, I'd lean towards not changing the directory right now to minimize any risks.

Shahroz16 (approved):
LGTM

levibostian (approved):

mrehan27:
Merging this first so commit history accurately reflects the changes, as this isn't directly related to changes for binary size comparison.

PR #291, opened 2024-03-27
ci: sdk binary size report

mrehan27:

### Changes

- Added workflow to comment SDK binary size change summary on every PR
- Updated workflow to commit SDK size report for every commit merged to `main` before being released

### DIFF
--- a/.github/workflows/deploy-sdk.yml
+++ b/.github/workflows/deploy-sdk.yml
@@ -26,10 +26,7 @@ jobs:
         uses: levibostian/setup-sd@add-file-extension # Using fork until upstream Action has bug fixed in it.
 
       - name: Generate SDK Size Report
-        run: ./gradlew generateSdkSizeReport -PsaveReportTo=reports
-
-      - name: Stage SDK Size Report
-        run: git add reports/
+        run: ./gradlew generateSdkSizeReport -PwriteReportTo=reports/sdk-binary-size.json
 
       # Semantic-release tool is used to:
       # 1. Determine the next semantic version for the software during deployment.
--- a/.github/workflows/deploy-sdk.yml
+++ b/.github/workflows/deploy-sdk.yml
@@ -25,6 +25,9 @@ jobs:
         # uses: kenji-miyake/setup-sd@v1
         uses: levibostian/setup-sd@add-file-extension # Using fork until upstream Action has bug fixed in it.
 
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy-sdk.yml
```diff
@@ -25,6 +25,12 @@ jobs:
         # uses: kenji-miyake/setup-sd@v1
         uses: levibostian/setup-sd@add-file-extension # Using fork until upstream Action has bug fixed in it.
 
+      - name: Generate SDK Size Report
+        run: ./gradlew generateSdkSizeReport -PsaveReportTo=reports
+
+      - name: Stage SDK Size Report
+        run: git add reports/
+
```
mrehan27, line 33: I haven't been able to test this as this only runs when we prepare for release. So would appreciate more eyes on it.

About the workflow, the change is to create and store report in `main` branch whenever a release is made. [Added more details about it on Linear](https://linear.app/customerio/issue/MBL-157/track-the-binary-size-of-our-android-sdk#comment-5bc03dee).

mrehan27:
[Test PR](https://github.com/customerio/customerio-android/pull/292#issuecomment-2024878264) to verify changes for adding comment when binary size changes.

.github/workflows/sdk-binary-size.yml
```diff
@@ -0,0 +1,67 @@
...
+        with:
+          ref: ${{ github.base_ref }}
+
+      - name: Generate report for base branch
+        run: |
+          ./gradlew publishToMavenLocal
+          ./gradlew generateSdkSizeReport
+        env:
+          # Use local version to make sure the report is generated for the current changes
+          MODULE_VERSION: 'local'
```
Shahroz16, line 26: if we have `IS_DEVELOPMENT` true, wouldn't it automatically set the version to local?

mrehan27: Looking at the code, this seems correct. I haven't tried it yet, will update and see if it works fine 👍🏻

Shahroz16 (approved):

.github/workflows/deploy-sdk.yml
Shahroz16: I would rather suggest move the `Stage SDK Size Report` part in the Semantic assets section we have, we already stage files in it.

PR #294, opened 2024-03-29
ci: fix for binary size report gradle task

mrehan27:

### Changes 🤦🏻 

- Added `publishToMavenLocal` so local version can be fetched


Shahroz16 (approved):
lol

PR #303, opened 2024-04-19
ci: sdk compatibility with older gradle versions

mrehan27:
helps: [MBL-259](https://linear.app/customerio/issue/MBL-259/improve-ci-actions-to-verify-gradle-compatibility-with-older-versions)

### Changes

- Add `gradle-compatibility-builds` job to verify SDK compilation with older versions of gradle
- Added `update-gradle-compatibility.sh` script to help making gradle related changes easier for older versions
- Updated versions of actions related to Android


.github/workflows/snapshot-release.yml
```diff
@@ -50,4 +52,121 @@ jobs:
             Version: `${{ steps.set-snapshot-version.outputs.VERSION }}`
             Repository: `https://s01.oss.sonatype.org/content/repositories/snapshots/`
           edit-mode: replace
-          token: ${{ secrets.GITHUB_TOKEN }}
\ No newline at end of file
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+  # This job depends on the publish job to finish so we can use the snapshot version to verify
+  # gradle compatibility of the sample apps with actual releases of the SDK with latest changes.
+  gradle-compatibility-builds:
```
mrehan27, line 59: Keeping this with snapshot build ensures we test apps with changes actually published. If we used local, we first need to publish locally with actual AGP version, switch AGP version and then build sample apps. Using snapshot build reduces this effort by not publishing SDK every time separately.

.github/workflows/snapshot-release.yml
```diff
@@ -50,4 +52,121 @@ jobs:
...
+      fail-fast: false
+      matrix:
+        # Matrix of sample apps and AGP versions to test
+        # List all sample apps here that you want to test with the AGP versions.
+        sample-app-with-agp:
+          - "java_layout_current"
+          - "kotlin_compose_current"
+          - "java_layout_7"
+          - "kotlin_compose_7"
+        include:
```
mrehan27, line 73: I couldn't think to better design this matrix, but for the combination we needed, I think this suits best.

scripts/update-gradle-compatibility.sh
```diff
@@ -0,0 +1,109 @@
+#!/bin/sh
+
+# Script that update files required for the Android Gradle Plugin version compatibility with
```
mrehan27, line 3: Switching gradle plugin requires some changes to compile, so I moved them to script so it is easier to maintain.

.github/workflows/snapshot-release.yml
```diff
@@ -50,4 +52,121 @@ jobs:
...
+            sample-app: "kotlin_compose"
+            cio-siteid-secret-key: "CUSTOMERIO_KOTLIN_WORKSPACE_SITE_ID"
+            cio-apikey-secret-key: "CUSTOMERIO_KOTLIN_WORKSPACE_API_KEY"
+            agp-version: "8.3.1"
+            # We don't need to specify more versions here, as we'll use the default values for actual gradle versions.
+          - sample-app-with-agp: "java_layout_7"
+            sample-app: "java_layout"
+            cio-siteid-secret-key: "CUSTOMERIO_JAVA_WORKSPACE_SITE_ID"
+            cio-apikey-secret-key: "CUSTOMERIO_JAVA_WORKSPACE_API_KEY"
+            agp-version: "7.4.1"
```
mrehan27, line 90: `minifyReleaseWithR8` does break with `7.2.0` so that verifies this does help us confirm if it works correctly with Gradle 7 (latest release) 🎉

Shahroz16 (approved):
Great comments!

PR #305, opened 2024-04-24
chore: add core module

mrehan27:

### Changes

- Added `core` module
- Updated gradle scripts and CI workflows


mrehan27:
Compare SDK binary size report is failing because the size of `core` module is currently in KBs (`908.4KB`). This is okay for this PR, as we do not need to support KB measurements. The size will increase to MBs once we add a few more classes, and the task should then work fine.

mrehan27:
@Shahroz16 Yes, we could reuse `base`, but I avoided doing so for following reasons:

- I believe `core` naming convention is more commonly used
- I wanted to avoid the ripple effects that `base` module might cause as is it currently used by other modules
- I'm still unsure whether we'll still need a lightweight separate `base` or just the `core` module eventually. So, I avoided making this decision too early to allow enough time to consider our options.

PR #306, opened 2024-04-24
refactor: base module and digraph classes to core

mrehan27:

### Changes

- Moved base module classes from `tracking` to `core`
- Moved base DiGraph class from `tracking` to `core` so we can create desired graphs in `core`

### DIFF
--- a/common-test/build.gradle
+++ b/common-test/build.gradle
@@ -13,6 +13,7 @@ android {
 }
 
 dependencies {
+    api project(":core")
     // since all modules in project inherit the base tracking SDK, add it here to our shared testing module as all modules in project should have access to this module.
     api project(':tracking')
 
--- a/common-test/src/main/java/io/customer/commontest/module/CustomerIOGenericModule.kt
+++ b/common-test/src/main/java/io/customer/commontest/module/CustomerIOGenericModule.kt
@@ -1,0 +1,8 @@
+package io.customer.commontest.module
+
+import io.customer.android.core.module.CustomerIOModule
+
+/**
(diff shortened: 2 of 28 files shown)


common-test/build.gradle
```diff
@@ -13,6 +13,7 @@ android {
 }
 
 dependencies {
+    api project(":core")
```
Shahroz16, line 16: interesting why does it need to be `api` ?

common-test/src/main/java/io/customer/commontest/module/CustomerIOGenericModule.kt
Shahroz16, line 1: do we need a new file for this typealias? think we can just add it in `CustomerIOModule` class? as that's in core too?

Shahroz16 (approved):
looks good, added minor suggestions

common-test/src/main/java/io/customer/commontest/module/CustomerIOGenericModule.kt
mrehan27: `CustomerIOModule` is in core module so we can't keep `CustomerIOGenericModule` internal here. Also, `CustomerIOGenericModule` is mainly used by tests for convenience. I moved it to `commontest` module so only tests can access it and it doesn't get shipped with release. That's why I kept this to separate class.

common-test/build.gradle
mrehan27: I think it can be implementation. But we are using all dependencies as `api` in `commontest` module, so I didn't change that. But yes, we can improve this in future.

PR #307, opened 2024-04-26
refactor: improve digraph support

mrehan27:
part of [MBL-273](https://linear.app/customerio/issue/MBL-273/enhance-digraph-for-improved-dependency-management)

### Changes

- Refactored `DiGraph` to simplify accessing and overriding dependencies
- Added tests to validate `DiGraph` functionality

### DIFF
--- a/core/src/main/kotlin/io/customer/android/core/di/DiGraph.kt
+++ b/core/src/main/kotlin/io/customer/android/core/di/DiGraph.kt
@@ -103,7 +103,7 @@ abstract class DiGraph {
 
     // TODO: Remove deprecated functions after all usages are removed.
     @Deprecated("Use newInstance or singleton instead", ReplaceWith("newInstance()"))
-    inline fun <reified DEP> override(): DEP? = overrides[DEP::class.java.simpleName] as? DEP
+    inline fun <reified DEP> override(): DEP? = overrides[DEP::class.java.name] as? DEP
 
     @Deprecated("Use singleton instead", ReplaceWith("singleton(newInstanceCreator)"))
     inline fun <reified INST : Any> getSingletonInstanceCreate(newInstanceCreator: () -> INST): INST {
--- a/core/src/main/kotlin/io/customer/android/core/di/DiGraph.kt
+++ b/core/src/main/kotlin/io/customer/android/core/di/DiGraph.kt
@@ -1,48 +1,111 @@
 package io.customer.android.core.di
 
 abstract class DiGraph {
+    /**
+     * Map of dependencies that can be overridden with mocks in test functions.
+     */
     val overrides: MutableMap<String, Any> = mutableMapOf()
 
     /**
-     * Call when you want to override a dependency with a mock in test functions.
-     * `di.overrideDependency(CustomerIOInstance::class.java, customerIOMock)`
-     *
-     * See `CustomerIOTest` (or any other class that is top-level of the module) for an example on use.
-     */
-    fun <DEP> overrideDependency(dependency: Class<DEP>, value: DEP) {
-        overrides[dependency.simpleName] = value as Any
-    }
-
-    /**
-     * Call in Digraph property getters to get allow mock overriding before trying to get actual instance.
-     * ```
-     * val fileStorage: FileStorage
-     *   get() = override() ?: FileStorage(siteId, context)
-     * ```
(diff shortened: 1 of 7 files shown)


core/src/main/kotlin/io/customer/android/core/di/DiGraph.kt
```diff
@@ -1,48 +1,111 @@
...
     val overrides: MutableMap<String, Any> = mutableMapOf()
 
     /**
-     * Call when you want to override a dependency with a mock in test functions.
-     * `di.overrideDependency(CustomerIOInstance::class.java, customerIOMock)`
-     *
-     * See `CustomerIOTest` (or any other class that is top-level of the module) for an example on use.
+     * Map of singleton instances of dependencies in the graph.
+     */
+    val singletons: MutableMap<String, Any> = mutableMapOf()
```
Shahroz16, line 12: why not just use `ConcurrentHashMap` we won't have to use `synchronized` then.

Shahroz16: same goes for `overrides`

core/src/main/kotlin/io/customer/android/core/di/DiGraph.kt
```diff
@@ -1,48 +1,111 @@
...
+     * Internal function to get or create a singleton instance of a dependency in the graph.
+     * This function is thread-safe and will only create a single instance of the dependency.
+     * This function is marked as public only because it is inlined in other public functions.
+     * Do not call this function directly. Instead, use the `singleton` function.
+     */
+    inline fun <reified Dependency : Any> getOrCreateSingletonInstance(
+        newInstanceCreator: () -> Dependency
+    ): Dependency {
+        // Use a synchronized block to prevent multiple threads from creating multiple instances of the singleton.
+        synchronized(lock = this) {
```
Shahroz16, line 34: we can get rid of this with suggestion above

core/src/main/kotlin/io/customer/android/core/di/DiGraph.kt
```diff
@@ -1,48 +1,111 @@
...
+    /**
+     * Internal function to get an overridden dependency in the graph.
+     * This function is marked as public only because it is inlined in other public functions.
+     * Do not call this function directly. Instead, use the `newInstance` or `singleton` functions
+     * to get the dependency.
      */
-    fun <DEP> overrideDependency(dependency: Class<DEP>, value: DEP) {
-        overrides[dependency.simpleName] = value as Any
+    inline fun <reified Dependency : Any> getOverriddenInstance(): Dependency? {
+        return overrides[Dependency::class.java.name] as? Dependency
```
Shahroz16, line 21: how do you feel about adding a method such as 

```
    private inline fun <reified Dependency : Any> dependencyKey(): String = Dependency::class.java.name

```

so we don't have to keep writing the whole `Dependency::class.java.name` thing over and over again ? :D 

usage could be something like

```
inline fun <reified Dependency : Any> getOverriddenInstance(): Dependency? {
        return overrides[dependencyKey<Dependency>()] as? Dependency
    }
```

Shahroz16 (approved):
added suggestions that i think can improve performance and readability

core/src/main/kotlin/io/customer/android/core/di/DiGraph.kt
mrehan27: This a good idea, but we probably require `computeIfAbsent` to avoid multiple instances here, which was added in [API Level 24](https://developer.android.com/reference/java/util/concurrent/ConcurrentHashMap#computeIfAbsent(K,%20java.util.function.Function%3C?%20super%20K,?%20extends%20V%3E)) unfortunately, while we are on 21 yet. So I have kept `synchronized` block` for now.

For `overrides`, I think we don't need it because we aren't getting and creating instances at same time from it.

core/src/main/kotlin/io/customer/android/core/di/DiGraph.kt
mrehan27: Responded above

core/src/main/kotlin/io/customer/android/core/di/DiGraph.kt
mrehan27: Good idea, added this function.

PR #311, opened 2024-04-29
chore: add CIODestination plugin

Shahroz16:
closes: https://linear.app/customerio/issue/MBL-271/add-ciodestination-in-datapipeline-module

Added CIODestination plugin to use Segment directly and send events directly.


datapipelines/src/main/java/io/customer/datapipelines/plugins/CustomerIODestination.kt
mrehan27, line 1: Might be good to place it in `kotlin` folder instead of `java`? As with new grouping, Android Studio will group it inside following view with `kotlin`.

Screenshot 2024-04-29 at 2 55 29 PM

datapipelines/src/main/java/io/customer/datapipelines/plugins/CustomerIODestination.kt
```diff
@@ -0,0 +1,127 @@
...
+import com.segment.analytics.kotlin.core.System
+import com.segment.analytics.kotlin.core.TrackEvent
+import com.segment.analytics.kotlin.core.platform.DestinationPlugin
+import com.segment.analytics.kotlin.core.platform.EventPipeline
+import com.segment.analytics.kotlin.core.platform.Plugin
+import com.segment.analytics.kotlin.core.platform.VersionedPlugin
+import com.segment.analytics.kotlin.core.platform.plugins.DestinationMetadataPlugin
+import com.segment.analytics.kotlin.core.platform.policies.CountBasedFlushPolicy
+import com.segment.analytics.kotlin.core.platform.policies.FlushPolicy
+import com.segment.analytics.kotlin.core.platform.policies.FrequencyFlushPolicy
```
mrehan27, line 20: I don't remember if we have project wide recommendation for `*` imports, but this definitely is good use case to introduce common settings to switch to `*` imports (can be considered later though).

datapipelines/src/main/java/io/customer/datapipelines/plugins/CustomerIODestination.kt
```diff
@@ -0,0 +1,127 @@
...
+import kotlinx.coroutines.launch
+import kotlinx.serialization.Serializable
+import sovran.kotlin.Subscriber
+
+@Serializable
+data class CustomerIOSettings(
```
mrehan27, line 26: Can we please add some docs to this? Hard for me to understand its usage in the code without any docs.

datapipelines/src/main/java/io/customer/datapipelines/plugins/CustomerIODestination.kt
```diff
@@ -0,0 +1,127 @@
...
+    var apiKey: String,
+    var apiHost: String? = null
+)
+
+class CustomerIODestination : DestinationPlugin(), VersionedPlugin, Subscriber {
```
mrehan27, line 31: Even though I full understand this class, I would still appreciate some docs it to explain its purpose, limitations and any other useful info that you think might be good for anyone reading this in future.

datapipelines/src/main/java/io/customer/datapipelines/plugins/CustomerIODestination.kt
Shahroz16: Good call, it automatically adds to `java` but yeah I would rename it to `kotlin` which is kind of annoying

datapipelines/src/main/java/io/customer/datapipelines/plugins/CustomerIODestination.kt
Shahroz16: I think ktlin doesnt recommend it might not let you do it by default.

mrehan27 (approved):

PR #312, opened 2024-04-29
chore: initialize data pipeline module

mrehan27:

### Changes

- Added `Logger` to static DI component
- Added basic implementation of `CustomerIO` and `CustomerIOInstance` in `core` module
- Added `CustomerIOBuilder` to allow building `CustomerIO` SDK using builder patter

### Example Usage

After the changes, the SDK can be initialized with data pipeline module using the following code:

```kotlin
import io.customer.android.sdk.CustomerIO
import io.customer.datapipelines.Builder

with(CustomerIO.Builder(applicationContext = this, cdpApiKey = "API_KEY")) {
    setLogLevel(CioLogLevel.DEBUG)
    addCustomerIOModule(ModuleMessagingInApp())
    addCustomerIOModule(ModuleMessagingPushFCM())
    build()
}
```

### Note

There are currently multiple classes for `CustomerIO` and `CustomerIOInstance`. This is to allow adding new code without breaking the existing tests and builds. Once we remove the dependency of older implementation of these classes, we can move newly added classes to desired package. The change will be smaller as it only needs package and import updates, but will allow us keeping both implementations side by side while we are adding new features.


core/src/main/kotlin/io/customer/android/sdk/CustomerIO.kt
```diff
@@ -0,0 +1,79 @@
+package io.customer.android.sdk
+
+import androidx.annotation.VisibleForTesting
+import io.customer.android.core.di.SDKComponent
+import io.customer.base.internal.InternalCustomerIOApi
+
+// TODO: Move this class and its dependencies (CustomerIOInstance) to the correct package.
+// We need to move this class to the right package to avoid breaking imports for the users of the SDK.
+// We have placed the class in the wrong package for now to avoid breaking the build.
+// Once old implementations are removed, we can move the class to the correct package.
```
mrehan27, line 10: The note is reminder to what I explained in PR description.

datapipelines/src/main/kotlin/io/customer/datapipelines/config/DataPipelinesModuleConfig.kt
```diff
@@ -0,0 +1,7 @@
+package io.customer.datapipelines.config
+
+import io.customer.android.core.module.CustomerIOModuleConfig
+
+class DataPipelinesModuleConfig(
```
mrehan27, line 5: Config changes will be covered later in config ticket

core/src/main/kotlin/io/customer/android/core/di/SDKComponent.kt
```diff
@@ -1,14 +1,21 @@
...
  * Object level DiGraph for the SDK. Provides a centralized way to manage all
  * dependencies in the SDK in a single place.
  * The object can be accessed from anywhere in the SDK and can be used to provide
  * a consistent and efficient way to access them throughout the SDK.
  */
+@Suppress("MemberVisibilityCanBePrivate", "unused")
 object SDKComponent : DiGraph() {
     val androidSDKComponent: AndroidSDKComponent? get() = getOrNull()
+    val buildEnvironment: BuildEnvironment get() = singleton { DefaultBuildEnvironment() }
+    val logger: Logger get() = singleton { LogcatLogger(buildEnvironment = buildEnvironment) }
```
Shahroz16, line 19: Curious about the usage of `singleton` here?

- Since the `object` itself is a singleton, any class instances that it contains as members are also created once and persist for the lifetime of the application, effectively behaving as part of the singleton. Why are we adding another singleton on top of that?

core/src/main/kotlin/io/customer/android/sdk/CustomerIO.kt
Shahroz16: > We need to move this class to the right package to avoid breaking imports for the users of the SDK.

We are going with a breaking change, this was part of what import previously? and what will be the new one?

core/src/main/kotlin/io/customer/android/sdk/CustomerIO.kt
```diff
@@ -0,0 +1,79 @@
...
+        ): CustomerIO {
+            val existingInstance = instance
+            if (existingInstance != null) {
+                SDKComponent.logger.error("CustomerIO instance is already initialized, skipping the initialization.")
+                return existingInstance
+            }
+            return CustomerIO(implementation = implementation).apply {
+                instance = this
+            }
+        }
```
Shahroz16, line 67: small performance improvement we could do is instead of marking the whole method with `@Synchronized`, we could just skip the null check as it doesn't require synchronization.  Synchronization is only necessary when the instance is being created

```
fun createInstance(implementation: CustomerIOInstance): CustomerIO {
            return instance ?: synchronized(this) {
                instance ?: CustomerIO(implementation).also {
                    instance = it
                }
            }
        }
 ```
 
 this is without log, but you can make a version with logger too.

File: datapipelines/src/main/kotlin/io/customer/datapipelines/CustomerIOBuilder.kt [hunk 5]
```diff
@@ -0,0 +1,52 @@
...
+        val androidSDKComponent = SDKComponent.registerAndroidSDKComponent(context = applicationContext)
+        val modules = SDKComponent.modules
+
+        // Initialize DataPipelinesModule with the provided configuration
+        val dataPipelinesConfig = DataPipelinesModuleConfig(cdpApiKey = cdpApiKey)
+        val dataPipelinesModule = DataPipelinesModule(androidSDKComponent, dataPipelinesConfig)
+
+        // Register DataPipelinesModule and all other modules
+        modules[DataPipelinesModule.MODULE_NAME] = dataPipelinesModule
+        modules.putAll(registeredModules.associateBy { module -> module.moduleName })
```
Shahroz16, line 41: how do you feel about moving this into its own method? make the method lighter and to the point

core/src/main/kotlin/io/customer/android/sdk/CustomerIO.kt
mrehan27: > We are going with a breaking change

Yes, but this does mean we want to break everything that worked before 😅 We should update the package only if we really want to, so I called it out make sure that's by choice and not forced.

> This was part of what import previously? and what will be the new one?

Sorry I didn't get the question here. Are you asking what was the import of `CustomerIO` before and after? If yes, it will be `io.customer.sdk` for both.

core/src/main/kotlin/io/customer/android/core/di/SDKComponent.kt
mrehan27: To make sure our tests don't break. I think the [earlier PR that introduced `singleton`](https://github.com/customerio/customerio-android/pull/307) could explain better about its usage.

core/src/main/kotlin/io/customer/android/sdk/CustomerIO.kt
mrehan27: Good call, I didn't realize that will acquire lock even if the instance is already created, updating.

mrehan27: I realized that's incorrect. We need to make sure we return instance from lock too. Example use case

- Two threads try to access instance together
- Both of them gets `null`
- One of them enters the lock to create and set new instance
- The other keeps waiting, but has already passed `null` check
- The first one leaves and second one enters the lock to create and set new instance resulting in two separate instances

So `@Synchronized` and `synchronized` block does the same thing here. I'll stick with `@Synchronized` as it is easier to read.

core/src/main/kotlin/io/customer/android/core/di/SDKComponent.kt
Shahroz16: In the earlier PR, the `SDKComponent` didn't have `buildEnvironment` or `logger`. My questions was mostly around why did we go reusing the instance of these classes and not creating a newer one? 

 For example, `DefaultBuildEnvironment()` is just returning a static config. It will be accessed once as well?

core/src/main/kotlin/io/customer/android/sdk/CustomerIO.kt
Shahroz16: I am okay with `@Synchronized`

Shahroz16 (approved):

core/src/main/kotlin/io/customer/android/core/di/SDKComponent.kt
mrehan27: Because we used `singleton` here, both `Logger` and `BuildEnvironment` will be created only once, regardless of how many times they are accessed. We could use `newInstance` for `BuildEnvironment` if we prefer, but I don’t think it will have much effect.

PR #313, opened 2024-04-30
test: updated DiGraphTest to unit test

mrehan27:
### Changes

- Updated `DiGraphTest` to run as unit test for quicker results
- Added `BaseUnitTest` class for simple unit tests

### DIFF
--- a/common-test/src/main/java/io/customer/commontest/BaseUnitTest.kt
+++ b/common-test/src/main/java/io/customer/commontest/BaseUnitTest.kt
@@ -1,0 +1,14 @@
+package io.customer.commontest
+
+import org.junit.After
+import org.junit.Before
+
+abstract class BaseUnitTest {
+    @Before
+    open fun setUp() {
+    }
+
+    @After
+    open fun tearDown() {
+    }
+}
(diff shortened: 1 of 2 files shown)


common-test/src/main/java/io/customer/commontest/BaseUnitTest.kt
```diff
@@ -0,0 +1,14 @@
+package io.customer.commontest
+
+import org.junit.After
+import org.junit.Before
+
+abstract class BaseUnitTest {
```
mrehan27, line 13: I'm okay if we decide to remove this, but we can't yet use `BaseTest` in simple unit test due to context requirement there. So I added this so we can track simple unit tests when we improve test structure in future.

core/src/test/java/io/customer/android/core/di/DiGraphTest.kt
```diff
@@ -195,10 +195,10 @@ class DiGraphTest : BaseTest() {
             getSingleton(givenInstanceThread1, 100)
         }
         val thread2 = thread(start = false) {
-            getSingleton(givenInstanceThread2, 200)
+            getSingleton(givenInstanceThread2, 500)
```
mrehan27, line 198: To make this more reliable on CI, increased the waiting time.

Shahroz16 (approved):
makes sense

PR #314, opened 2024-04-30
refactor: update packages and imports

Shahroz16:
This PR does the following:

- Update `core` package structure to match `tracking` to ensure imports don't break for customers
- Conflicting files (that are present in both tracking and core with same name) change its `name space` so sample apps can compile


core/src/androidTest/java/io/customer/android/core/util/LoggerTest.kt
mrehan27, line 1: I think we need to move this file from `android` to `core` too

core/src/test/java/io/customer/android/core/di/DiGraphTest.kt
mrehan27, line 1: This one too

PR #315, opened 2024-04-30
chore: use new sdk initialization in sample apps

mrehan27:

### Changes

- Updated sample apps to use new SDK initialization (old not yet removed to avoid breaking functionality)
- Added support in sample apps for setting CDP API Key
- Deep link support for CDP API Key
- Updated CI actions


samples/java_layout/src/main/java/io/customer/android/sample/java_layout/sdk/CustomerIORepository.java
```diff
@@ -50,6 +51,19 @@ public void initializeSdk(SampleApplication application) {
 
         // Finally, build to finish initializing the SDK.
         builder.build();
+
+        // TODO: Remove old builder and use new builder to initialize the SDK
+        // New method to initialize CustomerIO
+        // The new method should be called after the old method till the old method is removed
+        // This is because the push and in-app modules are still using properties only initialized in the old method
```
levibostian, line 58: Good comment. 

Are we tracking `TODO`s anywhere in our tickets? 

Once we merge this, I am curious how we keep ourselves accountable to following up to tasks like this and also being able to measure the amount of effort involved in each `TODO` we create in the code. 

I greatly appreciate the effort in keeping the CI passing, including sample apps building. Seems worth it to me.

samples/kotlin_compose/src/main/java/io/customer/android/sample/kotlin_compose/MainApplication.kt
```diff
@@ -42,5 +43,23 @@ class MainApplication : Application() {
 
             build()
         }
+
+        // TODO: Remove old builder and use new builder to initialize the SDK
```
levibostian, line 47: same questions as before, `TODO`.

levibostian (approved):

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/ui/settings/SettingsActivity.java
```diff
@@ -75,6 +75,10 @@ private void parseLinkParams() {
         // https://www.java-sample.com/settings&site_id=xxx&api_key=yyy
```
Shahroz16, line 75: if you are updating the deep link to include `cdp_api_key`, maybe we need to add it here as well for reference?

Shahroz16 (approved):

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/sdk/CustomerIORepository.java
mrehan27: Great call. I've created a [ticket to address these TODOs](https://linear.app/customerio/issue/MBL-297/resolve-all-pending-todos-for-milestone-1) before we complete Milestone 1.

Regarding the efforts, I think it's challenging to set rules for it, but I've been trying to ensure that nothing breaks, all CI checks pass, and adding TODOs only when absolutely necessary.

Most of the TODOs so far involve straightforward code removal and renaming actions. I think we just need to be mindful of what we categorize as a TODO. If we notice that something added as a TODO requires a larger effort, we should highlight it and create a separate ticket for it.

samples/kotlin_compose/src/main/java/io/customer/android/sample/kotlin_compose/MainApplication.kt
mrehan27: Responded above

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/ui/settings/SettingsActivity.java
mrehan27: Good catch, updated

PR #317, opened 2024-05-01
chore: add configuration for datapipeline

Shahroz16:
closes: https://linear.app/customerio/issue/MBL-219/update-the-config-for-initialization

- Updates to `DataPipelineModule` and `CustomerIOBuilder` to add the missing configs
- Added tests
- Moved `CustomerIOBuilderTest` to unit test to speed up


core/src/main/kotlin/io/customer/sdk/data/model/Region.kt
Shahroz16, line 1: its in core now

datapipelines/build.gradle
```diff
@@ -37,4 +37,8 @@ dependencies {
     api project(":core")
     implementation(Dependencies.segment)
     implementation(Dependencies.kotlinxSerializationJson)
+
+    api Dependencies.mockito
+    api Dependencies.mockitoKotlin
+    api Dependencies.robolectric
```
Shahroz16, line 43: need roboelectric for application instance in unit test

datapipelines/src/main/kotlin/io/customer/datapipelines/DataPipelinesModule.kt
```diff
@@ -18,13 +23,30 @@ internal constructor(
 ) : CustomerIOModule<DataPipelinesModuleConfig>, CustomerIOInstance {
     override val moduleName: String = MODULE_NAME
 
-    private val analytics: Analytics
+    private val errorLogger = object : ErrorHandler {
```
Shahroz16, line 26: to put logs under our tag

datapipelines/build.gradle
mrehan27, line 43: Shouldn't these dependencies be `testImplementation`?

```suggestion
    testImplementation(Dependencies.mockito)
    testImplementation(Dependencies.mockitoKotlin)
    testImplementation(Dependencies.robolectric)
```

datapipelines/src/main/kotlin/io/customer/datapipelines/DataPipelinesModule.kt
mrehan27: Can you please add this as comment?

datapipelines/src/main/kotlin/io/customer/datapipelines/DataPipelinesModule.kt
```diff
@@ -18,13 +23,30 @@ internal constructor(
...
+    }
+
+    private val analytics: Analytics = Analytics(
+        writeKey = moduleConfig.cdpApiKey,
+        context = androidSDKComponent.applicationContext
+    ) {
+        flushAt = moduleConfig.flushAt
+        flushInterval = moduleConfig.flushInterval
+        flushPolicies = moduleConfig.flushPolicies
+        autoAddSegmentDestination = false
```
mrehan27, line 39: Would be great if can leave a comment here why keeping it `false` is important.

datapipelines/src/test/java/io/customer/datapipelines/sdk/CustomerIOBuilderTest.kt
```diff
@@ -0,0 +1,135 @@
...
+import org.mockito.kotlin.never
+import org.mockito.kotlin.verify
+import org.mockito.kotlin.whenever
+import org.robolectric.Robolectric
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
+class CustomerIOBuilderTest : BaseUnitTest() {
+
+    private val application: Application = Robolectric.buildActivity(Activity::class.java).setup().get().application
```
mrehan27, line 28: I think this should be moved to some base class as we'll keep needing this for several tests. I'm okay to not changes it in this PR, we can cover this in Setup Testing Environment ticket. But adding note so we don't miss this.

datapipelines/src/test/java/io/customer/datapipelines/sdk/CustomerIOBuilderTest.kt
```diff
@@ -0,0 +1,135 @@
...
+import org.junit.runner.RunWith
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.never
+import org.mockito.kotlin.verify
+import org.mockito.kotlin.whenever
+import org.robolectric.Robolectric
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
+class CustomerIOBuilderTest : BaseUnitTest() {
```
mrehan27, line 26: I think there are a few tests missing:

- Validating default values
- Validating setting regions (both `US` and `EU`) works correctly and as expected

mrehan27 (commented):
Looks mostly good, but I think we are missing some important tests. Once added, I feel we are good to merge this.

datapipelines/build.gradle
Shahroz16: 🤦

Shahroz16: 🤦

datapipelines/src/test/java/io/customer/datapipelines/sdk/CustomerIOBuilderTest.kt
Shahroz16: This is robolectric specific, not sure if its wise to make our base test robolectric, either we create another robolectric base class that extends base class or let each class add this if needed.

mrehan27: Yes, I meant to have another base class to help similar cases, we can look into this later with test changes though, no action required for this PR.

mrehan27 (approved):
Looks good now 👍🏻

PR #319, opened 2024-05-06
refactor: moved sdk classes to datapipelines module

mrehan27:

### Background

Since Kotlin extensions do not work the same way as Swift, keeping `CustomerIO` SDK files in `core` module and adding extensions in the `datapipelines` module poses challenges for customers. Kotlin users need to import an extension for every method, while the syntax for Java users becomes even more complicated. To address this, we decided to move `CustomerIO` class to `datapipelines` module for now. With the planned introduction of `EventBus`, we do not foresee the need for usage of `CustomerIO` directly from `core` elsewhere. In case the need arise in future, we can reevaluate and find alternatives to maintain better customer experience when integrating the SDK.

### Changes

- Moved `CustomAttributes` to `core` so it can be reused by all modules
- Moved `CustomerIO` and `CustomerIOInstance` classes from `core` module to `datapipelines` module
- Updated API classes


datapipelines/src/main/kotlin/io/customer/sdk/CustomerIOInstance.kt
Shahroz16, line 1: We are returning the `CustomerIO` class from the builder, is this also required for Java/Kotlin to work properly?

Shahroz16 (approved):

samples/sample-app.gradle
```diff
@@ -58,7 +58,7 @@ dependencies {
     } else {
         // Stable releases dependency, use published versions directly
         implementation "io.customer.android:tracking:$cioSDKVersion"
-//        implementation "io.customer.android:datapipelines:$cioSDKVersion"
+        implementation "io.customer.android:datapipelines:$cioSDKVersion"
```
mrehan27, line 61: @Shahroz16 This fixes the snapshot build.

datapipelines/src/main/kotlin/io/customer/sdk/CustomerIOInstance.kt
mrehan27: Discussed on call, we added `DataPipelineInstance` to keep this option open.

PR #320, opened 2024-05-06
chore: add logger to analytics in datapipelines module

mrehan27:

### Changes

- Added custom logger to `analytics` to ensure that log configurations align with the SDK
- Updated sample apps to set `logLevel` when initializing the SDK with new SDK builder


Shahroz16 (approved):

Shahroz16:
@mrehan27 the snapshots is failing

mrehan27:
@Shahroz16 Yeah, I noticed it too. Even though `:samples:java_layout:assembleRelease` works fine locally, but it seems to fail on CI since we started using it in sample app. Could it be because we have temporarily tweaked the package of `CustomerIO` or `CustomerIOInstance` without moving them to match package? 🤔 I can try pushing that if that fixes it for us.

mrehan27:
Merging this, and moving build failing discussion to #319 since builds are failing there too and it might be more relevant to fix the cause.

PR #322, opened 2024-05-06
feat: expose public APIs to identify and clear profiles

mrehan27:

### Changes

- Implemented and exposed public APIs to identify profiles with or without attributes and to clear profiles
- Implemented and exposed public APIs to set profile attributes
- Updated sample apps

#### Kotlin

```kotlin
CustomerIO.instance().identify(userId = email)
CustomerIO.instance().identify(userId = email, traits = mapOf("name" to name))
CustomerIO.instance().profileAttributes = mapOf("age" to age)
CustomerIO.instance().clearIdentify()
```

#### Java

```java
CustomerIO.instance().identify(email)
CustomerIO.instance().identify(email, attributes)
CustomerIO.instance().setProfileAttributes(attributes)
CustomerIO.instance().clearIdentify()
```

### Not Included

Tests are not included in this PR as we need to develop the base test structure. However, tests will be added in upcoming PRs soon.


datapipelines/src/main/kotlin/io/customer/datapipelines/DataPipelinesModule.kt
```diff
@@ -75,6 +77,84 @@ internal constructor(
...
+            if (userId != null) {
+                commonIdentify<Nothing>(userId = userId, traitsMap = value)
+            } else {
+                logger.debug("No user profile found, updating traits for anonymous user ${analytics.anonymousId()}")
+                analytics.identify(traits = value)
+            }
+        }
+
+    override fun identify(userId: String) {
+        commonIdentify<Nothing>(userId = userId)
```
mrehan27, line 139: `commonIdentify<Nothing>` might not be very clean, but this does help having a common method to identify profiles for all cases. Since this is internal, I'm happy continue using it the same way.

datapipelines/src/main/kotlin/io/customer/sdk/DataPipelineInstance.kt
```diff
@@ -1,9 +1,69 @@
...
+     * Identify a customer (aka: Add or update a profile).
+     * [Learn more](https://customer.io/docs/identifying-people/) about identifying a customer in Customer.io
+     * Note: You can only identify 1 profile at a time in your SDK. If you call this function multiple times,
+     * the previously identified profile will be removed. Only the latest identified customer is persisted.
+     *
+     * @param userId ID you want to assign to the customer.
+     * This value can be an internal ID that your system uses or an email address.
+     * [Learn more](https://customer.io/docs/api/#operation/identify)
+     */
+    fun identify(userId: String)
```
mrehan27, line 27: To keep our SDK compatible with Java, we have to add two separate methods to the interface as using `JvmOverloads` and default values isn't possible with interface at the moment.

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/sdk/CustomerIORepository.java
```diff
@@ -103,11 +103,11 @@ private void configureSdk(CustomerIO.Builder builder, final CustomerIOSDKConfig
     }
 
     public void identify(@NonNull String email, @NonNull Map<String, String> attributes) {
-        CustomerIO.instance().identify(email, attributes);
```
mrehan27, line 106: Due to this removal, apps may no longer send push token and events to Fly as profile will no longer be identified to older SDK. But I think its okay as it'll keep the logs clean and easier to remove old code later.

samples/kotlin_compose/src/main/java/io/customer/android/sample/kotlin_compose/ui/login/LoginViewModel.kt
```diff
@@ -62,9 +62,9 @@ class LoginViewModel @Inject constructor(
     ) {
         viewModelScope.launch(Dispatchers.IO) {
             userRepository.login(email = email, name = name, isGuest = isGuest)
-            CustomerIO.instance().identify(
```
mrehan27, line 65: Same as [above](https://github.com/customerio/customerio-android/pull/322/files#r1591125113)

datapipelines/src/main/kotlin/io/customer/sdk/DataPipelineInstance.kt
Shahroz16: I think we don't have to stick to interface, lets just utilize abstract class and add `JvmOverloads`

datapipelines/src/main/kotlin/io/customer/datapipelines/DataPipelinesModule.kt
Shahroz16: hmm i think there exists a world where we don't have to utilize `Nothing` in the first place.

datapipelines/src/main/kotlin/io/customer/sdk/DataPipelineInstance.kt
Shahroz16, line 1: Let's simplify this interface by doing following:
- Move to abstract class, to cater `JVM overloads` plus. `inline` methods which we can use to to allow us to find out the serialization strategy 
- Helps `DataPipelineModule` to have only implementation of 1 use-case that's needed.

Shahroz16 (changes requested):

Shahroz16:
As discussed on the call, here are the PR suggestion ideas in a separate [branch](shahroz/mbl-214-identify-profile-alternative).

mrehan27:
@Shahroz16 Updated as discussed, changes are slightly bigger due to code movement in different files.

datapipelines/src/main/kotlin/io/customer/sdk/DataPipelineInstance.kt
```diff
@@ -1,9 +1,95 @@
...
+     * [Learn more](https://customer.io/docs/identifying-people/) about identifying a customer in Customer.io
+     * Note: You can only identify 1 profile at a time in your SDK. If you call this function multiple times,
+     * the previously identified profile will be removed. Only the latest identified customer is persisted.
+     *
+     * @param userId Identifier you want to assign to the customer.
+     * This value can be an internal ID that your system uses or an email address.
+     * [Learn more](https://customer.io/docs/api/#operation/identify)
+     * @param traits Map of <String, Any> to be added
+     */
+    fun identify(userId: String, traits: CustomAttributes) {
```
Shahroz16, line 64: it would be great if we can just add that its added for java

Shahroz16 (approved):

PR #325, opened 2024-05-09
refactor: instrumentation test

Shahroz16:
changes:

- fixes instrumentation test
- speed up the test by caching snapshots

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,7 +33,7 @@ jobs:
         if: ${{ always() }} # if running tests fails, we still want to parse the test results
 
   instrumentation-test:
-    runs-on: macos-12
+    runs-on: ubuntu-latest
     timeout-minutes: 45
     strategy:
       # Add more samples here as they are added to the project.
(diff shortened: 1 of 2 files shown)


samples/kotlin_compose/src/androidTest/java/io/customer/android/sample/kotlin_compose/MemoryLeakageInstrumentedTest.kt
```diff
@@ -83,6 +85,22 @@ class MemoryLeakageInstrumentedTest {
             build()
         }
 
+        CustomerIOBuilder(
```
levibostian, line 88: Is this change in scope? It seems unrelated to speeding up the CI. No need to reply, just a thought.

.github/workflows/test.yml
levibostian, line 1: Overall, these changes seem really valuable. I would love to add this change to the codebase. 

For maintainability sake, I think we need some documentation to help us understand what each step is doing. Why we need to do a step and how we are doing it. So if in 6 months someone besides yourself needs to fix something, they have the information they need to do it on their own. 

Up to you on how to document this all. I am OK with using comments inside of this file. I do consider this a blocker for this PR as even I don't understand what some of these steps are doing.

.github/workflows/test.yml
```diff
@@ -40,22 +42,48 @@ jobs:
...
+        uses: gradle/actions/setup-gradle@v3
+
+      - name: AVD cache
+        uses: actions/cache@v3
+        id: avd-cache
+        with:
+          path: |
+            ~/.android/avd/*
+            ~/.android/adb*
+          key: avd-31
```
levibostian, line 61: When using `actions/cache`, you need to use both `key` and `restore-keys`. Otherwise the cache will be saved once but never updated again. 

I recently used this action in iOS codebase. [Maybe this code could be a helpful reference for you?](https://github.com/customerio/customerio-ios/blob/490593e38d1bfa010b4361b9a4e1bc50c34a4230/.github/actions/main-sample-app-build/action.yml#L31-L62)

levibostian (commented):
:chefs-kiss: anytime we can speed up the CI. Thanks!!

samples/kotlin_compose/src/androidTest/java/io/customer/android/sample/kotlin_compose/MemoryLeakageInstrumentedTest.kt
mrehan27: Not speeding up, but I think it is to fix this test as some SDK calls in sample app now rely on new SDK initialization. And if we don't initialize new instance, the code may keep breaking and tests will fail.

mrehan27 (approved):
Thanks for making these changes 👍🏻

PR #326, opened 2024-05-10
chore: allow override data pipeline instance

mrehan27:

### Changes

- Added common method to update Analytics configurations from module configuration so it can be reused by tests
- Allowed overriding DataPipelines instance for testing purposes only
- Reset static DI graph component

### DIFF
--- a/datapipelines/src/main/kotlin/io/customer/sdk/CustomerIOBuilder.kt
+++ b/datapipelines/src/main/kotlin/io/customer/sdk/CustomerIOBuilder.kt
@@ -59,6 +59,9 @@ class CustomerIOBuilder(
     // Configuration options required for migration from earlier versions
     private var migrationSiteId: String? = null
 
+    // Overrides the default DataPipelines module instance with custom implementation
+    private var overrideDataPipelineInstance: ((DataPipelinesModuleConfig) -> CustomerIO)? = null
+
     /**
      * Specifies the log level for the SDK.
      * Default value is [CioLogLevel.ERROR].
@@ -153,6 +156,17 @@ class CustomerIOBuilder(
         return this
     }
 
+    /**
+     * Overrides the default [CustomerIO] instance with a custom implementation.
+     * This is useful for testing only and should not be used in production.
+     */
+    internal fun setOverrideDataPipelineInstance(
+        overrideDataPipelineInstance: (DataPipelinesModuleConfig) -> CustomerIO
+    ): CustomerIOBuilder {
+        this.overrideDataPipelineInstance = overrideDataPipelineInstance
+        return this
+    }
+
     fun build(): CustomerIO {
         // Register AndroidSDKComponent to fulfill the dependencies required by the SDK modules
         val androidSDKComponent = SDKComponent.registerAndroidSDKComponent(context = applicationContext)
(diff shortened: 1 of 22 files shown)


datapipelines/src/main/kotlin/io/customer/sdk/CustomerIOBuilder.kt
```diff
@@ -59,6 +59,9 @@ class CustomerIOBuilder(
     // Configuration options required for migration from earlier versions
     private var migrationSiteId: String? = null
 
+    // Overrides the default DataPipelines module instance with custom implementation
+    private var overrideDataPipelineInstance: ((DataPipelinesModuleConfig) -> CustomerIO)? = null
```
Shahroz16, line 63: I think if we were to utilize delegate from DiGraph we won't have to override manually?

datapipelines/src/main/kotlin/io/customer/sdk/android/CustomerIO.kt
```diff
@@ -64,21 +66,14 @@ class CustomerIO private constructor(
         }
     }
 
-    private val analytics: Analytics = Analytics(
+    internal val analytics: Analytics = overrideAnalytics ?: Analytics(
```
Shahroz16, line 69: same concept goes here as mentioned above

datapipelines/src/main/kotlin/io/customer/sdk/CustomerIOBuilder.kt
mrehan27: Yes, this is a good suggestion. The only problem is, we need `DataPipelinesModuleConfig` to create test instance. We can think of options to solve this problem though, but I think using `DiGraph` can simplify a lot of code in this PR. I'll try this out and get back soon.

mrehan27: I added some updates to `SDKComponent` as we discussed on call earlier, let me know if there are any concerns.

datapipelines/src/test/java/io/customer/datapipelines/extensions/DiGraphExt.kt
```diff
@@ -0,0 +1,14 @@
...
+import io.customer.sdk.core.di.SDKComponent
+
+/**
+ * Registers the analytics factory to the SDKComponent.
+ * The key should match identifier defined in [SDKComponent.analyticsFactory].
+ * The method is placed in test sources to avoid exposing it to release builds.
+ */
+fun SDKComponent.registerAnalyticsFactory(
+    factory: (moduleConfig: DataPipelinesModuleConfig) -> Analytics
+) = registerDependency(identifier = "analyticsFactory") { factory }
```
Shahroz16, line 14: lets make the `analyticsFactory` an enum? to make sure we don't misspell

Shahroz16 (approved):

datapipelines/src/test/java/io/customer/datapipelines/extensions/DiGraphExt.kt
mrehan27: Updated 👍🏻

PR #327, opened 2024-05-10
test: add data pipeline interaction test

mrehan27:

### Changes

- Updated test dependencies in the DataPipelines module
- Added tests to `DataPipelinesInteractionTests` to validate changes
- Added base `UnitTest` to the DataPipelines module
- Set up `analytics` instance with the test environment for easier testing
- Added useful extensions for testing with `analytics`


buildSrc/src/main/kotlin/io.customer/android/Dependencies.kt
```diff
@@ -31,6 +31,8 @@ object Dependencies {
     const val gradleNexusPublishPlugin =
         "io.github.gradle-nexus:publish-plugin:${Versions.GRADLE_NEXUS_PUBLISH_PLUGIN}"
     const val junit4 = "junit:junit:${Versions.JUNIT4}"
+    const val junitBom = "org.junit:junit-bom:${Versions.JUNIT_BOM}"
```
Shahroz16, line 34: if we are adding `junitbom` do we need to specify version for `junit4` above?

datapipelines/src/test/java/io/customer/datapipelines/core/UnitTest.kt
```diff
@@ -0,0 +1,75 @@
...
+        analytics.add(outputReaderPlugin)
+    }
+
+    protected open fun createModuleInstance(
+        cdpApiKey: String = TEST_CDP_API_KEY,
+        applyConfig: CustomerIOBuilder.() -> Unit = {}
+    ): CustomerIO {
+        val builder = CustomerIOBuilder(mock(), cdpApiKey)
+        // Disable adding destination to analytics instance so events are not sent to the server by default
+        builder.setAutoAddCustomerIODestination(false)
```
Shahroz16, line 47: I believe there is an `enabled` flag too, somewhere, could utilizing that be more "near to use-case"?

datapipelines/src/test/java/io/customer/datapipelines/extensions/JsonExt.kt
Shahroz16, line 1: nice !!!!

datapipelines/src/test/java/io/customer/datapipelines/utils/Analytics.kt
```diff
@@ -0,0 +1,72 @@
...
+
+fun createTestAnalyticsInstance(moduleConfig: DataPipelinesModuleConfig): Analytics {
+    val configuration = createAnalyticsConfig(moduleConfig = moduleConfig)
+    configuration.application = "Test"
+    return object : Analytics(configuration, TestCoroutineConfiguration()) {}
+}
+
+fun Analytics.clearPersistentStorage() {
+    val writeKey = configuration.writeKey
+    File("/tmp/analytics-kotlin/$writeKey").deleteRecursively()
```
Shahroz16, line 71: where are we creating the temp storage?

datapipelines/src/test/java/io/customer/datapipelines/utils/Plugins.kt
```diff
@@ -0,0 +1,37 @@
...
+
+val OutputReaderPlugin.lastEvent: BaseEvent?
+    get() = allEvents.lastOrNull()
+
+val OutputReaderPlugin.identifyEvents: List<IdentifyEvent>
+    get() = allEvents.filterIsInstance<IdentifyEvent>()
+val OutputReaderPlugin.screenEvents: List<ScreenEvent>
+    get() = allEvents.filterIsInstance<ScreenEvent>()
+val OutputReaderPlugin.trackEvents: List<TrackEvent>
+    get() = allEvents.filterIsInstance<TrackEvent>()
```
Shahroz16, line 37: instead of repeating for every type and possibly adding more, would this be easier? open to keeping the current as well, if that improves readability. 

```suggestion
inline fun <reified T : BaseEvent> getEventsOfType(): List<T> {
        return allEvents.filterIsInstance<T>()
    }
```

buildSrc/src/main/kotlin/io.customer/android/Dependencies.kt
mrehan27: Yes because we are not modifying tests in `common-test` module in this ticket. When we work on those tests, we can surely remove this.

datapipelines/src/test/java/io/customer/datapipelines/core/UnitTest.kt
mrehan27: I followed similar approach in iOS, we can try `enabled` too but that fails for tests at the moment. We might need to mock more changes for it to work.

datapipelines/src/test/java/io/customer/datapipelines/utils/Analytics.kt
mrehan27: Aah yeah, that was missed I think, we can remove this method now.

datapipelines/src/test/java/io/customer/datapipelines/utils/Plugins.kt
mrehan27: I think it just kills the concept of these extensions. I added this to get rid of specifying classes. Otherwise, `plugin.allEvents.filterIsInstance<TrackEvent>` and `plugin.getEventsOfType<TrackEvent>()` is almost same to me and `getEventsOfType` looks unnecessary here. So I'll stick to current implementation for ease of casting and not specifying these classes every time.

datapipelines/src/test/java/io/customer/datapipelines/utils/Analytics.kt
mrehan27: Correction, we do need this in cleanup to remove files created in previous tests. So yes, analytics is creating this storage in test instance.

datapipelines/src/test/java/io/customer/datapipelines/core/UnitTest.kt
mrehan27: Looking at `analytics` code, I don't think enabled is going to work for tests

```kotlin
fun process(event: BaseEvent) {
        if (!enabled) return
```

datapipelines/src/test/java/io/customer/datapipelines/utils/Analytics.kt
Shahroz16: We are relying on the path that's being created when the storage provider is created, we should ideally provide our own path different from the provider. The application would utilize an Android storage provider so its now an issue for now

datapipelines/src/test/java/io/customer/datapipelines/core/UnitTest.kt
Shahroz16: yeah it seems like its stopping processing of events too.

Shahroz16 (approved):

datapipelines/src/test/java/io/customer/datapipelines/utils/Analytics.kt
mrehan27: Agreed. We can consider this improvement for future.

PR #329, opened 2024-05-12
feat: added screen APIs

Shahroz16:
closes: https://linear.app/customerio/issue/MBL-216/track-a-screen-view-manually

- Added public APIs to track screen events
- Update sample apps to utilize the new APIs

Tests are going to follow in the next PR when the base setup PR is merged.

### DIFF
--- a/datapipelines/src/main/kotlin/io/customer/sdk/DataPipelineInstance.kt
+++ b/datapipelines/src/main/kotlin/io/customer/sdk/DataPipelineInstance.kt
@@ -154,6 +154,57 @@ abstract class DataPipelineInstance : CustomerIOInstance {
     }
 
     /**
+     * The screen methods represents screen views in your mobile apps
+     *
+     * @param name A name for the screen.
+     * @param properties Additional details about the screen.
+     * @see [Learn more](https://customer.io/docs/cdp/sources/source-spec/screen-spec/)
+     */
+    @JvmOverloads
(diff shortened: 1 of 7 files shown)


datapipelines/src/main/kotlin/io/customer/sdk/DataPipelineInstance.kt
```diff
@@ -153,6 +153,57 @@ abstract class DataPipelineInstance : CustomerIOInstance {
...
+    }
+
+    /**
+     * The screen methods represents screen views in your mobile apps
+     *
+     * @param name A name for the screen.
+     * @param properties Additional details about the screen in Map <String, Any> format.
+     * @see [Learn more](https://customer.io/docs/cdp/sources/source-spec/screen-spec/)
+     */
+    fun screen(name: String, properties: CustomAttributes) {
```
mrehan27, line 175: I think we decided to use `title` here instead of `name` to match iOS and analytics?

```suggestion
    fun screen(title: String, properties: CustomAttributes) {
```

Shahroz16: hmm, I tried to follow https://customer.io/docs/cdp/sources/source-spec/screen-spec/ but seems like there is a discrepancy.

mrehan27 (approved):

PR #330, opened 2024-05-14
test: add data pipeline json test

mrehan27:

### Changes

- Added code and extensions to read and validate changes from storage
- Added tests to `DataPipelinesCompatibilityTests` to validate changes


Shahroz16 (approved):
Looks good, we should ideally move away from using `attributes` to `properties`

PR #332, opened 2024-05-14
test: identify user profile

mrehan27:

### Changes

- Added more tests to `DataPipelinesInteractionTests` and `DataPipelinesCompatibilityTests` for identify profile
- Moved `OutputReaderPlugin` out of `UnitTest` to avoid being misused
- Added helper extensions and `UserTraits` class to test package

### DIFF
--- a/datapipelines/src/test/java/io/customer/datapipelines/DataPipelinesInteractionTests.kt
+++ b/datapipelines/src/test/java/io/customer/datapipelines/DataPipelinesInteractionTests.kt
@@ -23,6 +23,8 @@ import org.amshove.kluent.shouldNotBeNull
 import org.junit.Test
 
 class DataPipelinesInteractionTests : UnitTest() {
+    //region Setup test environment
+
     private lateinit var outputReaderPlugin: OutputReaderPlugin
 
     override fun initializeModule() {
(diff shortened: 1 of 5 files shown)


datapipelines/src/test/java/io/customer/datapipelines/DataPipelinesInteractionTests.kt
```diff
@@ -54,4 +70,99 @@ class DataPipelinesInteractionTests : UnitTest() {
...
+        sdkInstance.identify(givenIdentifier)
+
+        analytics.userId().shouldBeNull()
+        analytics.traits() shouldBeEqualTo emptyJsonObject
+
+        outputReaderPlugin.allEvents.shouldBeEmpty()
+    }
+
+    @Test
+    fun identify_givenEmptyIdentifier_givenProfileAlreadyIdentified_expectRequestIgnored() {
```
Shahroz16, line 139: what happens if we previously identified a profile with traits, and then identified with empty identifier and empty traits, can we add a test for it?

Shahroz16 (approved):

datapipelines/src/test/java/io/customer/datapipelines/DataPipelinesInteractionTests.kt
mrehan27: Sure we can, added.

PR #337, opened 2024-05-16
chore: kotlin compose ui updates for data pipelines

mrehan27:
part of [MBL-299](https://linear.app/customerio/issue/MBL-299/update-sample-app-to-display-new-config-values)

### Changes

- Update Kotlin Compose UI to use new naming convention for DataPipelines
- Updated variable naming to match new implementation


samples/kotlin_compose/src/main/java/io/customer/android/sample/kotlin_compose/data/repositories/PreferenceRepository.kt
```diff
@@ -31,10 +31,10 @@ class PreferenceRepositoryImp(private val dataStore: DataStore<Preferences>) :
         dataStore.edit { preferences ->
             preferences[CDP_API_KEY] = configuration.cdpApiKey
             preferences[SITE_ID] = configuration.siteId
-            preferences[API_KEY] = configuration.apiKey
-            configuration.trackUrl?.let { preferences[TRACK_API_URL_KEY] = it }
-            preferences[BACKGROUND_QUEUE_SECONDS_DELAY] = configuration.backgroundQueueSecondsDelay
-            preferences[BACKGROUND_QUEUE_MIN_NUM_TASKS] = configuration.backgroundQueueMinNumTasks
+            configuration.apiHost?.let { preferences[API_HOST_KEY] = it }
+            configuration.cdnHost?.let { preferences[CDN_HOST_KEY] = it }
```
levibostian, line 35:
```suggestion
            preferences[API_HOST_KEY] = configuration.apiHost
```

I was expecting to see this, so that you could set the preferences value back to null if you wish. 

Is there a reason to first check for null?

samples/kotlin_compose/src/main/java/io/customer/android/sample/kotlin_compose/ui/settings/SettingsViewModel.kt
```diff
@@ -56,23 +59,23 @@ class SettingsViewModel @Inject constructor(
 
     private fun validateCustomTrackUrl(trackUrl: String?): String {
```
levibostian, line 60: nit-picky, but I think using "track" in the naming here could be confusing? Since we are no longer using word "track".

samples/kotlin_compose/src/main/java/io/customer/android/sample/kotlin_compose/ui/settings/SettingsViewModel.kt
```diff
@@ -56,23 +59,23 @@ class SettingsViewModel @Inject constructor(
...
-        }
-        // Passed all checks, return empty string
-        return ""
+
+        return runCatching {
+            val uri = Uri.parse(trackUrl)
+            // Since SDK does not support custom schemes, we manually append http:// to the URL
+            // So the URL is considered invalid if it ends with a slash, contains a scheme, query or fragment
+            return@runCatching when {
+                uri.scheme != null -> "URL should not include 'http://' or 'https://'."
```
levibostian, line 71: I think it's good for the app to allow the flexibility of setting a scheme. Is there a reason why we are changing this behavior from before?

samples/kotlin_compose/src/main/java/io/customer/android/sample/kotlin_compose/data/models/Configuration.kt
```diff
@@ -1,47 +1,29 @@
...
-    }
-    builder.autoTrackDeviceAttributes(this.trackDeviceAttributes)
-    builder.autoTrackScreenViews(this.trackScreen)
-    return builder
-}
-
 fun Configuration.setValuesFromBuilder(builder: CustomerIOBuilder): CustomerIOBuilder {
     builder.setMigrationSiteId(this.siteId)
-    builder.setAutoTrackDeviceAttributes(this.trackDeviceAttributes)
+    this.apiHost?.let { builder.setApiHost(it) }
```
levibostian, line 21:
```suggestion
    builder.setApiHost(this.apiHost) 
```

Similar comment as another - I think it's a good thing if we allow setting null or empty as an apihost, yeah?

levibostian (approved):

samples/kotlin_compose/src/main/java/io/customer/android/sample/kotlin_compose/data/repositories/PreferenceRepository.kt
mrehan27: `preferences[API_HOST_KEY]` actually doesn't allow `null` values, so it has to be non-null

samples/kotlin_compose/src/main/java/io/customer/android/sample/kotlin_compose/ui/settings/SettingsViewModel.kt
mrehan27: Yeah that was a miss 🤦🏻 Removed track from name here.

samples/kotlin_compose/src/main/java/io/customer/android/sample/kotlin_compose/ui/settings/SettingsViewModel.kt
mrehan27: The reason is the way analytics utilizing the values of [`apiHost`](https://github.com/segmentio/analytics-kotlin/blob/d2b3f183125efcd113de9fa0485795900162489b/core/src/main/java/com/segment/analytics/kotlin/core/HTTPClient.kt#L154) and [`cdnHost`](https://github.com/segmentio/analytics-kotlin/blob/d2b3f183125efcd113de9fa0485795900162489b/core/src/main/java/com/segment/analytics/kotlin/core/HTTPClient.kt#L143). To be more flexible here, we actually have to do a bit more parsing before passing the value matching what analytics need.

So I kept it simple to pass the exact value because

- This is sample app and not SDK API
- It is easier to write without `http` prefix than with it 😄

samples/kotlin_compose/src/main/java/io/customer/android/sample/kotlin_compose/data/models/Configuration.kt
mrehan27: The builder doesn't allow this today in SDK because the idea is, if you need to set `null`, don't call the setter as default value is already `null`. Setter should only be called if there is a need to set non-null value.

PR #338, opened 2024-05-16
chore: prevent error thrown when image URL cannot be composed because empty string

levibostian:
Fixes: https://linear.app/customerio/issue/MBL-308/error-javanetmalformedurlexception-no-protocol-on-receive-rich

Customer shared with us a push that they composed in Fly using the rich push composer, that does not have an image specified in the editor, but a "image" field gets added to the push payload with an empty string. The SDK handled when the "image" property is null, but not empty.

Customer is not experiencing a crash of the app, but they are getting stacktraces reported to them. So, this commit is a chore instead of a fix.

Testing:
* I added a new automated test. The test passes with the bug fix code, fails without it.

### DIFF
--- a/messagingpush/src/main/java/io/customer/messagingpush/CustomerIOPushNotificationHandler.kt
+++ b/messagingpush/src/main/java/io/customer/messagingpush/CustomerIOPushNotificationHandler.kt
@@ -240,11 +240,14 @@ internal class CustomerIOPushNotificationHandler(
         )
     }
 
-    private fun addImage(
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun addImage(
         imageUrl: String,
         builder: NotificationCompat.Builder,
         body: String
     ) = runBlocking {
+        if (imageUrl.isEmpty()) return@runBlocking
+
         val style = NotificationCompat.BigPictureStyle()
             .bigLargeIcon(null)
             .setSummaryText(body)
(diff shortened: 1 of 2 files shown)


messagingpush/src/main/java/io/customer/messagingpush/CustomerIOPushNotificationHandler.kt
```diff
@@ -240,11 +240,14 @@ internal class CustomerIOPushNotificationHandler(
...
     }
 
-    private fun addImage(
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun addImage(
         imageUrl: String,
         builder: NotificationCompat.Builder,
         body: String
     ) = runBlocking {
+        if (imageUrl.isEmpty()) return@runBlocking
```
mrehan27, line 249: I think we should avoid running into `runBlocking` when the URL is null or empty. We can exit early without needing to involve the code in `runBlocking`.

Optional suggestion: I would prefer if we had a single `isNullOrBlank` check for `imageUrl` instead of checking for `null` before calling this method and then for `empty` inside the method. I'm okay with moving the check either before calling this or inside the method.

mrehan27 (changes requested):
Just a few suggestions. Thanks for making these changes and adding test to validate it.

levibostian:
Converting to draft as we may not end up merging this but instead making a change in frontend/backend.

levibostian:
We made a backend change to fix this issue for the customer so this PR is not needed. 

There could still be some value in merging this PR, but that would mean we should make the change on iOS, too. The value in merging is little so I'll close this.

PR #340, opened 2024-05-21
refactor: moved device related classes to core

mrehan27:
part of [MBL-217](https://linear.app/customerio/issue/MBL-217/device-management)

### Changes

- Moved `ApplicationStore`, `BuildStore`, `DeviceStore` and `Client` to core module so it can be reused by native and wrapper Data Pipelines module to get device info
- Moved autogenerated `Version` file to core module and fixed related scripts


core/src/main/kotlin/io/customer/sdk/core/di/AndroidSDKComponent.kt
```diff
@@ -1,15 +1,35 @@
...
+    val client: Client
 ) : DiGraph() {
     val applicationContext: Context
         get() = newInstance { context.applicationContext }
+    val buildStore: BuildStore
+        get() = newInstance { BuildStoreImpl() }
+    val applicationStore: ApplicationStore
+        get() = newInstance { ApplicationStoreImpl(context = applicationContext) }
+    val deviceStore: DeviceStore
+        get() = singleton {
```
Shahroz16, line 28: if BuildStore is being recreated, maybe it makes sense to recreate DeviceStore as well?

PR #341, opened 2024-05-21
feat: track device with attributes

mrehan27:
part of [MBL-217](https://linear.app/customerio/issue/MBL-217/device-management)

### Changes

- Implemented and exposed public APIs to track device with attributes
- Added `GlobalPreferenceStore` to store common attributes e.g. device token

### Example Usage

```kotlin
CustomerIO.instance().registeredDeviceToken
CustomerIO.instance().deviceAttributes = mapOf("name" to "Brand", "os" to 33)
CustomerIO.instance().registerDeviceToken(token)
CustomerIO.instance().deleteDeviceToken
```

### Not Included

Tests and sample app updates are not included in this PR, will be added in upcoming PRs soon.

### DIFF
--- a/core/src/main/kotlin/io/customer/sdk/data/store/GlobalPreferenceStore.kt
+++ b/core/src/main/kotlin/io/customer/sdk/data/store/GlobalPreferenceStore.kt
@@ -1,0 +1,36 @@
+package io.customer.sdk.data.store
+
+import android.content.Context
+import androidx.core.content.edit
+
+/**
+ * Store for global preferences that are not tied to a specific api key, user
+ * or any other entity.
+ */
+interface GlobalPreferenceStore {
+    fun saveDeviceToken(token: String)
+    fun getDeviceToken(): String?
+
+    fun clearAll()
+}
+
+internal class GlobalPreferenceStoreImpl(
+    context: Context
+) : PreferenceStore(context), GlobalPreferenceStore {
+
+    override val prefsName: String by lazy {
+        "io.customer.sdk.${context.packageName}"
+    }
+
+    override fun saveDeviceToken(token: String) = prefs.edit {
+        putString(KEY_DEVICE_TOKEN, token)
+    }
+
+    override fun getDeviceToken(): String? = prefs.read {
+        getString(KEY_DEVICE_TOKEN, null)
+    }
+
+    companion object {
+        private const val KEY_DEVICE_TOKEN = "device_token"
+    }
+}
--- a/core/src/main/kotlin/io/customer/sdk/data/store/PreferenceStore.kt
+++ b/core/src/main/kotlin/io/customer/sdk/data/store/PreferenceStore.kt
@@ -1,0 +1,33 @@
+package io.customer.sdk.data.store
+
+import android.content.Context
+import android.content.SharedPreferences
+
+/**
+ * Base preference repository that can be reused among different preference repositories.
+ */
+internal abstract class PreferenceStore(val context: Context) {
+    abstract val prefsName: String
+
+    internal val prefs: SharedPreferences
+        get() = context.applicationContext.getSharedPreferences(
+            prefsName,
+            Context.MODE_PRIVATE
+        )
+
+    /**
+     * Clear all preferences stored in associated preference file asynchronously.
+     */
+    open fun clearAll() {
+        prefs.edit().clear().apply()
+    }
+}
+
+/**
+ * Retrieves value from the preference file using the provided action.
+ * This method handles any exceptions that might occur during the retrieval
+ * process and returns null in case of an exception.
+ */
+inline fun <Value> SharedPreferences.read(
+    action: SharedPreferences.() -> Value?
+): Value? = runCatching { action(this) }.getOrNull()
--- a/datapipelines/src/main/kotlin/io/customer/datapipelines/plugins/ContextPlugin.kt
+++ b/datapipelines/src/main/kotlin/io/customer/datapipelines/plugins/ContextPlugin.kt
@@ -1,0 +1,33 @@
+package io.customer.datapipelines.plugins
+
+import com.segment.analytics.kotlin.core.Analytics
+import com.segment.analytics.kotlin.core.BaseEvent
+import com.segment.analytics.kotlin.core.platform.Plugin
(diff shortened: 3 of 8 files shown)


core/src/main/kotlin/io/customer/sdk/data/store/GlobalPreferenceStore.kt
```diff
@@ -0,0 +1,36 @@
...
+interface GlobalPreferenceStore {
+    fun saveDeviceToken(token: String)
+    fun getDeviceToken(): String?
+
+    fun clearAll()
+}
+
+internal class GlobalPreferenceStoreImpl(
+    context: Context
+) : PreferenceStore(context), GlobalPreferenceStore {
```
mrehan27, line 19: This is same as [`SitePreferenceRepository` in tracking](https://github.com/customerio/customerio-android/blob/c8cbb12eea9502a2ad8be4560e365d194124f59d/sdk/src/main/java/io/customer/sdk/repository/preference/SitePreferenceRepository.kt) today. The naming is updated to be more flexible.

core/src/main/kotlin/io/customer/sdk/data/store/PreferenceStore.kt
```diff
@@ -0,0 +1,33 @@
+package io.customer.sdk.data.store
+
+import android.content.Context
+import android.content.SharedPreferences
+
+/**
+ * Base preference repository that can be reused among different preference repositories.
+ */
+internal abstract class PreferenceStore(val context: Context) {
```
mrehan27, line 9: same as [`BasePreferenceRepository` in tracking](https://github.com/customerio/customerio-android/blob/c8cbb12eea9502a2ad8be4560e365d194124f59d/sdk/src/main/java/io/customer/sdk/repository/preference/BasePreferenceRepository.kt).

datapipelines/src/main/kotlin/io/customer/sdk/android/CustomerIO.kt
```diff
@@ -157,11 +186,70 @@ class CustomerIO private constructor(
...
+            return
+        }
+
+        val existingDeviceToken = contextPlugin.deviceToken
+        if (existingDeviceToken != null && existingDeviceToken != token) {
+            logger.debug("token has been refreshed, deleting old token to avoid registering same device multiple times")
+            deleteDeviceToken()
+        }
+
+        // TODO: Append auto tracked device attributes here
```
mrehan27, line 233: This will be covered in [MBL-215](https://linear.app/customerio/issue/MBL-215/automatic-device-attributes-collection-and-user-agent-updates)

datapipelines/src/main/kotlin/io/customer/datapipelines/plugins/ContextPlugin.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    fun updateDeviceProperties(deviceToken: String?, attributes: CustomAttributes) {
+        this.deviceToken = deviceToken
+        this.attributes = attributes
+    }
+
+    override fun execute(event: BaseEvent): BaseEvent {
+        deviceToken?.let { token ->
+            // Device token is expected to be attached to device in context
+            event.putInContextUnderKey("device", "token", token)
```
Shahroz16, line 29: I think this needs to be optional, if `device.token` already exists, should we still override it? this could get problematic in migration or in case device token changes in real time.

datapipelines/src/main/kotlin/io/customer/datapipelines/plugins/ContextPlugin.kt
```diff
@@ -0,0 +1,33 @@
...
+import com.segment.analytics.kotlin.core.BaseEvent
+import com.segment.analytics.kotlin.core.platform.Plugin
+import com.segment.analytics.kotlin.core.utilities.putInContextUnderKey
+import io.customer.sdk.data.model.CustomAttributes
+
+/**
+ * Plugin class responsible for updating the context properties in events
+ * tracked by Customer.io SDK.
+ */
+class ContextPlugin(override var analytics: Analytics) : Plugin {
```
Shahroz16, line 13: we dont need to add `analytics` here as it would be added and set by timeline itself.

datapipelines/src/main/kotlin/io/customer/sdk/android/CustomerIO.kt
```diff
@@ -157,11 +186,70 @@ class CustomerIO private constructor(
...
+
+        logger.debug("resetting user profile")
         analytics.reset()
     }
 
+    override val registeredDeviceToken: String?
+        get() = globalPreferenceStore.getDeviceToken()
+
+    override var deviceAttributes: CustomAttributes
+        get() = emptyMap()
```
Shahroz16, line 202: if possible lets check if analytics is holding these in memory/disk and we can just return those?

datapipelines/src/main/kotlin/io/customer/sdk/android/CustomerIO.kt
```diff
@@ -157,11 +186,70 @@ class CustomerIO private constructor(
...
+
+    override fun registerDeviceToken(deviceToken: String) {
+        if (deviceToken.isBlank()) {
+            logger.debug("device token cannot be blank. ignoring request to register device token")
+            return
+        }
+
+        logger.info("registering device token $deviceToken for user profile: $registeredUserId")
+
+        logger.debug("storing device token to device storage")
```
Shahroz16, line 215:
```suggestion
        logger.info("storing and registering device token $deviceToken for user profile: $registeredUserId")
```

datapipelines/src/main/kotlin/io/customer/sdk/android/CustomerIO.kt
mrehan27: If analytics is holding it somewhere, it will be good to return, but doesn't look like so. So let's stick to this for now.

datapipelines/src/main/kotlin/io/customer/datapipelines/plugins/ContextPlugin.kt
mrehan27: Good call, I think we should consider this with migration changes. I'll find ticket and update it on Linear too.

Shahroz16 (approved):
👍

PR #344, opened 2024-05-23
chore: sample apps api keys with prefix

mrehan27:
### Background

I realized it's harder to connect both of these sample apps to different workspaces because of the way we currently fetch keys. So I made some changes that allow us to provide optional different keys for the sample apps, with fallback to current format so we can reuse the same keys as well. These changes shouldn't break anything on CI or on local machines for any developer.

### Changes

- Updated the sample apps' `build.gradle` script to prefer keys with prefix
- If the prefix is not provided or prefix keys are missing, it falls back to keys without prefix
- Both sample apps are updated to mention their prefix

### Example `local.properties` File

```
# Used by both sample apps
apiKey=API_KEY
siteId=SITE_ID

# Ignored as both sample apps have overridden the value
cdpApiKey=CDP_API_KEY

# Used by Java Layout only
javaLayout_cdpApiKey=JAVA_CDP

# Used by Kotlin Compose only
kotlinCompose_cdpApiKey=KOTLIN_CDP
```

### DIFF
--- a/samples/java_layout/build.gradle
+++ b/samples/java_layout/build.gradle
@@ -4,6 +4,7 @@ plugins {
 }
 
 // Include Customer.io SDK dependencies and common gradle properties for sample apps
+ext.appConfigKeyPrefix = 'javaLayout_'
 apply from: "${rootDir}/samples/sample-app.gradle"
 
 android {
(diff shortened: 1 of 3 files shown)


Shahroz16 (approved):
can be helpful, thanks

PR #346, opened 2024-05-23
chore: fix memory leaks in data pipelines module

mrehan27:
### Changes

- Used new logger reference in nested log helper objects to avoid memory leaks
- Reset static DIGraph in `clearInstance`


Shahroz16 (approved):
Beautiful 🙇

PR #348, opened 2024-05-24
chore: separate non-test classes into support package

mrehan27:
### Background

I realized it was hard to differentiate between test files and files containing helper code for writing tests. So I separated them to make it easier to differentiate.

### Changes

- Moved all helper non-test classes to the `support` package
- Ensured all test files match the package of the class they are testing


Shahroz16 (commented):
How do you feel about these support classes being in a separate module? if number of support files keeps on increasing (even right now they are more than the tests), it would make it super easy to differentiate between test and support module.

mrehan27:
@Shahroz16 I don't think having a separate module just for helper test classes will help. I believe it will complicate our project structure and limit the ability to access internal classes within them. So I see more cons than pros in having a different module for this case. Can you please list the benefits you think this option will provide?

Shahroz16 (approved):
Happy to move forward with this, if this looks more of a hassle right now, my only concern is we have 11 support files as compared to 2 actual test, which is fine because we need these classes. But if the trajectory remains the same and we have all classes public already then I can see more value it containing all support classes together to not over crowd the data pipeline classes.

PR #349, opened 2024-05-27
chore: use mockk instead of mockito in data pipelines

mrehan27:

### Background

I noticed we have been using both `Mockito` and `MockK` in Data Pipelines module. While this isn't necessarily an issue, we should only use both if needed. For functionalities that can be handled by both, we should stick to our preferred choice for consistency. I prefer `MockK` over `Mockito` because:

- Our project codebase is mostly Kotlin
- `MockK` is more Kotlin-friendly and easier to integrate
- It has a DSL-like syntax and is easy to adopt in Kotlin
- It allows mocking/spying methods for final classes

I haven't removed `Mockito` from the module yet, I only updated it to see how it works. If everything goes well, we can later remove `Mockito` from the module/project.

### Changes

- Used `MockK` for mocking objects instead of `Mockito`
- Replaced verification and mocking calls from `Mockito` with `MockK`

### DIFF
--- a/datapipelines/src/test/java/io/customer/datapipelines/sdk/CustomerIOBuilderTest.kt
+++ b/datapipelines/src/test/java/io/customer/datapipelines/sdk/CustomerIOBuilderTest.kt
@@ -12,15 +12,14 @@ import io.customer.sdk.core.di.SDKComponent
 import io.customer.sdk.core.util.CioLogLevel
 import io.customer.sdk.data.model.Region
 import io.customer.sdk.extensions.random
+import io.mockk.every
+import io.mockk.mockk
+import io.mockk.verify
 import org.amshove.kluent.shouldBe
 import org.amshove.kluent.shouldNotBe
 import org.amshove.kluent.shouldNotBeEqualTo
 import org.junit.Test
 import org.junit.runner.RunWith
-import org.mockito.kotlin.mock
-import org.mockito.kotlin.never
-import org.mockito.kotlin.verify
-import org.mockito.kotlin.whenever
 import org.robolectric.Robolectric
 import org.robolectric.RobolectricTestRunner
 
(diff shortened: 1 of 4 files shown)


datapipelines/src/test/java/io/customer/datapipelines/sdk/CustomerIOBuilderTest.kt
```diff
@@ -36,53 +35,57 @@ class CustomerIOBuilderTest : BaseUnitTest() {
...
 
         createCustomerIOBuilder()
             .addCustomerIOModule(givenModule1)
             .addCustomerIOModule(givenModule2)
             .build()
 
-        verify(givenModule1).initialize()
-        verify(givenModule2).initialize()
+        verify(exactly = 1) { givenModule1.initialize() }
+        verify(exactly = 1) { givenModule2.initialize() }
```
Shahroz16, line 70:
```suggestion
        verifyCount {
            1 * { givenModule1.initialize() } // same as verify(exactly = 1) {  }
            1 * { givenModule2.initialize() } // same as verify(exactly = 1) { }
        }
```

datapipelines/src/test/java/io/customer/datapipelines/support/core/UnitTest.kt
```diff
@@ -48,21 +49,29 @@ abstract class UnitTest {
...
+            context = mockk(relaxed = true),
             client = Client.Android(Version.version)
         )
         // Mock global preference store to avoid reading/writing to shared preferences
-        globalPreferenceStore = mock()
-        androidSDKComponent.overrideDependency(GlobalPreferenceStore::class.java, globalPreferenceStore)
+        globalPreferenceStore = mockk<GlobalPreferenceStore>(relaxUnitFun = true).also { instance ->
+            androidSDKComponent.overrideDependency(GlobalPreferenceStore::class.java, instance)
+        }
+        every { globalPreferenceStore.getDeviceToken() } returns null
```
Shahroz16, line 67: why do we need this? it wasn't before?

Shahroz16 (approved):
Can we please update mockk version too, if we are relying on it solely? `verifyCount` is only available in newer version. `1.13.11` is latest.

datapipelines/src/test/java/io/customer/datapipelines/support/core/UnitTest.kt
mrehan27: What we were previously doing with `Mockito` was actually using relaxed mocks. Here, I used `relaxUnitFun = true` so for functions expected to return any non-unit value, we have to explicitly mention them. So we have more control over them and can have clarity on test results with the expected behavior of these functions.

datapipelines/src/test/java/io/customer/datapipelines/sdk/CustomerIOBuilderTest.kt
mrehan27: This does sound good, but unfortunately, it requires newer version of Kotlin as it gives the following error:

> /Users/mrehan/.gradle/caches/transforms-3/b2c3058d9c2ad25669f5c8e3f7c7515a/transformed/jetified-kotlin-stdlib-common-1.9.10.jar!/META-INF/kotlin-stdlib-common.kotlin_module: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.9.0, expected version is 1.7.1.

Merging this for now. We can update this later when we upgrade the Kotlin version.

PR #351, opened 2024-05-29
chore: allow modifying config in data pipelines test setup

mrehan27:

### Changes

- Moved the creation of `AndroidSDKComponent` earlier with `CustomerIOBuilder` initialization so tests don't have to explicitly create it and can reuse it to override dependencies before initializing the module.
- Added `JUnitTest` and `RobolectricTest` classes to avoid rewriting setup code and to make writing both types of tests easier
- Updated `UnitTest` for more flexibility. Tests should avoid directly inheriting it and use `JUnitTest` or `RobolectricTest` instead
- Added `TestConfiguration` to wrap setup parameters in a single class for flexibility
- Added a DSL-like option for convenient initialization of `TestConfiguration`
- Fixed parent classes for tests

### DIFF
--- a/datapipelines/src/test/java/io/customer/datapipelines/support/core/TestConfiguration.kt
+++ b/datapipelines/src/test/java/io/customer/datapipelines/support/core/TestConfiguration.kt
@@ -1,0 +1,55 @@
+package io.customer.datapipelines.support.core
+
+import com.segment.analytics.kotlin.core.Analytics
+import io.customer.sdk.CustomerIOBuilder
+
+/**
+ * Test configuration for setting up the CustomerIO SDK and analytics instance.
+ *
+ * @property cdpApiKey CDP API key to be used for testing
+ * @property sdkConfig used to configure the CustomerIO SDK instance for each test separately
+ * @property configurePlugins used to configure plugins for analytics instance before
+ * we add more plugins to it in CustomerIO initialization
+ */
+class TestConfiguration private constructor(
+    val cdpApiKey: String,
+    val sdkConfig: CustomerIOBuilder.() -> Unit,
+    val configurePlugins: Analytics.() -> Unit
+) {
+    class Builder {
+        private var cdpApiKey: String = TEST_CDP_API_KEY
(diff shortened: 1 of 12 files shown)


datapipelines/src/test/java/io/customer/datapipelines/support/stubs/UnitTestLogger.kt
```diff
@@ -0,0 +1,37 @@
...
+        }
+    }
+
+    override fun error(message: String) {
+        runIfMeetsLogLevelCriteria(CioLogLevel.ERROR) {
+            println("[ERROR]: $message")
+        }
+    }
+
+    private fun runIfMeetsLogLevelCriteria(levelForMessage: CioLogLevel, block: () -> Unit) {
```
Shahroz16, line 32: too complicated for a test? 😄 maybe let's turn it on by default ? and not log when not needed?

datapipelines/src/test/java/io/customer/datapipelines/support/core/TestConfiguration.kt
```diff
@@ -0,0 +1,55 @@
...
+        fun build(): TestConfiguration = TestConfiguration(
+            cdpApiKey = cdpApiKey,
+            sdkConfig = sdkConfig,
+            configurePlugins = configurePlugins
+        )
+
+        companion object {
+            const val TEST_CDP_API_KEY = "TESTING_API_KEY"
+        }
+    }
```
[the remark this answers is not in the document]
Shahroz16: lets just use DSL !! 🪨

Shahroz16 (approved):
just few comments to make implementation lighter

datapipelines/src/test/java/io/customer/datapipelines/support/stubs/UnitTestLogger.kt
mrehan27: Agreed. But turning on make logs a bit messy. Simplified it more for testing.

PR #352, opened 2024-05-29
test: auto track device attributes

mrehan27:

### Changes

- Fixed tests breaking in `DataPipelinesCompatibilityTests` for device registration
- Added tests for auto track device attributes in `DeviceAttributesTests`
- Updated `DeviceStoreStub` and helper extensions to help writing tests

### DIFF
--- a/datapipelines/src/test/java/io/customer/datapipelines/DeviceAttributesTests.kt
+++ b/datapipelines/src/test/java/io/customer/datapipelines/DeviceAttributesTests.kt
@@ -1,0 +1,131 @@
+package io.customer.datapipelines
+
+import android.Manifest
+import android.content.pm.PackageManager
+import com.segment.analytics.kotlin.android.plugins.AndroidContextPlugin
+import io.customer.datapipelines.support.core.RobolectricTest
+import io.customer.datapipelines.support.core.TestConfiguration
+import io.customer.datapipelines.support.core.testConfiguration
+import io.customer.datapipelines.support.extensions.deviceToken
+import io.customer.datapipelines.support.extensions.encodeToJsonValue
+import io.customer.datapipelines.support.utils.OutputReaderPlugin
+import io.customer.datapipelines.support.utils.TestConstants
+import io.customer.datapipelines.support.utils.trackEvents
+import io.customer.sdk.extensions.random
+import io.mockk.every
+import kotlinx.serialization.json.booleanOrNull
+import kotlinx.serialization.json.intOrNull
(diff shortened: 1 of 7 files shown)


datapipelines/src/test/java/io/customer/datapipelines/DeviceAttributesTests.kt
```diff
@@ -0,0 +1,126 @@
...
+    fun track_givenAutoTrackDeviceAttributesDisabled_expectNoDeviceAttributesInProperties() {
+        setupTestEnvironment(
+            testConfiguration {
+                sdkConfig {
+                    // Disable auto tracking of device attributes
+                    setAutoTrackDeviceAttributes(false)
+                }
+                configurePlugins {
+                    // Add Android context plugin so that device attributes can be tracked by analytics
+                    add(AndroidContextPlugin())
```
Shahroz16, line 65: doesn't it get added by default?

datapipelines/src/test/java/io/customer/datapipelines/DeviceAttributesTests.kt
```diff
@@ -0,0 +1,126 @@
...
+import org.amshove.kluent.shouldNotBeNullOrBlank
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+
+@RunWith(RobolectricTestRunner::class)
+@Config(manifest = Config.NONE)
+class DeviceAttributesTests : RobolectricTest() {
```
Shahroz16, line 34: i think one test missing would be having both custom attribute and these built-in device attribute and make sure we get all of them.

datapipelines/src/test/java/io/customer/datapipelines/DeviceAttributesTests.kt
mrehan27: Unfortunately not in tests. It is because the plugin is added when using constructor for Android. But since we want to override schedulers with test ones, we have to use Java constructor which by default doesn't add this plugin and we have to include it manually.

datapipelines/src/test/java/io/customer/datapipelines/DeviceAttributesTests.kt
mrehan27: Good call, added test for it

Shahroz16 (approved):
👍

PR #354, opened 2024-05-30
chore: added event types

Shahroz16:
closes: [MBL-344 : Event types](https://linear.app/customerio/issue/MBL-344/event-types)

- added event types
- updated tests to cater new types

### DIFF
--- a/core/src/main/kotlin/io/customer/sdk/communication/Event.kt
+++ b/core/src/main/kotlin/io/customer/sdk/communication/Event.kt
@@ -4,18 +4,18 @@ import java.util.Date
 import java.util.UUID
 
 /**
- * Base class containing common properties for all events.
- */
-abstract class BaseEvent(
-    open val storageId: String = UUID.randomUUID().toString(),
-    open val params: Map<String, String> = emptyMap(),
-    open val timestamp: Date = Date()
-)
-
-/**
  * Class to represent an event that can be published and subscribed to using [EventBus].
  */
-sealed class Event : BaseEvent() {
+sealed class Event {
+    // Unique identifier for the event
+    open val storageId: String = UUID.randomUUID().toString()
+
+    // Additional metadata associated with the event
+    open val params: Map<String, String> = emptyMap()
+
+    // Timestamp of the event
+    open val timestamp: Date = Date()
+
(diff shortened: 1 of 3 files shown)


core/src/test/kotlin/io/customer/sdk/core/di/EventBusTest.kt
```diff
@@ -27,43 +27,43 @@ class EventBusTest : BaseUnitTest() {
...
-        val testEvent = Event.DummyEvent("Test Message")
+        val testEvent = Event.ProfileIdentifiedEvent("Test Message")
         println("Publishing event: $testEvent")
         eventBus.publish(testEvent)
 
         delay(100) // Give some time for the event to be collected
 
         events.size shouldBeEqualTo 1
-        events[0].message shouldBeEqualTo testEvent.message
+        (events[0] as Event.ProfileIdentifiedEvent).identifier shouldBeEqualTo testEvent.identifier
```
mrehan27, line 42: Instead of casting directly, I think we should have something like this

```suggestion
        events
          .shouldHaveSingleItem()
          .shouldBeInstanceOf<Event.ProfileIdentifiedEvent>()
          .identifier shouldBeEqualTo testEvent.identifier
```

core/src/main/kotlin/io/customer/sdk/communication/Event.kt
```diff
@@ -3,22 +3,44 @@ package io.customer.sdk.communication
...
+    open val timestamp: Date = Date()
+)
+
 /**
  * Class to represent an event that can be published and subscribed to using [EventBus].
  */
-sealed class Event {
-    // Unique identifier for the event
-    val storageId: String = UUID.randomUUID().toString()
+sealed class Event : BaseEvent() {
```
mrehan27, line 18: Why do we need BaseEvent? I think we can have something like this instead

```suggestion
sealed class Event {
    open val storageId: String = UUID.randomUUID().toString(),
    open val params: Map<String, String> = emptyMap(),
    open val timestamp: Date = Date()
```

mrehan27 (approved):
Minor comments, rest looks good 👍🏻

PR #355, opened 2024-05-31
fix: in-app message displayed on wrong page

mrehan27:
part of [MBL-353](https://linear.app/customerio/issue/MBL-353/android-fix-in-app-message-displayed-on-wrong-page)

### Changes

- Revalidated current route just before making the in-app message visible
- If the message to be displayed has page rules that don't match the current route, dismiss the message instead of displaying it using `dismissMessage`


levibostian:
Code LGTM. 

I suggest that before merging, [check this test case I wrote](https://linear.app/customerio/issue/MBL-353/android-fix-in-app-message-displayed-on-wrong-page#comment-99f1512d) and make sure it passes with this implementation.

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebViewClientInterceptor.kt
```diff
@@ -0,0 +1,14 @@
+package io.customer.messaginginapp.gist.presentation.engine
+
+import android.webkit.WebView
+
+/**
+ * WebViewClient interceptor for EngineWebView
+ * This interface is used to intercept the WebViewClient events while testing
+ */
+internal interface EngineWebViewClientInterceptor {
```
mrehan27, line 9: Added for tests only

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebViewInterface.kt
```diff
@@ -15,13 +16,34 @@ internal data class EngineWebEvent(
...
+        isAttachedToWebView = false
+        // Remove JavaScript interface from WebView
+        webView.removeJavascriptInterface(JAVASCRIPT_INTERFACE_NAME)
+    }
+
     @JavascriptInterface
     fun postMessage(message: String) {
+        // Check if the interface is attached to WebView
+        // If not, do not process the message
+        if (!isAttachedToWebView) {
```
mrehan27, line 43: `removeJavascriptInterface` method doesn't work as I expected and keeps sending messages here. So I added flag to prevent sending calls to native when we know that the message has been cancelled.

mrehan27:
PR is updated and ready for review.

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistSdk.kt
```diff
@@ -107,12 +115,57 @@ object GistSdk : Application.ActivityLifecycleCallbacks {
...
+     * Cancels any active message being loaded if the page rule does not match the new route.
+     */
+    private fun cancelActiveMessage() {
+        val currentMessage = gistModalManager.currentMessage
+        val currentMessageProperties = currentMessage?.let { message ->
+            GistMessageProperties.getGistProperties(message)
+        }
+
+        // If there is no active message or the message does not have a route rule, we don't need to do anything
+        if (currentMessage == null || currentMessageProperties?.routeRule == null) {
```
Shahroz16, line 159: we need to also check for regex, in case the route rule isn't null but regex matches we don't need to cancel it.

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
```diff
@@ -55,11 +70,12 @@ internal class EngineWebView @JvmOverloads constructor(
...
                 it.settings.textZoom = 100
                 it.setBackgroundColor(Color.TRANSPARENT)
-                it.addJavascriptInterface(EngineWebViewInterface(this), "appInterface")
+                engineWebViewInterface.attach(webView = it)
 
                 it.webViewClient = object : WebViewClient() {
                     override fun onPageFinished(view: WebView, url: String?) {
-                        view.loadUrl("javascript:window.parent.postMessage = function(message) {window.appInterface.postMessage(JSON.stringify(message))}")
+                        view.loadUrl("javascript:window.parent.postMessage = function(message) {window.${EngineWebViewInterface.JAVASCRIPT_INTERFACE_NAME}.postMessage(JSON.stringify(message))}")
+                        engineWebViewClientInterceptor?.onPageFinished(view, url)
```
Shahroz16, line 78: future refactor, we can extract the testing client from one used in code.

Shahroz16 (changes requested):
There are couple of refactors needed but that's mostly on Gist part which we can add in future update, but major blocker is when we cancel the message

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistSdk.kt
mrehan27: Good catch, updated PR

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
mrehan27: It surely would make the code look much nicer, but that needs a bit more refactoring of these classes which I avoided in this PR to avoid scope creep.

messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/listeners/Queue.kt
```diff
@@ -250,6 +249,16 @@ class Queue : GistListener {
...
     override fun onMessageDismissed(message: Message) {}
 
+    override fun onMessageCancelled(message: Message) {
+        val gistProperties = GistMessageProperties.getGistProperties(message)
+        val routeRule = gistProperties.routeRule
+        Log.i(
+            GIST_TAG,
+            "Message cancelled. Adding message back to queue with route rule: $routeRule"
+        )
+        addMessageToLocalStore(message)
```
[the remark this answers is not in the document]
mrehan27: Updated this to match implementation with iOS

Shahroz16 (commented):
Think the mulitple messages in queue changes aren't added in as well

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistModalActivity.kt
```diff
@@ -138,6 +153,22 @@ class GistModalActivity : AppCompatActivity(), GistListener, GistViewListener {
         }
     }
 
+    override fun onMessageCancelled(message: Message) {
+        Log.i(GIST_TAG, "Message Cancelled")
+        currentMessage?.let { currentMessage ->
+            if (currentMessage.instanceId == message.instanceId) {
```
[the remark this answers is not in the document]
mrehan27: This is a good point, but looking at Android code, we are mostly using `instanceId` in all other methods of same listener. So I would lean towards not making the change in this PR, and consider this in future improvement.

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistModalManager.kt
```diff
@@ -47,6 +47,12 @@ internal class GistModalManager : GistListener {
         }
     }
 
+    override fun onMessageCancelled(message: Message) {
+        if (message.instanceId == currentMessage?.instanceId) {
```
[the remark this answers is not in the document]
mrehan27: Same as above, we are mostly using `instanceId` today in Android

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistSdk.kt
```diff
@@ -280,6 +341,7 @@ interface GistListener {
     fun embedMessage(message: Message, elementId: String)
     fun onMessageShown(message: Message)
     fun onMessageDismissed(message: Message)
+    fun onMessageCancelled(message: Message)
```
levibostian, line 344: Is this interface part of the public API for customers? 

I am OK adding this method if it's for internal code, only.

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistView.kt
```diff
@@ -30,6 +30,10 @@ class GistView @JvmOverloads constructor(
     private var firstLoad: Boolean = true
     var listener: GistViewListener? = null
 
+    // Indicates if the message is visible to user or not
+    internal val isEngineVisible: Boolean
+        get() = engineWebView?.alpha == 1.0f
```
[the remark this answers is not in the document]
levibostian: I agree, communicating with the UI layer (compared to business layer) is the most reliable way to determine if a message is truly shown or not. 

To double-check, when the message is dismissed and when cancelled, is the alpha not 1.0?

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
```diff
@@ -36,6 +38,19 @@ internal class EngineWebView @JvmOverloads constructor(
         } catch (e: Exception) {
             Log.e(GIST_TAG, "Error while creating EngineWebView: ${e.message}")
         }
+        // Get WebViewClientInterceptor from GistSdk directly
+        engineWebViewClientInterceptor = GistSdk.engineWebViewClientInterceptor
```
levibostian, line 42: Instead of a strong reference being held in this class, would it be better memory management to change the class property to a getter that accesses the singleton each time requested? 

``kt
private var engineWebViewClientInterceptor: EngineWebViewClientInterceptor?
  get() = GistSdk.engineWebViewClientInterceptor
```

If we need a strong reference for a specific reason, a comment added would be helpful to make sure we always do that.

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistSdk.kt
Shahroz16: It's `internal` only

Shahroz16 (approved):
Looks good 😅

PR #356, opened 2024-06-06
chore: integrated event bus datapipeline

Shahroz16:
closes: [MBL-345 : Integrate Event Bus in Data Pipelines](https://linear.app/customerio/issue/MBL-345/integrate-event-bus-in-data-pipelines)

integrated `EventBus` with DataPipeline module and added tests.

### DIFF
--- a/core/src/main/kotlin/io/customer/sdk/communication/EventBus.kt
+++ b/core/src/main/kotlin/io/customer/sdk/communication/EventBus.kt
@@ -4,12 +4,9 @@ import io.customer.sdk.core.di.SDKComponent
 import kotlin.reflect.KClass
 import kotlin.reflect.cast
 import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
-import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.SharedFlow
-import kotlinx.coroutines.flow.filter
 import kotlinx.coroutines.flow.filterIsInstance
 import kotlinx.coroutines.flow.mapNotNull
 import kotlinx.coroutines.launch
@@ -39,7 +36,7 @@ class EventBusImpl(
 
     val jobs = mutableListOf<Job>()
 
-    val scope: CoroutineScope = SDKComponent.androidSDKComponent?.scopeProvider?.eventBusScope ?: CoroutineScope(Dispatchers.Default + SupervisorJob())
+    val scope: CoroutineScope = SDKComponent.scopeProvider.eventBusScope
 
     override fun publish(event: Event) {
         scope.launch {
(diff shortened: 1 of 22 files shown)


core/src/main/kotlin/io/customer/sdk/communication/EventBus.kt
```diff
@@ -32,8 +39,7 @@ class EventBusImpl(
 
     val jobs = mutableListOf<Job>()
 
-    val scope: CoroutineScope = SDKComponent.androidSDKComponent?.scopeProvider?.eventBusScope
-        ?: CoroutineScope(Dispatchers.Default + SupervisorJob())
+    val scope: CoroutineScope = SDKComponent.androidSDKComponent?.scopeProvider?.eventBusScope ?: CoroutineScope(Dispatchers.Default + SupervisorJob())
```
mrehan27, line 42: I think there should be consistent way to get scope here. As with this code, if the EventBus is somehow accessed before SDK is initialized, the scope would be different.

I just checked, and I think we can safely move the scope to `SDKComponent` for guaranteed access. Is there a reason for restricting it to `AndroidSDKComponent` that I missed? 🤔

core/src/main/kotlin/io/customer/sdk/communication/EventBus.kt
```diff
@@ -55,4 +61,17 @@ class EventBusImpl(
         jobs.forEach { it.cancel() }
         jobs.clear()
     }
+
+    override fun <T : Event> subscribe(type: KClass<T>, action: suspend (T) -> Unit): Job {
+        val job = scope.launch {
+            flow.filter { type.isInstance(it) }.mapNotNull { it.safeCast(type) }.collect { event ->
```
mrehan27, line 67: Feels like unnecessary looping. Can we do something like this?

```suggestion
            flow.mapNotNull { it.safeCast(type) }.collect { event ->
```

core/src/main/kotlin/io/customer/sdk/communication/EventBus.kt
```diff
@@ -55,4 +61,17 @@ class EventBusImpl(
...
+                action(event)
+            }
+        }
+        jobs.add(job)
+        return job
+    }
+    private fun <T : Event> Event.safeCast(type: KClass<T>): T? {
+        return if (type.isInstance(this)) type.cast(this) else null
```
mrehan27, line 75: Can't we use Kotlin's predefined safeCast here like this?

```suggestion
        return type.safeCast(this)
```

Or may be remove this method and just use Kotlin one 🤔 

```kotlin
IdentifyEvent::class.safeCast(event)
```

mrehan27 (approved):
Mostly looks good, just some comments that we should address before merging.

core/src/main/kotlin/io/customer/sdk/communication/EventBus.kt
Shahroz16: Makes sense, I wanted to keep it in `AndroidSDKComponent` so the modules who don't need it won't initialize then `AndroidSDKComponent` and we won't have this dependency added to them for no reason. But scope and dispatchers seems more of reusable component so happy to move there.

core/src/main/kotlin/io/customer/sdk/communication/EventBus.kt
Shahroz16: good catch, it was left from when I was force casting 😄

core/src/main/kotlin/io/customer/sdk/communication/EventBus.kt
Shahroz16: I used it first but then I saw this
```
@WasExperimental(ExperimentalStdlibApi::class)
```
and thought it said experimental but it said `was` :D

PR #357, opened 2024-06-11
test: in-app message displayed on wrong page

mrehan27:
part of [MBL-353](https://linear.app/customerio/issue/MBL-353/android-fix-in-app-message-displayed-on-wrong-page)

### Changes

- Updated `BaseIntegrationTest` to allow adding modules when initializing the SDK
- Added tests to validate that in-app messages should not be displayed on the wrong page
- Added Gist helper extensions to allow mocking the behavior of actual in-app message


messaginginapp/src/androidTest/java/io/customer/messaginginapp/support/GistEngineWebViewClientInterceptor.kt
```diff
@@ -0,0 +1,140 @@
...
+            return
+        }
+
+        val engineMessage = EngineWebMessage(
+            gist = EngineWebEvent(
+                instanceId = driver.message.instanceId,
+                method = "routeLoaded",
+                parameters = buildMap {
+                    put("route", route)
+                }
```
Shahroz16, line 91:
```suggestion
                parameters = mapOf("route" to route)
```

`buildMap` seems overkill considering the use-case? 😄

Shahroz16 (approved):
This can be alot lighter but given the way webclient is setup right now, this does the job ! thank you for adding this.

PR #362, opened 2024-06-19
chore: replace tracking hooks with event bus

Shahroz16:
closes: [MBL-348 : Remove tracking requests](https://linear.app/customerio/issue/MBL-348/remove-tracking-requests)

- Removes hooks and other tracking usage with EventBus
- Adds internal dependency of digraph to SDKComponent


datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
```diff
@@ -115,6 +115,15 @@ class CustomerIO private constructor(
 
         // subscribe to journey events emitted from push/in-app module to send them via data pipelines
         subscribeToJourneyEvents()
+        // if profile is already identified, republish identifier for late-added modules.
+        postProfileAlreadyIdentified()
+    }
+
+    private fun postProfileAlreadyIdentified() {
+        // TODO: Need to check SitePreferences for userId and then publish event
```
mrehan27, line 123: Curious why we need to check SitePreferences? Isn't `analytics.userId()` doing the same? 🤔

messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
```diff
@@ -85,38 +79,47 @@ internal constructor(
...
+            gistProvider.clearUserToken()
+        }
     }
 
+    // TODO: Remove config and replace it with moduleConfig
     private fun initializeGist(config: CustomerIOConfig) {
         gistProvider.initProvider(
-            application = diGraph.context.applicationContext as Application,
+            // TODO: This should not be nullable
+            application = diGraph.androidSDKComponent?.applicationContext as Application,
```
mrehan27, line 120: Similar to [our earlier discussion](https://github.com/customerio/customerio-android/pull/359#discussion_r1638943077). Outside of this PR though, but I think we need to improve how we use `AndroidSDKComponent` in our modules as they all have public constructors like you pointed out.

For this line, I think we need to use safe cast?
```suggestion
            application = diGraph.androidSDKComponent?.applicationContext as? Application,
```

Or may be use `let` or early exit here as `GistProvider` may not allow nullable Application instance.

mrehan27 (approved):
Looks good 👍🏻

messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
Shahroz16: `AndroidSDKComponent` needs to be non-nullable otherwise we will keep on hitting this but as you pointed out, future update

datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
Shahroz16: `analytics.userId()` will be set only when you call `identify` but we can have a case when you have an identifier stored that will only be in `SitePreferences` so we need to adjust the migration so that it always gets called before this method, until then we need to probably rely on SitePreferences

PR #363, opened 2024-06-21
chore: inapp config builder

mrehan27:
part of: [MBL-377](https://linear.app/customerio/issue/MBL-377/decouple-tests-from-tracking)

### Changes

- Updated `MessagingInAppModuleConfig.Builder` to include required options (i.e. `siteId` and `region`)
- Removed unused code from `ModuleMessagingInApp`
- Updated sample apps
- Minor improvements in related code

### DIFF
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
@@ -1,70 +1,30 @@
 package io.customer.messaginginapp
 
 import android.app.Application
-import androidx.annotation.VisibleForTesting
 import io.customer.messaginginapp.di.gistProvider
-import io.customer.sdk.CustomerIOConfig
-import io.customer.sdk.android.CustomerIO
 import io.customer.sdk.communication.Event
 import io.customer.sdk.communication.subscribe
 import io.customer.sdk.core.di.SDKComponent
 import io.customer.sdk.core.module.CustomerIOModule
 import io.customer.sdk.data.request.MetricEvent
-import io.customer.sdk.di.CustomerIOComponent
 
-class ModuleMessagingInApp
-@VisibleForTesting
-internal constructor(
-    override val moduleConfig: MessagingInAppModuleConfig = MessagingInAppModuleConfig.default(),
-    private val overrideDiGraph: CustomerIOComponent?
+class ModuleMessagingInApp(
+    config: MessagingInAppModuleConfig
 ) : CustomerIOModule<MessagingInAppModuleConfig> {
-    @JvmOverloads
-    @Deprecated(
-        "organizationId no longer required and will be removed in future",
-        replaceWith = ReplaceWith("constructor(config: MessagingInAppModuleConfig)")
-    )
-    constructor(
-        organizationId: String,
-        config: MessagingInAppModuleConfig = MessagingInAppModuleConfig.default()
-    ) : this(
-        moduleConfig = config,
-        overrideDiGraph = null
-    )
-
-    @JvmOverloads
-    constructor(
-        config: MessagingInAppModuleConfig = MessagingInAppModuleConfig.default()
-    ) : this(
-        moduleConfig = config,
-        overrideDiGraph = null
-    )
-
-    override val moduleName: String = ModuleMessagingInApp.moduleName
+    override val moduleName: String = MODULE_NAME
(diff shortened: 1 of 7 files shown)


messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
```diff
@@ -1,70 +1,30 @@
...
 import io.customer.sdk.data.request.MetricEvent
-import io.customer.sdk.di.CustomerIOComponent
 
-class ModuleMessagingInApp
-@VisibleForTesting
-internal constructor(
-    override val moduleConfig: MessagingInAppModuleConfig = MessagingInAppModuleConfig.default(),
-    private val overrideDiGraph: CustomerIOComponent?
+class ModuleMessagingInApp(
+    config: MessagingInAppModuleConfig
```
mrehan27, line 12: I used `config` as customers who were using named parameters will need to change otherwise. Also, I think `config` here is simpler than `moduleConfig` for customers. But open to use `moduleConfig` directly if we think that's not adding much value.

messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
```diff
@@ -1,70 +1,30 @@
...
     private val logger = diGraph.logger
 
-    private val oldDiGraph: CustomerIOComponent
-        get() = overrideDiGraph ?: CustomerIO.instance().diGraph
-
-    private val config: CustomerIOConfig
-        get() = oldDiGraph.sdkConfig
-
-    companion object {
-        const val moduleName: String = "MessagingInApp"
```
mrehan27, line 59: Renamed and moved to bottom of class for consistency

messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
```diff
@@ -1,70 +1,30 @@
...
-        overrideDiGraph = null
-    )
-
-    @JvmOverloads
-    constructor(
-        config: MessagingInAppModuleConfig = MessagingInAppModuleConfig.default()
-    ) : this(
-        moduleConfig = config,
-        overrideDiGraph = null
-    )
```
mrehan27, line 40: Removed unused and deprecated constructors

messaginginapp/src/main/java/io/customer/messaginginapp/di/DIGraphMessagingInApp.kt
```diff
@@ -13,7 +14,10 @@ internal val SDKComponent.gistApiProvider: GistApi
 internal val SDKComponent.gistProvider: InAppMessagesProvider
     get() = newInstance<InAppMessagesProvider> { GistInAppMessagesProvider(gistApiProvider) }
 
-fun SDKComponent.inAppMessaging(): ModuleMessagingInApp {
-    return modules[ModuleMessagingInApp.moduleName] as? ModuleMessagingInApp
+// We need to add extension functions to the CustomerIO class as this is how
+// customers will interact with in-app messaging module.
+@Suppress("UnusedReceiverParameter")
+fun CustomerIO.inAppMessaging(): ModuleMessagingInApp {
+    return SDKComponent.modules[ModuleMessagingInApp.MODULE_NAME] as? ModuleMessagingInApp
```
mrehan27, line 21: The comment above should help explain reason for reverting it back to CustomerIO extension

mrehan27: Reverted change as I realized we'll be removing `CustomerIO` dependency from in-app where the code will create problem. We need to improve how customers interact with these modules, but that's a different problem and we'll solve in another PRs.

messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
Shahroz16: Lets make a follow up ticket to remove from docs, if they are still there?

Shahroz16 (approved):

messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
mrehan27: Sure, will do 👍🏻

PR #364, opened 2024-06-21
chore: use MockK instead of Mockito for in-app messages tests

mrehan27:
part of: [MBL-377](https://linear.app/customerio/issue/MBL-377/decouple-tests-from-tracking)

### Changes

- Moved `messaginginapp` tests from `sharedTest` package to `test` package for convenience
- Updated `messaginginapp` tests to use `MockK` instead of `Mockito` and `EventBus` for hooks
- Moved `UnitTestLogger` to the `commontest` module for reuse
- Fixed `EventBus` scope for tests
- Added `UnitTest` and `JUnitTest` classes to `messaginginapp` tests for more convenient test writing
- Minor fixes in related code


messaginginapp/build.gradle
```diff
@@ -51,4 +51,25 @@ dependencies {
     implementation Dependencies.okhttpLoggingInterceptor
     testImplementation Dependencies.androidxTestJunit
     androidTestImplementation Dependencies.junit4
```
Shahroz16, line 53: are these all still needed?

Shahroz16 (approved):

messaginginapp/build.gradle
mrehan27: Good call, removed

PR #365, opened 2024-06-24
chore: clean up tracking dependency from push module

mrehan27:
part of: [MBL-335](https://linear.app/customerio/issue/MBL-335/send-fcm-token-to-eventbus)

### Changes

- Updated code to reduce `push` dependency on `tracking` module
- Removed SDK initialization checks
- Extensions and improvements in DI graphs
- Code comments and minor updates

### Not Included

Tests updates are not included in this PR, will be added in upcoming PRs soon.


core/src/main/kotlin/io/customer/sdk/core/di/AndroidSDKComponent.kt
```diff
@@ -20,8 +21,10 @@ class AndroidSDKComponent(
     val context: Context,
     val client: Client
 ) : DiGraph() {
+    val application: Application
+        get() = newInstance<Application> { context.applicationContext as Application }
```
mrehan27, line 25: Ideally, application context should always be an instance of Application. So I don't think we need to handle any nullable scenario here, but open to ideas.

core/src/main/kotlin/io/customer/sdk/core/di/SDKComponent.kt
```diff
@@ -20,13 +20,26 @@ import io.customer.sdk.core.util.SdkScopeProvider
...
+    val logger: Logger
+        get() = singleton<Logger> { LogcatLogger(buildEnvironment = buildEnvironment) }
+
+    // Communication dependencies
+    val eventBus: EventBus
+        get() = singleton<EventBus> { EventBusImpl() }
+    val dispatchersProvider: DispatchersProvider
+        get() = newInstance<DispatchersProvider> { SdkDispatchers() }
+    val scopeProvider: ScopeProvider
+        get() = newInstance<ScopeProvider> { SdkScopeProvider(dispatchersProvider) }
```
mrehan27, line 42: Just grouped these as it was becoming hard to read and added explicit types to avoid unwanted test results.

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOCloudMessagingReceiver.kt
```diff
@@ -27,8 +26,6 @@ class CustomerIOCloudMessagingReceiver : BroadcastReceiver() {
         val extras = intent.extras
         // Ignore event if no data was received in extras
         if (extras == null || extras.isEmpty) return
-        // If CustomerIO instance isn't initialized, we cannot process the notification
-        val sdkInstance = context.getSDKInstanceOrNull() ?: return
```
mrehan27, line 31: This is important because it changes how metrics work. Unfortunately, it is tough to test with native as it is mostly utilized in wrappers. But because EventBus replays event, delayed SDK initialization shouldn't be issue here IMO.

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingService.kt
```diff
@@ -58,9 +57,6 @@ open class CustomerIOFirebaseMessagingService : FirebaseMessagingService() {
             remoteMessage: RemoteMessage,
             handleNotificationTrigger: Boolean = true
         ): Boolean {
-            // if CustomerIO instance isn't initialized, we can't handle the notification
-            val sdkInstance = context.getSDKInstanceOrNull() ?: return false
```
mrehan27, line 62: Same as above.

messagingpush/src/main/java/io/customer/messagingpush/processor/PushMessageProcessorImpl.kt
```diff
@@ -89,7 +88,7 @@ internal class PushMessageProcessorImpl(
         if (moduleConfig.autoTrackPushEvents) {
             eventBus.publish(
                 Event.TrackPushMetricEvent(
-                    event = MetricEvent.delivered.name,
+                    event = Metric.Delivered.name,
```
mrehan27, line 91: @Shahroz16 Do we need to use `Metric.Delivered.name.lowercase(Locale.ENGLISH)` instead for this and similar changes? Or will it work without the transformation?

messagingpush/src/main/java/io/customer/messagingpush/provider/FCMTokenProviderImpl.kt
```diff
@@ -5,7 +5,14 @@ import com.google.android.gms.common.ConnectionResult
 import com.google.android.gms.common.GoogleApiAvailability
 import com.google.firebase.messaging.FirebaseMessaging
 import io.customer.sdk.core.di.SDKComponent
-import io.customer.sdk.device.DeviceTokenProvider
+
+/**
+ *  Responsible for token generation and validity
+ */
+interface DeviceTokenProvider {
```
mrehan27, line 12: Moved from `tracking` module

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOCloudMessagingReceiver.kt
Shahroz16: We should be able to verify this with a test as well

Shahroz16: lets a todo here? for test

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingService.kt
Shahroz16: same as above

messagingpush/src/main/java/io/customer/messagingpush/processor/PushMessageProcessorImpl.kt
Shahroz16: I think we need a class that lives in `core` with the format that CDP expects, so then we can just use name easily everywhere

Shahroz16 (approved):
added some minor suggestions, could be done in another PR if needed

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOCloudMessagingReceiver.kt
mrehan27: Sure, will try adding test for it. Adding todo for now.

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOFirebaseMessagingService.kt
mrehan27: Added todo

messagingpush/src/main/java/io/customer/messagingpush/processor/PushMessageProcessorImpl.kt
mrehan27: Added extension method

PR #368, opened 2024-06-26
chore: move lifecycle observer to core module

mrehan27:
part of: [MBL-337](https://linear.app/customerio/issue/MBL-337/track-push-metrics)

### Changes

- Added `ActivityLifecycleCallbacks` observer to `core` module so any module can listen to it
- Exposed lifecycle flow so observers can leverage it for listening to new events and last emitted event
- Updated `push` and `in-app` module to observe newly added lifecycle callback from `core`
- Minor improvements in DI graph

### Not Included

Tests updates are not included in this PR, will be updated in another ticket specific to test structure


core/src/main/kotlin/io/customer/sdk/core/util/ScopeProvider.kt
```diff
@@ -5,9 +5,12 @@ import kotlinx.coroutines.SupervisorJob
...
 interface ScopeProvider {
     val eventBusScope: CoroutineScope
+    val lifecycleListenerScope: CoroutineScope
 }
 
 class SdkScopeProvider(private val dispatchers: DispatchersProvider) : ScopeProvider {
     override val eventBusScope: CoroutineScope
         get() = CoroutineScope(dispatchers.default + SupervisorJob())
+    override val lifecycleListenerScope: CoroutineScope
+        get() = CoroutineScope(dispatchers.default + SupervisorJob())
```
mrehan27, line 15: Although we are accessing these coroutines only once, I still think we should update them as follows to avoid recreating them every time.

```
override val eventBusScope: CoroutineScope = CoroutineScope(dispatchers.default + SupervisorJob())
override val lifecycleListenerScope: CoroutineScope = CoroutineScope(dispatchers.default + SupervisorJob())
```

core/src/main/kotlin/io/customer/sdk/lifecycle/CustomerIOActivityLifecycleCallbacks.kt
```diff
@@ -0,0 +1,86 @@
...
+import kotlinx.coroutines.flow.SharedFlow
+import kotlinx.coroutines.launch
+
+/**
+ * SDK component to listen to lifecycle events of all activities.
+ * The class is responsible for registering and unregistering the lifecycle callbacks.
+ * SDK clients can subscribe to lifecycle events of all activities using the [subscribe] function
+ * and should avoid listening to events from ActivityLifecycleCallbacks directly.
+ */
+class CustomerIOActivityLifecycleCallbacks : Application.ActivityLifecycleCallbacks {
```
mrehan27, line 20: This is supposed to be the only `ActivityLifecycleCallbacks` listener in SDK, rest of the modules can listen to and use replay to get last event.

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistSdk.kt
```diff
@@ -134,6 +135,20 @@ object GistSdk : Application.ActivityLifecycleCallbacks {
...
     }
 
+    private fun subscribeToLifecycleEvents() {
+        SDKComponent.activityLifecycleCallbacks.subscribe { events ->
+            events.filter { state ->
+                state.event == Lifecycle.Event.ON_RESUME || state.event == Lifecycle.Event.ON_PAUSE
+            }.collect { state ->
+                when (state.event) {
+                    Lifecycle.Event.ON_RESUME -> onActivityResumed(state.activity)
+                    Lifecycle.Event.ON_PAUSE -> onActivityPaused(state.activity)
```
mrehan27, line 145: These are not on main thread now, but looking at code, I think we don't need them to be on main specifically? 🤔

Shahroz16 (commented):
Majorly looks good because going further in review, the thing i am worried about is, we are adding this functionality in `common` because feature modules use it. So if someone just intends to use us for `CDP/Tracking` we are still observing lifecycle even though no module would be subscribing to it?

mrehan27:
@Shahroz16 Good point, but I'm not much concerned about it because

- This is [same as in `main` today](https://github.com/customerio/customerio-android/blob/bfd8e28158c56e97e74f6ab8634200756b6545c3/sdk/src/main/java/io/customer/sdk/CustomerIO.kt#L367), so we are not adding anything extra with the change
- The number of customer not using any of other feature modules is significantly low, so the use case is very small

We can still check on modules count and skip listener if needed. But I see this being more error-prone and adding less value, so would not suggest to go with it.

core/src/main/kotlin/io/customer/sdk/core/di/SDKComponent.kt
```diff
@@ -23,6 +24,9 @@ object SDKComponent : DiGraph() {
     // Static map to store all the modules registered with the SDK
     val modules: MutableMap<String, CustomerIOModule<*>> = mutableMapOf()
 
+    // Android specific dependencies
```
Shahroz16, line 27: should this be in `AndroidSDKComponent`?

Shahroz16 (approved):

core/src/main/kotlin/io/customer/sdk/core/di/SDKComponent.kt
mrehan27:
```suggestion
    // Android component dependencies
```

PR #371, opened 2024-07-01
refactor: update test file paths for migration

mrehan27:

### Changes

- Updated class names and moved classes to the paths where they are expected to end up with final changes for improved testing
- This will help us maintain file history
- This will simplify upcoming PRs by showing content changes instead of showing files deleted and added due to bigger changes in a single file
- Moved classes out of `sharedTest` package as we'll be removing them. Only classes that will no longer be needed after upcoming changes remain, and they will be removed later.

### Note

The project and tests compile with these changes, so we can safely merge this into the feature branch to simplify PR reviews and avoid conflicts for upcoming changes.


common-test/src/main/java/io/customer/commontest/extensions/IntExtensions.kt
```diff
@@ -0,0 +1,9 @@
+package io.customer.commontest.extensions
+
+import java.util.Random
+
+fun Int.Companion.random(min: Int, max: Int): Int {
```
mrehan27, line 5: The extension was available in production code of `tracking` module. Since we only need this for tests, I moved them up here to `commontest` module.

Since `tracking` module uses in production code as well, the imports are not changed for `tracking` and only updated for rest of the modules.

common-test/src/main/java/io/customer/commontest/extensions/StringExtensions.kt
```diff
@@ -0,0 +1,9 @@
+package io.customer.commontest.extensions
+
+fun String.Companion.random(length: Int): String {
```
mrehan27, line 3: Same as above extension.

datapipelines/src/test/java/io/customer/datapipelines/testutils/core/UnitTestDelegate.kt
```diff
@@ -25,7 +25,7 @@ import io.mockk.spyk
  * by child classes which allows running tests with or without depending on Android
  * context and resources.
  */
-abstract class UnitTest {
+abstract class UnitTestDelegate {
```
mrehan27, line 28: This is renamed to delegate because we cannot inherit multiple classes. So future updates to base test classes in `datapipelines` will utilize it through composition to avoid code duplication and simplify base setup.

messagingpush/src/sharedTest/java/io/customer/messagingpush/PlaceholderTest.kt
mrehan27, line 1: Removed it as it looked like not needed anymore.

Shahroz16 (approved):
looks good

PR #372, opened 2024-07-01
chore: decouple tracking tests

mrehan27:

### Changes

- Moved all classes from `sharedTest` to `test` so we can safely remove `sharedTest` from gradle script in upcoming PRs
- Copied some files from `commontest` to `tracking` so the `commontest` package can be updated without breaking `tracking` tests

### Note

The project and tests compile with these changes, so we can safely merge this into the feature branch to simplify PR reviews and avoid conflicts for upcoming changes.


common-test/src/main/java/io/customer/commontest/core/BaseTest.kt
```diff
@@ -51,7 +50,6 @@ abstract class BaseTest {
         get() = di.jsonAdapter
 
     protected lateinit var mockWebServer: MockWebServer
-    protected lateinit var dateUtilStub: DateUtilStub
```
mrehan27, line 54: Removed so this can safely be moved to `tracking`, as we no longer rely on this for dates.

sdk/src/androidTest/java/io/customer/sdk/util/commontest/BaseTest.kt
```diff
@@ -0,0 +1,141 @@
...
+import org.junit.After
+import org.junit.Before
+import retrofit2.HttpException
+import retrofit2.Response
+
+/**
+ * Base class for a unit test class to subclass. If you want to create integration tests, use [RobolectricTest].
+ * Meant to provide convenience to test classes with properties and functions tests may use.
+ */
+abstract class BaseTest {
```
mrehan27, line 30: This is copy of what `BaseTest` is today in `commontest`. This is copied so we can updated `BaseTest` without breaking `tracking` module tests.

sdk/src/androidTest/java/io/customer/sdk/util/commontest/DateUtilStub.kt
```diff
@@ -0,0 +1,20 @@
+package io.customer.sdk.util.commontest
+
+import io.customer.base.extenstions.getUnixTimestamp
+import io.customer.sdk.util.DateUtil
+import java.util.*
+
+/**
+ * Convenient alternative to mocking [DateUtil] in your test since the code is boilerplate.
+ */
+class DateUtilStub : DateUtil {
```
mrehan27, line 10: Moved to `tracking` as no longer needed by other modules

sdk/src/test/java/io/customer/sdk/commontest/BaseTest.kt
```diff
@@ -0,0 +1,141 @@
...
+import org.junit.After
+import org.junit.Before
+import retrofit2.HttpException
+import retrofit2.Response
+
+/**
+ * Base class for a unit test class to subclass. If you want to create integration tests, use [RobolectricTest].
+ * Meant to provide convenience to test classes with properties and functions tests may use.
+ */
+abstract class BaseTest {
```
mrehan27, line 30: Because all files moved to `test` and we are no longer using `sharedTest`, we need to copy all files twice. One for `androidTest` and other for `test`. This is only for tracking module. Rest of the modules will leverage `commontest` to write common code.

Shahroz16 (approved):

PR #373, opened 2024-07-02
chore: decouple common-test from tracking

mrehan27:
part of [MBL-394](https://linear.app/customerio/issue/MBL-394/decouple-common-tests-from-tracking-module)

### Background

Since the changes in this ticket will cause some modules' tests to fail compilation until the changes are complete, this PR is created to hold all changes for the ticket until we merge it into base feature branch. All future PRs will be created on top of it. Once the stack is complete and merged, we'll merge this into the base feature branch so all modules can compile and run successfully.

### Changes

- Renamed `mockk` dependency to `mockK`.

### PR Stack

- [x] #374 
- [x] #377 
- [x] #378 
- [x] #379 
- [x] #380

PR #374, opened 2024-07-02
refactor: common-test base test classes

mrehan27:

### Background

Previously, `common-test` reduced a lot of boilerplate code for tests by providing base test classes. However, since we have now decoupled our modules from `tracking` and made several updates, base classes needed to be updated so they can be leveraged in all modules. This PR addresses these updates to ensure `common-test` can benefit all modules in the codebase today. The goals include:

- Decouple from `tracking` module and depend only on `core` module
- Include base classes that are not strongly tied to SDK initialization and can be reused by all modules, including independent feature modules
- Simplify setting up tests while remaining flexible

### Changes

- Simplified `test` and `androidTest` dependencies
- Removed concept of `sharedTest` so tests can be run and debugged individually
- Introduced test configuration to simplify modifying test setup options
- Introduced more base test classes to reduce code weight in base test classes
- Simplified overriding dependencies in DI graphs
- Added extensions to help write test code

#### Providing Test Configuration

```kotlin
testConfigurationDefault {
    diGraph {
        sdk { overrideDependency<Logger>(UnitTestLogger()) }
        android { overrideDependency<Application>(TestApplication()) }
    }
}
```

#### Combining Test Configurations

```kotlin
override fun setup(testConfig: TestConfig) {
    super.setup(defaultTestConfiguration + testConfig)
}
```

#### Overriding Dependencies

```kotlin
diGraph.overrideDependency<Logger>(UnitTestLogger())
```

#### Verifying Mock Calls

```kotlin
verifyOnce { mock.methodName() }
verifyNever { mock.methodName() }
verifyNoInteractions { mock }
```

### Notes

See [parent PR](https://github.com/customerio/customerio-android/pull/373) for more details and context.


base/build.gradle
```diff
@@ -32,6 +32,7 @@ dependencies {
     implementation Dependencies.coroutinesCore
     implementation Dependencies.androidxAnnotations
 
+    testImplementation platform(Dependencies.junitBom)
```
mrehan27, line 35: We might not `base` module in future, but that's something to evaluate later. Added this to fix for now.

core/src/main/kotlin/io/customer/sdk/core/di/DiGraph.kt
```diff
@@ -105,8 +105,9 @@ abstract class DiGraph {
      * `di.overrideDependency(CustomerIOInstance::class.java, customerIOMock)`
      * See `CustomerIOTest` (or any other class that is top-level of the module) for an example on use.
      */
-    fun <Dependency : Any> overrideDependency(dependency: Class<Dependency>, value: Dependency) {
-        overrides[dependency.name] = value as Any
+    inline fun <reified Dependency : Any> overrideDependency(value: Dependency, identifier: String? = null) {
```
mrehan27, line 108: Because we only use this in tests, I tried to simplify this too.

base/build.gradle
```diff
@@ -32,6 +32,6 @@ dependencies {
     implementation Dependencies.coroutinesCore
     implementation Dependencies.androidxAnnotations
 
-    testImplementation Dependencies.junit4
+    testImplementation "junit:junit:4.13.2"
```
mrehan27, line 35: We might not `base` module in future, but that's something to evaluate later. Added this to fix for now.

common-test/src/main/java/io/customer/commontest/config/DefaultTestConfig.kt
```diff
@@ -0,0 +1,32 @@
+package io.customer.commontest.config
+
+/**
+ * Default implementation of [TestConfig] that contains a list of test arguments
+ * and DI graph configuration.
+ * The configuration can be combined with other configurations using the [plus] operator.
+ */
+class DefaultTestConfig internal constructor(
```
Shahroz16, line 8: `TestConfigBuilder` seems to be doing same thing? maybe we can simplify by just utilizing it and not adding another layer on top of it?

Shahroz16: something like 

```
class TestConfigBuilder {
    private val arguments = mutableListOf<TestArgument>()
    private val diGraph = DIGraphConfiguration()

    fun argument(arg: TestArgument) {
        arguments.add(arg)
    }

    fun diGraph(block: DIGraphConfiguration.() -> Unit) {
        diGraph.apply(block)
    }

    fun build(): TestConfig = object : TestConfig {
        override val arguments: List<TestArgument> = this@TestConfigBuilder.arguments
        override val diGraph: DIGraphConfiguration = this@TestConfigBuilder.diGraph

        override fun plus(other: TestConfig): TestConfig = TestConfigBuilder().apply {
            arguments.addAll(this@TestConfig.arguments)
            arguments.addAll(other.arguments)
            diGraph.plus(other.diGraph)
        }.build()
    }
}

// DSL function for creating TestConfig
fun testConfig(block: TestConfigBuilder.() -> Unit): TestConfig =
    TestConfigBuilder().apply(block).build()
```

common-test/src/main/java/io/customer/commontest/config/TestConfigExtensions.kt
```diff
@@ -0,0 +1,58 @@
...
+ * If more than one argument of type [Arg] is found, throws an [IllegalArgumentException].
+ */
+inline fun <reified Arg : TestArgument> TestConfig.argumentOrNull(): Arg? {
+    val args = arguments.filterIsInstance<Arg>()
+    return when {
+        args.isEmpty() -> return null
+        args.size == 1 -> args[0]
+        else -> throw IllegalArgumentException("Expected exactly one argument of type ${Arg::class.java.simpleName}, but found ${args.size}")
+    }
+}
```
Shahroz16, line 33:
```suggestion
inline fun <reified Arg : TestArgument> TestConfig.argumentOrNull(): Arg? =
    arguments.filterIsInstance<Arg>().singleOrNull()
```

common-test/src/main/java/io/customer/commontest/extensions/MockKExtensions.kt
Shahroz16, line 1:
```
object MockKExtensions {
    inline fun verifyOnce(verifyBlock: MockKVerificationScope.() -> Unit) = 
        verify(exactly = 1, verifyBlock = verifyBlock)

    inline fun verifyNever(verifyBlock: MockKVerificationScope.() -> Unit) = 
        verify(exactly = 0, verifyBlock = verifyBlock)

    fun verifyNoInteractions(vararg mocks: Any) = 
        verify { mocks.toList() wasNot Called }
}
```

`verifyOnce` etc are very common names and might conflict with others

common-test/src/main/java/io/customer/commontest/core/BaseTest.kt
```diff
@@ -1,137 +1,54 @@
...
-
-    // override in test class to override SDK config for all test functions in the class.
-    protected open fun setupConfig() = createConfig()
-
-    // default @Before for tests using a default SDK config set
-    @Before
-    open fun setup() {
-        setup(cioConfig = setupConfig())
+    protected open fun teardown() {
+        clearAllComponents()
```
Shahroz16, line 36: might as well just call from the method, why do we need another method? 

```
SDKComponent.reset()
clearAllMocks()
```

Shahroz16 (approved):
great additions, minor suggestions and improvements

common-test/src/main/java/io/customer/commontest/config/DIGraphConfiguration.kt
```diff
@@ -0,0 +1,29 @@
...
+        private set
+    var androidSDKComponent: ConfigDSL<AndroidSDKComponent> = {}
+        private set
+
+    operator fun plus(other: DIGraphConfiguration): DIGraphConfiguration {
+        val copy = DIGraphConfiguration()
+        copy.sdkComponent = sdkComponent + other.sdkComponent
+        copy.androidSDKComponent = androidSDKComponent + other.androidSDKComponent
+        return copy
+    }
```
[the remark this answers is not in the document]
mrehan27: Yes, updated

common-test/src/main/java/io/customer/commontest/config/DefaultTestConfig.kt
mrehan27: This is one nice way to look at this 😄 I think it might be less flexible, but as long as it works, I have no strong preference. We can simply update implementation here later if need, the callers should not be affected.

common-test/src/main/java/io/customer/commontest/config/TestConfigExtensions.kt
mrehan27: Aah, I kept looking for similar method and literally forgot about this one 🤦🏻 Thanks for sharing

common-test/src/main/java/io/customer/commontest/extensions/MockKExtensions.kt
mrehan27: Good call. I wanted to add global functions for ease of use, so updated as follows

```
assertCalledOnce
assertCalledNever
assertNoInteractions
```

common-test/src/main/java/io/customer/commontest/core/BaseTest.kt
mrehan27: Yeah, removed

common-test/src/main/java/io/customer/commontest/config/DefaultTestConfig.kt
mrehan27: I tried this, and plus operator needs builder to be concrete which doesn't work well with DataPipelineTestConfig. So leaving it as it is for now, we can improve this later after datapipeline updates if needed.

PR #375, opened 2024-07-02
chore: updated constructor to not account for task builder

Shahroz16:
closes: [MBL-396 : Revisit Push Module Config to Eliminate Redundant Configurations](https://linear.app/customerio/issue/MBL-396/revisit-push-module-config-to-eliminate-redundant-configurations)

improves customer experience and dev experience.

### DIFF
--- a/messagingpush/src/main/java/io/customer/messagingpush/data/communication/CustomerIOPushNotificationCallback.kt
+++ b/messagingpush/src/main/java/io/customer/messagingpush/data/communication/CustomerIOPushNotificationCallback.kt
@@ -20,7 +20,9 @@ interface CustomerIOPushNotificationCallback {
     fun onNotificationClicked(
         payload: CustomerIOParsedPushPayload,
         context: Context
-    ): Unit? = null
+    ): Unit? {
+        return null
+    }
 
     /**
      * Called when all attributes for the notification has been set by the SDK
--- a/messagingpush/src/main/java/io/customer/messagingpush/data/communication/CustomerIOPushNotificationCallback.kt
+++ b/messagingpush/src/main/java/io/customer/messagingpush/data/communication/CustomerIOPushNotificationCallback.kt
@@ -2,7 +2,6 @@ package io.customer.messagingpush.data.communication
 
 import android.content.Context
 import androidx.core.app.NotificationCompat
-import androidx.core.app.TaskStackBuilder
 import io.customer.messagingpush.data.model.CustomerIOParsedPushPayload
 
 interface CustomerIOPushNotificationCallback {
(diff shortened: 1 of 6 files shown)


messagingpush/src/main/java/io/customer/messagingpush/data/communication/CustomerIOPushNotificationCallback.kt
```diff
@@ -11,27 +10,27 @@ interface CustomerIOPushNotificationCallback {
...
      * open metrics tracking
      */
-    fun createTaskStackFromPayload(
-        context: Context,
-        payload: CustomerIOParsedPushPayload
-    ): TaskStackBuilder? = null
+    fun onNotificationClicked(
+        payload: CustomerIOParsedPushPayload,
+        context: Context
+    ): Unit? = null
```
mrehan27, line 23: If I'm thinking right, we're here trying to simplify so if customers just implement the function, it means we should not process further. To stop any further action, they should return `null` explicitly. I like the idea though, but I think this could be confusing and error-prone. Can we stick to simple `boolean` instead for simplification? 🤔 

Might not be very good example though, but [`onInterceptTouchEvent`](https://developer.android.com/reference/android/view/ViewGroup#onInterceptTouchEvent(android.view.MotionEvent)) and many others use `boolean` for similar cases. And I think we should stick with boolean too 😄

messagingpush/src/main/java/io/customer/messagingpush/processor/PushMessageProcessorImpl.kt
```diff
@@ -139,13 +139,13 @@ internal class PushMessageProcessorImpl(
...
         val notificationCallback = moduleConfig.notificationCallback
-        val taskStackFromPayload = notificationCallback?.createTaskStackFromPayload(
+        val wasClicked = notificationCallback?.onNotificationClicked(
             context = activityContext,
             payload = payload
         )
-        if (taskStackFromPayload != null) {
+
+        if (wasClicked != null) {
             logger.info("Notification target overridden by createTaskStackFromPayload, starting new stack for link $deepLink")
```
mrehan27, line 148: Might need to improve this, something like this? 🤔 

```suggestion
            logger.info("Notification target overridden by onNotificationClicked, link $deepLink handled by host app")
```

messagingpush/src/main/java/io/customer/messagingpush/data/communication/CustomerIOPushNotificationCallback.kt
Shahroz16: I don't mind it being `boolean`, but i feel like its confusing. `onInterceptTouchEvent` true or false both provide a different meaning. In our case, it means, if user want to handle the `notification clicked` they are going to provide us with action to perform. 

Otherwise, we do what we do. They would only need to provide `null` if they first up, start overriding the optional callback on purpose and then wants to tell us, to still not cater for what they provided 😄

mrehan27: If you say so, I'm okay with either option 😄

PR #377, opened 2024-07-03
refactor: core module test classes

mrehan27:

### Changes

- Removed test dependencies from module's gradle file as they are now included in the common gradle file for test dependencies
- Updated tests to match package of their respective classes
- Updated tests to work with `common-test` updates

### Test Status

All tests for `core` module are passing ✅

### Notes

See [parent PR](https://github.com/customerio/customerio-android/pull/373) for more details and context.

### DIFF
--- a/core/src/test/java/io/customer/sdk/core/di/DiGraphTest.kt
+++ b/core/src/test/java/io/customer/sdk/core/di/DiGraphTest.kt
@@ -1,6 +1,7 @@
 package io.customer.sdk.core.di
 
-import io.customer.commontest.core.BaseUnitTest
+import io.customer.commontest.config.TestConfig
+import io.customer.commontest.core.JUnit5Test
 import java.util.concurrent.CountDownLatch
 import java.util.concurrent.TimeUnit
 import kotlin.concurrent.thread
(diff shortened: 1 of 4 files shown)


core/src/test/java/io/customer/sdk/core/di/DiGraphTest.kt
```diff
@@ -231,3 +232,6 @@ class DiGraphTest : BaseUnitTest() {
         diGraph.getOrNull<Float>().shouldBeNull()
     }
 }
+
+// Custom data class to test dependency injection with the same class name.
+private data class Pair(val value: Int)
```
Shahroz16, line 237: why is the naming `Pair`?, should we name it something that more suitable here and then wont have to use `kotlin.Pair` as well to avoid conflict

Shahroz16 (approved):

core/src/test/java/io/customer/sdk/core/di/DiGraphTest.kt
mrehan27: That's exactly the test is, it should not fail if class names are same 😅

PR #382, opened 2024-07-05
chore: prefer existing device token

mrehan27:

### Changes

- Updated `ContextPlugin` to avoid rewriting token if one already exists
- This is to prevent data corruption with the old token already added to JSON
- Moved helpful extension from tests to production code that helps append the token to event JSON
- Added tests to validate changes

### DIFF
--- a/datapipelines/src/main/kotlin/io/customer/datapipelines/plugins/PluginExtensions.kt
+++ b/datapipelines/src/main/kotlin/io/customer/datapipelines/plugins/PluginExtensions.kt
@@ -1,29 +1,62 @@
 package io.customer.datapipelines.plugins
 
 import com.segment.analytics.kotlin.core.BaseEvent
+import kotlinx.serialization.json.JsonArray
+import kotlinx.serialization.json.JsonElement
 import kotlinx.serialization.json.JsonObject
 import kotlinx.serialization.json.JsonPrimitive
-import kotlinx.serialization.json.jsonObject
-import kotlinx.serialization.json.jsonPrimitive
 
 /**
- * Gets the value at nested path in the JSON object.
- * Example: `{"a": {"b": "c"}}.findAtPath("a.b")` returns c
+ * Gets the value(s) at nested path in the JSON object.
+ * Handles arrays by returning all matching elements.
+ * Example1: `{"a": [{"b": "c"}, {"e": "f"}]}.findAtPath("a.b")` returns [JsonPrimitive("c")]
+ * Example2: `{"a": {"b": "c"}}.findAtPath("a.b")` returns [JsonPrimitive("c")]
  */
-internal fun JsonObject.findAtPath(path: String): JsonPrimitive? = kotlin.runCatching {
-    // Split the path into keys
-    val keys = path.split(".")
-    // Start with the current JSON object
-    var currentElement = this
-    // Traverse the JSON object to find the value at the path
-    for (key in keys.dropLast(1)) {
-        // If the key does not exist, return null
-        currentElement = currentElement[key]?.jsonObject ?: return null
-    }
-    // Return the value at the last key in the path
-    return currentElement[keys.last()]?.jsonPrimitive
-}.getOrNull()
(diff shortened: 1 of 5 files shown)


datapipelines/src/main/kotlin/io/customer/datapipelines/plugins/PluginExtensions.kt
```diff
@@ -0,0 +1,29 @@
...
+
+import com.segment.analytics.kotlin.core.BaseEvent
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.JsonPrimitive
+import kotlinx.serialization.json.jsonObject
+import kotlinx.serialization.json.jsonPrimitive
+
+/**
+ * Gets the value at nested path in the JSON object.
+ * Example: `{"a": {"b": "c"}}.getNestedString("a.b")` returns c
```
Shahroz16, line 11:
```suggestion
 * Example: `{"a": {"b": "c"}}.findAtPath("a.b")` returns c
```

typo?

datapipelines/src/main/kotlin/io/customer/datapipelines/plugins/PluginExtensions.kt
Shahroz16, line 1: Good addition but it doesn't support JsonArray? understand we don't need it for our use-case but if we are creating an extension, might as well just add support for Json array 

```
/**
 * Gets the value(s) at nested path in the JSON object.
 * Handles arrays by returning all matching elements.
 * Example: `{"a": [{"b": "c"}, {"b": "d"}]}.findAtPath("a.b")` returns [JsonPrimitive("c"), JsonPrimitive("d")]
 */
fun JsonObject.findAtPath(path: String): List<JsonPrimitive> {
    // Split the path into individual keys
    val keys = path.split('.')
    // Start the recursive search
    return findAtPathRecursive(this, keys)
}

/**
 * Recursively searches for values at the specified path in the JSON element.
 * @param element The current JSON element being processed.
 * @param keys The remaining path keys to process.
 * @return A list of JsonPrimitives found at the path.
 */
private fun findAtPathRecursive(element: JsonElement, keys: List<String>): List<JsonPrimitive> {
    // Base case: If no more keys, check if the element is a JsonPrimitive
    if (keys.isEmpty()) {
        return if (element is JsonPrimitive) listOf(element) else emptyList()
    }

    // Extract the first key and the remaining keys
    val key = keys.first()
    val remainingKeys = keys.drop(1)

    return when (element) {
        is JsonObject -> {
            // If element is a JsonObject, look up the next key
            val nextElement = element[key]
            // If the key exists, continue the recursive search
            if (nextElement != null) findAtPathRecursive(nextElement, remainingKeys) else emptyList()
        }
        is JsonArray -> {
            // If element is a JsonArray, apply the search to each element
            element.flatMap { findAtPathRecursive(it, keys) }
        }
        else -> {
            // If element is neither JsonObject nor JsonArray, return an empty list
            emptyList()
        }
    }
}

/**
 * Extension function for BaseEvent to search in its context at a specified path.
 * @param path The dot-separated path to search for.
 * @return A list of JsonPrimitives found at the path.
 */
fun BaseEvent.findInContextAtPath(path: String): List<JsonPrimitive> =
    context.findAtPath(path)
```

Shahroz16 (approved):
looks good

datapipelines/src/main/kotlin/io/customer/datapipelines/plugins/PluginExtensions.kt
mrehan27: This sounds nice Shahroz. thanks for sharing this. The only concern I have is, I'm slightly unclear on how to use it. Can you please help me include some example usage for it in docs like we did for `findAtPath`? I can then include it in this PR before merging.

Shahroz16: Added a commit with the method and updated method description

PR #383, opened 2024-07-09
chore: move background queue from tracking to migration

Shahroz16:
helps: https://linear.app/customerio/issue/MBL-395/migration-for-stored-data

Moving part of BQ needed for migration in migration module so we can deprecate tracking.


mrehan27 (approved):
My assumption is that these files are copied from `tracking` module to make it easier to get started on `migration` and track changes from actual `tracking` implementation. The final implementation will be much lighter than this. If that's true, I'm okay with it and approve the changes.

PR #384, opened 2024-07-09
chore: fix memory leak in lifecycle listener

mrehan27:
### Changes

- Used `WeakReference` instead of `replay` to prevent holding strong activity references and avoid memory leaks
- Exposed `currentLifecycleState` which wrappers can use to get current state and act accordingly

### Testing

`MemoryLeakageInstrumentedTest` which was previously failing now passes after the changes.

### DIFF
--- a/core/src/main/kotlin/io/customer/sdk/lifecycle/CustomerIOActivityLifecycleCallbacks.kt
+++ b/core/src/main/kotlin/io/customer/sdk/lifecycle/CustomerIOActivityLifecycleCallbacks.kt
@@ -5,6 +5,7 @@ import android.app.Application
 import android.os.Bundle
 import androidx.lifecycle.Lifecycle
 import io.customer.sdk.core.di.SDKComponent
+import java.lang.ref.WeakReference
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.flow.MutableSharedFlow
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/io/customer/sdk/lifecycle/CustomerIOActivityLifecycleCallbacks.kt
```diff
@@ -18,9 +19,13 @@ import kotlinx.coroutines.launch
  * and should avoid listening to events from ActivityLifecycleCallbacks directly.
  */
 class CustomerIOActivityLifecycleCallbacks : Application.ActivityLifecycleCallbacks {
-    private val lifecycleEvents = MutableSharedFlow<LifecycleStateChange>(replay = 1)
+    private val lifecycleEvents = MutableSharedFlow<LifecycleStateChange>()
     private val subscriberScope = SDKComponent.scopeProvider.lifecycleListenerScope
 
+    // Stores last emitted lifecycle state to provide current state to new subscribers.
+    // The reference is weak to avoid memory leaks.
+    private var currentLifecycleStateRef: WeakReference<LifecycleStateChange>? = null
```
mrehan27, line 27: This is only added for wrappers. If we want to skip it for now and update later when working on wrappers, I'm okay with it.

Shahroz16, line 27: why dont we just use `state flow` if we want to have the last state stored?

mrehan27: Because the problem actually lies in strong reference. Whether we use `StateFlow` or `SharedFlow` with replay, the memory leakage tests fail. My understanding is that it fails because we hold strong reference to activity in `LifecycleStateChange` class, and if for any reason, we hold  strong reference to `LifecycleStateChange`, it causes a memory leak.

So `WeakReference` here is used to have similar behavior to `StateFlow`, but with a weak reference to prevent potential memory leaks.

Shahroz16: That makes sense but i would make a case that we should rather keep a weak reference of `Activity` in first place? and then even if `StateFlow/Shared flow` is keeping a strong reference we would not care.

Shahroz16: I created another [PR](https://github.com/customerio/customerio-android/pull/385) with the strategy mentioned above, if it sounds good lets merge that otherwise, we can keep it otherwise we can go with this one.

mrehan27: Thanks Shahroz for creating the PR, added some comments on it.

mrehan27:
Closing this as we discussed this on call and decided to merge a mix of both changes in #385

PR #385, opened 2024-07-09
chore: memory leakage for application lifecycle

Shahroz16:
alternate implementation for memory leakage.

### DIFF
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -313,6 +313,7 @@ public final class io/customer/sdk/events/TrackMetric$Push : io/customer/sdk/eve
 
 public final class io/customer/sdk/lifecycle/CustomerIOActivityLifecycleCallbacks : android/app/Application$ActivityLifecycleCallbacks {
 	public fun <init> ()V
+	public final fun currentLifecycleState ()Lio/customer/sdk/lifecycle/LifecycleStateChange;
 	public fun onActivityCreated (Landroid/app/Activity;Landroid/os/Bundle;)V
 	public fun onActivityDestroyed (Landroid/app/Activity;)V
 	public fun onActivityPaused (Landroid/app/Activity;)V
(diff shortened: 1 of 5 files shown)


mrehan27 (commented):
What other benefits do you see for this approach besides using `StateFlow` instead of caching by ourselves?

Asking because I had considered this approach already but didn't go for it because:

- Weak reference of `activity` doesn't guarantee that listeners will always execute their action completely because `activity` can be `null` sometimes. Even though the chance is very rare for recently emitted values, the possibility is there, making debugging difficult
- None of our listeners hold `activity`, so they cannot leak context despite strong reference
- Both our listeners today require `activity` to perform desired actions. Having no `activity` makes listeners useless despite them observing, filtering and collecting all values

Shahroz16:
> What other benefits do you see for this approach besides using `StateFlow` instead of caching by ourselves?
> 
> Asking because I had considered this approach already but didn't go for it because:
> 
> * Weak reference of `activity` doesn't guarantee that listeners will always execute their action completely because `activity` can be `null` sometimes. Even though the chance is very rare for recently emitted values, the possibility is there, making debugging difficult
> * None of our listeners hold `activity`, so they cannot leak context despite strong reference
> * Both our listeners today require `activity` to perform desired actions. Having no `activity` makes listeners useless despite them observing, filtering and collecting all values

My reasoning for this approach:

- We are not relying on custom implementation of state flow, we are using the built in one. Which you mentioned too. 
- `Activity` should never be `null` it would be only `null` which it has been garbage collected, If the activity has been garbage collected then, it would have been `null` with another variable holding that value as well
- Similarly, if activity is garbage collected, none of our listener would/should require it either. I can't think of a case, where activity needs to be garbage collected and our callback would need to require it. If you can think of one, please let me know. 
- Regardless of weak reference, like you said, SharedFlow (previous implementation) wasn't holding the strong reference to activity, if the activity was to be garbage collect, what would have happened then in case of our callback?

mrehan27 (approved):
Looks good, just some small comments

PR #388, opened 2024-07-11
chore: add site pref repo to migration module

mrehan27:

### Changes

- Updated core `PreferenceStore` to allow extending in other modules
- Refactored and added `SitePreferenceRepository` to `migration` module to fetch values added by `tracking` module in older releases


Shahroz16 (approved):
Looks good

PR #390, opened 2024-07-12
refactor: move event names to constants

mrehan27:

### Changes

- Added `EventNames` to hold constants for event names so they can be reused
- Updated code and tests to use newly added constants

### DIFF
--- a/datapipelines/src/main/kotlin/io/customer/datapipelines/util/EventNames.kt
+++ b/datapipelines/src/main/kotlin/io/customer/datapipelines/util/EventNames.kt
@@ -1,0 +1,11 @@
+package io.customer.datapipelines.util
+
+/**
+ * Event names to identify specific events in data pipelines so they can be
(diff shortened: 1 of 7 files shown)


datapipelines/src/main/kotlin/io/customer/datapipelines/util/EventNames.kt
```diff
@@ -0,0 +1,11 @@
...
+
+/**
+ * Event names to identify specific events in data pipelines so they can be
+ * reflected on Journeys.
+ */
+internal object EventNames {
+    const val DEVICE_UPDATE = "Device Created or Updated"
+    const val DEVICE_DELETE = "Device Deleted"
+    const val METRIC_DELIVERY = "Report Delivery Event"
+}
```
mrehan27, line 11: We'll need to reuse them in migration

Shahroz16 (approved):

PR #391, opened 2024-07-12
chore: parse migration tasks

mrehan27:

### Changes

- Added `MigrationProcessor` so Queue can communicate with Data Pipelines module
- Simplified `QueueRunner` by forwarding parsing responsibility to `JsonAdapter`
- Merged all data classes into one `MigrationTask` so it easier to understand and manage
- Updated `JsonAdapter` to parse core JSONObject to `MigrationTask`
- Removed unused code and extensions from migration module queue


tracking-migration/src/main/java/io/customer/tracking/migration/request/MigrationTask.kt
```diff
@@ -0,0 +1,60 @@
...
+ * Interface for the migration tasks that need to be processed by the migration processor.
+ * The migration tasks are used to migrate the data from the old tracking implementation
+ * to the new data pipelines.
+ * Each task represents specific action so it can be processed accordingly.
+ */
+sealed interface MigrationTask {
+    val timestamp: Long
+    val identifier: String?
+
+    data class IdentifyProfile(
```
mrehan27, line 15: Replacement for `IdentifyProfileQueueTaskData`

tracking-migration/src/main/java/io/customer/tracking/migration/request/MigrationTask.kt
```diff
@@ -0,0 +1,60 @@
...
+        override val timestamp: Long,
+        override val identifier: String,
+        val attributes: JSONObject
+    ) : MigrationTask
+
+    data class TrackEvent(
```
mrehan27, line 21: Replacement for `TrackEventQueueTaskData`

tracking-migration/src/main/java/io/customer/tracking/migration/request/MigrationTask.kt
```diff
@@ -0,0 +1,60 @@
...
+        override val timestamp: Long,
+        override val identifier: String,
+        val event: String,
+        val type: String,
+        val properties: JSONObject
+    ) : MigrationTask
+
+    data class TrackPushMetric(
```
mrehan27, line 29: Replacement for `Metric`

tracking-migration/src/main/java/io/customer/tracking/migration/request/MigrationTask.kt
```diff
@@ -0,0 +1,60 @@
...
+        override val timestamp: Long,
+        override val identifier: String? = null,
+        val deliveryId: String,
+        val deviceToken: String,
+        val event: String
+    ) : MigrationTask
+
+    data class TrackDeliveryEvent(
```
mrehan27, line 37: Replacement for `DeliveryEvent`

tracking-migration/src/main/java/io/customer/tracking/migration/request/MigrationTask.kt
```diff
@@ -0,0 +1,60 @@
...
+        override val timestamp: Long,
+        override val identifier: String? = null,
+        val deliveryType: String,
+        val deliveryId: String,
+        val event: String,
+        val metadata: JSONObject
+    ) : MigrationTask
+
+    data class RegisterDeviceToken(
```
mrehan27, line 46: Replacement for `RegisterPushNotificationQueueTaskData`

tracking-migration/src/main/java/io/customer/tracking/migration/request/MigrationTask.kt
```diff
@@ -0,0 +1,60 @@
...
+        override val timestamp: Long,
+        override val identifier: String,
+        val token: String,
+        val platform: String,
+        val lastUsed: Long,
+        val attributes: JSONObject
+    ) : MigrationTask
+
+    data class DeletePushToken(
```
mrehan27, line 55: Replacement for `DeletePushNotificationQueueTaskData`

tracking-migration/src/main/java/io/customer/tracking/migration/util/JsonAdapter.kt
```diff
@@ -11,4 +22,101 @@ class JsonAdapter {
...
+
+        val type = task.stringOrNull("type")?.let { enumValueOfOrNull<QueueTaskType>(it) }
+        requireNotNull(type) { "Queue task type is invalid for $task. Could not run task." }
+
+        val taskJson = runCatching { JSONObject(data) }.getOrElse {
+            throw RuntimeException("Queue task data is invalid for $task. Could not run task.")
+        }
+        val timestamp = taskJson.longOrNull("timestamp") ?: Date().time
+
+        val result = when (type) {
```
mrehan27, line 44: All field names added here for parsing are taken from deleted models to make sure we parse JSON correctly.

tracking-migration/src/main/java/io/customer/tracking/migration/util/JsonAdapter.kt
```diff
@@ -11,4 +22,101 @@ class JsonAdapter {
...
+                    event = eventJson.requireField("name") {
+                        val value = stringOrNull(it)
+                        remove(it)
+                        return@requireField value
+                    },
+                    type = eventJson.requireField("type") {
+                        val value = stringOrNull(it)
+                        remove(it)
+                        return@requireField value
+                    },
```
Shahroz16, line 68: we can probably extract it a separate method? 

```
 private fun JSONObject.requireAndRemove(key: String): String {
        val value = this.requireField<String>(key)
        this.remove(key)
        return value
    }
```

tracking-migration/src/main/java/io/customer/tracking/migration/util/JsonAdapter.kt
```diff
@@ -11,4 +22,101 @@ class JsonAdapter {
...
+                )
+            }
+
+            QueueTaskType.TrackEvent -> {
+                val eventJson = taskJson.requireField("event") { jsonObjectOrNull(it) }
+
+                MigrationTask.TrackEvent(
+                    timestamp = eventJson.longOrNull("timestamp") ?: timestamp,
+                    identifier = taskJson.requireField("identifier") { stringOrNull(it) },
+                    event = eventJson.requireField("name") {
```
Shahroz16, line 59: it would become 
```
                    event = eventJson.requireAndRemove("name"),
```

tracking-migration/src/main/java/io/customer/tracking/migration/util/JsonAdapter.kt
```diff
@@ -11,4 +22,101 @@ class JsonAdapter {
...
+                        val value = stringOrNull(it)
+                        remove(it)
+                        return@requireField value
+                    },
+                    type = eventJson.requireField("type") {
```
Shahroz16, line 64:
```
                    event = eventJson.requireAndRemove("type"),
```

tracking-migration/src/main/java/io/customer/tracking/migration/util/JsonAdapter.kt
```diff
@@ -11,4 +22,101 @@ class JsonAdapter {
...
+                    metadata = payloadJson.jsonObjectOrNull("metadata") ?: JSONObject()
+                )
+            }
+        }
+
+        return Result.success(result)
+    }
+}
+
+private fun <T> JSONObject?.requireField(key: String, getter: JSONObject.(String) -> T?): T {
```
Shahroz16, line 120: think this can be simplified, 

```
private inline fun <reified T> JSONObject.requireField(key: String): T {
    return when (T::class) {
        String::class -> this.stringOrNull(key)
        else -> this.opt(key)
    } as? T ?: throw IllegalArgumentException("Required key '$key' is missing or has wrong type in $this")
}
```

now we can just use, e.g for json object
```
val payloadJson = taskJson.requireField<JSONObject>("payload")
```
for string
```
deliveryType = taskJson.requireField("type"),
```

we can check for type in runtime

tracking-migration/src/main/java/io/customer/tracking/migration/util/JsonAdapter.kt
mrehan27: Updated

tracking-migration/src/main/java/io/customer/tracking/migration/util/JsonAdapter.kt
mrehan27: Good call. The only concern I have is it will suppress any type that the callers tries to parse and is not currently supported as the whole code is in `runCatching`. I have updated it to more generic implementation and added docs to reduce the possibility of this, but we should still be careful when using this so that we don't keep suppressing it unknowingly.

Shahroz16 (approved):

PR #392, opened 2024-07-12
chore: process migrated tasks

mrehan27:

### Changes

- Implemented `MigrationProcessor` in Data Pipelines module
- Added `MigrationAssistant` to help migrating old tasks
- Requested migration using `MigrationAssistant` on SDK initialization
- Added `MigrationSDKComponent` so it is easier to satisfy dependencies in migration module
- Extensions to transform `JSONObject` to `JsonObject` recursively


datapipelines/src/main/kotlin/io/customer/datapipelines/extensions/JsonExtensions.kt
```diff
@@ -0,0 +1,48 @@
...
+ * it to JsonObject for processing.
+ */
+fun JSONObject.toJsonObject(): JsonObject = buildJsonObject {
+    val keys = keys()
+    while (keys.hasNext()) {
+        val key = keys.next()
+        val value = get(key)
+        put(key, value.toSerializableJson())
+    }
+}
```
Shahroz16, line 25:
```suggestion
fun JSONObject.toJsonObject(): JsonObject = buildJsonObject {
    for (key in keys()) {
        put(key, opt(key).toSerializableJson())
    }
}
```

`get` if you want error to be thrown

datapipelines/src/main/kotlin/io/customer/datapipelines/extensions/JsonExtensions.kt
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Transforms [JSONArray] to [JsonArray] by leveraging kotlinx.serialization.json.
+ */
+fun JSONArray.toJsonArray(): JsonArray = buildJsonArray {
+    for (i in 0 until length()) {
+        val value = get(i)
+        add(value.toSerializableJson())
+    }
+}
```
Shahroz16, line 35:
```suggestion
fun JSONArray.toJsonArray(): JsonArray = buildJsonArray {
    for (i in 0 until length()) {
        add(opt(i).toSerializableJson())
    }
}
```

datapipelines/src/main/kotlin/io/customer/datapipelines/extensions/JsonExtensions.kt
```diff
@@ -0,0 +1,48 @@
...
+ * The method is recursive and can handle nested JSONObjects and JSONArrays.
+ */
+private fun Any?.toSerializableJson(): JsonElement {
+    return when (this) {
+        null -> JsonNull
+        is JSONObject -> this.toJsonObject()
+        is JSONArray -> this.toJsonArray()
+        else -> this.toJsonElement()
+    }
+}
```
Shahroz16, line 48:
```suggestion
private fun Any?.toSerializableJson(): JsonElement {
    return when (this) {
        null -> JsonNull
        is JSONObject -> toJsonObject()
        is JSONArray -> toJsonArray()
        else -> toJsonElement()
    }
}
```

Shahroz16: if we have custom classes with serializers than `Json.encodeToJsonElement()` might be safer than `toJsonElement()`

datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
```diff
@@ -48,7 +60,7 @@ class CustomerIO private constructor(
     androidSDKComponent: AndroidSDKComponent,
     override val moduleConfig: DataPipelinesModuleConfig,
     overrideAnalytics: Analytics? = null
-) : CustomerIOModule<DataPipelinesModuleConfig>, DataPipelineInstance() {
+) : CustomerIOModule<DataPipelinesModuleConfig>, DataPipelineInstance(), MigrationProcessor {
```
Shahroz16, line 63: can we probably create a separate class for it? `CustomerIO` class is already too big 😅

tracking-migration/src/main/java/io/customer/tracking/migration/MigrationAssistant.kt
```diff
@@ -0,0 +1,45 @@
...
+        migrationSiteId = migrationSiteId
+    )
+    private val sitePreferences: SitePreferenceRepository = migrationSDKComponent.sitePreferences
+    private val queue: Queue = migrationSDKComponent.queue
+
+    /**
+     * Starts the migration process by migrating the existing tracking data to the new
+     * data pipelines implementation using provided migration processor.
+     */
+    fun migrate() {
```
Shahroz16, line 29: can we call it `init` and avoid a method call? this class will only be initialized if `siteId` exists

Shahroz16 (approved):
Minor improvements suggestions

datapipelines/src/main/kotlin/io/customer/datapipelines/extensions/JsonExtensions.kt
mrehan27: Updated

datapipelines/src/main/kotlin/io/customer/datapipelines/extensions/JsonExtensions.kt
mrehan27: Updated

datapipelines/src/main/kotlin/io/customer/datapipelines/extensions/JsonExtensions.kt
mrehan27: We are actually using `com.segment.analytics.kotlin.core.utilities.toJsonElement` here which has support for many other types and calling `Json.encodeToJsonElement()` internally where needed. The only two types that we needed and weren't support were `org.json.JSONObject` and `org.json.JSONArray` so I added wrapper to it.

datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
mrehan27: I was thinking the same initially, let me try to decouple it and share if I see any challenges in it.

mrehan27: I moved code to `TrackingMigrationProcessor`. Let me know how if it looks good to you now.

tracking-migration/src/main/java/io/customer/tracking/migration/MigrationAssistant.kt
mrehan27: I avoid it as it reduces code readability. I added some comments and a helper method to improve readability with code in init block.

Issue #361, opened 2024-06-18
ANR in EngineWebView initialization


remi-ollivier:
**SDK version:** 3.8.1 

**Environment:** Production

**Are logs available?** 

```
#00  pc 0x0000000001c03f7c  [anon:libwebview reservation]
  #01  pc 0x0000000002691133  [anon:libwebview reservation]
  #02  pc 0x000000000002cbad  /apex/com.android.runtime/bin/linker (__dl__ZL10call_arrayIPFviPPcS1_EEvPKcPT_jbS5_+204)
  #03  pc 0x000000000001ffee  [anon:linker_alloc]
  at android.webkit.WebViewLibraryLoader.nativeLoadWithRelroFile (Native method)
  at android.webkit.WebViewLibraryLoader.loadNativeLibrary (WebViewLibraryLoader.java:214)
  at android.webkit.WebViewFactory.getProviderClass (WebViewFactory.java:427)
  at android.webkit.WebViewFactory.getProvider (WebViewFactory.java:262)
  at android.webkit.WebView.getFactory (WebView.java:2576)
  at android.webkit.WebView.ensureProviderCreated (WebView.java:2570)
  at android.webkit.WebView.setOverScrollMode (WebView.java:2638)
  at android.view.View.<init> (View.java:5363)
  at android.view.View.<init> (View.java:5515)
  at android.view.ViewGroup.<init> (ViewGroup.java:702)
  at android.widget.AbsoluteLayout.<init> (AbsoluteLayout.java:56)
  at android.webkit.WebView.<init> (WebView.java:413)
  at android.webkit.WebView.<init> (WebView.java:355)
  at android.webkit.WebView.<init> (WebView.java:337)
  at android.webkit.WebView.<init> (WebView.java:324)
  at android.webkit.WebView.<init> (WebView.java:314)
  at io.customer.messaginginapp.gist.presentation.engine.EngineWebView.<init> (EngineWebView.kt:34)
  at io.customer.messaginginapp.gist.presentation.engine.EngineWebView.<init> (EngineWebView.kt:20)
  at io.customer.messaginginapp.gist.presentation.GistView.<init> (GistView.kt:27)
  at io.customer.messaginginapp.gist.presentation.GistSdk.init$lambda$0 (GistSdk.kt:106)
  at android.os.Handler.handleCallback (Handler.java:938)
  at android.os.Handler.dispatchMessage (Handler.java:99)
  at android.os.Looper.loop (Looper.java:223)
[...]

PR #393, opened 2024-07-15
fix: remove gist pre-loading

Shahroz16:
With our in-app rendering move from flutter to html, we don't need to pre-load flutter engine to download assets. 

more context here: https://customerio.slack.com/archives/C036FTE6UJ0/p1721046751535539
resolves the ANR issue too: https://github.com/customerio/customerio-android/issues/361


mrehan27 (approved):
So happy to see this update 😄 🎉

PR #394, opened 2024-07-15
chore: added device token migration

Shahroz16:
Added device token migration from `sitePreference` to `globalPreference`

### DIFF
--- a/datapipelines/src/main/kotlin/io/customer/datapipelines/migration/TrackingMigrationProcessor.kt
+++ b/datapipelines/src/main/kotlin/io/customer/datapipelines/migration/TrackingMigrationProcessor.kt
@@ -26,6 +26,7 @@ internal class TrackingMigrationProcessor(
     migrationSiteId: String
 ) : MigrationProcessor {
     private val logger: Logger = SDKComponent.logger
+    private val globalPreferenceStore = SDKComponent.android().globalPreferenceStore
 
     // Start the migration process in init block to start migration as soon as possible
     // and to avoid any manual calls to replay migration.
@@ -43,6 +44,11 @@ internal class TrackingMigrationProcessor(
 
     override fun processProfileMigration(identifier: String): Result<Unit> = runCatching {
         dataPipelineInstance.identify(userId = identifier)
+    }
+
+    override fun processDeviceMigration(oldDeviceIdentifier: String) = runCatching {
+        logger.debug("Migrating existing device with token: $oldDeviceIdentifier")
+        globalPreferenceStore.saveDeviceToken(oldDeviceIdentifier)
     }
 
     override suspend fun processTask(task: MigrationTask): Result<Unit> = runCatching {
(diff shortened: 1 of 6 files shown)


tracking-migration/src/main/java/io/customer/tracking/migration/MigrationAssistant.kt
```diff
@@ -36,6 +37,7 @@ class MigrationAssistant private constructor(
             // Re-identify old profile to new implementation
             logger.debug("Migrating existing profile...")
             migrateExistingProfile()
+            migrateExistingDevice()
```
mrehan27, line 40: Because we use device token in identify method, shouldn't we be saving device token before identifying profile? 🤔

tracking-migration/src/main/java/io/customer/tracking/migration/MigrationAssistant.kt
```diff
@@ -55,6 +57,16 @@ class MigrationAssistant private constructor(
         }
     }
 
+    private fun migrateExistingDevice() {
+        // If there is no old device token, then either the device was never added or it was already migrated
+        val oldDeviceIdentifier = sitePreferences.getDeviceToken() ?: return
+
+        logger.debug("Migrating existing device with token: $oldDeviceIdentifier")
+        globalPreferenceStore.saveDeviceToken(oldDeviceIdentifier)
```
mrehan27, line 65: While this should work fine, but storing token in new `globalPreferenceStore` should be responsibility of `MigrationProcessor` and not `MigrationAssistant`. `MigrationAssistant` is currently responsible for fetching and removing existing data only. Asking it to process and save in new store is violating the responsibility in my opinion and this should be handed over to `MigrationProcessor` instead.

datapipelines/src/main/kotlin/io/customer/datapipelines/migration/TrackingMigrationProcessor.kt
```diff
@@ -45,6 +46,11 @@ internal class TrackingMigrationProcessor(
         dataPipelineInstance.identify(userId = identifier)
     }
 
+    override fun processDeviceMigration(oldDeviceIdentifier: String) = runCatching {
```
mrehan27, line 49: We should fix this warning before merging

```
The corresponding parameter in the supertype 'MigrationProcessor' is named 'token'. This may cause problems when calling this function with named arguments
```

tracking-migration/src/main/java/io/customer/tracking/migration/MigrationAssistant.kt
```diff
@@ -55,6 +57,16 @@ class MigrationAssistant private constructor(
         }
     }
 
+    private fun migrateExistingDevice() {
+        // If there is no old device token, then either the device was never added or it was already migrated
+        val oldDeviceIdentifier = sitePreferences.getDeviceToken() ?: return
+
+        if (migrationProcessor.processDeviceMigration(oldDeviceIdentifier).isSuccess) {
+            // remove the old device token to prevent updating GlobalPreferenceStore again
```
mrehan27, line 65:
```suggestion
            // remove the old device token to prevent migrating device token again
```

mrehan27 (approved):
Looks good 👍🏻

PR #395, opened 2024-07-16
chore: fix device token in migration tasks

mrehan27:

### Changes

- Added `TrackingMigrationPlugin` to append device token and platform in event context
- Updated code to add and remove plugin as needed

### DIFF
--- a/datapipelines/src/main/kotlin/io/customer/datapipelines/migration/TrackingMigrationPlugin.kt
+++ b/datapipelines/src/main/kotlin/io/customer/datapipelines/migration/TrackingMigrationPlugin.kt
@@ -1,0 +1,34 @@
+package io.customer.datapipelines.migration
+
+import com.segment.analytics.kotlin.core.Analytics
+import com.segment.analytics.kotlin.core.BaseEvent
+import com.segment.analytics.kotlin.core.platform.Plugin
+import com.segment.analytics.kotlin.core.utilities.putInContextUnderKey
+
+/**
+ * Plugin responsible for adding device context extras to the events.
+ * This is because the device context extras cannot be added directly to the events
+ * and need to be added at plugin level.
+ */
+class TrackingMigrationPlugin : Plugin {
+    override val type: Plugin.Type = Plugin.Type.Before
+    override lateinit var analytics: Analytics
+
+    private val deviceContextExtras: MutableMap<BaseEvent, Map<String, String>> = mutableMapOf()
+
+    fun addDeviceContextExtras(event: BaseEvent, extras: Map<String, String>) {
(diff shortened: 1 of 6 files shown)


datapipelines/src/main/kotlin/io/customer/datapipelines/migration/TrackingMigrationPlugin.kt
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Plugin responsible for adding device context extras to the events.
+ * This is because the device context extras cannot be added directly to the events
+ * and need to be added at plugin level.
+ */
+class TrackingMigrationPlugin : Plugin {
+    override val type: Plugin.Type = Plugin.Type.Before
+    override lateinit var analytics: Analytics
+
+    private val deviceContextExtras: MutableMap<BaseEvent, Map<String, String>> = mutableMapOf()
```
mrehan27, line 17: This might not be the best idea but we don't really have any identifier for an event that doesn't change. Happy to try out other options if there are more ideas.

datapipelines/src/main/kotlin/io/customer/datapipelines/migration/TrackingMigrationProcessor.kt
```diff
@@ -89,24 +104,26 @@ internal class TrackingMigrationProcessor(
                     put("token", task.token)
                     put("properties", task.attributes.toJsonObject())
                 }
-            ).apply {
-                putInContextUnderKey("device", "token", task.token)
-                putInContextUnderKey("device", "type", "android")
-            }
+            ).addDeviceContextExtras(
+                mapOf("token" to task.token, "type" to "android")
```
mrehan27, line 108: Not sure if we need `type` here now but iOS does this so added here 🤔

datapipelines/src/main/kotlin/io/customer/datapipelines/migration/TrackingMigrationPlugin.kt
Shahroz16: The only other idea i could think of is to probably add them in properties with a unique name and then in the plugin move them from properties to context, that would probably save from keeping the map.

Shahroz16 (approved):

datapipelines/src/main/kotlin/io/customer/datapipelines/migration/TrackingMigrationPlugin.kt
mrehan27: That's one good idea to think about. Not sure about the challenges it comes with, but will surely try this out later to reduce cost of keeping this map.

PR #396, opened 2024-07-16
chore: memory leakage fix

Shahroz16:
Store weak reference of lambda received for event bus action so that we are not leaking memory.


core/src/test/java/io/customer/sdk/communication/EventBusTest.kt
```diff
@@ -103,7 +103,7 @@ class EventBusTest : JUnit5Test() {
         println("Publishing event: $testEvent")
         eventBus.publish(testEvent)
 
-        delay(100) // Give some time for the event to be collected
+        delay(200) // Give some time for the event to be collected
```
Shahroz16, line 106: since I added `synchronized(jobs)` before inserting jobs, had to increase the delay because time it takes could get longer now.

mrehan27 (approved):
Looks good 👍🏻

PR #398, opened 2024-07-16
chore: copy tracking classes to test migration

mrehan27:

### Changes

This PR copies all data classes from the deprecated `tracking` module to test directory of `migration` module. This allows us to write automated tests that closely match original data generated for those classes. The classes are copied to test directory to avoid including any changes in release build. Since `tracking` module used `Moshi` to parse data classes to JSON, the dependency is copied as well but only as `testImplementation` to allow writing tests that provide more confidence.

### DIFF
--- a/tracking-migration/src/test/java/io/customer/sdk/data/request/Device.kt
+++ b/tracking-migration/src/test/java/io/customer/sdk/data/request/Device.kt
@@ -1,0 +1,20 @@
+package io.customer.sdk.data.request
+
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+import io.customer.sdk.data.model.CustomAttributes
+import java.util.Date
+
+@JsonClass(generateAdapter = true)
+data class Device(
+    @field:Json(name = "id") val token: String,
+    val platform: String = "android",
+    @field:Json(name = "last_used") val lastUsed: Date?, // nullable to cater for `lastUsed` field in older versions of the SDK
+    @field:Json(name = "lastUsed") val lastUsedDeprecated: Date? = null, // additional field to cater for `lastUsed` field in older versions of the SDK
+    val attributes: CustomAttributes
+)
+
+@JsonClass(generateAdapter = true)
+internal data class DeviceRequest(
+    val device: Device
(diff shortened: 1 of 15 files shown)


tracking-migration/src/test/java/io/customer/sdk/data/request/Device.kt
```diff
@@ -0,0 +1,20 @@
...
+import com.squareup.moshi.JsonClass
+import io.customer.sdk.data.model.CustomAttributes
+import java.util.Date
+
+@JsonClass(generateAdapter = true)
+data class Device(
+    @field:Json(name = "id") val token: String,
+    val platform: String = "android",
+    @field:Json(name = "last_used") val lastUsed: Date?, // nullable to cater for `lastUsed` field in older versions of the SDK
+    @field:Json(name = "lastUsed") val lastUsedDeprecated: Date? = null, // additional field to cater for `lastUsed` field in older versions of the SDK
```
mrehan27, line 13: Added this for convenience

Shahroz16 (commented):
I am not sure i fully understand the use-case for this? at the end of the day these classes generate a `string` which we stored on file. Why can't we just get string and add those as mocks? even when we run test they would be on a limited use-cases of events and we can get mocked strings for all of those possible cases.

mrehan27:
@Shahroz16 Yes, we do need strings eventually. However the strings must be valid JSONs for these data models with multiple data combinations. We don't have any script or other helping tool at the moment to generate the JSON, so using these classes while writing tests was quickest win.

While I agree that we may not need to add more test cases to `migration` module for these data classes in future, having these data classes allows us to create JSON quickly and conveniently at this moment. Once we are done with including all tests for migration, we can extract the strings and remove these classes in future. But for now, I think having these data classes is more valuable than copying JSON strings.

Since these changes are only included in migration tests, this should not have any impact on customers in any of our future releases.

mrehan27:
Discussed this on call and we decided to continue with json strings.

Closing this as we no longer need the changes. PR with json strings has been created to address this. Please refer to #402  for further updates.

PR #399, opened 2024-07-17
chore: fix migration json adapter

mrehan27:

### Changes

- Update `MigrationTask` to always have an `identifier` (update for metric events mainly)
- Fixed minor parsing issues in migration `JsonAdapter`

### DIFF
--- a/datapipelines/src/main/kotlin/io/customer/datapipelines/migration/TrackingMigrationProcessor.kt
+++ b/datapipelines/src/main/kotlin/io/customer/datapipelines/migration/TrackingMigrationProcessor.kt
@@ -123,7 +123,7 @@ internal class TrackingMigrationProcessor(
         }
 
         trackEvent.timestamp = task.timestamp.toString()
-        task.identifier?.let { trackEvent.userId = it }
+        trackEvent.userId = task.identifier
 
         logger.debug("processing migrated task: $trackEvent")
         analytics.process(trackEvent)
--- a/tracking-migration/src/main/java/io/customer/tracking/migration/util/JsonAdapter.kt
+++ b/tracking-migration/src/main/java/io/customer/tracking/migration/util/JsonAdapter.kt
@@ -1,9 +1,9 @@
 package io.customer.tracking.migration.util
 
+import io.customer.base.extenstions.getUnixTimestamp
 import io.customer.sdk.core.util.enumValueOfOrNull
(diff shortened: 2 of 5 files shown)


datapipelines/src/main/kotlin/io/customer/datapipelines/migration/TrackingMigrationProcessor.kt
```diff
@@ -123,7 +123,7 @@ internal class TrackingMigrationProcessor(
         }
 
         trackEvent.timestamp = task.timestamp.toString()
-        task.identifier?.let { trackEvent.userId = it }
+        trackEvent.userId = task.identifier
```
mrehan27, line 126: [iOS uses `anonymousId` today](https://github.com/customerio/customerio-ios/blob/2f4e4e4d07bedffc251b0f0a19b3e2bd55b0375d/Sources/DataPipeline/DataPipelineImplementation.swift#L355), we can do the same by updating it as something like follows:

```kotlin
if ((trackEvent as? TrackEvent)?.event == EventNames.METRIC_DELIVERY) {
    trackEvent.anonymousId = task.identifier
} else {
    trackEvent.userId = task.identifier
}
```

But I think setting `userId` should work for us and keep things simple.

tracking-migration/src/main/java/io/customer/tracking/migration/util/JsonAdapter.kt
```diff
@@ -41,7 +41,7 @@ class JsonAdapter {
         val taskJson = runCatching { JSONObject(data) }.getOrElse {
             throw RuntimeException("Queue task data is invalid for $task. Could not run task.")
         }
-        val timestamp = taskJson.longOrNull("timestamp") ?: Date().time
+        val timestamp = taskJson.longOrNull("timestamp") ?: Date().getUnixTimestamp()
```
Shahroz16, line 44: is this because queue keep them in seconds or server expects them in seconds?

mrehan27: It is because that's how we [parse Date in JsonAdapter in `main` today](https://github.com/customerio/customerio-android/blob/f1c1ceedfb053d04a8e4c2d2f6e26ccbb84a1e29/sdk/src/main/java/io/customer/sdk/data/moshi/adapter/UnixDateAdapter.kt#L31)

PR #401, opened 2024-07-17
test: migration assistant class

mrehan27:

### Changes

- Moved `CoroutineScope` in `MigrationSDKComponent` so it can be overridden in tests
- Updated `MigrationAssistant` internal constructor to allow overriding `MigrationSDKComponent` for tests
- Added base `JUnitTest` class in migration tests for convenience
- Added tests for `MigrationAssistant`


Shahroz16 (approved):
looks good

PR #404, opened 2024-07-19
chore: fixes and updates

Shahroz16:
- moved from weak-reference to app lifecycle for memory leakage
- rebased with `main`
- segment version bump


core/src/main/kotlin/io/customer/sdk/communication/EventBus.kt
```diff
@@ -44,6 +45,18 @@ class EventBusImpl(
...
     }
 
+    // Remove all subscriptions when the application is destroyed
+    init {
+        SDKComponent.activityLifecycleCallbacks.subscribe { events ->
+            events
+                .filter { state ->
+                    state.event == Lifecycle.Event.ON_DESTROY
+                }.collect { _ ->
+                    removeAllSubscriptions()
```
mrehan27, line 55: If I'm reading it correctly, it'll remove all subscriptions once any activity is destroyed and not the application. So for single activity apps like Kotlin sample, it'll work. But for multiple activities apps like Java sample, it'll fail.

I think we need to find a way to keep track of activities and remove subscriptions only when all activities are destroyed. Also, I'd be happy to chat about other possible solutions we can try to avoid relying on this callback.

mrehan27 (changes requested):

mrehan27 (approved):

PR #405, opened 2024-07-19
test: migration json adapter invalid data

mrehan27:

### Changes

- Added more tests for `JsonAdapter` to validate error scenarios
- Added basic tests for running `Queue`
- Added helper methods, base test class and sample json for invalid data to facilitate writing tests

### Note

Queue runner tests will be added separately later.

### DIFF
--- a/common-test/src/main/java/io/customer/commontest/extensions/AssertionExtensions.kt
+++ b/common-test/src/main/java/io/customer/commontest/extensions/AssertionExtensions.kt
@@ -8,8 +8,9 @@ inline fun <reified E : Throwable> Result<Any>.shouldBeFailure(): E {
         "Expected failure but the result was successful"
     }
 
-    return assertNotNull(exceptionOrNull() as? E) {
-        "Expected exception of type ${E::class} but was ${exceptionOrNull()?.let { it::class } ?: "null"}"
+    val exception = exceptionOrNull()
+    return assertNotNull(exception as? E) {
+        "Expected exception of type ${E::class} but was ${exception?.let { it::class } ?: "null"}"
     }
 }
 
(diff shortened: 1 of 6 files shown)


tracking-migration/src/test/java/io/customer/tracking/migration/testutils/data/MigrationDataSet.kt
```diff
@@ -224,3 +224,68 @@ sealed interface DeletePushNotificationQueueTaskData : MigrationDataSet {
         """.trimIndent()
     }
 }
+
+sealed interface InvalidMigrationDataSet : MigrationDataSet {
```
Shahroz16, line 228:
```
        fun createJsonString(storageId: String, type: String, data: String, totalRuns: Int): String = """
            {
              "storageId": "$storageId",
              "type": "$type",
              "data": $data,
              "runResults": {
                "totalRuns": $totalRuns
              }
            }
        """.trimIndent()
    }
```

its a one off file so doesn't really matter, but something like this could have made life a little easier 😅

Shahroz16 (approved):

common-test/src/main/java/io/customer/commontest/extensions/AssertionExtensions.kt
```diff
@@ -0,0 +1,19 @@
...
+
+    val exception = exceptionOrNull()
+    if (exception == null) {
+        throw AssertionError("Expected exception of type ${E::class} but was null")
+    } else if (exception !is E) {
+        throw AssertionError("Expected exception of type ${E::class} but was ${exception::class}")
+    }
+
+    return exception
+}
```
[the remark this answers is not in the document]
mrehan27: Updated with a bit more tweaks

tracking-migration/src/test/java/io/customer/tracking/migration/testutils/data/MigrationDataSet.kt
mrehan27: Yeah, but if we have done this in initial PR 😄 Since there are a lot of similar json samples in this file, I would stick with current implementation for now as all json samples will need to be updated to use this method. We can improve this later to look more clean.

PR #406, opened 2024-07-22
chore: updated implementation to pass enrichment

Shahroz16:
With latest update of Segment, we can pass enrichment along with event, so we don't have to utilize the plugin anymore.


datapipelines/src/main/kotlin/io/customer/datapipelines/migration/TrackingMigrationProcessor.kt
```diff
@@ -67,69 +56,86 @@ internal class TrackingMigrationProcessor(
...
     override suspend fun processTask(task: MigrationTask): Result<Unit> = runCatching {
-        val trackEvent = when (task) {
-            is MigrationTask.IdentifyProfile -> IdentifyEvent(
-                userId = task.identifier,
-                traits = task.attributes.toJsonObject()
-            )
+        data class MigrationEventData(
+            val trackEvent: BaseEvent,
+            val enrichmentClosure: EnrichmentClosure? = null
+        )
```
mrehan27, line 62: Why do we need to create the class every time processTask is called? 🤔 

We can make this class private if we don't want to expose to other modules, or use simple pair/map or similar class if we want to avoid adding new class just for migration.

tracking-migration/src/main/java/io/customer/tracking/migration/MigrationAssistant.kt
```diff
@@ -43,7 +43,6 @@ class MigrationAssistant internal constructor(
             queue.run()
             // Log completion of migration
             logger.debug("Migration completed successfully")
-            migrationProcessor.onMigrationCompleted()
```
mrehan27, line 46: Can you please fix `MigrationAssistantTest` too?

mrehan27 (approved):
Great update, glad we now have simpler workaround. Just a few comments that we should address before merging.

PR #408, opened 2024-07-22
test: queue run request

mrehan27:

### Changes

- Added tests for `QueueRunRequest`
- Added more `runCatching` calls to reduce crash possibilities
- Helper methods and classes for convenience


tracking-migration/src/main/java/io/customer/tracking/migration/queue/QueueRunRequest.kt
```diff
@@ -14,52 +14,61 @@ internal class QueueRunRequestImpl internal constructor(
...
-        val tasksToRun = inventory.toMutableList()
+        runCatching {
+            logger.debug("queue starting to run tasks...")
+            val tasksToRun = queueStorage.getInventory().toMutableList()
+            while (tasksToRun.isNotEmpty()) {
+                // get the next task to run using the query criteria
+                val currentTaskMetadata = queryRunner.getNextTask(tasksToRun)
+                if (currentTaskMetadata == null) {
+                    logger.debug("all queue tasks have been migrated or failed to run. Exiting queue run.")
+                    break
```
Shahroz16, line 25: codecov points out we are not testing the `break` case? it would be good to add it

tracking-migration/src/test/java/io/customer/tracking/migration/testutils/stubs/QueueStorageStub.kt
```diff
@@ -0,0 +1,60 @@
...
+    override fun get(taskStorageId: String): QueueTask? {
+        return inventory.find { it.taskPersistedId == taskStorageId }
+    }
+
+    override fun delete(taskStorageId: String): QueueModifyResult {
+        return inventory
+            .find { it.taskPersistedId == taskStorageId }
+            ?.let { task ->
+                inventory.remove(task)
+            } ?: false
```
Shahroz16, line 58:
```suggestion
        return inventory.removeIf { it.taskPersistedId == taskStorageId }
```

Shahroz16 (approved):

tracking-migration/src/main/java/io/customer/tracking/migration/queue/QueueRunRequest.kt
mrehan27: Added test case to cover this line too

PR #409, opened 2024-07-23
test: tracking migration processor

mrehan27:

### Changes

- Added tests for `TrackingMigrationProcessor`
- Fixed issues identified in `TrackingMigrationProcessor` while writing tests:
  - If device token does not exist in data pipelines, register the device
  - If device token already exists and is the same, do nothing
  - If device token already exists and is different, delete received token as existing token should be the most recent one
  - If profile is not identified in data pipelines, identify profile with given identifier
  - If profile is already identified, skip identify request
  - Matched `timestamp` format with what is expected by `analytics`
  - Updated `timestamp` and `userId` to ensure no other plugin overrides them later
- Updated `SDKComponentKeys` and added extensions to `SDKComponent` to allow overriding `TrackingMigrationProcessor`
- Added `SegmentInstantFormatter` for convenient date formatting in the desired format expected by `analytics`
- Added helper methods and extensions to facilitate writing newly added tests

### DIFF
--- a/datapipelines/src/main/kotlin/io/customer/datapipelines/migration/TrackingMigrationProcessor.kt
+++ b/datapipelines/src/main/kotlin/io/customer/datapipelines/migration/TrackingMigrationProcessor.kt
@@ -166,8 +166,8 @@ internal class TrackingMigrationProcessor(
 
         val enrichmentClosure: (event: BaseEvent) -> Unit = { event ->
             eventData.enrichmentClosure?.invoke(event)
-            // Update the event with the task identifier and timestamp at the end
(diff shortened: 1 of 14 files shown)


datapipelines/src/main/kotlin/io/customer/datapipelines/migration/TrackingMigrationProcessor.kt
mrehan27, line 1: The file includes bug fixes and some important changes

Shahroz16 (approved):
Looks good

PR #410, opened 2024-07-23
chore: added rules for retrofit

Shahroz16:
- Removed pro-guard rules from `sdk` and moved them to `in-app`


mrehan27 (approved):
Looks good 👍🏻

PR #415, opened 2024-07-24
fix: android user agent version

Shahroz16:
There was an issue in CI workflow where it commits the version file, but woudn't checkout to it before pushing it to maven central.

### DIFF
--- a/.github/workflows/deploy-sdk.yml
+++ b/.github/workflows/deploy-sdk.yml
@@ -150,7 +150,10 @@ jobs:
     if: needs.deploy-git-tag.outputs.new_release_published == 'true' # only run if a git tag was made.
     runs-on: ubuntu-latest
     steps:
+      - name: Checkout git tag that got created in previous step
       - uses: actions/checkout@v4
+        with:
+          ref: ${{ needs.deploy-git-tag.outputs.new_release_version }}
       - uses: ./.github/actions/setup-android
       - name: Push to Sonatype servers
         run: MODULE_VERSION=${{ needs.deploy-git-tag.outputs.new_release_version }} ./scripts/deploy-code.sh


mrehan27 (approved):
Looks good to me, hope this goes well 🤞🏻

levibostian (approved):
This change is [almost identical to the logic in iOS SDK where we checkout the git tag made before pushing to cocoapods](https://github.com/customerio/customerio-ios/blob/2f4e4e4d07bedffc251b0f0a19b3e2bd55b0375d/.github/workflows/deploy-sdk.yml#L186-L187). 

Looking at the file, I think it's a safe change but it could fail if `needs.deploy-git-tag.outputs.new_release_version` returns nil or something. I wonder if it's worth [adding the if conditional like we do in iOS](https://github.com/customerio/customerio-ios/blob/2f4e4e4d07bedffc251b0f0a19b3e2bd55b0375d/.github/workflows/deploy-sdk.yml#L185)?

Shahroz16:
> This change is [almost identical to the logic in iOS SDK where we check the git tag made before pushing to cocoapods](https://github.com/customerio/customerio-ios/blob/2f4e4e4d07bedffc251b0f0a19b3e2bd55b0375d/.github/workflows/deploy-sdk.yml#L186-L187).
> 
> Looking at the file, I think it's a safe change but it could fail if `needs.deploy-git-tag.outputs.new_release_version` returns nil or something. I wonder if it's worth [adding the if conditional like we do in iOS](https://github.com/customerio/customerio-ios/blob/2f4e4e4d07bedffc251b0f0a19b3e2bd55b0375d/.github/workflows/deploy-sdk.yml#L185)?

not fully sure why we are using this check, I cant recall but the best bet seems to be if we wanted to make a release using workflow, hence hesitant about adding it today.

PR #420, opened 2024-08-08
feat: public api for dismiss message

mrehan27:
### Background

Since we decoupled all modules from `CustomerIO` class and moved them to `core` module, modules were moved to the `SDKComponent` class. This broke some public APIs accessed through `SDKComponent` and made some of our documentation outdated (e.g., dismissing in-app messaging). To fix this, we had two options:

#### Option 1 - Expose in-app module through `CustomerIO` object again.

I think this is the most feasible option, so I opted for this.

#### Option 2 - Ask customers to use `SDKComponent` instead.

I don't think recommending customers to use `SDKComponent` is a very good idea, so I opted for option 1.

### Changes

- Exposed in-app module through the `CustomerIO` object for customer ease
- Exposed in-app module through module object directly to help Java customers
- Marked `SDKComponent` methods exposing these modules as internal

### Resources

- [Docs link that is currently broken](https://customer.io/docs/sdk/android/in-app/in-app-event-listeners/#dismiss-in-app-message)
- [Slack thread for the request](https://customerio.slack.com/archives/C02580R06/p1723131682532329)

### DIFF
--- a/messaginginapp/api/messaginginapp.api
+++ b/messaginginapp/api/messaginginapp.api
@@ -21,9 +21,11 @@ public final class io/customer/messaginginapp/ModuleMessagingInApp : io/customer
 	public synthetic fun getModuleConfig ()Lio/customer/sdk/core/module/CustomerIOModuleConfig;
 	public fun getModuleName ()Ljava/lang/String;
 	public fun initialize ()V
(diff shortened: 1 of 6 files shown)


Shahroz16 (approved):
LGTM

PR #427, opened 2024-08-26
chore: refactor Gist SDK interfacing

Shahroz16:

Changes:
- Removed unneeded interfacing for Gist SDK
- Update to Gist SDK to observe state changes/callbacks and utilize them


core/src/main/kotlin/io/customer/sdk/data/store/GlobalPreferenceStore.kt
```diff
@@ -10,7 +10,11 @@ import androidx.core.content.edit
 interface GlobalPreferenceStore {
     fun saveDeviceToken(token: String)
     fun getDeviceToken(): String?
-
+    fun saveUserId(token: String)
```
mrehan27, line 13: I think we shouldn't be using global store for in-app purposes. This can be greatly misleading and can pose challenges to anyone reading the codebase in future. I think we should either have a consistent global store managed by module identifying and keeping track of profiles (i.e. `datapipelines`). or create separate store for in-app to isolate it from other modules.

messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
```diff
@@ -12,82 +14,40 @@ import io.customer.sdk.events.Metric
...
-) : CustomerIOModule<MessagingInAppModuleConfig> {
+) : CustomerIOModule<MessagingInAppModuleConfig>, GistListener {
     override val moduleName: String = MODULE_NAME
     override val moduleConfig: MessagingInAppModuleConfig = config
 
     private val diGraph: SDKComponent = SDKComponent
     private val eventBus = diGraph.eventBus
-    private val gistProvider by lazy { diGraph.gistProvider }
+    private val gistProvider: GistProvider
+        get() = diGraph.gistProvider
```
mrehan27, line 24: I see a lot similar usages for `get` in the PR, is there a reason for preferring `get()` over following:

```
private val gistProvider: GistProvider = diGraph.gistProvider
```

Particularly when `gistProvider` is singleton and is used so many times in the class, I think using `get()` is slightly inefficient here 🤓

messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
```diff
@@ -12,82 +14,40 @@ import io.customer.sdk.events.Metric
...
 
-    private fun initializeGist() {
-        gistProvider.initProvider(
-            application = diGraph.android().application,
-            siteId = moduleConfig.siteId,
-            region = moduleConfig.region.code
-        )
-    }
-
     companion object {
```
mrehan27, line 51: This may sound a bit nitpicky 😁 But would appreciate if we could avoid placing `companion object` in middle of the class, as it makes the code harder to read IMO. It would be better to place it either at the bottom (preferred) or at the top of the class so it's easier to locate.

messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
```diff
@@ -97,4 +57,52 @@ class ModuleMessagingInApp(
                 ?: throw IllegalStateException("ModuleMessagingInApp not initialized")
         }
     }
+
+    override fun embedMessage(message: Message, elementId: String) {
+        moduleConfig.eventListener?.messageShown(InAppMessage.getFromGistMessage(message))
```
mrehan27, line 62: That's one place where using `get()` would be a good choice in my opinion 😅 

```kotlin
private val eventListener: InAppEventListener?
    get() = moduleConfig.eventListener
```

messaginginapp/src/main/java/io/customer/messaginginapp/di/DIGraphMessagingInApp.kt
```diff
@@ -1,24 +1,30 @@
...
+internal val SDKComponent.gistQueue: GistQueue
+    get() = singleton<GistQueue> { Queue() }
 
-internal val SDKComponent.gistProvider: InAppMessagesProvider
-    get() = newInstance<InAppMessagesProvider> { GistInAppMessagesProvider(gistApiProvider) }
+internal val SDKComponent.inAppMessaging: ModuleMessagingInApp
+    get() = ModuleMessagingInApp.instance()
 
-@Suppress("UnusedReceiverParameter")
-fun CustomerIOInstance.inAppMessaging(): ModuleMessagingInApp {
```
mrehan27, line 19: Why did we remove this? This will break public facing API again for dismissing in-app message 🤐

messaginginapp/src/main/java/io/customer/messaginginapp/di/DIGraphMessagingInApp.kt
```diff
@@ -1,24 +1,30 @@
...
+    get() = singleton<GistProvider> {
+        GistSdk(
+            application = android().application,
+            siteId = moduleConfig.siteId,
+            dataCenter = moduleConfig.region.code
+        )
+    }
 internal val SDKComponent.inAppMessagingManager: InAppMessagingManager
-    get() = singleton<InAppMessagingManager> { InAppMessagingManager }
+    get() = singleton { InAppMessagingManager(inAppMessaging) }
```
mrehan27, line 30: Just to be safe

```suggestion
    get() = singleton<InAppMessagingManager> { InAppMessagingManager(inAppMessaging) }
```

messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/listeners/Queue.kt
```diff
@@ -22,18 +22,25 @@ import okhttp3.ResponseBody.Companion.toResponseBody
...
-    internal var shownMessageQueueIds = mutableSetOf<String>()
-    private val inAppMessagingManager: InAppMessagingManager = SDKComponent.inAppMessagingManager
+class Queue : GistQueue {
 
-    init {
-        GistSdk.addListener(this)
-    }
+    private val inAppMessagingManager: InAppMessagingManager
+        get() = SDKComponent.inAppMessagingManager
+    private val state: InAppMessagingState
```
mrehan27, line 35: As class level property, wouldn't `currentState` be more descriptive instead? 🤔 
```suggestion
    private val currentState: InAppMessagingState
```

messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/listeners/Queue.kt
```diff
@@ -22,18 +22,25 @@ import okhttp3.ResponseBody.Companion.toResponseBody
...
+    private val scope: CoroutineScope = SDKComponent.scopeProvider.lifecycleListenerScope
+    private val application: Context
+        get() = SDKComponent.android().applicationContext
 
     private val cacheSize = 10 * 1024 * 1024 // 10 MB
-    private val cacheDirectory by lazy { File(GistSdk.application.cacheDir, "http_cache") }
+    private val cacheDirectory by lazy { File(application.cacheDir, "http_cache") }
     private val cache by lazy { Cache(cacheDirectory, cacheSize.toLong()) }
 
     private fun saveToPrefs(context: Context, key: String, value: String) {
```
mrehan27, line 46: I think we don't need `context` here now since it is available at class level?

messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/listeners/Queue.kt
```diff
@@ -46,111 +53,87 @@ class Queue : GistListener {
...
 
-    internal fun fetchUserMessagesFromLocalStore() {
-        handleMessages(localMessageStore)
-    }
+            response
+        }.build()
 
-    internal fun clearUserMessagesFromLocalStore() {
-        localMessageStore.clear()
+        Retrofit.Builder().baseUrl(GistEnvironment.PROD.getGistQueueApiUrl()).addConverterFactory(GsonConverterFactory.create()).client(httpClient).build().create(GistQueueService::class.java)
```
mrehan27, line 105: I was expecting the environment to be used from state, is that a typo? Also, would appreciate splitting to multiple lines for better readability.

```suggestion
        Retrofit.Builder()
            .baseUrl(state.environment.getGistQueueApiUrl())
            .addConverterFactory(GsonConverterFactory.create())
            .client(httpClient)
            .build()
            .create(GistQueueService::class.java)
```

messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/listeners/Queue.kt
```diff
@@ -22,18 +22,25 @@ import okhttp3.ResponseBody.Companion.toResponseBody
...
 
-    init {
-        GistSdk.addListener(this)
-    }
+    private val inAppMessagingManager: InAppMessagingManager
+        get() = SDKComponent.inAppMessagingManager
+    private val state: InAppMessagingState
+        get() = inAppMessagingManager.getCurrentState()
+    private val logger: Logger = SDKComponent.logger
+    private val scope: CoroutineScope = SDKComponent.scopeProvider.lifecycleListenerScope
```
mrehan27, line 38: I'm assuming this will be replaced with dedicated scope in future PRs

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistSdk.kt
```diff
@@ -1,339 +1,122 @@
...
-        GistSdk.application = application
-        GistSdk.siteId = siteId
-        GistSdk.dataCenter = dataCenter
-        isInitialized = true
-        gistEnvironment = environment
-
-        subscribeToLifecycleEvents()
+    private fun onActivityPaused() {
+        logger.debug("Activity paused, stopping polling")
+        resetTimer()
```
mrehan27, line 55: I think this can break things. Previously, we stopped timer only when activities list was empty, this seems to stop on pause state for every activity? Also, if user never stays for 10 seconds, will this never fetch as on every screen change, the timer is stopped 🤔

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistSdk.kt
```diff
@@ -1,339 +1,122 @@
...
-        gistQueue = Queue()
-        gistModalManager = GistModalManager()
-        currentRoute = ""
+    private fun fetchInAppMessages(duration: Long) {
+        logger.debug("Starting polling with duration: $duration")
+        timer?.cancel()
+        // create a timer to run the task after the initial run
+        timer = timer(name = "GistPolling", daemon = true, period = 20000) {
+            gistQueue.fetchUserMessages()
+        }
```
mrehan27, line 77: I don't see `duration` used in timer here. Did I miss something? 👀

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistSdk.kt
```diff
@@ -1,339 +1,122 @@
...
-    internal fun handleGistAction(
-        message: Message,
-        currentRoute: String,
-        action: String,
-        name: String
-    ) {
-        for (listener in listeners) {
-            listener.onAction(message, currentRoute, action, name)
+    override fun setUserId(userId: String) {
+        if (state.userId == userId) {
```
mrehan27, line 107: Should we add similar check on route change too? 🤔

messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
Shahroz16: The benefit of having this is, on compile-time behavior, this declaration does not set any value
It simply defines a property with a custom-getter and usage of it is very limited so even if recomputes, the value is `Singleton` so it brings back the old value without creating a new instance. 

`GistProvider` takes in an `application` instance which is dependent on the SDK lifecycle, so want to only access its value when it's about to get used instead of accessing it as compile time. 

We can use `lazy` here too, but it does the exactly same thing,
- do not initialize at compile time
- stores the value in a parameter and then checks if it exists and returns it.

core/src/main/kotlin/io/customer/sdk/data/store/GlobalPreferenceStore.kt
Shahroz16: > I think we should either have a consistent global store managed by module identifying and keeping track of profiles (i.e. datapipelines)

These modules are independent and should not be bound to `datapipelines`, the idea of adding it in global was if the value is present there, no matter which source added it, other modules can utilize it too, if the value isn't there, they can rely on their own local values. 

I am okay with creating module specific preference store, but we would need to do it for every module, if that's okay, happy to create one.

messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
Shahroz16: I am not sure, how? `ModuleMessagingInApp` can't be created without `MessagingInAppModuleConfig` which takes `eventListener`. 

There's no way to modify the `eventListener` after ModuleMessagingInApp is initialized, so what benefit are we getting from using `get()` here?

messaginginapp/src/main/java/io/customer/messaginginapp/di/DIGraphMessagingInApp.kt
Shahroz16: good point

messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/listeners/Queue.kt
Shahroz16: yup

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistSdk.kt
Shahroz16: nope onResume, makes the network call first and then run the timer.

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistSdk.kt
Shahroz16: it was on middleware level, but added here as it would save load

core/src/main/kotlin/io/customer/sdk/data/store/GlobalPreferenceStore.kt
mrehan27: Discussed on call, we'll add module specific preference store to in-app module

messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
mrehan27: Discussed on call, we'll continue using `get` for classes accessing `AndroidSDKComponent` to keep them lazy and updated for others.

messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
mrehan27: Discussed on call, this isn't needed at the moment

mrehan27 (approved):
Looks good

Shahroz16:
Merging this will shift the prefs in another PR to reduce load

PR #428, opened 2024-08-26
chore: middleware and UI logic

Shahroz16:

changes:
- Added middleware
- Added changes in UI layers wrt to states
- Removed log events from classes


messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/listeners/Queue.kt
```diff
@@ -43,66 +43,87 @@ class Queue : GistQueue {
...
+            .baseUrl(GistEnvironment.PROD.getGistQueueApiUrl())
+            .addConverterFactory(GsonConverterFactory.create())
+            .client(httpClient)
+            .build()
+            .create(GistQueueService::class.java)
     }
 
-    override fun clearPrefs(context: Context) {
-        context.getSharedPreferences("network_cache", Context.MODE_PRIVATE).edit().clear().apply()
+    private fun handleResponse(response: okhttp3.Response, originalRequest: okhttp3.Request): okhttp3.Response {
```
mrehan27, line 81: I think keywords like `intercept` or `transform` would be more appropriate for this method and the methods it calls. I would expect a method with `handle` prefix to be something that executes a final action. However, this method seems more focused on intercepting and transforming the response.

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistModalActivity.kt
```diff
@@ -19,26 +19,33 @@ import io.customer.messaginginapp.gist.data.model.Message
...
-
-    // Indicates if the message is visible to user or not
+    // indicates if the message is visible to user or not
     internal val isEngineVisible: Boolean
         get() = binding.gistView.isEngineVisible
 
+    private var messagePosition: MessagePosition = MessagePosition.CENTER
+
+    private val currentMessageState: MessageState.Loaded?
+        get() = state.currentMessageState as? MessageState.Loaded
```
mrehan27, line 47: Shouldn't this be `currentLoadedMessageState` for clarity?

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistModalActivity.kt
```diff
@@ -52,21 +59,22 @@ class GistModalActivity : AppCompatActivity(), GistListener, GistViewListener, T
...
         binding = ActivityGistBinding.inflate(layoutInflater)
         setContentView(binding.root)
         val messageStr = this.intent.getStringExtra(GIST_MESSAGE_INTENT)
         val modalPositionStr = this.intent.getStringExtra(GIST_MODAL_POSITION_INTENT)
-        Gson().fromJson(messageStr, Message::class.java)?.let { messageObj ->
-            inAppMessagingManager.dispatch(InAppMessagingAction.LogEvent("GisModelActivity onCreate: $messageObj"))
-            currentMessage = messageObj
-            isCancelled = false
-            currentMessage?.let { message ->
+        val parsedMessage = kotlin.runCatching { Gson().fromJson(messageStr, Message::class.java) }.getOrNull()
```
mrehan27, line 67: I think we should handle onFailure as well in case something breaks here?

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistModalActivity.kt
```diff
@@ -77,32 +85,49 @@ class GistModalActivity : AppCompatActivity(), GistListener, GistViewListener, T
...
     override fun onPause() {
-        inAppMessagingManager.dispatch(InAppMessagingAction.LogEvent("GisModelActivity onPause"))
-        GistSdk.removeListener(this)
         super.onPause()
         overridePendingTransition(0, 0)
     }
 
     override fun finish() {
-        inAppMessagingManager.dispatch(InAppMessagingAction.LogEvent("GisModelActivity finish"))
+        logger.debug("GisModelActivity finish")
```
mrehan27, line 130: Not for this PR though, but noticed typo in `GisModelActivity` log in whole. Something that we can note and update later.

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistModalActivity.kt
```diff
@@ -112,37 +137,40 @@ class GistModalActivity : AppCompatActivity(), GistListener, GistViewListener, T
             animation.setTarget(binding.modalGistViewLayout)
             animation.start()
             animation.doOnEnd {
-                inAppMessagingManager.dispatch(InAppMessagingAction.LogEvent("GisModelActivity finish animation completed"))
+                logger.debug("GisModelActivity finish animation completed")
```
mrehan27, line 140: IIRC, we saw some challenges because dismiss was called, but animation was not complete, causing some unexpected behaviors. Shouldn't we have a state for clarity on being dismissed and being fully dismissed and wiped out for better control?

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistModalActivity.kt
mrehan27: Discussed on call, this is already handled. I just misread the code slightly.

messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessagingMiddlewares.kt
```diff
@@ -0,0 +1,224 @@
...
+import io.customer.sdk.core.di.SDKComponent
+import org.reduxkotlin.Store
+import org.reduxkotlin.middleware
+
+/**
+ * Middleware to log actions and state changes.
+ */
+internal fun loggerMiddleware() = middleware<InAppMessagingState> { store, next, action ->
+    SDKComponent.logger.debug("Store: action: ${action::class.simpleName}: $action")
+    SDKComponent.logger.debug("Store: state before reducer: ${store.state}")
```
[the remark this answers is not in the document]
Shahroz16: We can but its going to get too crowded

messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessagingMiddlewares.kt
```diff
@@ -0,0 +1,239 @@
...
+        next(action)
+    }
+}
+
+/**
+ * Middleware to handle Gist listener actions.
+ *
+ * @param gistListener The Gist listener.
+ */
+internal fun gistListenerMiddleware(gistListener: GistListener?) = middleware<InAppMessagingState> { store, next, action ->
```
mrehan27, line 212: We can fix lint warning by renaming store here

```
Parameter 'store' is never used, could be renamed to _
```

messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessagingState.kt
```diff
@@ -32,19 +32,53 @@ data class InAppMessagingState(
...
+    }
+    if (this.messagesInQueue != other.messagesInQueue) {
+        differences["messagesInQueue"] = Pair(this.messagesInQueue, other.messagesInQueue)
+    }
+    if (this.shownMessageQueueIds != other.shownMessageQueueIds) {
+        differences["shownMessageQueueIds"] = Pair(this.shownMessageQueueIds, other.shownMessageQueueIds)
+    }
+
+    return differences
+}
```
mrehan27, line 84: Since the comparison is mostly same, we can eliminate repetitive code and implement something like this:

```kotlin
fun InAppMessagingState.diff(other: InAppMessagingState): Map<String, Pair<Any?, Any?>> {
    val differences = mutableMapOf<String, Pair<Any?, Any?>>()

    fun <T> putIfDifferent(key: String, value: InAppMessagingState.() -> T) {
        val thisValue = value(this)
        val otherValue = value(other)
        if (thisValue != otherValue) {
            differences[key] = Pair(thisValue, otherValue)
        }
    }

    putIfDifferent("siteId") { siteId }
    putIfDifferent("environment") { environment }
    // other values

    return differences
}
```

We can use `Any` instead of `T` too to further simplify if needed.

mrehan27 (approved):
Looks good

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistView.kt
```diff
@@ -46,119 +50,120 @@ class GistView @JvmOverloads constructor(
...
-                endpoint = GistSdk.gistEnvironment.getEngineApiUrl(),
+                endpoint = store.environment.getEngineApiUrl(),
                 properties = message.properties
             )
             engineWebView?.setup(engineWebConfiguration)
         }
     }
 
     fun stopLoading() {
-        inAppMessagingManager.dispatch(InAppMessagingAction.EngineStopLoading)
```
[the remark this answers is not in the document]
mrehan27: This is dismissed state. I was asking for cancelled state, when the message is cancelled and dismissed because route has changed and we need to stop further loading and sending callbacks to customer.

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistModalActivity.kt
mrehan27: Correct. And IIRC there were some challenges faced with it, probably because we cleared current message reference before animation was complete, so 100ms could be tricky where there were multiple messages and the route changed in-between. Do you think it won't be an issue with new state implementation?

messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessagingState.kt
Shahroz16: I think it can be simplified even more than that, I wanted to keep it readable 😄 let me find another way

PR #429, opened 2024-08-26
chore: added tests

Shahroz16:

changes:
- Added/Updated tests
- Some refactor and PR suggestions


messaginginapp/src/test/java/io/customer/messaginginapp/InAppMessagingStoreTest.kt
```diff
@@ -0,0 +1,372 @@
...
+    override fun setup(testConfig: TestConfig) {
+        super.setup(
+            testConfigurationDefault {
+                diGraph {
+                    sdk {
+                        overrideDependency<ModuleMessagingInApp>(module)
+                    }
+                }
+            }
+        )
```
mrehan27, line 58: I think `attachToSDKComponent` does the same and this can be simplified to

```suggestion
        super.setup(testConfig)
```

messaginginapp/src/test/java/io/customer/messaginginapp/InAppMessagingStoreTest.kt
```diff
@@ -0,0 +1,372 @@
...
+            Message(queueId = "2", priority = 1),
+            Message(queueId = "3", priority = 3)
+        )
+
+        initializeAndSetUser()
+        manager.dispatch(InAppMessagingAction.ProcessMessageQueue(messages))
+
+        val state = manager.getCurrentState()
+        state.messagesInQueue.size shouldBeEqualTo 3
+        (state.currentMessageState as? MessageState.Processing)?.message?.queueId shouldBeEqualTo "2"
```
mrehan27, line 87: We should use `shouldBeInstanceOf` here instead

```suggestion
        state.currentMessageState
            .shouldBeInstanceOf<MessageState.Processing>()
            .message
            .queueId shouldBeEqualTo "2"
```

messaginginapp/src/test/java/io/customer/messaginginapp/InAppMessagingStoreTest.kt
```diff
@@ -0,0 +1,372 @@
...
+        val state = manager.getCurrentState()
+        state.messagesInQueue.size shouldBeEqualTo 2
+        state.messagesInQueue.any { it.queueId == "1" } shouldBe true
+        state.messagesInQueue.any { it.queueId == "2" } shouldBe true
+    }
+
+    @Test
+    fun givenMessageWithSpecificRouteRule_whenRouteChanges_thenMessageStateUpdatesAccordingly() = runTest {
+        initializeAndSetUser()
+        val message = Message(queueId = "1", properties = mapOf("gist" to mapOf("routeRuleAndroid" to "home")))
```
mrehan27, line 110: I believe `routeRuleAndroid` is supposed to be a regex? We have `pageRuleContains` and `pageRuleEquals` extensions to create regex matching values received from server. I suggest we used them instead here for more accurate results.

```suggestion
        val message = Message(queueId = "1", properties = mapOf("gist" to mapOf("routeRuleAndroid" to pageRuleEquals("home"))))
```

messaginginapp/src/test/java/io/customer/messaginginapp/InAppMessagingStoreTest.kt
```diff
@@ -0,0 +1,372 @@
...
+
+        manager.dispatch(InAppMessagingAction.ProcessMessageQueue(listOf(message)))
+        manager.dispatch(InAppMessagingAction.SetPageRoute("home"))
+
+        var state = manager.getCurrentState()
+        state.currentRoute shouldBe "home"
+        (state.currentMessageState as? MessageState.Processing)?.message?.queueId shouldBe "1"
```
mrehan27, line 117: This and all similar castings should be replaced with `shouldBeInstanceOf`

messaginginapp/src/test/java/io/customer/messaginginapp/InAppMessagingStoreTest.kt
```diff
@@ -0,0 +1,372 @@
...
+
+        val currentState = state.currentMessageState
+        currentState shouldBeInstanceOf MessageState.Dismissed::class.java
+        (currentState as MessageState.Dismissed).message.queueId shouldBe "1"
+    }
+
+    @Test
+    fun givenMultipleMessagesWithDifferentRouteRules_whenRouteChanges_thenCorrectMessageIsDisplayed() = runTest {
+        initializeAndSetUser()
+        val homeMessage = Message(queueId = "1", properties = mapOf("gist" to mapOf("routeRuleAndroid" to "home")))
```
mrehan27, line 132: Same comment for all `routeRuleAndroid` values in the test, should leverage `pageRuleContains` and `pageRuleEquals`

messaginginapp/src/test/java/io/customer/messaginginapp/InAppMessagingStoreTest.kt
```diff
@@ -0,0 +1,372 @@
...
+        manager.dispatch(InAppMessagingAction.ProcessMessageQueue(listOf(homeMessage, profileMessage, generalMessage)))
+        manager.dispatch(InAppMessagingAction.SetPageRoute("home"))
+
+        // verify general message is displayed first (as it has no route rule)
+        var state = manager.getCurrentState()
+        val messageBeingDisplayed = (state.currentMessageState as? MessageState.Processing)?.message
+        messageBeingDisplayed?.queueId shouldBe "3"
+
+        // make the message visible and then dismiss it
+        manager.dispatch(InAppMessagingAction.DisplayMessage(messageBeingDisplayed!!))
```
mrehan27, line 146: By using `shouldBeInstanceOf`, we should be able to remove all not-null assertion operators in the test

messaginginapp/src/test/java/io/customer/messaginginapp/MessagingInAppIntegrationTest.kt
mrehan27, line 1: There are some extensions left in `GistExtensions.kt` that can also be removed as they were only used by this test file

messaginginapp/src/test/java/io/customer/messaginginapp/ModuleMessagingInAppTest.kt
```diff
@@ -148,4 +108,100 @@ internal class ModuleMessagingInAppTest : JUnitTest() {
         // verify that the module's dismissMessage method was called
         assertCalledOnce { inAppMessagesProviderMock.dismissMessage() }
     }
+
+    @Test
+    fun `when screen viewed event occurs, gist provider sets current route`() {
```
mrehan27, line 113: I know this is a common naming convention used in Android, but it seems inconsistent with our codebase. I would appreciate if we could follow the same naming convention at least across the file.

mrehan27 (approved):
No major blockers, just some small improvements that we should address before merging.

messaginginapp/src/test/java/io/customer/messaginginapp/ModuleMessagingInAppTest.kt
Shahroz16: yeah that was added by mistake

messaginginapp/src/test/java/io/customer/messaginginapp/InAppMessagingStoreTest.kt
Shahroz16: hmm makes sense, i would rather move the module here as well, otherwise doesn't make at top

PR #430, opened 2024-08-28
chore: store for in-app module

Shahroz16:

changes:
- adds specific store for in-app module


messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/listeners/Queue.kt
```diff
@@ -17,12 +18,26 @@ import okhttp3.Cache
...
 import okhttp3.OkHttpClient
 import okhttp3.ResponseBody.Companion.toResponseBody
+import retrofit2.Response
 import retrofit2.Retrofit
 import retrofit2.converter.gson.GsonConverterFactory
+import retrofit2.http.Body
+import retrofit2.http.POST
+import retrofit2.http.Path
+
+interface GistQueueService {
```
mrehan27, line 28: Any special reason for moving it into same file? I'm okay with either, but keeping such classes separate help keeping code files small IMO.

mrehan27 (approved):
Looks good

messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/listeners/Queue.kt
Shahroz16: Previous path didn't make sense, `/data/repository/GistQueueService.kt` and its a very small interface, not even being used in multiple clients, so didn't made a lot of sense for it be in its own file.

PR #432, opened 2024-08-28
chore: added more tests

Shahroz16:


messaginginapp/src/test/java/io/customer/messaginginapp/InAppPreferenceStoreTest.kt
```diff
@@ -0,0 +1,91 @@
...
+
+@RunWith(RobolectricTestRunner::class)
+class InAppPreferenceStoreTest : IntegrationTest() {
+
+    private lateinit var inAppPreferenceStore: InAppPreferenceStoreImpl
+    private lateinit var context: Context
+
+    override fun setup(testConfig: TestConfig) {
+        super.setup(testConfig)
+        context = ApplicationProvider.getApplicationContext()
```
mrehan27, line 23: We can use `applicationMock` instead from base class here? 🤔

mrehan27 (approved):
Looks good 👍🏻

PR #436, opened 2024-09-09
fix: updates tracking keys

Shahroz16:
closes: [MBL-516: Android SDK (4x?) is not tracking responses for in-app message clicks / not tracking all click data](https://linear.app/customerio/issue/MBL-516/android-sdk-4x-is-not-tracking-responses-for-in-app-message-clicks-not)

`action_name`, `action_value` were being used in tracking API calls, whereas data pipeline required the keys to be `actionName` and `actionValue`


mrehan27 (approved):
Looks good. Although I understand why the change was needed, I would appreciate some details in PR description or linked ticket. This will help anyone reading it in the future understand why this change was necessary and why the names are different from the ones used with tracking API.

PR #437, opened 2024-09-10
chore: ignore embed and cleanup properties

Shahroz16:
changes:
- ignore embedded messages that were blocking the queue from being processed
- move GistProperties as part of `Message`


messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/model/Message.kt
```diff
@@ -22,53 +22,55 @@ data class Message(
...
         }
+        return GistProperties(routeRule = routeRule, elementId = elementId, campaignId = campaignId, position = position, persistent = persistent)
+    }
+
+    override fun toString(): String {
+        return "Message(messageId=$messageId, instanceId=$instanceId, priority=$priority, queueId=$queueId, properties=$gistProperties"
     }
 }
+
+fun Message.getRouteRule(): String? {
```
mrehan27, line 74: This can probably be removed now?

mrehan27 (approved):
Looks good 👍🏻

PR #439, opened 2024-09-19
chore: use meta-data for user agent client

mrehan27:
part of [MBL-522](https://linear.app/customerio/issue/MBL-522/user-agent-rn-android)

### Changes

- Moved `Client` class out of `AndroidSDKComponent` constructor and made it a singular dependency, allowing for decoupling
- Simplified `Client` class by removing sealed implementation
- Updated `Client` class to create user-agent client based on meta-data provided in `AndroidManifest`
- Added and used extensions to fetch meta-data from `AndroidManifest` and fixed the deprecated method warning for `ApplicationInfo`
- Added tests to verify the changes
- Removed `ClientArgument` from `TestArgument` as it is no longer needed

### Benefits of this approach

- `Client` is no longer dependent on SDK initialization
- `Client` will always remain the same, regardless of which classes initialize the SDK and register `AndroidSDKComponent`
- `Client` can still be overridden in `AndroidSDKComponent`

### User agent sent from Android SDK

```
Customer.io Android Client/4.2.0 (Google Pixel 6; 34) io.customer.android.sample.java_layout/1.0
```

### User agent sent from React Native SDK (Pending changes)

```
Customer.io ReactNative Client/3.8.0 (Google sdk_gphone64_arm64; 33) io.customer.rn_sample.fcm/1.0
```

### DIFF
--- a/common-test/src/main/java/io/customer/commontest/core/BaseTest.kt
+++ b/common-test/src/main/java/io/customer/commontest/core/BaseTest.kt
@@ -45,6 +45,6 @@ abstract class BaseTest {
         val application = testConfig.argumentOrNull<ApplicationArgument>()?.value ?: return
         val client = testConfig.argumentOrNull<ClientArgument>()?.value ?: return
 
-        testConfig.configureAndroidSDKComponent(SDKComponent.registerAndroidSDKComponent(application, client))
+        testConfig.configureAndroidSDKComponent(SDKComponent.registerAndroidSDKComponent(application))
     }
 }
--- a/common-test/src/main/java/io/customer/commontest/core/BaseTest.kt
+++ b/common-test/src/main/java/io/customer/commontest/core/BaseTest.kt
@@ -1,7 +1,6 @@
 package io.customer.commontest.core
 
 import io.customer.commontest.config.ApplicationArgument
-import io.customer.commontest.config.ClientArgument
 import io.customer.commontest.config.TestConfig
 import io.customer.commontest.config.argumentOrNull
 import io.customer.commontest.config.configureAndroidSDKComponent
@@ -43,7 +42,6 @@ abstract class BaseTest {
      */
     private fun registerAndroidSDKComponent(testConfig: TestConfig) {
         val application = testConfig.argumentOrNull<ApplicationArgument>()?.value ?: return
-        val client = testConfig.argumentOrNull<ClientArgument>()?.value ?: return
 
         testConfig.configureAndroidSDKComponent(SDKComponent.registerAndroidSDKComponent(application))
     }
--- a/core/src/androidTest/AndroidManifest.xml
+++ b/core/src/androidTest/AndroidManifest.xml
@@ -1,13 +1,22 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools">
 
     <application>
+        <!--
+        This is how wrapper SDKs can provide their own values for the SDK_SOURCE and SDK_VERSION meta-data.
+        tools:replace="android:value" is used to replace the value of the meta-data.
+        tools:node="replace" is used to replace the entire meta-data tag.
+        Either of these attributes can be used to resolve any conflicts that may arise when merging
+        multiple AndroidManifest.xml files with the same meta-data tags.
+        -->
         <meta-data
-            android:name="io.customer.sdk.android.core.USER_AGENT"
-            android:value="TestUserAgent" />
+            android:name="io.customer.sdk.android.core.SDK_SOURCE"
+            android:value="TestUserAgent"
+            tools:replace="android:value" />
         <meta-data
             android:name="io.customer.sdk.android.core.SDK_VERSION"
-            android:value="1.3.5" />
+            android:value="1.3.5"
+            tools:node="replace" />
     </application>
-
 </manifest>
--- a/core/src/androidTest/AndroidManifest.xml
+++ b/core/src/androidTest/AndroidManifest.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <application>
+        <meta-data
+            android:name="io.customer.sdk.android.core.USER_AGENT"
+            android:value="TestUserAgent" />
+        <meta-data
+            android:name="io.customer.sdk.android.core.SDK_VERSION"
+            android:value="1.3.5" />
+    </application>
+
+</manifest>
--- a/core/src/androidTest/java/io/customer/sdk/data/store/AndroidManifestClientTest.kt
+++ b/core/src/androidTest/java/io/customer/sdk/data/store/AndroidManifestClientTest.kt
@@ -1,0 +1,15 @@
+package io.customer.sdk.data.store
+
+import io.customer.commontest.core.AndroidTest
+import io.customer.sdk.core.extensions.applicationMetaData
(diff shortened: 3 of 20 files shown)


core/src/androidTest/AndroidManifest.xml
```diff
@@ -0,0 +1,13 @@
...
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <application>
+        <meta-data
+            android:name="io.customer.sdk.android.core.USER_AGENT"
+            android:value="TestUserAgent" />
+        <meta-data
+            android:name="io.customer.sdk.android.core.SDK_VERSION"
+            android:value="1.3.5" />
+    </application>
```
mrehan27, line 11: This is only for tests

messagingpush/src/main/java/io/customer/messagingpush/CustomerIOPushNotificationHandler.kt
```diff
@@ -114,16 +114,7 @@ internal class CustomerIOPushNotificationHandler(
...
-            context.packageManager.getApplicationInfo(
-                context.packageName,
-                PackageManager.GET_META_DATA
-            )
-        } catch (ex: Exception) {
-            logger.error("Package not found ${ex.message}")
-            null
-        }
-        val appMetaData = applicationInfo?.metaData
+        val appMetaData = context.applicationMetaData()
```
mrehan27, line 117: Reused extension to fix deprecated warning

common-test/src/main/java/io/customer/commontest/core/BaseTest.kt
```diff
@@ -43,8 +42,7 @@ abstract class BaseTest {
      */
     private fun registerAndroidSDKComponent(testConfig: TestConfig) {
         val application = testConfig.argumentOrNull<ApplicationArgument>()?.value ?: return
-        val client = testConfig.argumentOrNull<ClientArgument>()?.value ?: return
```
mrehan27, line 46: We now no longer need Client to register and initialize `AndroidSDKComponent`

core/src/androidTest/AndroidManifest.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <application>
+        <meta-data
+            android:name="io.customer.sdk.android.core.USER_AGENT"
```
Shahroz16, line 6:
```suggestion
            android:name="io.customer.sdk.android.core.SDK_SOURCE"
```

core/src/main/kotlin/io/customer/sdk/data/store/Client.kt
```diff
@@ -1,83 +1,45 @@
...
-        sdkVersion: String
-    ) : Client(source = source, sdkVersion = sdkVersion)
-
     companion object {
-        internal const val SOURCE_ANDROID = "Android"
-        internal const val SOURCE_REACT_NATIVE = "ReactNative"
-        internal const val SOURCE_EXPO = "Expo"
-        internal const val SOURCE_FLUTTER = "Flutter"
+        private const val SOURCE_ANDROID = "Android"
+        internal const val META_DATA_USER_AGENT = "io.customer.sdk.android.core.USER_AGENT"
```
Shahroz16, line 21:
```suggestion
        internal const val META_DATA_SDK_SOURCE = "io.customer.sdk.android.core.SDK_SOURCE"
```

core/src/main/kotlin/io/customer/sdk/data/store/Client.kt
```diff
@@ -1,83 +1,45 @@
...
-            source.equals(
-                other = SOURCE_EXPO,
-                ignoreCase = true
-            ) -> Expo(sdkVersion = sdkVersion)
-            source.equals(
-                other = SOURCE_FLUTTER,
-                ignoreCase = true
-            ) -> Flutter(sdkVersion = sdkVersion)
-            else -> Other(source = source, sdkVersion = sdkVersion)
+        fun fromMetadata(metadata: Bundle?): Client {
```
Shahroz16, line 33: what would happen if both android SDK and react native SDK manifest have same values? shouldn't we override or handle that case?

core/src/androidTest/java/io/customer/sdk/data/store/AndroidManifestClientTest.kt
```diff
@@ -0,0 +1,15 @@
+package io.customer.sdk.data.store
+
+import io.customer.commontest.core.AndroidTest
+import io.customer.sdk.core.extensions.applicationMetaData
+import org.amshove.kluent.shouldBeEqualTo
+import org.junit.Test
+
+class AndroidManifestClientTest : AndroidTest() {
```
Shahroz16, line 8: it would be great if we can test meta data merging here

core/src/main/kotlin/io/customer/sdk/core/extensions/ContextExtensions.kt
```diff
@@ -0,0 +1,29 @@
...
+        )
+    } else {
+        @Suppress("DEPRECATION")
+        packageManager.getApplicationInfo(packageName, PackageManager.GET_META_DATA)
+    }
+
+    return@runCatching appInfo.metaData
+}.onFailure { ex ->
+    SDKComponent.logger.error("Failed to get application meta-data: ${ex.message}")
+}.getOrNull()
```
Shahroz16, line 29:
```suggestion
/**
 * Gets meta-data from AndroidManifest.xml file.
 *
 * @return The meta-data bundle from application info, or null if retrieval fails.
 */
fun Context.getApplicationMetaData(): Bundle? {
    return try {
        val appInfo = getApplicationInfo()
        appInfo.metaData
    } catch (e: PackageManager.NameNotFoundException) {
        SDKComponent.logger.error("Failed to get application meta-data: ${e.message}")
        null
    }
}

/**
 * Helper function to get ApplicationInfo based on Android version.
 */
@Suppress("DEPRECATION")
private fun Context.getApplicationInfo(): ApplicationInfo {
    return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
        packageManager.getApplicationInfo(
            packageName,
            PackageManager.ApplicationInfoFlags.of(PackageManager.GET_META_DATA.toLong())
        )
    } else {
        packageManager.getApplicationInfo(packageName, PackageManager.GET_META_DATA)
    }
}
```

I think this method is doing too much 😅, we can probably decouple it to make it more readable and keep it fail-proof?

Shahroz16 (commented):
Approach looks solid, great idea, some suggestions and 1 thing to validate before we finalize

core/src/androidTest/AndroidManifest.xml
mrehan27: Updated

core/src/main/kotlin/io/customer/sdk/data/store/Client.kt
mrehan27: Added values in common test to help verify manifest merger challenges

core/src/androidTest/java/io/customer/sdk/data/store/AndroidManifestClientTest.kt
mrehan27: Added multiple manifests, test will not change because the changes have to be made when merging Manifest

core/src/main/kotlin/io/customer/sdk/core/extensions/ContextExtensions.kt
mrehan27: This is a bit tricky because the messy code is mainly due to deprecation warnings and fetching application info the recommended way for different API levels, while try-catch is for application info not the meta-data. I’ve tried splitting it into smaller method that throws, breaking it into two halves. Let me know if this looks better.

Shahroz16 (approved):
Looks good

PR #441, opened 2024-09-24
fix: add userId for deleteToken

Shahroz16:
closes: [MBL-528: Android SDK should consistently add userId in delete token payload](https://linear.app/customerio/issue/MBL-528/android-sdk-should-consistently-add-userid-in-delete-token-payload)

Since event are asynchronous, we can have `userId` null before `deleteDevice` access the value of `userId`. So we should enrich the delete device event with `userId` so it always have that value.

### DIFF
--- a/datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
+++ b/datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
@@ -4,6 +4,7 @@ import androidx.annotation.VisibleForTesting
 import com.segment.analytics.kotlin.android.Analytics
 import com.segment.analytics.kotlin.core.Analytics
 import com.segment.analytics.kotlin.core.ErrorHandler
+import com.segment.analytics.kotlin.core.platform.EnrichmentClosure
 import com.segment.analytics.kotlin.core.platform.plugins.logger.LogKind
 import com.segment.analytics.kotlin.core.platform.plugins.logger.LogMessage
 import io.customer.base.internal.InternalCustomerIOApi
@@ -182,7 +183,8 @@ class CustomerIO private constructor(
     override fun <Traits> identify(
         userId: String,
         traits: Traits,
-        serializationStrategy: SerializationStrategy<Traits>
+        serializationStrategy: SerializationStrategy<Traits>,
+        enrichment: EnrichmentClosure?
     ) {
         if (userId.isBlank()) {
             logger.debug("Profile cannot be identified: Identifier is blank. Please retry with a valid, non-empty identifier.")
@@ -206,7 +208,8 @@ class CustomerIO private constructor(
         analytics.identify(
             userId = userId,
             traits = traits,
-            serializationStrategy = serializationStrategy
+            serializationStrategy = serializationStrategy,
+            enrichment = enrichment
         )
 
         if (isFirstTimeIdentifying || isChangingIdentifiedProfile) {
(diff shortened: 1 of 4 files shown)


datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
```diff
@@ -310,14 +323,18 @@ class CustomerIO private constructor(
             return
         }
 
-        track(EventNames.DEVICE_DELETE)
+        trackDelete(enrichmentClosure = enrichment)
+    }
+
+    private fun trackDelete(enrichmentClosure: EnrichmentClosure?) {
+        track(name = EventNames.DEVICE_DELETE, properties = emptyJsonObject, serializationStrategy = JsonAnySerializer.serializersModule.serializer(), enrichment = enrichmentClosure)
```
mrehan27, line 330: I see `trackDelete` only called once, so either this can be replaced with code directly, or should be more generic, something like following

```kotlin
private fun trackWithEmptyProperties(name:String, enrichmentClosure: EnrichmentClosure?) {
        track(name = name, properties = emptyJsonObject, serializationStrategy = JsonAnySerializer.serializersModule.serializer(), enrichment = enrichmentClosure)
}
```

Function name can surely be improved though

datapipelines/src/main/kotlin/io/customer/sdk/DataPipelineInstance.kt
```diff
@@ -127,12 +128,14 @@ abstract class DataPipelineInstance : CustomerIOInstance {
...
      * @param serializationStrategy strategy to serialize [properties]
+     * @param enrichment closure to enrich the generated event
      * @see [Learn more](https://customer.io/docs/cdp/sources/source-spec/track-spec/)
      */
     abstract fun <T> track(
         name: String,
         properties: T,
-        serializationStrategy: SerializationStrategy<T>
+        serializationStrategy: SerializationStrategy<T>,
+        enrichment: EnrichmentClosure? = null
```
mrehan27, line 138: I think this method can't be used directly from Java, so we probably don't need to add `@JvmOverloads`. However, since it's public, just wanted to double-check if you share the same thoughts, just to make sure we don't break any public APIs unknowingly.

datapipelines/src/main/kotlin/io/customer/sdk/DataPipelineInstance.kt
```diff
@@ -253,5 +256,5 @@ abstract class DataPipelineInstance : CustomerIOInstance {
     /**
      * Delete the currently registered device token
      */
-    abstract fun deleteDeviceToken()
+    abstract fun deleteDeviceToken(enrichment: EnrichmentClosure? = null)
```
mrehan27, line 259: Because this one is public, I think we should add `@JvmOverloads` here just to be safe

datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
```diff
@@ -224,9 +228,9 @@ class CustomerIO private constructor(
      * Common method to track an event with traits.
      * All other track methods should call this method to ensure consistency.
      */
-    override fun <T> track(name: String, properties: T, serializationStrategy: SerializationStrategy<T>) {
+    override fun <T> track(name: String, properties: T, serializationStrategy: SerializationStrategy<T>, enrichment: EnrichmentClosure?) {
         logger.debug("track an event with name $name and attributes $properties")
-        analytics.track(name = name, properties = properties, serializationStrategy = serializationStrategy)
+        analytics.track(name = name, properties = properties, serializationStrategy = serializationStrategy, enrichment = enrichment)
     }
```
mrehan27, line 234: I think we could have avoided exposing `EnrichmentClosure` to public for now by adding private functions like following

```kotlin
private fun <T> track(name: String, properties: T, serializationStrategy: SerializationStrategy<T>, enrichment: EnrichmentClosure?)
private fun fun deleteDeviceToken(enrichment: EnrichmentClosure?)
```

And calling them with `null` or desired values from current `public` methods.

So unless we really want them to be exposed to public or wrapper SDKs, I'm okay with this. But if not, may be we can try not changing public APIs for this?

And if we do so, rest of the suggestions can be ignored as they will no longer be needed if we don't change any public apis.

Shahroz16: i think it would be a suited solution for now.

mrehan27 (approved):
Looks good 👍🏻

PR #443, opened 2024-09-25
chore: Update JavaLayout sample app name to "Native"

mahmoud-elmorabea:
Completes [Update app name to “Android Native”](https://linear.app/customerio/issue/MBL-558/update-app-name-to-android-native)

## Context
This PR changes the Android native sample app name to be "Native". This makes it easier to differentiate the app in the device launcher from other sample apps.

I've chosen to make the name "Native" instead of "Android Native" to avoid it being truncated.

| Truncated version  | Non-truncated version |
| ------------- | ------------- |
| app-name-after-truncated | app-name-after  |

## Final result
| Before  | After |
| ------------- | ------------- |
| app-name-before | app-name-after |

### DIFF
--- a/samples/java_layout/src/main/res/values/strings.xml
+++ b/samples/java_layout/src/main/res/values/strings.xml
@@ -1,5 +1,5 @@
 <resources>
-    <string name="app_name">JavaLayout</string>
+    <string name="app_name">Native</string>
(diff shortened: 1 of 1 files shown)


Shahroz16 (approved):
🎊 🎊

PR #449, opened 2024-10-02
chore: Cleanup unused API_KEY from sample apps

mahmoud-elmorabea:
This PR removes the legacy `API_KEY` for the sample apps where it is no longer used. This change affects both `JavaLayout` and `KotlinCompose` sample apps.


Shahroz16 (changes requested):
Looks good, but I think we might have missed a few spots where `apiKey` is being added but not utilized mostly in workflows and probably gradle tasks/files etc. 

Or if its just for sample apps and you plan to make another one for those, happy to approve.

Shahroz16 (approved):
Looks good, and the failing test is due to emulator not booting to test memory leakges, so doesn't really matter here. Feel free to merge.

PR #451, opened 2024-10-07
chore: ensure Android component is initialized from services

mrehan27:
### Changes

- Renamed `registerAndroidSDKComponent` to `ensureAndroidComponent` for clearer purpose
- Called the renamed method from all entry points (push services, activities) to ensure Android component is always initialized before accessing any of its dependencies
- Removed unused, deprecated methods from `DiGraph`, as they were intended for SDK use only

### DIFF
--- a/core/src/main/kotlin/io/customer/sdk/core/di/DiGraph.kt
+++ b/core/src/main/kotlin/io/customer/sdk/core/di/DiGraph.kt
@@ -133,18 +133,4 @@ abstract class DiGraph {
         overrides.clear()
         singletons.clear()
     }
-
-    // TODO: Remove deprecated functions after all usages are removed.
-    @Deprecated("Use overrideDependency<Dependency> instead", ReplaceWith("overrideDependency<Dependency>(value)"))
-    fun <Dependency : Any> overrideDependency(dependency: Class<Dependency>, value: Dependency) {
-        overrides[dependency.name] = value as Any
-    }
-
-    @Deprecated("Use newInstance or singleton instead", ReplaceWith("newInstance()"))
-    inline fun <reified DEP : Any> override(): DEP? = overrides[dependencyKey<DEP>(identifier = null)] as? DEP
-
-    @Deprecated("Use singleton instead", ReplaceWith("singleton(newInstanceCreator)"))
-    inline fun <reified INST : Any> getSingletonInstanceCreate(newInstanceCreator: () -> INST): INST {
-        return getOrCreateSingletonInstance(identifier = null, newInstanceCreator = newInstanceCreator)
-    }
 }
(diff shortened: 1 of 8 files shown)


core/src/main/kotlin/io/customer/sdk/core/di/DiGraph.kt
```diff
@@ -133,18 +133,4 @@ abstract class DiGraph {
         overrides.clear()
         singletons.clear()
     }
-
-    // TODO: Remove deprecated functions after all usages are removed.
-    @Deprecated("Use overrideDependency<Dependency> instead", ReplaceWith("overrideDependency<Dependency>(value)"))
-    fun <Dependency : Any> overrideDependency(dependency: Class<Dependency>, value: Dependency) {
```
mrehan27, line 139: Removed these as they were deprecated and are no longer used by any DI graph in the SDK

core/src/main/kotlin/io/customer/sdk/core/di/SDKComponentExt.kt
```diff
@@ -9,8 +9,10 @@ import android.content.Context
 /**
  * Create and register an instance of AndroidSDKComponent with the provided context,
  * only if it is not already initialized.
+ * This function should be called from all entry points of the SDK to ensure that
+ * AndroidSDKComponent is initialized before accessing any of its dependencies.
  */
-fun SDKComponent.registerAndroidSDKComponent(
```
mrehan27, line 13: Since the method is intended for internal SDK use only, I renamed it to provide more clarity on its purpose and function.

core/src/main/kotlin/io/customer/sdk/core/di/SDKComponentExt.kt
```diff
@@ -9,8 +9,10 @@ import android.content.Context
...
+fun SDKComponent.ensureAndroidComponent(
```
Shahroz16, line 15:
```suggestion
fun SDKComponent.setupAndroidComponent(
```

since we are not ensuring but rather initializing or reusing ?

Shahroz16 (approved):

core/src/main/kotlin/io/customer/sdk/core/di/SDKComponentExt.kt
mrehan27: Happy renaming this, but `setup` is a bit confusing for me. If `AndroidSDKComponent` is already initialized, it simply returns existing instance. However if it's not initialized, it initializes the component and then returns new instance, ensuring `AndroidSDKComponent` is always initialized after this call and created only once. 🤔

mrehan27: Discussed internally, we want to stick with `setupAndroidComponent`

PR #453, opened 2024-10-10
chore: Improve public settings screen UX

mahmoud-elmorabea:
This PR is one of series of PRs that update the UX of the settings page for the Java_Layout sample app.

PR stack:
- Create a new internal settings screen -> #452
- **Update public settings screen -> This PR**
- Update theme to new brand -> #454

Notes:
- I will collect all changes in a base branch `update-settings-page-ux` and then merge them as one `chore` to the main branch
- Choosing `chore` not to trigger an SDK release when this is merged

The general idea of the improvements is to have our settings screen map as close as possible to the configuration of our SDK when it's being initialized. You can see how closely the fields in this public settings screen map to the config:

### The UI was tested on multiple devices
| Phone - API 35 | Phone - API 21 | Tablet |
| ------------- | ------------- | ------------- |
|   |  |  |
|   |  |  |

### Video of how it works
[public-settings-page.webm](https://github.com/user-attachments/assets/47828932-a4fe-4526-8c21-197b8e1f5e7a)


samples/java_layout/src/main/java/io/customer/android/sample/java_layout/ui/settings/SettingsActivity.java
```diff
@@ -77,30 +74,22 @@ private void parseLinkParams() {
...
-        ViewUtils.prepareForAutomatedTests(binding.flushAtTextInput, R.string.acd_flush_at_input);
-        ViewUtils.prepareForAutomatedTests(binding.trackScreensSwitch, R.string.acd_track_screens_switch);
-        ViewUtils.prepareForAutomatedTests(binding.trackDeviceAttributesSwitch, R.string.acd_track_device_attributes_switch);
-        ViewUtils.prepareForAutomatedTests(binding.debugModeSwitch, R.string.acd_debug_mode_switch);
-        ViewUtils.prepareForAutomatedTests(binding.saveButton, R.string.acd_save_settings_button);
-        ViewUtils.prepareForAutomatedTests(binding.restoreDefaultsButton, R.string.acd_restore_default_settings_button);
+        ViewUtils.prepareForAutomatedTests(binding.settingsCdpApiKeyLabel, R.string.acd_cdp_api_key_input);
+        ViewUtils.prepareForAutomatedTests(binding.settingsSiteIdKeyLabel, R.string.acd_site_id_input);
+        ViewUtils.prepareForAutomatedTests(binding.settingsSaveButton, R.string.acd_save_settings_button);
+        ViewUtils.prepareForAutomatedTests(binding.settingsRestoreDefaultsButton, R.string.acd_restore_default_settings_button);
```
mahmoud-elmorabea, line 92: @Shahroz16 @mrehan27 I couldn't find those UI tests, do I need to update something for the new fields?

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/ui/settings/SettingsActivity.java
```diff
@@ -111,52 +100,20 @@ private void setupViews() {
...
-        });
+        binding.settingsSaveButton.setOnClickListener(view -> saveSettings());
+        binding.settingsRestoreDefaultsButton.setOnClickListener(view -> updateIOWithConfig(CustomerIOSDKConfig.getDefaultConfigurations()));
+        ViewUtils.clearErrorWhenTextedEntered(binding.settingsCdpApiKeyLabel, binding.settingsCdpApiKeyLayout);
+        ViewUtils.clearErrorWhenTextedEntered(binding.settingsSiteIdKeyLabel, binding.settingsSiteIdKeyLayout);
     }
 
     private void setupObservers() {
-
-        binding.deviceTokenTextInput.setText(CustomerIO.instance().getRegisteredDeviceToken());
```
mahmoud-elmorabea, line 130: Moved to internal settings screen

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/ui/settings/SettingsActivity.java
```diff
@@ -111,52 +100,20 @@ private void setupViews() {
...
-
         settingsViewModel.getSDKConfigObservable().observe(this, config -> {
             binding.progressIndicator.hide();
             updateIOWithConfig(config);
             parseLinkParams();
         });
     }
 
-    private boolean isHostURLInvalid(String url) {
```
mahmoud-elmorabea, line 139: Moved to internal settings screen

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/ui/settings/SettingsActivity.java
scotttwittrockcio: The device token location should move to the main screen under the email.

mahmoud-elmorabea: @scotttwittrockcio maybe we can add it there for convenience but also keep it in the internal settings?

scotttwittrockcio: I don't think it's needed in both places. It's a really useful piece of information when testing push messages. It helps me quickly validate the device id matches the one in fly, and the one I'm using to send a test push to.

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/ui/settings/SettingsActivity.java
mrehan27: The tests are available in a separate Appium repository. We haven't maintained them in a while though, but having unique and consistent values across all sample apps should be sufficient for now.

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/sdk/CustomerIORepository.java
```diff
@@ -72,21 +68,23 @@ private void configureSdk(CustomerIOBuilder builder, final CustomerIOSDKConfig s
...
         }
 
-        if (sdkConfig.getFlushAt() != null) {
-            builder.flushAt(sdkConfig.getFlushAt());
-        }
-        if (sdkConfig.getFlushInterval() != null) {
-            builder.flushInterval(sdkConfig.getFlushInterval());
+        if (sdkConfig.isTestModeEnabled()) {
+            builder.flushInterval(1);
+            builder.flushAt(1);
```
mrehan27, line 73: Have we previously decided on these values? Flushing every 1 second seems too aggressive. Given that `flushAt` is already set to `1`, the SDK will likely flush after each event. So setting `flushInterval` to `1` second seems unnecessary. I suggest increasing the `flushInterval` to `5` or preferably `10` seconds to avoid the SDK repeatedly checking for events every second without reason.

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/sdk/CustomerIORepository.java
```diff
@@ -72,21 +68,23 @@ private void configureSdk(CustomerIOBuilder builder, final CustomerIOSDKConfig s
...
+        } else {
+            if (sdkConfig.getFlushAt() != null) {
+                builder.flushAt(sdkConfig.getFlushAt());
+            }
+            if (sdkConfig.getFlushInterval() != null) {
+                builder.flushInterval(sdkConfig.getFlushInterval());
```
mrehan27, line 79: Based on attached screenshots, it seems we no longer have an option to set `flushAt` and `flushInterval` anywhere in the app. If that's the case, we should remove this entirely from sample app as it's no longer needed.

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/data/model/CustomerIOSDKConfig.java
```diff
@@ -98,22 +119,29 @@ public static Map<String, String> toMap(@NonNull CustomerIOSDKConfig config) {
     private final Integer flushInterval;
     @Nullable
     private final Integer flushAt;
-    @Nullable
-    private final Boolean screenTrackingEnabled;
-    @Nullable
-    private final Boolean deviceAttributesTrackingEnabled;
-    @Nullable
-    private final Boolean debugModeEnabled;
+    private final boolean screenTrackingEnabled;
```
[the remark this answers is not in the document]
mahmoud-elmorabea: It wasn't clear to me that it was intentional so I just wanted to keep it consistent with the new values. Do we still want to keep them as nullable values? Having two getters for each field returning nullable boolean and primitive boolean was too confusing.

Do we still want to differentiate between default values and user values? Not sure if we actually use that distinction anywhere though

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/sdk/CustomerIORepository.java
mahmoud-elmorabea: You are absolutely right, that makes sense. Maybe we can leave the `flushInterval` at default and only depend on `flushAt` to send events immediately as they happen.

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/sdk/CustomerIORepository.java
mahmoud-elmorabea: That's absolutely true, will remove them!

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/data/model/CustomerIOSDKConfig.java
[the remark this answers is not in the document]
mrehan27: I agree that using object classes adds some management overhead, but primitive types will always resolve to either `true` or `false` which can't fully capture the essence of SDK default values. Although it only becomes relevant if SDK's default value changes. I can't think of any other major use cases where it would matter, so as long as saving and restoring values features are working fine, I think using primitives is fine and won't really make a difference in most cases.

mrehan27 (commented):
Everything looks good, except for what Scott suggested in his comment. Looks like device token is missing from public settings screen. Is there a separate PR for that change?

mahmoud-elmorabea:
@mrehan27 Yes, I will have follow up PRs for changes to login page and other parts

mrehan27 (approved):

PR #456, opened 2024-10-24
feat: support for large in-app messages

Shahroz16:
closes: [MBL-547: Bug: Large in-app messages not getting delivered](https://linear.app/customerio/issue/MBL-547/bug-large-in-app-messages-not-getting-delivered)

changes:
- Make JS post request for sending is message payload rather than query params
- Persistent messages are now added to `shownMessageQueueIds` only after dismissal, aligning with how we handle in-app message viewed logs in our middleware today


messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
```diff
@@ -77,83 +75,75 @@ internal class EngineWebView @JvmOverloads constructor(
...
+            }
 
-                it.webViewClient = object : WebViewClient() {
-                    override fun onPageFinished(view: WebView, url: String?) {
-                        view.loadUrl("javascript:window.parent.postMessage = function(message) {window.${EngineWebViewInterface.JAVASCRIPT_INTERFACE_NAME}.postMessage(JSON.stringify(message))}")
+            it.webViewClient = object : WebViewClient() {
+                override fun onPageFinished(view: WebView, url: String?) {
+                    val script = """
+                    window.postMessage($jsonString, '*');
+                """.trim()
```
mrehan27, line 101: I understand this is passing information to Web, but if you have any additional context about the script that could help future readers, it would be great to add that as a comment here.

Also, previously the URL sometimes helped us debug which message we were trying to load among other things. With that now being removed from URL, do you think there's anything additional we should log here to help future debugging, or do you feel the existing improved logs will suffice in most cases?

stephen-pope-customer-io (approved):
This is great!

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
Shahroz16: open to ideas, i think the state logs would let us know what message is about to be displayed?

messaginginapp/src/main/java/io/customer/messaginginapp/gist/GistEnvironment.kt
```diff
@@ -10,7 +10,7 @@ enum class GistEnvironment : GistEnvironmentEndpoints {
     DEV {
         override fun getGistQueueApiUrl() = "https://gist-queue-consumer-api.cloud.dev.gist.build"
         override fun getEngineApiUrl() = "https://engine.api.dev.gist.build"
-        override fun getGistRendererUrl() = "https://renderer.gist.build/2.0"
+        override fun getGistRendererUrl() = "https://renderer.gist.build/beta"
```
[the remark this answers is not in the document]
Shahroz16: It's updated, was waiting for stable version to be released :)

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
```diff
@@ -77,83 +75,76 @@ internal class EngineWebView @JvmOverloads constructor(
...
+            it.setBackgroundColor(Color.TRANSPARENT)
+
+            findViewTreeLifecycleOwner()?.lifecycle?.addObserver(this) ?: run {
+                logger.error("Lifecycle owner not found, attaching interface to WebView manually")
+                engineWebViewInterface.attach(webView = it)
+            }
 
-                it.webViewClient = object : WebViewClient() {
-                    override fun onPageFinished(view: WebView, url: String?) {
-                        view.loadUrl("javascript:window.parent.postMessage = function(message) {window.${EngineWebViewInterface.JAVASCRIPT_INTERFACE_NAME}.postMessage(JSON.stringify(message))}")
```
BernardGatt, line 102: I think this is what's breaking the bridge

PR #457, opened 2024-10-25
chore: Create workflow to build sample app with latest SDK release

mahmoud-elmorabea:
### Motivation
At the moment we don't have any sample apps that is built by pulling our SDK releases from Maven Central. That is the closest form to what our customers experience. So we want to start building those sample apps to:
- Make sure that the sample apps build fine with the released version on Maven Central
- Allow our internal Tech support, Customer success and other stakeholders to test and reproduce customers experience with our SDK

### Proposal
- New testing group on Firebase has been created only for sample apps with released SDK versions called `public`
- The new workflow can be triggered manually or with Github API
- A follow up to this change, we will trigger this workflow after the `Deploy SDK` workflow runs successfully


.github/workflows/build-sample-app-for-sdk-release.yml
```diff
@@ -0,0 +1,49 @@
...
+          bundler-cache: true
+
+      - name: Setup local.properties file for sample app
+        env:
+          SDK_VERSION: ${{ steps.latest-sdk-version-step.outputs.LATEST_TAG }}
+        run: |
+          touch "samples/local.properties"
+          echo "cdpApiKey=${{ secrets.CUSTOMERIO_JAVA_WORKSPACE_CDP_API_KEY }}" >> "samples/local.properties"
+          echo "siteId=${{ secrets.CUSTOMERIO_JAVA_WORKSPACE_SITE_ID }}" >> "samples/local.properties"
+          echo "sdkVersion=${{ env.SDK_VERSION }}" >> "samples/local.properties"
```
mahmoud-elmorabea, line 36: The sample app build scripts already check for `sdkVersion` [here](https://github.com/customerio/customerio-android/blob/main/samples/sample-app.gradle#L67) and this will cause the public release to be used instead of local project

samples/fastlane/Fastfile
```diff
@@ -57,5 +57,46 @@ platform :android do
       groups: test_groups,
       release_notes: build_notes
     )
-  end 
+  end
+
+  lane :build_sample_app_for_sdk_release do |values|
```
mahmoud-elmorabea, line 62: I opted for creating a new lane since the existing one is already fairly complicated. Also if we decide to move to a single app, we can just deprecate the older lane.

.github/workflows/build-sample-app-for-sdk-release.yml
Shahroz16, line 1: I think this can be easily integrated with the existing `build-sample-apps` workflow. 
- Add a workflow_dispatch with optional SDK version input
- You can add a new job

```
get-sdk-version:
  if: ${{ github.event_name == 'workflow_dispatch' }}
  outputs:
   sdk_version: ${{ steps.get-version.outputs.version }}
```
- Build configuration can be dynamic something like

```
- name: Deploy development build via Fastlane
  uses: maierj/fastlane-action@v3.1.0
  with:
    lane: ${{ github.event_name == 'workflow_dispatch' && 'android build_sample_app_for_sdk_release' || 'android build' }}
    options: ${{ github.event_name == 'workflow_dispatch' && format('{{"sdk_version":"{0}"}}', needs.get-sdk-version.outputs.sdk_version) || '' }}
```
- Similarly, `local.properties` can be update via
```
 if [ "${{ github.event_name }}" == "workflow_dispatch" ]; then
      echo "sdkVersion=${{ needs.get-sdk-version.outputs.sdk_version }}" >> "samples/local.properties"
    fi
```

samples/fastlane/Fastfile
Shahroz16: Hmm, I'm unsure about this because again, both lanes have 70% of the same code, apart from some hardcoded values in this lane. I don't mind separate lane, but then could we probably move the shared code in methods that both lanes can utilize? 

Also, could you point out the complicated parts in the other lane so we can see how we can make them less complicated in future? 

for example, something like this, ensure, that when you are testing a feature branch we don't show the version of the SDK in user agent, but rather the branch its being tested on. 

```
    UI.important("Updating the SDK's source code version to non-production version. This allows the sample apps to show the SDK version at runtime for app user to better understand the version of the SDK they are running.")
    sh("../../../scripts/update-version.sh \"#{new_app_version}.#{new_build_number}\"")
```

Shahroz16 (changes requested):
Thanks for the implementation! While both workflows/lanes serve different purposes, I notice there's significant code duplication that could lead to several issues:
such as maintenance burden and being error prone. 

I have added couple of suggestions, let me know if they are helpful. Even if we don't move towards a more configurable approach we should at least remove duplication and use shared resources.

samples/fastlane/Fastfile
```diff
@@ -57,5 +55,39 @@ platform :android do
...
+     # Apps use `build.gradle` or `build.kts`. This wildcard lookup is used to get the absolute path to one of these files present
+     path_to_app_gradle_file = File.expand_path(Dir.glob("../build.*").first, Dir.pwd)
+
+     new_app_version = values[:sdk_version]
+
+    build_sample_app(
+        sample_app_name: 'java_layout',
+        app_version: new_app_version,
+        firebase_test_groups: 'public',
```
scotttwittrockcio, line 88: Can we publish to two groups here? It would be good to publish to 'Stable builds' in addition to 'Public' until we get all SDK's moved over to the 'public' group. Thoughts? If the effort is much more than adding another group to an array, let's chat as it might not be worth it.

mahmoud-elmorabea: Sure thing, added it!

mahmoud-elmorabea:
@Shahroz16 I've extract common code in both Fastlane and the workflow and it's now shared.

.github/workflows/build-sample-app-for-sdk-release.yml
```diff
@@ -0,0 +1,17 @@
...
+      use_latest_sdk_version: true
+    secrets:
+      CUSTOMERIO_JAVA_WORKSPACE_CDP_API_KEY: ${{ secrets.CUSTOMERIO_JAVA_WORKSPACE_CDP_API_KEY }}
+      CUSTOMERIO_JAVA_WORKSPACE_SITE_ID: ${{ secrets.CUSTOMERIO_JAVA_WORKSPACE_SITE_ID }}
+      CUSTOMERIO_KOTLIN_WORKSPACE_CDP_API_KEY: ${{ secrets.CUSTOMERIO_KOTLIN_WORKSPACE_CDP_API_KEY }}
+      CUSTOMERIO_KOTLIN_WORKSPACE_SITE_ID: ${{ secrets.CUSTOMERIO_KOTLIN_WORKSPACE_SITE_ID }}
+      ANDROID_SIGNING_ALIAS: ${{ secrets.ANDROID_SIGNING_ALIAS }}
+      ANDROID_SIGNING_KEY_PASSWORD: ${{ secrets.ANDROID_SIGNING_KEY_PASSWORD }}
+      ANDROID_SIGNING_STORE_PASSWORD: ${{ secrets.ANDROID_SIGNING_STORE_PASSWORD }}
+      FIREBASE_APP_DISTRIBUTION_SERVICE_ACCOUNT_CREDS_B64: ${{ secrets.FIREBASE_APP_DISTRIBUTION_SERVICE_ACCOUNT_CREDS_B64 }}
```
Shahroz16, line 17:
```suggestion
    uses: ./.github/workflows/reusable_build_sample_apps.yml
    secrets: inherit
```

we can probably simplify it by [inherting](https://docs.github.com/en/actions/sharing-automations/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow) the secrets?

.github/workflows/reusable_build_sample_apps.yml
```diff
@@ -0,0 +1,113 @@
+name: Reusable build sample apps workflow
+
+on:
+  workflow_call:
```
Shahroz16, line 4: we can add 
```
secrets:
      inherit: true
 ```

File: .github/workflows/build-sample-apps.yml [hunk 6]
```diff
@@ -20,105 +20,43 @@ jobs:
...
-        # the variables APP_BUILD_NUMBER, APP_VERSION_STRING are created when fastlane runs "build". 
-        body: |
-          * ${{ matrix.sample-app }}: `${{ env.APP_VERSION_STRING }} (${{ env.APP_BUILD_NUMBER }})`
-        edit-mode: append # append new line to the existing PR comment to build a list of all sample app builds. 
+    if: ${{ always() }} # do not skip running this step if update-pr-comment does not run
+    needs: [ update-pr-comment ] # wait for PR comment to be created saying new builds are being made.
+    uses: ./.github/workflows/reusable_build_sample_apps.yml
+    with:
+      use_latest_sdk_version: false
+    secrets:
```
Shahroz16, line 53: similar to other workflow comment

Shahroz16 (approved):
Looks good just one suggestion regarding secrets simplified usage

.github/workflows/build-sample-app-for-sdk-release.yml
mahmoud-elmorabea: Great suggestion! Will do.

PR #461, opened 2024-11-04
chore: move to v2 for user queue request and user agent in headers

Shahroz16:
closes: https://linear.app/customerio/issue/MBL-495/move-to-v2-apis-of-gist-and-add-sdk-reporting

changes:
- Upgrade user queue endpoint path to v2
- Added SDK version and platform in request headers

### DIFF
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/listeners/Queue.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/listeners/Queue.kt
@@ -26,7 +26,7 @@ import retrofit2.http.POST
 import retrofit2.http.Path
 
 interface GistQueueService {
-    @POST("/api/v1/users")
+    @POST("/api/v2/users")
     suspend fun fetchMessagesForUser(@Body body: Any = Object()): Response<List<Message>>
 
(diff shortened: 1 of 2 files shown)


messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/listeners/Queue.kt
```diff
@@ -69,6 +69,8 @@ class Queue : GistQueue {
                 val networkRequest = originalRequest.newBuilder()
                     .addHeader(NetworkUtilities.CIO_SITE_ID_HEADER, state.siteId)
                     .addHeader(NetworkUtilities.CIO_DATACENTER_HEADER, state.dataCenter)
+                    .addHeader(NetworkUtilities.CIO_CLIENT_PLATFORM, SDKComponent.android().client.source.lowercase())
+                    .addHeader(NetworkUtilities.CIO_CLIENT_VERSION, SDKComponent.android().client.sdkVersion)
```
mrehan27, line 73: Did you verify if React Native Data Pipelines support is sending expected source and version correctly? If yes, then we're good with the change.

mrehan27 (approved):
Just one question, rest looks good.

PR #469, opened 2024-11-28
fix: Fixes issue with legacy renderer bridge

BernardGatt:
Reintroduces bridge function to pass messages between SDK and Renderer

Fixes issue: https://linear.app/customerio/issue/MBL-721/mobile-in-app-crashes-after-updating-from-version-3x-to-44-of-android

### DIFF
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
@@ -99,6 +99,7 @@ internal class EngineWebView @JvmOverloads constructor(
                     // post message to webview with the configuration data so that the message can be rendered
                     val script = """
                     window.postMessage($jsonString, '*');
+                    window.parent.postMessage = function(message) {window.${EngineWebViewInterface.JAVASCRIPT_INTERFACE_NAME}.postMessage(JSON.stringify(message))}
                 """.trim()
                     view.evaluateJavascript(script) { result ->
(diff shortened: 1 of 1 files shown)


messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
```diff
@@ -99,6 +99,7 @@ internal class EngineWebView @JvmOverloads constructor(
                     // post message to webview with the configuration data so that the message can be rendered
                     val script = """
                     window.postMessage($jsonString, '*');
+                    window.parent.postMessage = function(message) {window.${EngineWebViewInterface.JAVASCRIPT_INTERFACE_NAME}.postMessage(JSON.stringify(message))}
```
Shahroz16, line 102:
```suggestion
                    val script = """
    // Post the JSON message to the current frame's listeners
    // Ensures internal JavaScript communication via window.addEventListener('message') remains functional
    window.postMessage($jsonString, '*');
    
    // Override window.parent.postMessage to route messages to the native Android interface
    // This is necessary only for legacy message because WebView can only attach one native interface 
    // (via addJavascriptInterface),
    // and we have already added it as ${EngineWebViewInterface.JAVASCRIPT_INTERFACE_NAME}.
    window.parent.postMessage = function(message) {
        window.${EngineWebViewInterface.JAVASCRIPT_INTERFACE_NAME}.postMessage($jsonString);
    }
""".trim()
```

Shahroz16 (approved):

PR #474, opened 2024-12-13
feat: add option to disable screen view usage

mrehan27:
part of: [MBL-755](https://linear.app/customerio/issue/MBL-755/add-config-to-android-sdk)

### Changes

- Added `screenViewUse` enum option to allow disable sending screen view events to server and use them locally only
- Added `ScreenFilterPlugin` to filter events based on the `screenViewUse` configuration
- Added tests to ensure functionality

### `ScreenView` Options

- `Analytics` -> Sends all screen events to the server (same as before) - Default behavior
- `InApp` -> Retains screen events locally for in-app use only

### Sample Usage

```kotlin
CustomerIOBuilder(...)
    .screenViewUse(ScreenView.InApp)
    .build()
```

### DIFF
--- a/datapipelines/src/main/kotlin/io/customer/datapipelines/config/ScreenView.kt
+++ b/datapipelines/src/main/kotlin/io/customer/datapipelines/config/ScreenView.kt
@@ -5,7 +5,7 @@ package io.customer.datapipelines.config
  */
 enum class ScreenView {
     /**
-     * Screen view events are sent to our back end servers for analytics purposes.
+     * Screen view events are sent to destinations for analytics purposes.
      */
     Analytics,
 
--- a/datapipelines/src/main/kotlin/io/customer/datapipelines/config/ScreenView.kt
+++ b/datapipelines/src/main/kotlin/io/customer/datapipelines/config/ScreenView.kt
@@ -3,26 +3,36 @@ package io.customer.datapipelines.config
 /**
  * Enum class to define how CustomerIO SDK should handle screen view events.
  */
-enum class ScreenView {
+sealed class ScreenView(val name: String) {
     /**
      * Screen view events are sent to destinations for analytics purposes.
+     * They are also used to display in-app messages based on page rules.
      */
-    Analytics,
+    object All : ScreenView(name = "all")
 
     /**
-     * Screen view events are kept on device only. They are used to display in-app messages based on page rules. Events are not sent to our back end servers.
+     * Screen view events are kept on device only. They are used to display in-app messages based on
+     * page rules. Events are not sent to our back end servers.
      */
-    InApp;
+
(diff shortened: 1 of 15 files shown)


Shahroz16 (approved):

datapipelines/src/main/kotlin/io/customer/datapipelines/config/ScreenView.kt
```diff
@@ -0,0 +1,28 @@
+package io.customer.datapipelines.config
+
+/**
+ * Enum class to define how CustomerIO SDK should handle screen view events.
+ */
+enum class ScreenView {
```
mahmoud-elmorabea, line 6: Does this mean it's a mutually exclusive choice? I can either choose analytics or in-app? But what if a customer needs both?

datapipelines/src/main/kotlin/io/customer/datapipelines/plugins/ScreenFilterPlugin.kt
```diff
@@ -0,0 +1,27 @@
...
+import com.segment.analytics.kotlin.core.ScreenEvent
+import com.segment.analytics.kotlin.core.platform.EventPlugin
+import com.segment.analytics.kotlin.core.platform.Plugin
+import io.customer.datapipelines.config.ScreenView
+
+/**
+ * Plugin to filter screen events based on the configuration provided by customer app.
+ * This plugin is used to filter out screen events that should not be processed further.
+ */
+class ScreenFilterPlugin(private val screenViewUse: ScreenView) : EventPlugin {
```
mahmoud-elmorabea, line 14: Is this meant for only internal use or are we intentionally exposing it as public API? I don't think we should commit to adding this style of filtering till we discuss this widely and how it can affect the rest of the API (filtering events and other things) and how it can play with other SDKs as well.

datapipelines/src/main/kotlin/io/customer/datapipelines/config/ScreenView.kt
mahmoud-elmorabea, line 6: Curious why we are opting for making this an enum instead of:
- Boolean to allow screen analytics
- If in app is enabled, we always use screen views for in-app

mahmoud-elmorabea (commented):
I think we should minimize public API changes and limit it to work around the issue we have at the moment and leave larger API changes to a later point where we re-think having parts of our SDK enabled using a comprehensive mechanism.

datapipelines/src/main/kotlin/io/customer/datapipelines/plugins/ScreenFilterPlugin.kt
mrehan27: Updated to internal

PR #476, opened 2024-12-15
chore: handle the case of empty identifier

Shahroz16:
closes: [MBL-738: Android - Handle device registration when switching from profile with empty string/null identifier.](https://linear.app/customerio/issue/MBL-738/android-handle-device-registration-when-switching-from-profile-with)

handles the case, where if profile attributes are added before the identifying the profile, `userId` was set to empty string.

### DIFF
--- a/datapipelines/src/test/java/io/customer/datapipelines/DataPipelinesInteractionTests.kt
+++ b/datapipelines/src/test/java/io/customer/datapipelines/DataPipelinesInteractionTests.kt
@@ -73,6 +73,29 @@ class DataPipelinesInteractionTests : JUnitTest() {
     }
 
     @Test
+    fun identify_givenProfileAttributesAndNoIdentifier_expectSetNewProfileWithoutAttributes() {
+        val givenIdentifier = String.random
+
+        analytics.userId().shouldBeNull()
+        every { globalPreferenceStore.getDeviceToken() } returns String.random
+
+        sdkInstance.profileAttributes = mapOf("first_name" to "Dana", "ageInYears" to 30)
+        analytics.userId() shouldBe ""
+        sdkInstance.identify(givenIdentifier)
+
+        analytics.userId().shouldBeEqualTo(givenIdentifier)
+        analytics.traits().shouldBeEqualTo(emptyJsonObject)
+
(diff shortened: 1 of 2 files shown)


mrehan27 (approved):

datapipelines/src/test/java/io/customer/datapipelines/DataPipelinesInteractionTests.kt
```diff
@@ -72,6 +72,29 @@ class DataPipelinesInteractionTests : JUnitTest() {
...
 
+    @Test
+    fun identify_givenProfileAttributesAndNoIdentifier_expectSetNewProfileWithoutAttributes() {
+        val givenIdentifier = String.random
+
+        analytics.userId().shouldBeNull()
+        every { globalPreferenceStore.getDeviceToken() } returns String.random
+
+        sdkInstance.profileAttributes = mapOf("first_name" to "Dana", "ageInYears" to 30)
+        analytics.userId() shouldBe ""
```
mahmoud-elmorabea, line 83: I might be missing something but isn't setting the ID to empty string is what we are trying to fix?

Shahroz16: We are handling the case where it would set the ID to empty, we can't stop from setting the id to empty. When you set profile attributes before identifying, it sets identifier to empty string in the Segment SDK. 

So this test verifies that, when we set profile attributes before, we assert identifier is being set to an empty string and then when we identify it later on, it doesn't effect the flow because our code is handling the empty string along with `null`.

mahmoud-elmorabea: So basically we want `currentlyIdentifiedProfile` to be `null` instead of empty string and that fixes the issue?

Shahroz16: Yes, because that's the default case. When there is no user identified, identifier should be null. So now if it's empty (edge case) we also consider it as null.

mahmoud-elmorabea (approved):

PR #478, opened 2024-12-26
chore: Create a plugin that tracks application lifecycle foreground event

mahmoud-elmorabea:
Completes [MBL-526
](https://linear.app/customerio/issue/MBL-526/add-application-foregrounded-lifecycle-event-in-android-sdk)

This PR adds the missing application lifecycle event `Application Foregrounded`.

### Why are we not using `Application.ActivityLifecycleCallbacks`?
Our wrapper SDKs are initialized later in the app lifecycle, so they miss the initial `onCreate()` -> `onStart()` -> `onResume()` for the first activity in the application. That makes `Acitvity` lifecycle callbacks not usable for us.

### Some notes about the implementation
- Our SDKs already pull `AndroidX` lifecycle dependency `2.6.1`, so that's why I choose to include it with that specific version
- In ReactNative the `Plugin#setup()` method is not called on the main thread which causes the `lifecycle#addObserver()` call to not work, that's why I'm explicitly running it on UI thread

##### Flutter

https://github.com/user-attachments/assets/ce2cdfd1-1ef3-412c-8fa4-597f14dbaa82

##### RN

https://github.com/user-attachments/assets/10f801cd-abf5-415c-8fa5-1b9181d720f5

Here are some of the events of Fly for the what's going on in the video
Screenshot 2024-12-26 at 2 54 07 PM

### DIFF
--- a/datapipelines/src/test/java/io/customer/datapipelines/plugins/AutomaticApplicationLifecycleTrackingPluginTest.kt
+++ b/datapipelines/src/test/java/io/customer/datapipelines/plugins/AutomaticApplicationLifecycleTrackingPluginTest.kt
@@ -1,0 +1,92 @@
+package io.customer.datapipelines.plugins
+
+import android.app.Activity
+import com.segment.analytics.kotlin.core.Analytics
+import io.customer.datapipelines.testutils.core.JUnitTest
+import io.mockk.every
+import io.mockk.just
+import io.mockk.mockk
+import io.mockk.runs
+import io.mockk.verify
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+
+class AutomaticApplicationLifecycleTrackingPluginTest : JUnitTest() {
+
+    private val mockAnalytics = mockk<Analytics>()
+
+    private val subject = AutomaticApplicationLifecycleTrackingPlugin()
+
+    @BeforeEach
+    fun beforeEach() {
+        every { mockAnalytics.track(any()) } just runs
+        subject.setup(mockAnalytics)
+    }
+
+    @Test
+    fun `GIVEN app is closed WHEN app is opened THEN ApplicationForegrounded is tracked`() {
+        simulateAppOpened()
+
+        verify(exactly = 1) { mockAnalytics.track("Application Foregrounded") }
+    }
+
+    @Test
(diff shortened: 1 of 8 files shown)


mrehan27 (commented):
The only concern I have is regarding the behavior in wrappers and scenarios like push clicks or in-app message displays. Started slack thread to discuss this further. Rest look good.

datapipelines/src/test/java/io/customer/datapipelines/sdk/AndroidLifecyclePluginTests.kt
```diff
@@ -76,8 +76,9 @@ class AndroidLifecyclePluginTests : IntegrationTest() {
         val tracks = mutableListOf<TrackEvent>()
         verify { mockPlugin.track(capture(tracks)) }
         // 1. Application Installed
-        // 2. Application Opened
-        assertEquals(2, tracks.size)
+        // 2. Application Foregrounded
+        // 3. Application Opened
+        assertEquals(3, tracks.size)
```
Shahroz16, line 81: would we always get `opened` and `foregrounded`? because looking at the implementation they both are following the same strategy?

datapipelines/src/test/java/io/customer/datapipelines/plugins/AutomaticApplicationLifecycleTrackingPluginTest.kt
```diff
@@ -0,0 +1,95 @@
...
+        val mockLifecycle = mockk<Lifecycle>()
+        every { mockProcessLifecycleOwner.lifecycle } returns mockLifecycle
+        every { mockLifecycle.addObserver(capture(lifecycleObserverCaptor)) } just runs
+
+        every { mockAnalytics.track(any()) } just runs
+        subject.setup(mockAnalytics)
+    }
+
+    @Test
+    fun `WHEN application lifecycle moves to CREATED THEN Application Foregrounded is NOT tracked`() {
```
Shahroz16, line 43: I like this format too 😢 but can we update it to follow the current naming format with `_` to keep consistency? 

We can collectively move to it, if we decide later on.

Shahroz16 (commented):
Looks good, just one change

Shahroz16 (approved):

PR #479, opened 2024-12-26
chore: Fix in-app messages overlay color ignored from message content

mahmoud-elmorabea:
## Motivation
When sending an in-app message to the mobile apps, the specified `Overlay Color` in the in-app message editor is not reflected on the message shown on the app.

## Root cause
The value `overlayColor` sent to the mobile SDK from the server along with the in-app message payload is ignored and not used to set the in-app message overlay color.

Instead a hard-coded color of `#33000000` is used for all message.

## Testing the fix

Here's a version of the fix with slow animations (10X the normal duration) to make it easy to verify the fix is working

https://github.com/user-attachments/assets/5d9a7ef0-9da9-4a2f-a749-0a3ae132b0ac

Here's a version with the real animation speed

https://github.com/user-attachments/assets/1cb6d556-466a-42a3-9251-1bd6f3e02105


mahmoud-elmorabea:
If you notice that there is no closing animation when dismissing the modal, you are right! The fix for that is in the following PR #480

messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/model/Message.kt
```diff
@@ -13,7 +13,8 @@ data class GistProperties(
     val elementId: String?,
     val campaignId: String?,
     val position: MessagePosition,
-    val persistent: Boolean
+    val persistent: Boolean,
+    val overlayColor: String?
```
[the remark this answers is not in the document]
Shahroz16: also would be consistent to `MessagePosition`

messaginginapp/src/main/java/io/customer/messaginginapp/gist/utilities/ModalAnimationUtil.kt
```diff
@@ -0,0 +1,66 @@
...
+    const val FALLBACK_COLOR_STRING = "#33000000"
+
+    private const val TRANSLATION_ANIMATION_DURATION = 150L
+    private const val ALPHA_ANIMATION_DURATION = 150L
+    private const val COLOR_ANIMATION_DURATION = 100L
+
+    private val logger = SDKComponent.logger
+
+    fun createAnimationSetInFromTop(target: View, overlayEndColor: String): AnimatorSet {
+        return createEnterAnimation(target, overlayEndColor, -100f)
```
Shahroz16, line 21:
```suggestion
        target.alpha = 0f
        target.translationY = -100f
        return createEnterAnimation(target, overlayEndColor, -100f)
```

We can prevent frame drop here by setting initial state

messaginginapp/src/main/java/io/customer/messaginginapp/gist/utilities/ModalAnimationUtil.kt
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    fun createAnimationSetInFromBottom(target: View, overlayEndColor: String): AnimatorSet {
+        return createEnterAnimation(target, overlayEndColor, 100f)
```
Shahroz16, line 25:
```suggestion
        target.alpha = 0f
        target.translationY = 100f
        return createEnterAnimation(target, overlayEndColor, 100f)
```

similar to above

messaginginapp/src/main/java/io/customer/messaginginapp/gist/utilities/ModalAnimationUtil.kt
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    private fun createEnterAnimation(
+        target: View,
+        overlayEndColor: String,
+        translationYStart: Float
+    ): AnimatorSet {
+        val translationYAnimator = ObjectAnimator.ofFloat(target, "translationY", translationYStart, 0f).apply {
```
Shahroz16, line 33:
```suggestion
        val translationYAnimator = ObjectAnimator.ofFloat(
            target, 
            View.TRANSLATION_Y, 
            0f, 
            translationYEnd
        ).apply {
```

messaginginapp/src/main/java/io/customer/messaginginapp/gist/utilities/ModalAnimationUtil.kt
```diff
@@ -0,0 +1,66 @@
...
+            duration = TRANSLATION_ANIMATION_DURATION
+        }
+        val alphaAnimator = ObjectAnimator.ofFloat(target, "alpha", 0f, 1f).apply {
```
Shahroz16, line 36:
```suggestion
      val alphaAnimator = ObjectAnimator.ofFloat(
            target, 
            View.ALPHA, 
            1f, 
            0f
        )
```

Shahroz16: lets use View property constants instead of string literals to avoid reflection

messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/model/Message.kt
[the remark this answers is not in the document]
mahmoud-elmorabea: I kinda think color default values belong in the presentation logic not at the model so I prefer to have it here. Also defining this as non-null might be a bit misleading. But I see the point about consistency with that other field as well.

I can move it to the model if you feel strongly about it.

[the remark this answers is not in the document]
Shahroz16: > Also defining this as non-null might be a bit misleading

This is interesting, from the code it feels like we do indeed set an overlay with default color when user has not provided its override? if thats the case, then technically its not nullable? 

Either way, I don't feel strongly about it but I do think, the two things that make me think it make be more easier to have this at data layer other than the presentation
- We won't have to check for null values, every time. Because we are first going to get the message, check for nullable properties, and then check again for nullable overlay. 
- Consistency with other fields. 
 
 So a non-blocking opinion :)

Shahroz16 (approved):

messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/model/Message.kt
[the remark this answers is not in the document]
mahmoud-elmorabea: I see your point, I just think it's confusing for someone looking at the activity, if we set the default at data layer, they would think the value is coming from server while in fact it might not be.

They would have to know that this value has a hard-coded fallback at the parsing layer.

I vote we keep the default at the activity for clarity.

PR #480, opened 2024-12-26
chore: Fix closing animation for in-app message modal

mahmoud-elmorabea:
This PR applied the same fix as #479 but for the closing animation of the in-app message modal activity.

## Motivation
As I was applying the fix in #479, I noticed a two small issues with the closing animation for the in-app message modal activity:
- The closing animation is also using a hard-coded color and not the color from the message payload
- The logic to animate closing the modal was not being called at all

## The fix
- Using the same color of the overlay background as start for the closing animation
- Making sure we are calling the overridden `finish()` method that has the animation logic instead of `super.finish()`

## Testing
Here's a version of the fix with slow animations (10X the normal duration) to make it easy to verify the fix is working

https://github.com/user-attachments/assets/49fa90cc-bcd3-4078-a92a-70db3c595f17

Here's a version with the real animation speed

https://github.com/user-attachments/assets/4ff5654c-5457-42f7-be38-4f5ee37948dc


messaginginapp/src/main/java/io/customer/messaginginapp/gist/utilities/ModalAnimationUtil.kt
```diff
@@ -54,6 +63,32 @@ internal object ModalAnimationUtil {
...
+        val backgroundColor = extractBackgroundColor(target)
+        val backgroundColorAnimator = ObjectAnimator.ofArgb(target, "backgroundColor", parseColorSafely(backgroundColor, fallbackColor), Color.TRANSPARENT).apply {
+            duration = COLOR_ANIMATION_DURATION
+            startDelay = 0
+        }
+        val colorSet = AnimatorSet().apply {
+            play(backgroundColorAnimator)
+        }
+
+        val translationYAnimator = ObjectAnimator.ofFloat(target, "translationY", 0f, translationYEnd).apply {
```
Shahroz16, line 77: same as previous PR, `View.TRANSLATION_Y` instead of `translationY`

messaginginapp/src/main/java/io/customer/messaginginapp/gist/utilities/ModalAnimationUtil.kt
```diff
@@ -54,6 +63,32 @@ internal object ModalAnimationUtil {
...
+            duration = TRANSLATION_ANIMATION_DURATION
+        }
+        val alphaAnimator = ObjectAnimator.ofFloat(target, "alpha", 1f, 0f).apply {
```
Shahroz16, line 80: `View.ALPHA`

Shahroz16 (commented):
Great find !!

Shahroz16 (approved):

PR #483, opened 2025-01-06
chore: fix instrumentation test re-runs

Shahroz16:
Currently when the PR is created the instrumentation tests would run perfectly, but on any changes to PR causing re-run of action would result in the emulator not booting and resulting in instrumentation tests not being run. 

This PR fixes that issue.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -68,7 +68,7 @@ jobs:
             ~/.android/adb*
           key: avd-${{ matrix.api-level }}-v2
 
-      # Create AVD and generate snapshot for caching
+      # Create AVD and generate snapshot for caching, let check itt
       - name: Create AVD and generate snapshot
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -71,33 +68,36 @@ jobs:
...
-      - name: create AVD and generate snapshot for caching
+      # Create AVD and generate snapshot for caching
+      - name: Create AVD and generate snapshot
         if: steps.avd-cache.outputs.cache-hit != 'true'
         uses: reactivecircus/android-emulator-runner@v2
         with:
           api-level: ${{ matrix.api-level }}
           arch: x86_64
+          target: google_apis
+          ram-size: 4096M
```
Shahroz16, line 79: addition of these seems to have done the trick

mrehan27 (approved):
Looking at number of actions for this PR, looks like it is working good 👍🏻

PR #485, opened 2025-01-07
fix: Fix in-app messages overlay background color being ignored from message payload

mahmoud-elmorabea:
Reintroduces #479 with a parsing fix so that colors are shown as expected for message overlay background color.


Shahroz16 (approved):
Focused on `parseColor` which looks good, other I am assuming is just re-revert so should be good.

PR #493, opened 2025-02-10
fix: improve push delivery for Android 14 and above

Shahroz16:
closes: https://linear.app/customerio/issue/MBL-856/report-push-notification-metric-issues-instantly-dont-add-to-queue

changes:
- Save configuration and share via core
- Add HTTP layer in push utilizing saved config
- PushDeliveryTracker, utilizing HTTP client 
- Move classes/dependencies from Datapipeline to Core for other modules to be used

### DIFF
--- a/core/src/main/kotlin/io/customer/sdk/data/store/GlobalPreferenceStore.kt
+++ b/core/src/main/kotlin/io/customer/sdk/data/store/GlobalPreferenceStore.kt
@@ -2,6 +2,8 @@ package io.customer.sdk.data.store
 
 import android.content.Context
 import androidx.core.content.edit
+import io.customer.sdk.data.model.Settings
+import kotlinx.serialization.json.Json
 
 /**
  * Store for global preferences that are not tied to a specific api key, user
@@ -9,7 +11,9 @@ import androidx.core.content.edit
  */
 interface GlobalPreferenceStore {
     fun saveDeviceToken(token: String)
+    fun saveSettings(value: String)
     fun getDeviceToken(): String?
+    fun getSettings(): Settings?
     fun removeDeviceToken()
     fun clear(key: String)
     fun clearAll()
@@ -27,13 +31,27 @@ internal class GlobalPreferenceStoreImpl(
         putString(KEY_DEVICE_TOKEN, token)
     }
 
+    override fun saveSettings(value: String) = prefs.edit {
+        putString(KEY_CONFIG_SETTINGS, value)
+    }
+
     override fun getDeviceToken(): String? = prefs.read {
         getString(KEY_DEVICE_TOKEN, null)
+    }
+
+    override fun getSettings(): Settings? = prefs.read {
+        runCatching {
+            Json.decodeFromString(
+                Settings.serializer(),
+                getString(KEY_CONFIG_SETTINGS, null) ?: return null
+            )
+        }.getOrNull()
     }
 
     override fun removeDeviceToken() = clear(KEY_DEVICE_TOKEN)
 
     companion object {
         private const val KEY_DEVICE_TOKEN = "device_token"
+        private const val KEY_CONFIG_SETTINGS = "config_settings"
     }
 }
--- a/core/src/main/kotlin/io/customer/sdk/data/store/GlobalPreferenceStore.kt
+++ b/core/src/main/kotlin/io/customer/sdk/data/store/GlobalPreferenceStore.kt
@@ -11,7 +11,7 @@ import kotlinx.serialization.json.Json
  */
 interface GlobalPreferenceStore {
     fun saveDeviceToken(token: String)
-    fun saveSettings(value: String)
+    fun saveSettings(value: Settings)
     fun getDeviceToken(): String?
     fun getSettings(): Settings?
     fun removeDeviceToken()
@@ -31,8 +31,8 @@ internal class GlobalPreferenceStoreImpl(
         putString(KEY_DEVICE_TOKEN, token)
     }
 
-    override fun saveSettings(value: String) = prefs.edit {
-        putString(KEY_CONFIG_SETTINGS, value)
+    override fun saveSettings(value: Settings) = prefs.edit {
+        putString(KEY_CONFIG_SETTINGS, Json.encodeToString(Settings.serializer(), value))
     }
 
     override fun getDeviceToken(): String? = prefs.read {
--- a/core/src/main/kotlin/io/customer/sdk/util/EventNames.kt
+++ b/core/src/main/kotlin/io/customer/sdk/util/EventNames.kt
@@ -1,10 +1,10 @@
-package io.customer.datapipelines.util
+package io.customer.sdk.util
 
 /**
  * Event names to identify specific events in data pipelines so they can be
  * reflected on Journeys.
  */
-internal object EventNames {
+object EventNames {
     const val DEVICE_UPDATE = "Device Created or Updated"
     const val DEVICE_DELETE = "Device Deleted"
     const val METRIC_DELIVERY = "Report Delivery Event"
--- a/datapipelines/api/datapipelines.api
+++ b/datapipelines/api/datapipelines.api
@@ -174,6 +174,7 @@ public final class io/customer/sdk/CustomerIO : io/customer/sdk/DataPipelineInst
 	public fun getProfileAttributes ()Ljava/util/Map;
 	public fun getRegisteredDeviceToken ()Ljava/lang/String;
 	public fun getUserId ()Ljava/lang/String;
+	public final fun getWriteKey ()Ljava/lang/String;
 	public fun identify (Ljava/lang/String;Ljava/lang/Object;Lkotlinx/serialization/SerializationStrategy;)V
 	public fun initialize ()V
 	public static final fun instance ()Lio/customer/sdk/CustomerIO;
--- a/datapipelines/api/datapipelines.api
+++ b/datapipelines/api/datapipelines.api
@@ -174,7 +174,6 @@ public final class io/customer/sdk/CustomerIO : io/customer/sdk/DataPipelineInst
 	public fun getProfileAttributes ()Ljava/util/Map;
 	public fun getRegisteredDeviceToken ()Ljava/lang/String;
 	public fun getUserId ()Ljava/lang/String;
-	public final fun getWriteKey ()Ljava/lang/String;
 	public fun identify (Ljava/lang/String;Ljava/lang/Object;Lkotlinx/serialization/SerializationStrategy;)V
 	public fun initialize ()V
 	public static final fun instance ()Lio/customer/sdk/CustomerIO;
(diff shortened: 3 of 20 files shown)


gradlew
Shahroz16, line 1: ahhh need to remove these changes

core/src/main/kotlin/io/customer/sdk/data/store/GlobalPreferenceStore.kt
```diff
@@ -27,13 +31,27 @@ internal class GlobalPreferenceStoreImpl(
         putString(KEY_DEVICE_TOKEN, token)
     }
 
+    override fun saveSettings(value: String) = prefs.edit {
```
mahmoud-elmorabea, line 34: Does it make sense for this method to accept `Settings` so that the encoding/decoding is encapsulated within this store?

core/src/main/kotlin/io/customer/sdk/util/EventNames.kt
```diff
@@ -1,10 +1,10 @@
-package io.customer.datapipelines.util
+package io.customer.sdk.util
 
 /**
  * Event names to identify specific events in data pipelines so they can be
  * reflected on Journeys.
  */
-internal object EventNames {
+object EventNames {
```
mahmoud-elmorabea, line 7: Curious as to why we moved this to core? It seems like those events are very related to Data Pipelines

datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
```diff
@@ -173,6 +175,14 @@ class CustomerIO private constructor(
         logger.debug("CustomerIO SDK initialized with DataPipelines module.")
         // Migrate unsent events from previous version
         migrateTrackingEvents()
+
+        // save settings to storage
+        analytics.configuration.let { config ->
```
mahmoud-elmorabea, line 180: If this configuration object is available through analytics, what's the main reason for storing it locally? Is it for cases when the SDK hasn't been initialized yet?

messagingpush/src/main/java/io/customer/messagingpush/PushDeliveryTracker.kt
```diff
@@ -0,0 +1,56 @@
...
+import io.customer.messagingpush.di.httpClient
+import io.customer.sdk.core.di.SDKComponent
+import io.customer.sdk.util.EventNames
+import org.json.JSONObject
+
+interface PushDeliveryTracker {
+    fun trackMetric(token: String, event: String, deliveryId: String, onComplete: ((Result<Unit>) -> Unit?)? = null)
+}
+
+class PushDeliveryTrackerImpl : PushDeliveryTracker {
```
mahmoud-elmorabea, line 14: Do any of these types need to be public?

messagingpush/src/main/java/io/customer/messagingpush/PushDeliveryTracker.kt
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * Tracks a metric by performing a single POST request with JSON.
+     * Returns a `Result<Unit>`.
+     */
+    override fun trackMetric(
+        token: String,
+        event: String,
+        deliveryId: String,
+        onComplete: ((Result<Unit>) -> Unit?)?
```
mahmoud-elmorabea, line 27: Do we need a callback here? I'm not sure if the caller can do much in case of failure?

messagingpush/src/main/java/io/customer/messagingpush/processor/PushMessageProcessorImpl.kt
```diff
@@ -86,6 +92,14 @@ internal class PushMessageProcessorImpl(
...
         if (moduleConfig.autoTrackPushEvents) {
+            // Track delivered metrics via http
+            pushDeliveryTracker.trackMetric(
+                event = Metric.Delivered.name,
+                deliveryId = deliveryId,
+                token = deliveryToken
+            )
+
+            // Track delivered metrics via event bus
             eventBus.publish(
```
mahmoud-elmorabea, line 103: Does this mean that this event will be tracked twice if the app is open or not restricted by OS to run in the background?

datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
mahmoud-elmorabea, line 180: Should we add a test for this?

messagingpush/api/messagingpush.api
```diff
@@ -54,6 +54,19 @@ public final class io/customer/messagingpush/ModuleMessagingPushFCM : io/custome
...
+
+public final class io/customer/messagingpush/PushDeliveryTracker$DefaultImpls {
+	public static synthetic fun trackMetric$default (Lio/customer/messagingpush/PushDeliveryTracker;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Lkotlin/jvm/functions/Function1;ILjava/lang/Object;)V
+}
+
+public final class io/customer/messagingpush/PushDeliveryTrackerImpl : io/customer/messagingpush/PushDeliveryTracker {
+	public fun <init> ()V
+	public fun trackMetric (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Lkotlin/jvm/functions/Function1;)V
+}
+
```
mahmoud-elmorabea, line 69: Do we need those to be exposed publicly?

datapipelines/api/datapipelines.api
```diff
@@ -174,6 +174,7 @@ public final class io/customer/sdk/CustomerIO : io/customer/sdk/DataPipelineInst
 	public fun getProfileAttributes ()Ljava/util/Map;
 	public fun getRegisteredDeviceToken ()Ljava/lang/String;
 	public fun getUserId ()Ljava/lang/String;
+	public final fun getWriteKey ()Ljava/lang/String;
```
mahmoud-elmorabea, line 177: I don't see this method actually in `DataPipelinesModuleConfig`, I might be missing something though

mahmoud-elmorabea (changes requested):

core/src/main/kotlin/io/customer/sdk/util/EventNames.kt
Shahroz16: Because multiple modules are using it

datapipelines/api/datapipelines.api
Shahroz16: Good catch, it was an unused variable added to `CustomerIO` class.

datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
Shahroz16: analytics is a dependency/part of `data pipeline` module, while `push module` doesn't rely on it. It just needs `writeKey and endpoint` that it can use to make the HTTP request.

messagingpush/api/messagingpush.api
Shahroz16: I thought about it, made them public when I was adding test for it. But I am going to revisit it.

messagingpush/src/main/java/io/customer/messagingpush/PushDeliveryTracker.kt
Shahroz16: I thought about it, made them public when I was adding test for it. But I am going to revisit it.

messagingpush/src/main/java/io/customer/messagingpush/PushDeliveryTracker.kt
Shahroz16: Added it, in case we wanted to do some logging based on it or perform any operation based on its result.

messagingpush/src/main/java/io/customer/messagingpush/processor/PushMessageProcessorImpl.kt
Shahroz16: Yes, thats true. But i think its safer this way. Duplicate delivery is doesn't reflect on journeys etc but it safe guard from the fact that HTTP request can fail and we shouldn't just rely on it. The other way it becomes of offline queue so even if it fails, we have it in queue to be retried.

core/src/main/kotlin/io/customer/sdk/data/store/GlobalPreferenceStore.kt
Shahroz16: Think it does make sense, i wanted to avoid strong coupling of serialization with storage, but since string format could it make it generic and prone to break, it probably wiser to make this method stricter.

datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
Shahroz16: Let me try adding test for these

messagingpush/src/main/java/io/customer/messagingpush/processor/PushMessageProcessorImpl.kt
mahmoud-elmorabea: So Journey's will handle this as a single delivery if we report it twice? My understanding is that this will count twice against the same delivery ID

Shahroz16: > My understanding is that this will count twice against the same delivery ID

I am not 100% sure I get this? there were will be 2 requests in cdp and two might to journeys, but where is the count ?

datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
mrehan27: Would be nice to add these tests for the `EU` region as well

messagingpush/src/main/java/io/customer/messagingpush/network/HTTPClient.kt
```diff
@@ -0,0 +1,120 @@
...
+     *   - `Result.success(responseBody)` for 2xx response codes
+     *   - `Result.failure(exception)` for network errors or non-2xx codes
+     */
+    fun request(params: HttpRequestParams, onComplete: (Result<String>) -> Unit)
+}
+
+internal class HttpClientImpl : HttpClient {
+
+    private val connectTimeoutMs = 15_000
+    private val readTimeoutMs = 20_000
```
mrehan27, line 36: Do we really need such high timeout values here? 🤔

Shahroz16: nope not really, got these from iOS HTTP client and then kept it for consistency. We can lower it if needed.

mahmoud-elmorabea (approved):
Looks good! 💯  Thanks for addressing the comments 😇

messagingpush/src/main/java/io/customer/messagingpush/processor/PushMessageProcessorImpl.kt
mrehan27: The points make sense. However, I think we can still skip sending metrics to `EventBus` since the implementation aligns with iOS and has lower chance of failure due to network issues as push can only be delivered when the network is available. Or maybe we could push to `EventBus` only in case of failure? 🤔

That said, I do agree this ensures `delivered` metrics are always tracked even if the request fails. Since CDP already skips duplicate requests, that shouldn't be a concern. So I'm okay with either approach we decide to continue with.

mrehan27 (approved):

PR #494, opened 2025-02-12
chore: fix branch name fallback

mrehan27:
### Changes

- Fixed fallback value for `branchName` in primary sample app to match with other sample apps
- Updated workflow to add fallback for cases where tags are not fetched properly

### Screenshots

| Before |  After |
|-|-|
| img_before | img_after |


Shahroz16 (approved):

mahmoud-elmorabea (approved):
💯

PR #500, opened 2025-03-30
chore: fix screen view order

Shahroz16:
closes: [MBL-978: \[BUG\] Screenview current name incorrect when view changes rapidly](https://linear.app/customerio/issue/MBL-978/[bug]-screenview-current-name-incorrect-when-view-changes-rapidly)

changes: 
- updated the `CustomerIO` method calls to be `synchronized`
- move away from plugin for module updates


datapipelines/src/main/kotlin/io/customer/datapipelines/plugins/CustomerIODestination.kt
```diff
@@ -79,13 +81,11 @@ class CustomerIODestination : DestinationPlugin(), VersionedPlugin, Subscriber {
...
 
         // convert flushAt and flushIntervals into FlushPolicies
-        flushPolicies = if (analytics.configuration.flushPolicies.isEmpty()) {
+        flushPolicies = analytics.configuration.flushPolicies.ifEmpty {
             listOf(
                 CountBasedFlushPolicy(analytics.configuration.flushAt),
                 FrequencyFlushPolicy(analytics.configuration.flushInterval * 1000L)
             )
-        } else {
-            analytics.configuration.flushPolicies
```
Ahmed-Ali, line 88: how are we handling this case after the change?

Shahroz16: this was an outdated change from stacked PR.

datapipelines/src/test/java/io/customer/datapipelines/ConcurrentScreenViewTest.kt
```diff
@@ -0,0 +1,129 @@
...
+
+        // Give time for execution to complete
+        Thread.sleep(1000)
+
+        // Shutdown executor
+        executor.shutdown()
+        executor.awaitTermination(1, TimeUnit.SECONDS)
+
+        // Verify that each screen was recorded exactly once
+        screenNames.forEach { name ->
```
mrehan27, line 125: Shouldn't we also add a test to ensure the sequence of events in `EventBus` matches the sequence of `screen` method call?

mrehan27 (approved):
PR mostly looks good to me. I understand that we're ensuring one thing at a time. In addition to any PR comments, I'd still like us to verify the following before merging:

- Ensure `analytics` in `CustomerIO` class is no longer used directly anywhere else to prevent missing `EventBus` events. If possible, we should make `analytics` `private` to avoid future errors, though I understand that might be too much work for now and can be improved later.
- Confirm that all events are still posted to `EventBus` from all possible paths (e.g., `track` and other methods) after removing `DataPipelinePublishedEvents`.
- Ensure no unnecessary methods (like `trackImpl`) are exposed to customers. I checked and it looks good, but still worth confirming.
- I remember we offload processing using coroutines when tracking, but still good to verify that sending too many events at same time doesn't cause an ANR in the caller thread.

Once these are verified (or if they've already been checked), I believe we're good to merge and move forward. Thanks for all the efforts and great work on this! 👍🏻

datapipelines/src/test/java/io/customer/datapipelines/ConcurrentScreenViewTest.kt
Shahroz16: updated tests check sequence

Shahroz16:
> PR mostly looks good to me. I understand that we're ensuring one thing at a time. In addition to any PR comments, I'd still like us to verify the following before merging:
> 
> * Ensure `analytics` in `CustomerIO` class is no longer used directly anywhere else to prevent missing `EventBus` events. If possible, we should make `analytics` `private` to avoid future errors, though I understand that might be too much work for now and can be improved later.
> * Confirm that all events are still posted to `EventBus` from all possible paths (e.g., `track` and other methods) after removing `DataPipelinePublishedEvents`.
> * Ensure no unnecessary methods (like `trackImpl`) are exposed to customers. I checked and it looks good, but still worth confirming.
> * I remember we offload processing using coroutines when tracking, but still good to verify that sending too many events at same time doesn't cause an ANR in the caller thread.
> 
> Once these are verified (or if they've already been checked), I believe we're good to merge and move forward. Thanks for all the efforts and great work on this! 👍🏻

>  we should make `analytics` `private` to avoid future errors, though I understand that might be too much work for now and can be improved later.

yup, all out tests internally use `analytics` to find plugins and other validations. 

> Confirm that all events are still posted to `EventBus` from all possible paths (e.g., `track` and other methods) after removing `DataPipelinePublishedEvents`

The only way to access these methods is via the `CustomerIO` class, and they all call `trackImpl` internally which post on event bus. 

> Ensure no unnecessary methods (like `trackImpl`) are exposed to customers

Yup, they are protected methods, so they shouldn't be exposed to end users but rather implementation of this protocol. 

> I remember we offload processing using coroutines when tracking, but still good to verify that sending too many events at same time doesn't cause an ANR in the caller thread.

yup, a coroutine is launched bu Analytics for every method, I have added some tests too for this,

PR #502, opened 2025-04-04
fix: support null handling in nested structures

mrehan27:
part of: [MBL-955](https://linear.app/customerio/issue/MBL-955/[bug]-passing-nested-null-attribute-crashes-android-in-react-native)

### Changes

- Added `sanitizeNullsForJson()` extension to recursively replace all `null` values with `JsonNull` in `CustomAttributes` for `Map` and `List` before forwarding to `analytics`
- Added unit tests to validate the behavior of new extension

#### Sample Code

Crash on `main` before this fix:

```kotlin
CustomerIO.instance().track(
    "Add to Cart",
    mapOf(
        "item" to mapOf(
            "id" to "100",
            "name" to "Mobile",
            "discount" to null,
        ),
    )
)
```

### DIFF
--- a/datapipelines/src/main/kotlin/io/customer/datapipelines/extensions/JsonExtensions.kt
+++ b/datapipelines/src/main/kotlin/io/customer/datapipelines/extensions/JsonExtensions.kt
@@ -43,3 +43,16 @@ private fun Any?.toSerializableJson(): JsonElement {
         else -> this.toJsonElement()
     }
 }
+
+/**
+ * Recursively replaces all `null` values in nested structure with [JsonNull],
+ * preserving structure and type. Useful for safely converting to `JsonElement`
+ * using `Any.toJsonElement()` extension.
+ */
+@Suppress("UNCHECKED_CAST")
+internal fun <T> T?.sanitizeNullsForJson(): T = when (this) {
+    null -> JsonNull as T
+    is Map<*, *> -> this.mapValues { (_, v) -> v.sanitizeNullsForJson() } as T
+    is List<*> -> this.map { it.sanitizeNullsForJson() } as T
+    else -> this
+}
--- a/datapipelines/src/main/kotlin/io/customer/sdk/DataPipelineInstance.kt
+++ b/datapipelines/src/main/kotlin/io/customer/sdk/DataPipelineInstance.kt
@@ -67,7 +67,7 @@ abstract class DataPipelineInstance : CustomerIOInstance {
      */
     fun identify(userId: String, traits: CustomAttributes) {
         // Method needed for Java interop as inline doesn't work with Java
-        identify(userId = userId, traits = traits.sanitizeNullsForJson(), serializationStrategy = JsonAnySerializer.serializersModule.serializer())
+        identify(userId = userId, traits = traits, serializationStrategy = JsonAnySerializer.serializersModule.serializer())
     }
 
     /**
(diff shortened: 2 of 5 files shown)


datapipelines/src/main/kotlin/io/customer/datapipelines/extensions/JsonExtensions.kt
```diff
@@ -43,3 +43,16 @@ private fun Any?.toSerializableJson(): JsonElement {
...
+ * preserving structure and type. Useful for safely converting to `JsonElement`
+ * using `Any.toJsonElement()` extension.
+ */
+@Suppress("UNCHECKED_CAST")
+internal fun <T> T?.sanitizeNullsForJson(): T = when (this) {
+    null -> JsonNull as T
+    is Map<*, *> -> this.mapValues { (_, v) -> v.sanitizeNullsForJson() } as T
+    is List<*> -> this.map { it.sanitizeNullsForJson() } as T
+    else -> this
+}
```
mrehan27, line 58: Another variation I had in mind was following, but it looked overly complicated so I stick to current implementation for simplicity

```kotlin
internal fun <T> T?.sanitizeNullsForJsons(): T = when (this) {
    null -> JsonNull as T

    is Map<*, *> -> this.takeIf {
        it.values.none { v -> v == null || v is Map<*, *> || v is List<*> }
    } ?: this.mapValues { (_, v) -> v.sanitizeNullsForJson() } as T

    is List<*> -> this.takeIf {
        it.none { e -> e == null || e is Map<*, *> || e is List<*> }
    } ?: this.map { it.sanitizeNullsForJson() } as T

    else -> this
}
```

datapipelines/src/main/kotlin/io/customer/sdk/DataPipelineInstance.kt
```diff
@@ -66,7 +67,7 @@ abstract class DataPipelineInstance : CustomerIOInstance {
      */
     fun identify(userId: String, traits: CustomAttributes) {
         // Method needed for Java interop as inline doesn't work with Java
-        identify(userId = userId, traits = traits, serializationStrategy = JsonAnySerializer.serializersModule.serializer())
+        identify(userId = userId, traits = traits.sanitizeNullsForJson(), serializationStrategy = JsonAnySerializer.serializersModule.serializer())
```
mrehan27, line 70: The map eventually encodes to JSON using [`toJsonElement`](https://github.com/customerio/cdp-analytics-kotlin/blob/af97bf3e78519da7d9446eb3e9d507c84d3f94e5/core/src/main/java/com/segment/analytics/kotlin/core/utilities/JSON.kt#L396C9-L396C22) which is an extension on `Any` and not `Any?` resulting in crash. The sanitization here maps `null` items to `JsonNull` and prevent the crash. I used the extension only with `CustomAttributes` since `JsonObject` handles `null` on its own without any changes.

datapipelines/src/main/kotlin/io/customer/datapipelines/extensions/JsonExtensions.kt
mrehan27: The third option is to add the fix directly to `analytics`. That's also a good option if we want to try it out.

Shahroz16:
Closing this, as the work is going to be replaced by my new PR

PR #505, opened 2025-04-07
chore: update gradle to 8.10 and add kotlin support to Java sample

mrehan27:
part of: [MBL-990](https://linear.app/customerio/issue/MBL-990/sample-app-changes-for-android-xml-view-only)

### Changes

- Updated Gradle plugin version from `8.4` to `8.10` for better compatibility with newer versions of Android Studio
- Enabled Kotlin support in Java sample app as it's now our primary sample and will include more Kotlin code going forward


Shahroz16 (commented):
kotlin sample doesn't seem to compile, seems like a JDK issue, could this be a requirement for gradle update?

mrehan27:
Thanks for pointing it out @Shahroz16. There is indeed an issue with Kotlin sample locally. I'll close this one for now and revert to updating JDK on CLI. We can discuss any future improvements on slack with more details. Thanks 🙏🏻

PR #507, opened 2025-04-07
chore: add inline xml example to java sample

mrehan27:
part of: [MBL-990](https://linear.app/customerio/issue/MBL-990/sample-app-changes-for-android-xml-view-only)

### Changes

- Added `InlineExamplesActivity` to host inline message examples (Compose example will be added as fragment in the same activity in future)
- Created `AndroidXMLInlineExampleFragment` to show inline messages using Android XML layouts
- Added placeholder TextViews for inline message views which will be replaced with actual views once they're publicly available

### Screenshots

Screenshot_20250407_192314


Shahroz16 (approved):
lgtm

PR #508, opened 2025-04-08
fix: support null handling in nested structures

Shahroz16:
closes: [MBL-955](https://linear.app/customerio/issue/MBL-955/[bug]-passing-nested-null-attribute-crashes-android-in-react-native)

### Changes

- Added `sanitizeNullsForJson()` extension to recursively replace all `null` values with `JsonNull` in `CustomAttributes` for `Map` and `List` before forwarding to `analytics`
- Added unit tests to validate the behavior of new extension
- Added tests on `CustomerIO.instance` to validate end-to-end behavior
- Updated Datapipeline method to cater `Map<String,Any?>` so they can cater cases of both, `CustomeAttribute` and `mapof type <string, any?>` 
 
#### Sample Code

Crash on `main` before this fix:

```kotlin
CustomerIO.instance().track(
    "Add to Cart",
    mapOf(
        "item" to mapOf(
            "id" to "100",
            "name" to "Mobile",
            "discount" to null,
        ),
    )
)
```

### DIFF
--- a/datapipelines/src/main/kotlin/io/customer/sdk/DataPipelineInstance.kt
+++ b/datapipelines/src/main/kotlin/io/customer/sdk/DataPipelineInstance.kt
@@ -67,7 +67,7 @@ abstract class DataPipelineInstance : CustomerIOInstance {
      */
     fun identify(userId: String, traits: CustomAttributes) {
         // Method needed for Java interop as inline doesn't work with Java
-        identify(userId = userId, traits = traits.sanitizeNullsForJson(), serializationStrategy = JsonAnySerializer.serializersModule.serializer())
+        identify(userId = userId, traits = traits, serializationStrategy = JsonAnySerializer.serializersModule.serializer())
     }
 
     /**
@@ -89,7 +89,7 @@ abstract class DataPipelineInstance : CustomerIOInstance {
         serializationStrategy: SerializationStrategy<Traits>
     ) {
         synchronized {
-            identifyImpl(userId, traits, serializationStrategy)
+            identifyImpl(userId, traits.sanitizeNullsForJson(), serializationStrategy)
         }
     }
 
@@ -131,7 +131,7 @@ abstract class DataPipelineInstance : CustomerIOInstance {
      */
     fun track(name: String, properties: CustomAttributes) {
         // Method needed for Java interop as inline doesn't work with Java
-        track(name = name, properties = properties.sanitizeNullsForJson(), serializationStrategy = JsonAnySerializer.serializersModule.serializer())
+        track(name = name, properties = properties, serializationStrategy = JsonAnySerializer.serializersModule.serializer())
     }
 
     /**
@@ -152,7 +152,7 @@ abstract class DataPipelineInstance : CustomerIOInstance {
         serializationStrategy: SerializationStrategy<T>
     ) {
         synchronized {
-            trackImpl(name, properties, serializationStrategy)
+            trackImpl(name, properties.sanitizeNullsForJson(), serializationStrategy)
         }
     }
 
@@ -204,7 +204,7 @@ abstract class DataPipelineInstance : CustomerIOInstance {
      */
     fun screen(title: String, properties: CustomAttributes) {
         // Method needed for Java interop as inline doesn't work with Java
-        screen(title = title, properties = properties.sanitizeNullsForJson(), serializationStrategy = JsonAnySerializer.serializersModule.serializer())
+        screen(title = title, properties = properties, serializationStrategy = JsonAnySerializer.serializersModule.serializer())
     }
 
     /**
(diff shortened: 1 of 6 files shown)


datapipelines/src/main/kotlin/io/customer/sdk/DataPipelineInstance.kt
```diff
@@ -64,9 +65,9 @@ abstract class DataPipelineInstance : CustomerIOInstance {
      * [Learn more](https://customer.io/docs/api/#operation/identify)
      * @param traits Map of <String, Any> to be added
      */
-    fun identify(userId: String, traits: CustomAttributes) {
```
mahmoud-elmorabea, line 67: It looks like in the `CustomerIO` implementation the setter of profile attributes is calling `analytics.identify(traits)` directly in case of anonymous users, this will bypass the sanitization happening at the abstract `DataPipelineInstance`, right?

datapipelines/src/main/kotlin/io/customer/sdk/DataPipelineInstance.kt
```diff
@@ -64,9 +65,9 @@ abstract class DataPipelineInstance : CustomerIOInstance {
...
+    fun identify(userId: String, traits: Map<String, Any?>) {
         // Method needed for Java interop as inline doesn't work with Java
-        identify(userId = userId, traits = traits, serializationStrategy = JsonAnySerializer.serializersModule.serializer())
+        identify(userId = userId, traits = traits.sanitizeNullsForJson(), serializationStrategy = JsonAnySerializer.serializersModule.serializer())
```
mahmoud-elmorabea, line 70: Do we not think it's more appriopriate to do this sanitization at the implementation level `CustomerIO` class instead of the abstraction? 

If `CustomerIO` is called from any other place in the future directly, it will have the same problem

mahmoud-elmorabea (changes requested):
Approach looks solid! 💯 

Just wanted to double check on two main concerns:
- I think we might have missed at least one place using `CustomAttributes`, maybe it's worth doing a search for it to make sure we covered all. I only found one though
- I cross tested with iOS quickly and it seems like iOS omits the value all together rather than sending it null, but it's not consistent, have we cross checked behavior with iOS?

datapipelines/src/main/kotlin/io/customer/sdk/DataPipelineInstance.kt
Shahroz16: I initially went with that solution, but it didn't work out. why?
- If we try to sanitize data types other than a map, we are going to get classcast exception 
e.g
```
fun <Traits> identify(
        userId: String,
        traits: Traits,
        serializationStrategy: SerializationStrategy<Traits>
    ) {
        synchronized {
            identifyImpl(userId, traits.sanitizeNullsForJson(), serializationStrategy)
        }
    }
```

you are going to face this crash 

```
class java.util.LinkedHashMap cannot be cast to class kotlinx.serialization.json.JsonObject (java.util.LinkedHashMap is in module java.base of loader 'bootstrap'; kotlinx.serialization.json.JsonObject is in unnamed module of loader 'app')
```

Because:
	•	After calling `sanitizeNullsForJson()`, we end up with a plain Kotlin Map (LinkedHashMap) rather than a JsonObject.
	•	Our serializer (the SerializationStrategy) expects a JsonObject (or JsonElement), so it tries to cast the LinkedHashMap to JsonObject, causing the ClassCastException.
	•	In the case where we directly pass a concrete Map<String, Any?> to identify(...), the code uses a map-based serializer, so there is no cast to JsonObject and thus no crash.
	•	In contrast, when we call the generic identify(...) method, the serializer is inferred or supplied such that it’s expecting a JsonObject instead of a map, leading to the casting error.

Shahroz16:
> * I cross tested with iOS quickly and it seems like iOS omits the value all together rather than sending it null, but it's not consistent, have we cross checked behavior with iOS?

Thats a good callout, I agree with consistency but I also don't want to introduce a breaking change in terms of behavior. For example, currently `null` in the parent are treated as JsonNull and not skipped. 

If we add logic of skipping nulls, then even that current behavior is going to be changed and if anyone has any segments/events driven based on that current behavior, it could break for them

datapipelines/src/main/kotlin/io/customer/sdk/DataPipelineInstance.kt
mahmoud-elmorabea: Thanks for the clarification!

datapipelines/src/main/kotlin/io/customer/sdk/DataPipelineInstance.kt
mahmoud-elmorabea: @Shahroz16 Did we fix this one yet? I still see that the direct call to `analytics.identify(traits)` is not sanitized

Shahroz16: its showing up now

PR #510, opened 2025-04-09
refactor: rename GistView to base in-app view

mrehan27:
part of: [MBL-1027](https://linear.app/customerio/issue/MBL-1027/refactor-gistview-for-modal-and-inline-in-app-message-reuse)

### Changes

- Removed `Gist` naming convention from base views and listeners related to in-app message display:
  - Renamed `GistView` to `InAppMessageBaseView` - an abstract base class for both modal and inline in-app messages
  - Renamed `GistViewListener` to `InAppMessageViewEventsListener` for clarity
- Added `ModalInAppMessageView` as final view class for displaying and managing modal in-app message logic
- Replaced `GistView` with `ModalInAppMessageView` and used it for displaying modal in-app messages (no functional changes)

### DIFF
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/ui/InAppMessageBaseView.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/ui/InAppMessageBaseView.kt
@@ -1,4 +1,4 @@
-package io.customer.messaginginapp.gist.presentation
+package io.customer.messaginginapp.ui
 
 import android.content.ActivityNotFoundException
 import android.content.Context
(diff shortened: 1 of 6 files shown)


messaginginapp/src/main/java/io/customer/messaginginapp/ui/InAppMessageBaseView.kt
```diff
@@ -22,16 +22,23 @@ import io.customer.sdk.core.di.SDKComponent
...
 import java.nio.charset.StandardCharsets
 
-class GistView @JvmOverloads constructor(
+/**
+ * Base view class for displaying in-app messages. This class is responsible for managing common
+ * in-app message functionality, such as setting up the EngineWebView, handling tap events, etc.
+ * This class will be extended by specific in-app message views (e.g., ModalInAppMessageView)
+ * to handle specific message types.
+ */
+internal abstract class InAppMessageBaseView @JvmOverloads constructor(
```
mahmoud-elmorabea, line 31: 💯 for appropriate visibility update!

messaginginapp/src/main/java/io/customer/messaginginapp/ui/ModalInAppMessageView.kt
```diff
@@ -0,0 +1,14 @@
+package io.customer.messaginginapp.ui
+
+import android.content.Context
+import android.util.AttributeSet
+
+/**
+ * Final implementation of the InAppMessageHostView for displaying modal in-app messages.
+ * The view should be directly added to activity layout for displaying modal in-app messages.
+ */
+internal class ModalInAppMessageView @JvmOverloads constructor(
```
mahmoud-elmorabea, line 10: Since this is just a view and doesn't really have anything special or specific to being shown in modal, wondering why we'll need two sub-classes for model vs inline?

mahmoud-elmorabea (approved):
Looks great! 💯 

Thanks for keeping our API tight with the appropriate visibility levels! 🥇

messaginginapp/src/main/java/io/customer/messaginginapp/ui/ModalInAppMessageView.kt
mrehan27: #511 shows a brief example of how this could be helpful in extending the logic. I added this hierarchy to make it easier for us to have separate implementations and better optimize modal and inline in-app messages without affecting each other. Even if they remain empty for now, I believe they still add value for future enhancements without risking breaking things for the other type.

PR #511, opened 2025-04-09
refactor: base in-app message view

mrehan27:
part of: [MBL-1027](https://linear.app/customerio/issue/MBL-1027/refactor-gistview-for-modal-and-inline-in-app-message-reuse)

### Changes

- Renamed `listener` to `eventsListener` in `InAppMessageBaseView` for clarity and to avoid confusion with `listener` in `EngineWebView`
- Cleaned up and reorganized code in `InAppMessageBaseView` to simplify structure and reduce lint warnings
- Moved the following functionality to final child class `ModalInAppMessageView`:
  - Setting up `EngineWebView` during view initialization
  - `routeLoaded` logic for updating alpha
  - Remaining logic will be gradually moved to child classes as needed
- No other intentional logic changes in `InAppMessageBaseView`

### DIFF
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/ui/InAppMessageBaseView.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/ui/InAppMessageBaseView.kt
@@ -3,13 +3,12 @@ package io.customer.messaginginapp.ui
 import android.content.ActivityNotFoundException
 import android.content.Context
 import android.content.Intent
-import android.net.Uri
 import android.net.UrlQuerySanitizer
 import android.util.AttributeSet
 import android.util.Base64
-import android.util.DisplayMetrics
 import android.widget.FrameLayout
 import androidx.core.content.ContextCompat.startActivity
+import androidx.core.net.toUri
 import com.google.gson.Gson
 import io.customer.messaginginapp.di.inAppMessagingManager
 import io.customer.messaginginapp.gist.data.model.Message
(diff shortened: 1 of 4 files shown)


messaginginapp/src/main/java/io/customer/messaginginapp/ui/InAppMessageBaseView.kt
```diff
@@ -33,168 +32,215 @@ internal abstract class InAppMessageBaseView @JvmOverloads constructor(
...
     }
 
     fun stopLoading() {
+        logViewEvent("Stopping EngineWebView loading")
         engineWebView?.stopLoading()
     }
 
     override fun tap(name: String, action: String, system: Boolean) {
+        val message = currentMessage ?: return
+        val route = currentRoute ?: return
```
mahmoud-elmorabea, line 109: Maybe we want to log here that we are ignoring a tap because those fields are null?

messaginginapp/src/main/java/io/customer/messaginginapp/ui/InAppMessageBaseView.kt
```diff
@@ -33,168 +32,215 @@ internal abstract class InAppMessageBaseView @JvmOverloads constructor(
...
+
         var shouldLogAction = true
-        currentMessage?.let { message ->
```
mahmoud-elmorabea, line 81: It would actually if we can plan for future changes to extract this logic into a pure class that we can test

mahmoud-elmorabea (approved):
Looks good! 💯

messaginginapp/src/main/java/io/customer/messaginginapp/ui/InAppMessageBaseView.kt
mrehan27: I believe it would be a rare case, but I'll take another look and add it later if needed.

messaginginapp/src/main/java/io/customer/messaginginapp/ui/InAppMessageBaseView.kt
mrehan27: Absolutely. That's something I'll keep an eye on and try extracting the code to make it as independently testable as possible.

PR #513, opened 2025-04-13
chore: added reducer and action

Shahroz16:
closes: https://linear.app/customerio/issue/MBL-1024/add-actions-and-reducers-to-update-the-states-of-inline-views

changes: 
- Added logic for reducer to updates states based on action

### DIFF
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessageReducer.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessageReducer.kt
@@ -12,9 +12,18 @@ val inAppMessagingReducer: Reducer<InAppMessagingState> = { state, action ->
 
         is InAppMessagingAction.ProcessMessageQueue -> state.copy(messagesInQueue = action.messages.toSet())
         is InAppMessagingAction.SetPollingInterval -> state.copy(pollInterval = action.interval)
-        is InAppMessagingAction.EngineAction.MessageLoadingFailed -> state.copy(currentMessageState = MessageState.Dismissed(action.message))
-        is InAppMessagingAction.LoadMessage -> state.copy(currentMessageState = MessageState.Loading(action.message))
+        is InAppMessagingAction.EngineAction.MessageLoadingFailed -> state.copy(modalMessageState = MessageState.Dismissed(action.message))
+        is InAppMessagingAction.LoadMessage -> state.copy(modalMessageState = MessageState.Loading(action.message))
         is InAppMessagingAction.Reset -> InAppMessagingState(siteId = state.siteId, dataCenter = state.dataCenter, environment = state.environment)
+        is InAppMessagingAction.EmbedMessages -> {
+            var newEmbeddedMessagesState = state.embeddedMessagesState
+            action.messages.forEach { message ->
+                message.elementId?.let { elementId ->
+                    newEmbeddedMessagesState = newEmbeddedMessagesState.addMessage(message, elementId)
+                }
+            }
+            state.copy(embeddedMessagesState = newEmbeddedMessagesState)
+        }
         is InAppMessagingAction.DisplayMessage -> {
             action.message.queueId?.let { queueId ->
                 // If the message should be tracked shown when it is displayed, add the queueId to shownMessageQueueIds.
@@ -24,11 +33,26 @@ val inAppMessagingReducer: Reducer<InAppMessagingState> = { state, action ->
                     state.shownMessageQueueIds
                 }
 
-                state.copy(
-                    currentMessageState = MessageState.Displayed(action.message),
-                    shownMessageQueueIds = shownMessageQueueIds,
-                    messagesInQueue = state.messagesInQueue.filterNot { it.queueId == queueId }.toSet()
-                )
+                // Remove the message from the queue
+                val filteredQueue = state.messagesInQueue.filterNot { it.queueId == queueId }.toSet()
+
+                if (action.message.isEmbedded) {
+                    // Update embedded message state
+                    val elementId = action.message.elementId ?: ""
+                    state.updateEmbeddedMessage(
+                        queueId = queueId,
+                        newState = InlineMessageState.Embedded(action.message, elementId),
+                        shownMessageQueueIds = shownMessageQueueIds,
+                        messagesInQueue = filteredQueue
+                    )
+                } else {
+                    // Update modal message state
+                    state.copy(
+                        modalMessageState = MessageState.Displayed(action.message),
+                        shownMessageQueueIds = shownMessageQueueIds,
+                        messagesInQueue = filteredQueue
+                    )
+                }
             } ?: state
         }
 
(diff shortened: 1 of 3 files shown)


messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessageReducer.kt
```diff
@@ -5,16 +5,44 @@ import org.reduxkotlin.Reducer
...
-        is InAppMessagingAction.SetPageRoute -> state.copy(currentRoute = action.route)
-        is InAppMessagingAction.SetUserIdentifier -> state.copy(userId = action.user)
-        is InAppMessagingAction.ClearMessageQueue -> state.copy(messagesInQueue = setOf())
-
-        is InAppMessagingAction.ProcessMessageQueue -> state.copy(messagesInQueue = action.messages.toSet())
-        is InAppMessagingAction.SetPollingInterval -> state.copy(pollInterval = action.interval)
-        is InAppMessagingAction.EngineAction.MessageLoadingFailed -> state.copy(currentMessageState = MessageState.Dismissed(action.message))
-        is InAppMessagingAction.LoadMessage -> state.copy(currentMessageState = MessageState.Loading(action.message))
-        is InAppMessagingAction.Reset -> InAppMessagingState(siteId = state.siteId, dataCenter = state.dataCenter, environment = state.environment)
+        // Simple state copy operations
```
mahmoud-elmorabea, line 8: Not sure if this comment adds value here

messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessageReducer.kt
```diff
@@ -5,16 +5,44 @@ import org.reduxkotlin.Reducer
...
+
+        is InAppMessagingAction.LoadMessage ->
+            state.copy(modalMessageState = MessageState.Loading(action.message))
+
+        // Using the reset method for cleaner implementation
+        is InAppMessagingAction.Reset -> state.reset()
+
+        is InAppMessagingAction.EmbedMessages -> {
+            // Handling embedding messages in a single loop for better performance
+            val newEmbeddedMessagesState = action.messages.fold(state.embeddedMessagesState) { acc, message ->
```
mahmoud-elmorabea, line 38: I know `acc` is the object where we are accumulating the messages being folded but since we are actually accumulating the folding into a state, maybe we can rename the closure's param to `state` instead of `acc`

messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessageReducer.kt
```diff
@@ -5,16 +5,44 @@ import org.reduxkotlin.Reducer
...
+                message.elementId?.let { elementId ->
```
mahmoud-elmorabea, line 39: Can the `InAppMessagingAction.EmbedMessages#messages` contain `Message`s that are not inline messages? If that's not expected maybe we can log a warning or error or smth?

Or maybe we can filter that list beforehand before starting to fold it into the new state?

messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessageReducer.kt
```diff
@@ -39,16 +82,44 @@ val inAppMessagingReducer: Reducer<InAppMessagingState> = { state, action ->
...
+                if (action.message.queueId != null) {
+                    // Update embedded message state if it has a queueId
+                    state.updateEmbeddedMessage(
+                        queueId = action.message.queueId,
+                        newState = InlineMessageState.Dismissed(action.message),
+                        shownMessageQueueIds = shownMessageQueueIds
+                    )
+                } else {
+                    // For embedded messages without queueId
+                    // Just return the state unchanged
```
mahmoud-elmorabea, line 96: What does it mean for an embedded message to not. have a `queueId`?

messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessageReducer.kt
```diff
@@ -39,16 +82,44 @@ val inAppMessagingReducer: Reducer<InAppMessagingState> = { state, action ->
...
+                )
+            }
         }
 
-        else -> state
+        is InAppMessagingAction.EngineAction.Tap ->
+            state // No state changes for tap action
+
+        is InAppMessagingAction.ReportError ->
+            state // No state changes for error reporting
```
mahmoud-elmorabea, line 112: Should we add a log here or smth?

messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessageReducer.kt
```diff
@@ -5,16 +5,44 @@ import org.reduxkotlin.Reducer
 
 val inAppMessagingReducer: Reducer<InAppMessagingState> = { state, action ->
```
mahmoud-elmorabea, line 6: Does this reducer have no tests? 😅

mrehan27 (approved):
Looks good. Not sure why lint is showing some weird warnings, couldn't see most of them when checked out branch locally 😅

messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessageReducer.kt
Shahroz16: `elementId` itself is nullable because we get it from `properties: Map<String, Any?>?` 

and we get the same model for both types, we can create more models just to differentiate but looks like a not worth the effort?

messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessageReducer.kt
Shahroz16: this is majorly for the case, if we allow users to add embedded messages locally and not be sent via backend, which is why its optional.

messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessageReducer.kt
Shahroz16: we are logging in middleware for this action.

messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessageReducer.kt
Shahroz16: `InAppMessagingStoreTest` tests it, what it does is it provides the action and then check if the state matches the expectation.

messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessageReducer.kt
mahmoud-elmorabea: No, I think it's okay to use same model and I know that on code level `elementId`is nullable.

I'm just saying can the list passed from `InAppMessagingAction.EmbedMessages` have messages without `elementId`? Would that be a bug for an embedded message to not have `elementId`? If so, maybe we want to log that case to make it visible while debugging instead of simply skipping it silently

mahmoud-elmorabea (approved):
Looks good! 👍

PR #515, opened 2025-04-15
chore: updates in middleware to handle both modal and inline messages 

Shahroz16:
closes: https://linear.app/customerio/issue/MBL-1025/middleware-logic-updates-to-act-differently-for-modal-vs-inline-views

changes:
- update in middleware to cate inline messages and emit them

### DIFF
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessagingMiddlewares.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessagingMiddlewares.kt
@@ -149,13 +149,46 @@ internal fun processMessages() = middleware<InAppMessagingState> { store, next, 
         val notShownMessages = action.messages
             .filter { message ->
                 // filter out the messages that are already shown
-                // and the messages that have an elementId because we are not handling embedded messages
-                message.queueId != null && !store.state.shownMessageQueueIds.contains(message.queueId) && message.gistProperties.elementId == null
+                message.queueId != null && !store.state.shownMessageQueueIds.contains(message.queueId)
             }
             .distinctBy(Message::queueId)
             .sortedWith(compareBy(nullsLast()) { it.priority })
 
-        val messageToBeShownWithProperties = notShownMessages.firstOrNull { message ->
+        val modalMessages = notShownMessages.filter { it.gistProperties.elementId == null }
+        val inlineMessages = notShownMessages.filter { it.gistProperties.elementId != null }
+
+        val isCurrentMessageDisplaying = store.state.modalMessageState is ModalMessageState.Displayed
+        val isCurrentMessageBeingProcessed = store.state.modalMessageState is ModalMessageState.Loading
+
+        // update the state with the messages in the queue that are not shown
+        // because in the next steps we will check if there is a message to be shown and display them
+        next(InAppMessagingAction.ProcessMessageQueue(notShownMessages))
+
+        // Handle embedded messages
+        val inLineMessagesToBeShown = inlineMessages
+            .filter { message ->
+                val routeRule = message.gistProperties.routeRule
+                val currentRoute = store.state.currentRoute
+                // Check if the message matches the current route
+                when {
+                    routeRule == null -> true
+                    currentRoute == null -> false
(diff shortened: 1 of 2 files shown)


messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessagingMiddlewares.kt
```diff
@@ -149,13 +149,46 @@ internal fun processMessages() = middleware<InAppMessagingState> { store, next,
...
+        val inLineMessagesToBeShown = inlineMessages
+            .filter { message ->
+                val routeRule = message.gistProperties.routeRule
+                val currentRoute = store.state.currentRoute
+                // Check if the message matches the current route
+                when {
+                    routeRule == null -> true
+                    currentRoute == null -> false
+                    else -> runCatching { routeRule.toRegex().matches(currentRoute) }.getOrNull() ?: true
+                }
```
mrehan27, line 177: Would be nice to move this to extension so route matching is done by single function for both modal and inline messages.

mrehan27 (approved):

messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessagingMiddlewares.kt
```diff
@@ -149,13 +149,46 @@ internal fun processMessages() = middleware<InAppMessagingState> { store, next,
...
+        // Handle embedded messages
+        val inLineMessagesToBeShown = inlineMessages
+            .filter { message ->
+                val routeRule = message.gistProperties.routeRule
+                val currentRoute = store.state.currentRoute
+                // Check if the message matches the current route
+                when {
+                    routeRule == null -> true
+                    currentRoute == null -> false
+                    else -> runCatching { routeRule.toRegex().matches(currentRoute) }.getOrNull() ?: true
```
mahmoud-elmorabea, line 176: Do we want to fallback to show a message if the regex matching fails?

messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessagingMiddlewares.kt
```diff
@@ -149,13 +149,46 @@ internal fun processMessages() = middleware<InAppMessagingState> { store, next,
...
+                }
+            }
+            .filter { message ->
+                // Ensure no duplicate embedded messages for the same elementId in the active state
+                val elementId = message.gistProperties.elementId ?: return@filter true
```
mahmoud-elmorabea, line 181: Messages without an element id would be included in the embedded messages?

messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessagingMiddlewares.kt
```diff
@@ -149,13 +149,46 @@ internal fun processMessages() = middleware<InAppMessagingState> { store, next,
...
+                val existingState = store.state.queuedInlineMessagesState.getMessage(elementId)
+                existingState !is InlineMessageState.Embedded
+            }
+
+        if (inLineMessagesToBeShown.isNotEmpty()) {
+            store.dispatch(InAppMessagingAction.EmbedMessages(inLineMessagesToBeShown))
+        }
+
+        // Handle modal messages
+        val messageToBeShown = modalMessages.firstOrNull { message ->
```
mahmoud-elmorabea, line 191: Maybe call this `modalMessagesToBeShown`?

messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessagingMiddlewares.kt
Shahroz16: we can add it

messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessagingMiddlewares.kt
Shahroz16: we already made sure of this earlier

```
val inlineMessages = notShownMessages.filter { it.gistProperties.elementId != null }
```

this is a filter on this list and done so we don't have to avoid do null checks

PR #516, opened 2025-04-16
chore: Trigger sample apps public builds after deployment is successful

mahmoud-elmorabea:
Closes: [MBL-1053](https://linear.app/customerio/issue/MBL-1053/auto-create-testbed-releases-for-every-sdk-release)

This PR triggers a sample apps public release build after the SDK has been deployed successfully so that internal stakeholders always have up to date versions to test with.

### Note
I am not entirely sure if the package will be immediately available after publishing before it can be used to build the sample app or we need a delay. Starting out with immediate trigger and we can revise later if the build fails.

### DIFF
--- a/.github/workflows/deploy-sdk.yml
+++ b/.github/workflows/deploy-sdk.yml
@@ -237,3 +237,9 @@ jobs:
           # Help on how to get the webhook URL: https://github.com/marketplace/actions/slack-send#setup-2
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_NOTIFY_RELEASES_WEBHOOK_URL }}
           SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK
+
+  publish-sample-apps-public-builds:
+    needs: deploy-sonatype
+    uses: ./.github/workflows/build-sample-app-for-sdk-release.yml
+    secrets: inherit
+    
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-sdk.yml
```diff
@@ -237,3 +237,9 @@ jobs:
           # Help on how to get the webhook URL: https://github.com/marketplace/actions/slack-send#setup-2
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_NOTIFY_RELEASES_WEBHOOK_URL }}
           SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK
+
+  publish-sample-apps-public-builds:
+    needs: deploy-sonatype
```
mrehan27, line 242: I rolled out a release and tested it. It wasn't instantly available after Deploy action completed and took around 5–10 minutes to become available, in case that helps.

mrehan27: I kept trying to pull it locally on my machine until it became available shortly after deployment action completed. I do agree that syncing across regions can take time, but I still doubt it's instantly available. That said, I'll approve so you can verify your assumptions as well.

mrehan27 (approved):
I don't think it'll be instantly available, but approving so you can try it out and validate the changes in actual workflow.

PR #517, opened 2025-04-17
fix: delete old device token on token refresh

mrehan27:
fixes: [MBL-1061](https://linear.app/customerio/issue/MBL-1061/android-sdk-455-device-refresh-removals)

### Changes

- Included old token in Device Delete event to prevent it from being overwritten by new token, avoiding accidental deletion of new device
- Updated token deletion logic during profile change to ensure the device is removed only from previous profile
- Added tests to verify the updated behavior

### DIFF
--- a/datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
+++ b/datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
@@ -9,6 +9,7 @@ import com.segment.analytics.kotlin.core.platform.EnrichmentClosure
 import com.segment.analytics.kotlin.core.platform.plugins.logger.LogKind
 import com.segment.analytics.kotlin.core.platform.plugins.logger.LogMessage
 import com.segment.analytics.kotlin.core.utilities.JsonAnySerializer
+import com.segment.analytics.kotlin.core.utilities.putInContextUnderKey
 import io.customer.base.internal.InternalCustomerIOApi
 import io.customer.datapipelines.config.DataPipelinesModuleConfig
 import io.customer.datapipelines.di.analyticsFactory
@@ -215,9 +216,11 @@ class CustomerIO private constructor(
 
         if (isChangingIdentifiedProfile) {
             logger.info("changing profile from id $currentlyIdentifiedProfile to $userId")
-            if (registeredDeviceToken != null) {
+            if (registeredDeviceToken != null && currentlyIdentifiedProfile != null) {
                 logger.debug("deleting device token before identifying new profile")
-                deleteDeviceToken()
+                deleteDeviceToken { event ->
+                    event?.apply { this.userId = currentlyIdentifiedProfile }
+                }
             }
         }
 
@@ -319,7 +322,9 @@ class CustomerIO private constructor(
         val existingDeviceToken = contextPlugin.deviceToken
         if (existingDeviceToken != null && existingDeviceToken != token) {
             logger.debug("token has been refreshed, deleting old token to avoid registering same device multiple times")
-            deleteDeviceToken()
+            deleteDeviceToken { event ->
+                event?.putInContextUnderKey("device", "token", existingDeviceToken)
+            }
         }
 
         val attributes = if (moduleConfig.autoTrackDeviceAttributes) {
--- a/datapipelines/src/test/java/io/customer/datapipelines/DataPipelinesStandardDispatcherTest.kt
+++ b/datapipelines/src/test/java/io/customer/datapipelines/DataPipelinesStandardDispatcherTest.kt
@@ -1,0 +1,175 @@
+package io.customer.datapipelines
+
+import com.segment.analytics.kotlin.core.EventType
+import io.customer.commontest.config.TestConfig
+import io.customer.commontest.extensions.random
+import io.customer.datapipelines.testutils.core.JUnitTest
+import io.customer.datapipelines.testutils.core.testConfiguration
+import io.customer.datapipelines.testutils.extensions.deviceToken
+import io.customer.datapipelines.testutils.extensions.shouldMatchTo
+import io.customer.datapipelines.testutils.utils.OutputReaderPlugin
+import io.customer.datapipelines.testutils.utils.identifyEvents
+import io.customer.datapipelines.testutils.utils.trackEvents
+import io.customer.sdk.core.di.SDKComponent
+import io.customer.sdk.data.store.DeviceStore
+import io.customer.sdk.data.store.GlobalPreferenceStore
+import io.customer.sdk.util.EventNames
+import io.mockk.every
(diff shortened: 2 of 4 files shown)


datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
```diff
@@ -215,9 +216,11 @@ class CustomerIO private constructor(
 
         if (isChangingIdentifiedProfile) {
             logger.info("changing profile from id $currentlyIdentifiedProfile to $userId")
-            if (registeredDeviceToken != null) {
+            if (registeredDeviceToken != null && currentlyIdentifiedProfile != null) {
                 logger.debug("deleting device token before identifying new profile")
-                deleteDeviceToken()
+                deleteDeviceToken { event ->
+                    event?.apply { this.userId = currentlyIdentifiedProfile }
+                }
```
mrehan27, line 223: The test works without this change as well. So this is mainly to add extra confidence. But if we feel it's not necessary, we can discuss and revert it if needed.

mahmoud-elmorabea (commented):
Can you please add steps on how to reproduce the issue and validate it's fixed?

datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
```diff
@@ -319,7 +322,9 @@ class CustomerIO private constructor(
         val existingDeviceToken = contextPlugin.deviceToken
         if (existingDeviceToken != null && existingDeviceToken != token) {
             logger.debug("token has been refreshed, deleting old token to avoid registering same device multiple times")
-            deleteDeviceToken()
+            deleteDeviceToken { event ->
+                event?.putInContextUnderKey("device", "token", existingDeviceToken)
```
mahmoud-elmorabea, line 326: Are we consuming this somewhere? I can only see it being added

datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
mahmoud-elmorabea: I am not sure I understand why this adds more confidence?

datapipelines/src/test/java/io/customer/datapipelines/testutils/core/JUnitTest.kt
```diff
@@ -4,15 +4,20 @@ import com.segment.analytics.kotlin.core.Analytics
...
  * for running unit tests.
  * This class should be used for running unit tests using JUnit.
  */
-abstract class JUnitTest : JUnit5Test() {
-    private val delegate: UnitTestDelegate = UnitTestDelegate(applicationInstance = "Test")
+abstract class JUnitTest(
+    @OptIn(ExperimentalCoroutinesApi::class)
+    dispatcher: TestDispatcher = UnconfinedTestDispatcher()
+) : JUnit5Test() {
+    protected val delegate: UnitTestDelegate = UnitTestDelegate("Test", dispatcher)
```
mahmoud-elmorabea, line 20: Can you shed some light on these generic unit testing changes?

datapipelines/src/test/java/io/customer/datapipelines/DataPipelinesStandardDispatcherTest.kt
```diff
@@ -0,0 +1,175 @@
...
+        // Run all pending coroutines to ensure analytics is initialized and ready to process events
+        @Suppress("OPT_IN_USAGE")
+        testScope.runCurrent()
+    }
+
+    //endregion
+    //region Device token
+
+    @Test
+    fun device_givenTokenRefreshed_expectDeleteAndRegisterDeviceToken() = runTest {
```
mahmoud-elmorabea, line 66: The test names make it very difficult to understand the inital state / action and result. Can we add //given //when //then comments in the test?

datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
mrehan27: Because `analytics` dispatches events in coroutines and immediately returns to the next line in flow. Since we clear user profile right after the delete device event, there's a chance the `userId` might not be attached if `analytics` doesn't wait for the previous events to finish. But so far, it does respect the order so things work as expected.

So in case, the events just before the profile clear aren't respected, the delete device event would still include `userId` with this change, as it has already captured `userId` before the profile was cleared.

datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
mrehan27: This is being added to the final event being sent.

datapipelines/src/test/java/io/customer/datapipelines/testutils/core/JUnitTest.kt
mrehan27: There are no functional changes. This only adds the option to override `TestDispatcher`. The rest of the tests aren't affected by this change.

datapipelines/src/test/java/io/customer/datapipelines/DataPipelinesStandardDispatcherTest.kt
mrehan27: I believe the test name is already descriptive, aligns with naming conventions we currently follow, and clearly explains how a token refresh should result in device delete and create events. The code is also split into smaller blocks to make the given/when conditions easy to follow. That said, if you feel the name doesn't explain enough, please feel free to suggest a new one.

mrehan27:
@mahmoud-elmorabea Updated the PR description with details on how I tested it. Also, the test `device_givenTokenRefreshed_expectDeleteAndRegisterDeviceToken` fails on `main` but passes with updated code in this PR which validates the fix as well.

datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
Shahroz16: Makes sense, but have we checked, if enrichment is happening before or after apply user meta data, for example we can be setting `userId` here but if user meta data is applied later it will overwrite it?

Shahroz16 (approved):
looks good, just added one question to be certain

datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
mrehan27: Yeah, looking at code today, the closure is [applied after Enrichment and before Destination plugins](https://github.com/customerio/cdp-analytics-kotlin/blob/af97bf3e78519da7d9446eb3e9d507c84d3f94e5/core/src/main/java/com/segment/analytics/kotlin/core/platform/Timeline.kt#L29-L36).

PR #518, opened 2025-04-21
chore: simplify in-app base view implementation

mrehan27:
part of: [MBL-1028](https://linear.app/customerio/issue/MBL-1028/create-base-inline-message-view-using-refactored-gistview)

### Changes

- Reverted `InAppMessageBaseView` to `public` to avoid the need for an extra wrapper view, which could introduce challenges with updating size and view attributes like `alpha`, etc. Keeping it `public` simplifies the implementation.
- Moved several methods to child `ModalInAppMessageView` that are not needed for inline in-app messages
- Renamed a few internal methods for clarity and added `@UiThread` annotations to signal that these should not be called from JS, helping minimize thread related exceptions.

### DIFF
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/ui/InAppMessageBaseView.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/ui/InAppMessageBaseView.kt
@@ -7,6 +7,9 @@ import android.net.UrlQuerySanitizer
 import android.util.AttributeSet
 import android.util.Base64
 import android.widget.FrameLayout
+import androidx.annotation.AttrRes
+import androidx.annotation.StyleRes
+import androidx.annotation.UiThread
 import androidx.core.content.ContextCompat.startActivity
 import androidx.core.net.toUri
 import com.google.gson.Gson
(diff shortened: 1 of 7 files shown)


messaginginapp/src/main/java/io/customer/messaginginapp/ui/InAppMessageBaseView.kt
```diff
@@ -27,22 +30,22 @@ import java.nio.charset.StandardCharsets
...
+
     protected var currentMessage: Message? = null
     protected var currentRoute: String? = null
 
     protected val logger = SDKComponent.logger
-    protected val inAppMessagingManager = SDKComponent.inAppMessagingManager
-    protected val store: InAppMessagingState
+    internal val inAppMessagingManager = SDKComponent.inAppMessagingManager
+    internal val store: InAppMessagingState
         get() = inAppMessagingManager.getCurrentState()
```
mahmoud-elmorabea, line 47: If we are changing visibility only for the sake of sub-classes, maybe we can protected instead? Otherwise, I don't think the view should expose its internal state

messaginginapp/src/main/java/io/customer/messaginginapp/ui/InAppMessageBaseView.kt
```diff
@@ -27,22 +30,22 @@ import java.nio.charset.StandardCharsets
  * This class will be extended by specific in-app message views (e.g., ModalInAppMessageView)
  * to handle specific message types.
  */
-internal abstract class InAppMessageBaseView @JvmOverloads constructor(
+abstract class InAppMessageBaseView @JvmOverloads constructor(
```
mahmoud-elmorabea, line 33: Agree with the note of not wrapping this in another view only for visibility, but why does it need to be visible outside of the messaging module?

mahmoud-elmorabea (changes requested):

messaginginapp/src/main/java/io/customer/messaginginapp/ui/InAppMessageBaseView.kt
mrehan27: Unfortunately, `protected` is technically more visible than `internal` for open/abstract classes. Because if a customer inherits the class, they can access those members. So Kotlin doesn't allow exposing `internal` members as `protected` in public open classes for that reason.

Screenshot 2025-04-22 at 4 49 25 PM

messaginginapp/src/main/java/io/customer/messaginginapp/ui/InAppMessageBaseView.kt
mrehan27: Because the final class has to be `public` so that customers can use it in their XML files. And Kotlin doesn't allow `public` classes to inherit from `internal` classes as that would expose internal types to the public. So unfortunately, without adding a wrapper, we have to make `InAppMessageBaseView` `public`.

Screenshot 2025-04-22 at 4 51 53 PM

PR #519, opened 2025-04-21
chore: implement inline in-app base view

mrehan27:
part of: [MBL-1028](https://linear.app/customerio/issue/MBL-1028/create-base-inline-message-view-using-refactored-gistview)

### Changes

- Added lightweight internal view extensions to animate size changes with optional callbacks
- Added `InlineInAppMessageViewListener` to support progress indicators and flexible handling for XML and Compose-based inline views
- Updated `InlineInAppMessageBaseView` to:
  - Show and hide inline in-app messages
  - Trigger callbacks for loading and displaying progress states
  - Update view size with animations for smoother experience

### DIFF
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/ui/InAppMessageViewExtensions.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/ui/InAppMessageViewExtensions.kt
@@ -1,10 +1,79 @@
 package io.customer.messaginginapp.ui
 
+import android.animation.Animator
+import android.animation.AnimatorListenerAdapter
+import android.animation.AnimatorSet
+import android.animation.ValueAnimator
 import android.util.DisplayMetrics
+import android.view.View
+import androidx.annotation.UiThread
+import androidx.core.view.updateLayoutParams
 
 /**
  * Converts the given size from dp to pixels based on the device's screen density.
  */
 internal fun InAppMessageBaseView.dpToPixels(size: Double): Int {
     return size.toInt() * context.resources.displayMetrics.densityDpi / DisplayMetrics.DENSITY_DEFAULT
+}
+
+/**
+ * Updates the layout parameters of the view with the specified width and height with animation
+ * and optional callbacks for start and end events.
+ */
+@UiThread
+internal fun View.animateViewSize(
+    widthInPx: Int? = null,
+    heightInPx: Int? = null,
+    duration: Long = resources.getInteger(android.R.integer.config_longAnimTime).toLong(),
+    onStart: (() -> Unit)? = null,
+    onEnd: (() -> Unit)? = null
+) {
+    if (widthInPx == null && heightInPx == null) {
+        onStart?.invoke()
+        onEnd?.invoke()
+        return
+    }
+
+    val animators = mutableListOf<Animator>()
+
+    widthInPx?.let { targetWidth ->
+        val animator = ValueAnimator.ofInt(width, targetWidth).apply {
+            this.duration = duration
+            addUpdateListener { update ->
+                updateLayoutParams { width = update.animatedValue as Int }
+            }
+        }
+        animators.add(animator)
+    }
+
+    heightInPx?.let { targetHeight ->
+        val animator = ValueAnimator.ofInt(height, targetHeight).apply {
+            this.duration = duration
+            addUpdateListener { update ->
+                updateLayoutParams { height = update.animatedValue as Int }
+            }
+        }
+        animators.add(animator)
+    }
+
+    AnimatorSet().apply {
+        playTogether(animators)
+        this.duration = duration
+        addListener(object : AnimatorListenerAdapter() {
+            override fun onAnimationStart(animation: Animator) {
+                super.onAnimationStart(animation)
+                onStart?.invoke()
+            }
+
+            override fun onAnimationEnd(animation: Animator) {
+                super.onAnimationEnd(animation)
+                updateLayoutParams {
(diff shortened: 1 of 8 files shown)


messaginginapp/src/main/java/io/customer/messaginginapp/ui/InlineInAppMessageBaseView.kt
```diff
@@ -0,0 +1,175 @@
...
+
+    var elementId: String? = null
+
+    init {
+        visibility = GONE
+        subscribeToStore()
+    }
+
+    private fun subscribeToStore() {
+        inAppMessagingManager.subscribeToState(
```
mahmoud-elmorabea, line 35: Do we not want to extract this to its own class that we can unit test?

messaginginapp/src/main/java/io/customer/messaginginapp/ui/InlineInAppMessageBaseView.kt
```diff
@@ -0,0 +1,175 @@
...
+
+        when (inlineMessageState) {
+            is InlineMessageState.ReadyToEmbed -> embedMessage(state = inlineMessageState)
+            is InlineMessageState.Dismissed -> dismissMessage(message = inlineMessageState.message) {}
+            is InlineMessageState.Embedded -> {
+                // We are already displaying or rendering a message. Do not show another message until
+                // the current message is closed. The main reason for this is when a message is tracked
+                // as "opened", the Gist backend will not return this message on the next fetch call.
+                // We want to continue showing a message even if the fetch no longer returns the message
+                // and the message is currently visible.
```
mahmoud-elmorabea, line 61: Do we want to add a log here to simplify debugging?

mrehan27: I don't think we really need logs here, [iOS doesn't have them either](https://github.com/customerio/customerio-ios/blob/78f11d6d3fc6cc6a4870dfc6b8e60f12d21d6d67/Sources/MessagingInApp/Views/GistInlineInAppMessageView.swift#L171). But I'll revisit and see if adding some logs here help with debugging, I'll add some later.

messaginginapp/src/main/java/io/customer/messaginginapp/ui/InlineInAppMessageBaseView.kt
mahmoud-elmorabea: I would that the logic being simple is a stronger case for writing tests for it, it should be easy 😇 

Tests help very well with adding and changing code, so when more is added or current is changed, it will be good to already have the tests not to start adding them then.

messaginginapp/src/main/java/io/customer/messaginginapp/ui/InlineInAppMessageBaseView.kt
mahmoud-elmorabea: It's just that when this event happens, this will simply be swallowed, no UI change and no log. If I am not seeing a message for some reason (maybe a bug) and this code path is executed, I have no way of knowing other than attaching a debugger.

> [iOS doesn't have them either](https://github.com/customerio/customerio-ios/blob/78f11d6d3fc6cc6a4870dfc6b8e60f12d21d6d67/Sources/MessagingInApp/Views/GistInlineInAppMessageView.swift#L171).

I think we should add it there as well 😄

messaginginapp/src/main/java/io/customer/messaginginapp/ui/InlineInAppMessageBaseView.kt
mrehan27: No disagreement here. I'm first focusing on wrapping up the view and making sure we've moved the right code into base, modal, and inline in-app views. Once that's done, we'll work on adding tests. So tests will come in upcoming PRs, the next few will be more focused on balancing the views though, and then we'll shift toward adding tests for them.

messaginginapp/src/main/java/io/customer/messaginginapp/ui/InlineInAppMessageBaseView.kt
mrehan27: This is just a subscriber. Every store update is already logged as shown below, so it shouldn't be swallowed and should already be visible in logs. That's why I think it isn't needed here. But if you still think it would help, let me know and I can include it in upcoming PRs.

```
embeddedMessagesState: EmbeddedMessagesState(messagesByElementId: ["inline": ReadyToEmbed(message: id: {message-id}, queueId: {queue-id}, deliveryId: none, elementId: inline)]),
```

Also to clarify, these logs are from iOS. Android currently isn't printing `InlineMessageState` inside `embeddedMessagesState`. But I've noted that and plan to fix it in store instead in upcoming PRs to align with iOS and provide more clarity.

messaginginapp/src/main/java/io/customer/messaginginapp/ui/InAppMessageViewExtensions.kt
```diff
@@ -1,10 +1,79 @@
...
+                updateLayoutParams {
+                    widthInPx?.let { width = it }
+                    heightInPx?.let { height = it }
+                }
+                onEnd?.invoke()
+            }
+        })
+        start()
+    }
+}
```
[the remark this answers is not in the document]
mrehan27: I think some of the suggested improvements weren't necessary, so I filtered them out and kept only the updates that I felt were adding improvements.

PR #520, opened 2025-04-22
chore: add public XML inline in-app message view

mrehan27:
closes: [MBL-1029](https://linear.app/customerio/issue/MBL-1029/create-and-expose-dedicated-inline-message-view-for-android-xml)

### Changes

- Added `InlineInAppMessageView`, the public facing view for embedding inline in-app messages in customer apps
- Implemented `InlineInAppMessageView` with:
  - Ability to set `elementId` via XML or programmatically in Kotlin/Java
  - Support for displaying `ProgressBar` as loading indicator
- Updated sample app to use the newly added `InlineInAppMessageView`

### Sample Usage

```XML
<io.customer.messaginginapp.ui.InlineInAppMessageView
    android:id="@+id/inline_in_app_message"
    android:layout_width="match_parent"
    android:layout_height="wrap_content"
    app:elementId="banner-message"/>
```

#### `elementId`

Identifier used to load the corresponding inline in-app message.

#### `progressTint` (Bonus)

Tint color for the progress bar. Defaults to `colorControlActivated` from theme with fallback to `Color.Gray`.

### DIFF
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/ui/InlineInAppMessageView.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/ui/InlineInAppMessageView.kt
@@ -36,12 +36,17 @@ class InlineInAppMessageView @JvmOverloads constructor(
 
     init {
         var progressColor = resolveThemeColor(android.R.attr.colorControlActivated, Color.GRAY)
-        context.obtainStyledAttributes(attrs, R.styleable.InlineInAppMessageView).use { ta ->
-            elementId = ta.getString(R.styleable.InlineInAppMessageView_elementId)
-            if (ta.hasValue(R.styleable.InlineInAppMessageView_progressTint)) {
-                progressColor =
-                    ta.getColor(R.styleable.InlineInAppMessageView_progressTint, progressColor)
+        val typedArray = context.obtainStyledAttributes(attrs, R.styleable.InlineInAppMessageView)
+        try {
+            elementId = typedArray.getString(R.styleable.InlineInAppMessageView_elementId)
+            if (typedArray.hasValue(R.styleable.InlineInAppMessageView_progressTint)) {
+                progressColor = typedArray.getColor(
+                    R.styleable.InlineInAppMessageView_progressTint,
+                    progressColor
+                )
             }
+        } finally {
+            typedArray.recycle()
         }
         setupProgressIndicator(progressColor)
         viewListener = this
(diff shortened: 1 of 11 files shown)


messaginginapp/src/main/java/io/customer/messaginginapp/ui/InlineInAppMessageView.kt
```diff
@@ -0,0 +1,89 @@
...
+                progressColor = typedArray.getColor(
+                    R.styleable.InlineInAppMessageView_progressTint,
+                    progressColor
+                )
+            }
+        } finally {
+            typedArray.recycle()
+        }
+        setupProgressIndicator(progressColor)
+        viewListener = this
```
Shahroz16, line 52:
```suggestion
 context.obtainStyledAttributes(attrs, R.styleable.InlineInAppMessageView).use { typedArray -> 
            elementId = typedArray.getString(R.styleable.InlineInAppMessageView_elementId)
            
            val defaultProgressColor = resolveThemeColor(android.R.attr.colorControlActivated, Color.GRAY)
            val progressColor = if (typedArray.hasValue(R.styleable.InlineInAppMessageView_progressTint)) {
                typedArray.getColor(R.styleable.InlineInAppMessageView_progressTint, defaultProgressColor)
            } else {
                defaultProgressColor
            }
            
            setProgressColor(progressColor)   
        }
```

Can we use `use` extension which automatically handles resource cleanup?

messaginginapp/src/main/java/io/customer/messaginginapp/ui/InlineInAppMessageView.kt
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    private fun setupProgressIndicator(@ColorInt progressColor: Int) {
+        progressIndicator.isIndeterminate = true
+        progressIndicator.layoutParams = LayoutParams(
+            LayoutParams.WRAP_CONTENT,
+            LayoutParams.WRAP_CONTENT,
+            Gravity.CENTER
+        )
+        runCatching {
```
Shahroz16, line 62: do we need this? can this be simplified?

Shahroz16 (approved):

messaginginapp/src/main/java/io/customer/messaginginapp/ui/InlineInAppMessageView.kt
mrehan27: I actually [used `use` earlier](https://github.com/customerio/customerio-android/pull/520/commits/2110db6df8c7beb1646315049ff496172feb9bff), but [lint kept failing](https://github.com/customerio/customerio-android/actions/runs/14595042785/job/40938814762) so reverted to a simpler approach.

messaginginapp/src/main/java/io/customer/messaginginapp/ui/InlineInAppMessageView.kt
mrehan27: Looking at Android docs, we don't really need `runCatching` here. But added it just to be safe when dealing with drawables since it doesn't do much harm even if it fails 😅

PR #521, opened 2025-04-23
fix: use SDK instead of analytics for tracking in plugin

Shahroz16:
Since we have made `CustomerIO` instance methods as the only source of tracking and not rely on plugins for updates, we need to move away from tracking via `analytics` in the plugins.


Shahroz16:
> Change looks good 💯
> 
> Good catch changing those usages to use the synchronized implementation in our SDK.
> 
> I have a question, that's not directly related to this PR and we don't have to deal with it here but although our SDK has synchronization on its own calls (which maybe helps with threading issues) but the underlying calls to Segment's `Analytics` still post these calls to different threads.
> 
> So ultimately those calls can still go out of order because the actual processing happens on different threads with no order guarantees, am i getting this right?

@mahmoud-elmorabea yes, that's correct, we have synchronized our internal calls to respect order. The processing of the calls is still not ordered. From my testing only the events tracked on same milliseconds were at times out of order in analytics but other than that all were pretty much in order even without synchronization.

mrehan27 (approved):

PR #522, opened 2025-04-23
chore: clean up EngineWebView and related improvements

mrehan27:
part of: [MBL-989](https://linear.app/customerio/issue/MBL-989/showhide-the-xml-based-view)

### Changes

- Moved `instanceId` in `Message` class from constructor to a property to avoid it affecting equality checks and triggering unnecessary store events
- Added `releaseResources` method to `EngineWebView` to properly clean up `WebView` after use
- Updated `InAppMessageState` to print as expected by overriding `toString` directly as data classes were previously overriding it implicitly
- Updated `InlineInAppMessageBaseView` to dispatch `DisplayMessage` and clean up `EngineWebView` when it is no longer needed

### DIFF
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/model/Message.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/model/Message.kt
@@ -25,6 +25,8 @@ data class Message(
     val queueId: String? = null,
     val properties: Map<String, Any?>? = null
 ) {
+    // Should be property and not constructor parameter so it isn't used in equals
+    // As messages are identified uniquely by their queueId and not instanceId
     val instanceId: String = UUID.randomUUID().toString()
     val gistProperties: GistProperties
         get() = convertToGistProperties()
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/model/Message.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/model/Message.kt
@@ -21,11 +21,11 @@ data class GistProperties(
 
 data class Message(
     val messageId: String = "",
-    val instanceId: String = UUID.randomUUID().toString(),
     val priority: Int? = null,
     val queueId: String? = null,
     val properties: Map<String, Any?>? = null
 ) {
+    val instanceId: String = UUID.randomUUID().toString()
     val gistProperties: GistProperties
         get() = convertToGistProperties()
 
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessagingState.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessagingState.kt
@@ -63,47 +63,33 @@ internal sealed class InlineMessageState {
     abstract val message: Message
 
     data class ReadyToEmbed(override val message: Message, val elementId: String) : InlineMessageState() {
-        override fun toString(): String {
-            return "ReadyToEmbed(message=${message.queueId}, elementId=$elementId)"
-        }
+        override fun toString() = "ReadyToEmbed(message=${message.queueId}, elementId=$elementId)"
     }
 
     data class Embedded(override val message: Message, val elementId: String) : InlineMessageState() {
-        override fun toString(): String {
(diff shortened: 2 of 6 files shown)


messaginginapp/src/test/java/io/customer/messaginginapp/testutils/extension/GistExtensions.kt
```diff
@@ -16,17 +14,6 @@ fun pageRuleContains(route: String): String = "^(.*$route.*)\$"
 
 fun pageRuleEquals(route: String): String = "^($route)\$"
 
-fun decodeOptionsString(options: String): Message {
```
mrehan27, line 19: This isn't used anywhere so instead of fixing `instanceId` here, I simply removed it

messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/model/Message.kt
```diff
@@ -21,11 +21,11 @@ data class GistProperties(
 
 data class Message(
     val messageId: String = "",
-    val instanceId: String = UUID.randomUUID().toString(),
     val priority: Int? = null,
     val queueId: String? = null,
     val properties: Map<String, Any?>? = null
 ) {
+    val instanceId: String = UUID.randomUUID().toString()
```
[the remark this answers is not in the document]
mrehan27: Added

messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessagingState.kt
```diff
@@ -62,29 +62,48 @@ internal fun InAppMessagingState.withUpdatedEmbeddedMessage(
...
-    data class Dismissed(override val message: Message) : InlineMessageState()
-
-    override fun toString(): String = when (this) {
-        is ReadyToEmbed -> "ReadyToEmbed(message=${message.queueId}, elementId=$elementId)"
-        is Embedded -> "Embedded(message=${message.queueId}, elementId=$elementId)"
-        is Dismissed -> "Dismissed(message=${message.queueId})"
+    data class ReadyToEmbed(override val message: Message, val elementId: String) : InlineMessageState() {
+        override fun toString(): String {
+            return "ReadyToEmbed(message=${message.queueId}, elementId=$elementId)"
+        }
```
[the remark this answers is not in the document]
mrehan27: Updated all toString methods

messaginginapp/src/main/java/io/customer/messaginginapp/gist/data/model/Message.kt
```diff
@@ -21,11 +21,13 @@ data class GistProperties(
...
 data class Message(
     val messageId: String = "",
-    val instanceId: String = UUID.randomUUID().toString(),
     val priority: Int? = null,
     val queueId: String? = null,
     val properties: Map<String, Any?>? = null
 ) {
+    // Should be property and not constructor parameter so it isn't used in equals
+    // As messages are identified uniquely by their queueId and not instanceId
+    val instanceId: String = UUID.randomUUID().toString()
```
mahmoud-elmorabea, line 30: I'm curious as to how is this different, don't both get assigned at relatively the same time when the object is created? or do we want to ensure this is unique and not passed same values from constructor?

mrehan27: We just want to ensure it isn't used in `equals` when comparing messages. Kotlin by default doesn't include [properties declared in class body for `equals`](https://kotlinlang.org/docs/data-classes.html#properties-declared-in-the-class-body) which helps us get the expected comparison behavior so messages aren't considered different just because `instanceId` is different.

mrehan27: Yeah, that's also an option. But in that case, we might also want to consider moving away from data class here altogether. I'll take another look and update in later PRs if needed.

mahmoud-elmorabea: Yea, even a better idea @mrehan27 ! Since we are kinda growing out of the regular usage of a data class, having it as data class with explicit equals might be more confusing.

PR #523, opened 2025-04-24
chore: Update Logger interface to align with unified design across platforms

mahmoud-elmorabea:
Closes: [MBL-1073](https://linear.app/customerio/issue/MBL-1073/implement-unified-logging-interface-on-android)

## Overview
This PR aligns the Customer IO Android SDK `Logger` interface to the [unified interfaces definitions](https://www.notion.so/custio/Mobile-SDK-unified-logging-interfaces-1dc4302f4c2b80d08bacd478ca8bbf7e).

The main change in the interface is that logging functions now include a `tag` that will allow us to easily filter logs in the future by module.

Some notable changes in this PR:
- I made sure that the same `Logger` interface was used to avoid having two versions and go through deprecation process
- I've intentionally made `tag` parameter optional for now so we don't have to change any existing code that uses the existing `Logger` interface
- The `Logger Dispatcher` closure was not updated to include the `tag` parameter, I will update it when I am working on updating the wrappers
- I've chosen to add all new parameters to the end of functions like `Logger#debug()`, so that we don't have to change any code of all modules using the interface

## A more testable logger
I've changed how the `Logger` is initialized:
- `CioLogLevel` is now a required parameter and its usage is an implementation details
- I've wrapped calls to Android's `Log` class in a `LogcatLogger` so we can test the `Logger` message format
- Added more tests that rely solely on the behavior of the logger rather than internal state

## Test plan
I've ran the app both from this branch and the main branch with debug logging enabled and compared the logs between both versions and all log points were there before and after the change.

- You can run the same test yourself and compare with any diff tool
- I am attaching two files with cleaned up logs if you want to compare them without running app on multiple branches

[after.txt](https://github.com/user-attachments/files/19893013/after.txt)
[before.txt](https://github.com/user-attachments/files/19893015/before.txt)


core/src/main/kotlin/io/customer/sdk/core/util/LogcatLogger.kt
```diff
@@ -0,0 +1,24 @@
+package io.customer.sdk.core.util
+
+import android.util.Log
+
+/**
+ * Convenience wrapper around Android's [Log] class to avoid unit testing static calls
+ */
+internal class LogcatLogger {
+    fun i(tag: String, msg: String) {
```
uros-mil, line 9: It would be better if our methods had a full name - `info` instead of `i`.

core/src/main/kotlin/io/customer/sdk/core/util/CioLogLevel.kt
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Log levels for Customer.io SDK logs
+ *
+ * @property priority Priority of the log level. The higher the value, the more verbose
+ * the log level.
+ */
+enum class CioLogLevel(val priority: Int) {
+    NONE(priority = 0),
+    ERROR(priority = 1),
```
uros-mil, line 11: What is the reason for not having WARN between INFO and ERROR?

core/src/main/kotlin/io/customer/sdk/core/util/Logger.kt
```diff
@@ -21,98 +14,7 @@ interface Logger {
...
-                CioLogLevel.INFO -> Log.i(TAG, message)
-                CioLogLevel.DEBUG -> Log.d(TAG, message)
-            }
-        }
-    }
-
-    companion object {
-        const val TAG = "[CIO]"
-    }
+    fun info(message: String, tag: String? = null)
```
uros-mil, line 17: I'm not sure if `tag` is a good idea.
My view of usage:
- logger created in root and provided to all modules/submodules
- Having some module define `tag` in all calls is prone to errors
- It produces duplication if every log in the In-App module has to have the `In-App` tag

Alternative:
- Remove tag from log calls
- Create a wrapper/decorator implementation that accepts a Logger and a tag to append
```
class WrapperLogger: Logger { 
    init(logger: Logger, tag: String) { ... }
... 
}
```

This way, each module could accept the logger, wrap it inside its LoggerWrapper with its tag, and forward further to other classes under it.
So In-App module will accept a logger with the default `[CIO]` tag, will add its `[In-App]` tag, and all components under it will use the logger. 
This will automatically produce `[CIO][In-App]` as a prefix.

This also allows sections under the module to further wrap/decorate Logger - `[CIO][In-App][Gist]`

uros-mil (changes requested):
Let's discuss questions on the next `Mobile Squad Comms Window`, e.g., on Monday.

core/src/main/kotlin/io/customer/sdk/core/util/LogcatLogger.kt
mahmoud-elmorabea: You are absolutely right, it's a bad habit we inherited from the Android platform where the default logger class uses letters instead of words, changed to align with the other logger!

common-test/src/main/java/io/customer/commontest/util/UnitTestLogger.kt
```diff
@@ -9,25 +9,24 @@ import io.customer.sdk.core.util.Logger
  * the log level set to help debug issues while running unit tests.
  */
 class UnitTestLogger : Logger {
```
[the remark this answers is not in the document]
mahmoud-elmorabea: The main idea is that our logger uses Android's `Log` class which has two challenges in unit tests:
- All its calls are static which doesn't play well in unit tests since we can't easily mock them (need special extra testing dependency for that)
- It only exists when running code on Android devices and not on JVMs (local machines and CI servers), so tests fail because it can't use that class

So the general practice is to replace it with a no-op logger or something like this to avoid tests failing

core/src/main/kotlin/io/customer/sdk/core/util/Logger.kt
mahmoud-elmorabea: That is exactly the plan! 

So in next PRs where I implement push notification logs, I will create `PushMessagingLogger` in the push messaging module that wraps our logger and passed the tag, this:
- Unifies all calls to send the tag once, instead of passing it from all call sites in the module
- Allows us to do things like `PushMessagingLogger.logPushMessageRecieved(Message)` in call site and encapsulate all log formatting inside the module logger

Let me know if that's what you had in mind!

mrehan27 (commented):
Since this is going to `main`, can you please also verify if the changes work fine with wrappers? Or will this be verified in wrappers ticket instead?

I'm expecting there won't be many changes needed for wrappers. So even if we have to release a small fix from native to wrappers, it shouldn't be blocked by wrapper work for these logger updates. Just wanted to confirm.

core/src/main/kotlin/io/customer/sdk/core/environment/BuildEnvironment.kt
Shahroz16, line 1: Previously, the fallback was, 

```
private val fallbackLogLevel
        get() = when {
            buildEnvironment.debugModeEnabled -> CioLogLevel.DEBUG
            else -> CioLogLevel.ERROR
        }
```

Now the fallback is always going to be `error`, regardless. This forces the developer to implement this in their code, where they would set debug options based on flavor.  

I kind of liked the flexibility we had, not totally opposed this idea either but wanted to raise it

core/src/main/kotlin/io/customer/sdk/core/di/SDKComponent.kt
```diff
@@ -40,10 +40,16 @@ object SDKComponent : DiGraph() {
...
-        get() = newInstance<BuildEnvironment> { DefaultBuildEnvironment() }
+    private var initializedLogger: Logger? = null
     val logger: Logger
-        get() = singleton<Logger> { LogcatLogger(buildEnvironment = buildEnvironment) }
+        get() {
+            val loggerVar = initializedLogger
+            return loggerVar ?: newInstance<Logger> { NoOpLogger }
+        }
+
+    fun initializeLogger(level: CioLogLevel) {
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Good suggestion! Will do that.

core/src/main/kotlin/io/customer/sdk/core/util/NoOpLogger.kt
```diff
@@ -0,0 +1,39 @@
...
+import java.lang.IllegalStateException
+
+/**
+ * A fallback in case the SDK failed to initialize [Logger] instance before any logging calls happen.
+ *
+ * This allows us to expose the logger as non-null type without assuming any default logging levels.
+ *
+ * If any logs of this class are seen, that means there is a bug with SDK initialization logic for the [Logger]
+ */
+internal object NoOpLogger : Logger {
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Good call! 

Falling back to fallback logger that uses the same implementation as `LoggerImpl` but prints an error log when it's used so we can catch this case if it happens.

datapipelines/src/main/kotlin/io/customer/datapipelines/config/DataPipelinesModuleConfig.kt
```diff
@@ -29,7 +30,9 @@ class DataPipelinesModuleConfig(
     // Configuration options required for migration from earlier versions
     val migrationSiteId: String? = null,
     // Determines how SDK should handle screen view events
-    val screenViewUse: ScreenView
+    val screenViewUse: ScreenView,
+    // SDK logging level passed through from core SDK config
+    val logLevel: CioLogLevel
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Not sure what SDK config means? The Datapipelines uses this config internally, and it was pulling it from the logger itself, I just passed it down through config instead because I think this is more appropriate

tracking-migration/src/test/java/io/customer/tracking/migration/MigrationAssistantTest.kt
```diff
@@ -38,6 +40,7 @@ class MigrationAssistantTest : JUnitTest() {
                         overrideDependency<SitePreferenceRepository>(mockk())
                         // Use relaxed unit mock instead of full relaxed mocks to minimize false positives
                         overrideDependency<Queue>(mockk(relaxUnitFun = true))
+                        overrideDependency<Logger>(UnitTestLogger())
```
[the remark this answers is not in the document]
mahmoud-elmorabea: I applied you suggestion and this is no longer needed.

core/src/main/kotlin/io/customer/sdk/core/environment/BuildEnvironment.kt
mahmoud-elmorabea: I don't think we should under the hood be pulling this info and using it, we expose a clearly defined config for log level, if customers want to control it, they should use the config.

mahmoud-elmorabea:
Once we discuss the overall approach and no more major changes will happen, I will test this with the wrappers and update the test plan!

datapipelines/src/main/kotlin/io/customer/datapipelines/config/DataPipelinesModuleConfig.kt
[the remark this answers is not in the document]
mrehan27: The logger itself is in core module and isn't tied to Data Pipelines module only. We currently initialize it through Data Pipelines since that's our only SDK entry point for now, so I'm okay with storing it here as well. But ideally, if we need to store `logLevel`, I think it would be better to keep it at same level as the logger itself (somewhere in core module like shared `SdkConfig` or something similar).

Also, if we're not using `logLevel` elsewhere, do we still need to store it? The logger is already initialized with desired value so it might not be needed anymore?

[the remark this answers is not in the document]
mahmoud-elmorabea: Well, the initialization and the builder live in the Datapipelines module (that's just how things are), they might not be related to Datapipelines functionality, but they live there.

You can see [here](https://github.com/customerio/customerio-android/pull/523/files) we use this value to set Segments loglevel, at this point the available config is data pipelines config, that's why I added it there. 

I don't want the logger to expose its internal state, those values should be grabbed from config, is there anyway you can suggest to initialize Segment's log level from the core SDK config?

Shahroz16 (commented):
> Once we discuss the overall approach and no more major changes will happen, I will test this with the wrappers and update the test plan!

thats the current state of this PR

core/src/main/kotlin/io/customer/sdk/core/util/FallbackLogger.kt
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * A fallback in case the SDK failed to initialize [Logger] instance before any logging calls happen.
+ *
+ * This allows us to expose the logger as non-null type without assuming any default logging levels.
+ *
+ * If any logs of this class are seen, that means there is a bug with SDK initialization logic for the [Logger]
+ */
+internal class FallbackLogger : LoggerImpl(SDKComponent.DEFAULT_LOG_LEVEL) {
+    init {
+        error(
```
Shahroz16, line 14: this will log error of not initialized for wrappers, a new error that the CS might not be aware of.

Shahroz16 (approved):
Seems okay, given we have finalized the following 
- `warn` is needed or not
- Wrappers are tested, we extend these logger in wrappers and we want to ensure they don't break. 
- SDK and Wrappers are able to log even before the SDK is initialized 
- If we have new logs/errors added, we let them know to CS teams about what they mean

mrehan27 (commented):
Can we please fix validation tasks? Looks like API files need to be updated. Rest looks good.

mrehan27 (approved):
Looks good

PR #525, opened 2025-04-27
chore: fetch tags before setting the version

Shahroz16:
in the last automatic publishing or sample app release, I noticed the wrong version is being used and it seems its because Git would only have access to the tags that were included in the initial checkout, which could be incomplete or outdated.

### DIFF
--- a/.github/workflows/reusable_build_sample_apps.yml
+++ b/.github/workflows/reusable_build_sample_apps.yml
@@ -95,6 +95,8 @@ jobs:
           echo "BRANCH_NAME=${{ github.head_ref || github.ref_name }}" >> $GITHUB_ENV
           COMMIT_HASH="${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}"
           echo "COMMIT_HASH=${COMMIT_HASH:0:7}" >> $GITHUB_ENV
+          # Get the latest tag, making sure we get all tags with --tags flag (in case tag was just created)
+          git fetch --tags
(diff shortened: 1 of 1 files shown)


.github/workflows/reusable_build_sample_apps.yml
```diff
@@ -95,6 +95,8 @@ jobs:
           echo "BRANCH_NAME=${{ github.head_ref || github.ref_name }}" >> $GITHUB_ENV
           COMMIT_HASH="${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}"
           echo "COMMIT_HASH=${COMMIT_HASH:0:7}" >> $GITHUB_ENV
+          # Get the latest tag, making sure we get all tags with --tags flag (in case tag was just created)
+          git fetch --tags
```
mahmoud-elmorabea, line 99: We could also do something like [this](https://github.com/customerio/customerio-expo-plugin/blob/main/.github/workflows/reusable_build_sample_apps.yml#L20-L23) that we have on other repositories.

But if this works, I'm okay with that as well

mahmoud-elmorabea (approved):

.github/workflows/reusable_build_sample_apps.yml
Shahroz16: We do seem have these [lines](https://github.com/customerio/customerio-android/blob/b359d190221f4893e211ec460a46a7fcbd8be3f1/.github/workflows/reusable_build_sample_apps.yml#L38-L42) in the workflow. Think its too early in the step maybe, thats the cached value is being returned

PR #526, opened 2025-04-28
chore: clean up WebView on dismiss and improve inline in-app handling

mrehan27:
part of: [MBL-1088](https://linear.app/customerio/issue/MBL-1088/persist-inline-in-app-message-on-orientation-change)

### Changes

- Updated `releaseResources` in `EngineWebView` to call `stopLoading` and `destroy` on `WebView` for more thorough cleanup
- Updated `InlineInAppMessageBaseView` to:
  - Destroy the view only on `onDetachedFromWindow`, and prevent destruction on temporary detachments like configuration changes
  - Automatically re-show message if the view was destroyed and recreated after a config change or temporary detach
- Improved naming, added code comments, cleaned up code structure, and added extensions to simplify logic
- Updated sample app to cover Kotlin property usage for inline in-app view

#### Background and Foreground Handling

1. Open Inline Examples screen
2. Send and wait for an inline message to appear
3. Put the app in the background
4. Bring it back to the foreground
5. Inline message should remain visible

#### Screen Rotation Handling

1. Open the Inline Examples screen
2. Send and wait for an inline message to appear
3. Rotate the screen (change device orientation)
4. Inline message should be displayed again

### DIFF
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/ui/InAppMessageViewExtensions.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/ui/InAppMessageViewExtensions.kt
@@ -4,6 +4,9 @@ import android.animation.Animator
 import android.animation.AnimatorListenerAdapter
 import android.animation.AnimatorSet
 import android.animation.ValueAnimator
+import android.app.Activity
+import android.content.Context
+import android.content.ContextWrapper
 import android.util.DisplayMetrics
 import android.util.TypedValue
 import android.view.View
@@ -11,6 +14,7 @@ import androidx.annotation.AttrRes
 import androidx.annotation.ColorInt
 import androidx.annotation.UiThread
 import androidx.core.view.updateLayoutParams
+import io.customer.sdk.core.di.SDKComponent
 
 /**
  * Converts the given size from dp to pixels based on the device's screen density.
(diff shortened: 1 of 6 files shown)


messaginginapp/src/main/java/io/customer/messaginginapp/ui/InAppMessageViewExtensions.kt
```diff
@@ -99,3 +103,39 @@ internal fun InAppMessageBaseView.resolveThemeColor(
...
+    } ?: true
+}
+
+/**
+ * Safely finds the Activity from a Context by unwrapping ContextWrappers.
+ * Limits depth to prevent infinite loops in case of badly implemented ContextWrapper
+ * chains.
+ * Default max depth is only 5 because we don't expect more than 3 wrappers in a normal case.
+ */
+internal fun Context.findActivity(maxDepth: Int = 5): Activity? {
```
mahmoud-elmorabea, line 124: I've tested both scenarios and this seems to work. I am not sure how robust this would be but at least it's capped at 5 iterations so even if this is broken for some corner case, it will just disable this behavior.

Have we considered just using `onSaveInstanceState` and `onRestoreInstanceState`? and we can just cleanup and detach webview if we are dismissing the view explicitly or another message is going to be displayed. But for the normal case of detaching, it would be automatically detached, right?

mahmoud-elmorabea (approved):
Looks good.

Tested both scenarios and they're working fine. Left one suggestion.

messaginginapp/src/main/java/io/customer/messaginginapp/ui/InAppMessageViewExtensions.kt
mrehan27: > if this is broken for some corner case, it will just disable this behavior

Agreed. But I believe breaking this is better than risking an infinite loop and causing ANRs. I don't expect it to go deeper than 2–3 levels, still capping it at 5 just to be safe. At the moment, this seems like the only reliable way to support Activity, Fragments, and Compose together.

> Have we considered just using `onSaveInstanceState` and `onRestoreInstanceState`

I actually tried implementing that first, but it didn't behave as expected when Fragments and Views were fully recreated. If customer apps don't carefully handle fragment state restoration, those saved states might not work reliably.

This solution has been working well for all use cases I tested so far, and saved state logic started to look redundant after this so I removed it. I still have that version in a local branch in case we need to revisit it 😅 But as of now, I haven't found a case where this current approach fails and saved state would help. So I'm going with this for now. That said, if we find a scenario where this breaks down, we should absolutely revisit state handling for better reliability.

> But for the normal case of detaching, it would be automatically detached, right?

The behavior for normal (non-persistent) messages is that they get dismissed on back button press but not when the screen is rotated or the app goes to background. I've added this case in PR description as well.

PR #527, opened 2025-04-28
chore: fix binary size measurement action

Shahroz16:
changes: 
- Updated dependencies to ensure no dependency is pulling `kotlin-stdlib-jdk7` and we only relying on `kotlin-stdlib:1.8.20`
- Due to that, I had to update changes that came with those dependency updates, for example `dokka`
- Updated kotlin sample app to make sure it keeps compiling. 
- Updated GitHub action for gradle setup

**The action is still failing because it's trying to run the same thing for the `main` branch, which still has this issue. Once this PR gets merged, this issue should go away.**


scripts/check-sdk-size.sh
Shahroz16, line 1: created this script to test locally, think its a useful addition

base/build.gradle
```diff
@@ -12,15 +13,29 @@ ext {
...
+    namespace 'io.customer.base'
+    defaultConfig {
+        consumerProguardFiles "consumer-rules.pro"
+    }
+    buildFeatures {
+        buildConfig = false
+    }
+    buildTypes {
+        release {
+            minifyEnabled false
```
mahmoud-elmorabea, line 30: I can see this aligns with the rest of the modules, but do we want to NOT mnifiy modules built with release?

.github/actions/setup-android/action.yml
```diff
@@ -1,20 +1,22 @@
...
+    - name: Install Android SDK
+      uses: android-actions/setup-android@9fc6c4e9069bf8d3d10b2204b1fb8f6ef7065407 # v3.2.2
+
+    - name: Verify Gradle scripts are valid
+      uses: gradle/actions/wrapper-validation@af1da67850ed9a4cedd57bfd976089dd991e2582 # v4.3.1
+
+    - name: Setup Gradle and cache dependencies
+      uses: gradle/actions/setup-gradle@af1da67850ed9a4cedd57bfd976089dd991e2582 # v4.3.1
+      with:
+        gradle-version: '8.14'
```
mahmoud-elmorabea, line 22: Could this be implicitly fixing the kotlin version mismatch we had before that was causing the issue?

mahmoud-elmorabea (changes requested):
I think the change does fix the problem which is definitely a good step forward but I'm not sure if actually addresses the root issue.

I see that by pulling any of the latest failures [example](https://github.com/customerio/customerio-android/actions/runs/14729114037/job/41338836943), I see that there is a kotlin version conflict, I can verify this by running `./gradlew <anyModule>:dependencies`

.github/actions/setup-android/action.yml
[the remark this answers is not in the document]
Shahroz16: good catch, even though they are seem to be pointing at `4.3.1` commits hash are different

Shahroz16: nope tried it, didnt fix it

base/build.gradle
```diff
@@ -2,8 +2,9 @@ import io.customer.android.Configurations
 import io.customer.android.Dependencies
 
 plugins {
-    id 'java-library'
-    id 'kotlin'
+    id 'com.android.library'
+    id 'kotlin-android'
+    id 'com.twilio.apkscale'
```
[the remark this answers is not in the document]
Shahroz16: think this was made java library in data pipelines release

base/build.gradle
Shahroz16: it usually does not make sense to enable minifyEnabled in a public SDK that’s meant to be consumed by other developers as we need the APIs to be readable additionally, Minifying your SDK won’t significantly reduce the consumer app size, especially since most consumer builds already use R8 on their own.

Shahroz16:
> I think the change does fix the problem which is definitely a good step forward but I'm not sure if actually addresses the root issue.
> 
> I see that by pulling any of the latest failures [example](https://github.com/customerio/customerio-android/actions/runs/14729114037/job/41338836943), I see that there is a kotlin version conflict, I can verify this by running `./gradlew <anyModule>:dependencies`

Kotlin version conflict was caused due to `java` vs `android` modules using different version of kotlin, I believe, if you check out the repo and run the `./gradlew :any-module:measureSize` in this branch vs main, you will see the version conflict error doesn't come.

.github/actions/setup-android/action.yml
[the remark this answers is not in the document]
mahmoud-elmorabea: I actually meant the `gradle-version: '8.14'` 😅  but it's good to check the other one as well 😄

mahmoud-elmorabea:
> Kotlin version conflict was caused due to java vs android modules using different version of kotlin, I believe, if you check out the repo and run the ./gradlew :any-module:measureSize in this branch vs main, you will see the version conflict error doesn't come.

I don't think so, so I checked out this branch and ran `./gradlew :core:dependencies --configuration releaseRuntimeClasspath`. I can see both `kotlin-stdlib:1.8.21` and `kotlin-stdlib:1.7.21` (which resolves to higher version).

The exact versions that are mentioned in the error for failed jobs:
`java.lang.RuntimeException: Duplicate class kotlin.collections.jdk8.CollectionsJDK8Kt found in modules kotlin-stdlib-1.8.21 (org.jetbrains.kotlin:kotlin-stdlib:1.8.21) and kotlin-stdlib-jdk8-1.7.21 (org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.7.21)`

I can see the latest run on this PR also fails with the same error.

Shahroz16:
> I don't think so, so I checked out this branch and ran `./gradlew :core:dependencies --configuration releaseRuntimeClasspath`. I can see both `kotlin-stdlib:1.8.21` and `kotlin-stdlib:1.7.21` (which resolves to higher version).
> 
> The exact versions that are mentioned in the error for failed jobs: `java.lang.RuntimeException: Duplicate class kotlin.collections.jdk8.CollectionsJDK8Kt found in modules kotlin-stdlib-1.8.21 (org.jetbrains.kotlin:kotlin-stdlib:1.8.21) and kotlin-stdlib-jdk8-1.7.21 (org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.7.21)`
> 
> I can see the latest run on this PR also fails with the same error.

I mentioned it in the PR description, 

> the action is still failing because its trying to run the same thing for main branch (for comparison chart), which still has this base as java, once this PR gets merged, this issue should go away.

base/build.gradle
Shahroz16: Yes, we are not testing the SDK when its minified, we do test, sample app in minified mode and any effect it might have on our SDK.

mahmoud-elmorabea (approved):
Looks good!

I can still see some dependencies using Kotlin 1.7 but if we run a regression to make sure SDK is working well, that should be good.

Shahroz16:
Screenshot 2025-05-02 at 10 06 19 AM
Tested the snapshot version, events seems to be going okay, and the decompiled code and docs seems to be in the same shape too.

PR #528, opened 2025-05-01
chore: fixed persistent issues and added tests

Shahroz16:
closes: https://linear.app/customerio/issue/MBL-1021/persistant-messages

changes:
- Fixed the issues with persistent dismissing for persistent messages
- Added tests for persistent messages

### DIFF
--- a/messaginginapp/src/test/java/io/customer/messaginginapp/state/InAppMessageReducerTest.kt
+++ b/messaginginapp/src/test/java/io/customer/messaginginapp/state/InAppMessageReducerTest.kt
@@ -39,7 +39,6 @@ class InAppMessageReducerTest : JUnitTest() {
 
     @Test
     fun displayMessage_whenNonPersistentMessage_expectMessageMarkedAsShown() {
-        // Given a non-persistent message
         val testMessage = createTestMessage(persistent = false)
 
         // Initial state with the message in queue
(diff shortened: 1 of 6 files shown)


messaginginapp/src/test/java/io/customer/messaginginapp/state/InAppMessageReducerTest.kt
```diff
@@ -0,0 +1,282 @@
...
+    private lateinit var initialState: InAppMessagingState
+
+    override fun setup(testConfig: TestConfig) {
+        super.setup(
+            testConfigurationDefault {
+                diGraph {
+                    sdk { }
+                }
+            }
+        )
```
mrehan27, line 29: I believe we can remove this empty config? 🤔

messaginginapp/src/test/java/io/customer/messaginginapp/state/InAppMessageReducerTest.kt
```diff
@@ -0,0 +1,282 @@
...
+
+        initialState = InAppMessagingState(
+            siteId = String.random,
+            dataCenter = String.random,
+            environment = GistEnvironment.PROD,
```
mrehan27, line 34: Unless we need to exact match this, we can probably use non prod environment here?

mrehan27 (approved):

messaginginapp/src/test/java/io/customer/messaginginapp/state/InAppMessageReducerTest.kt
Shahroz16: removed

messaginginapp/src/test/java/io/customer/messaginginapp/state/InAppMessageReducerTest.kt
Shahroz16: we match with the initial state and since its prod, its closer to reality

PR #531, opened 2025-05-03
chore: Update docs to publish SDK locally to Maven Local

mahmoud-elmorabea:
Please checkout the read me added for more about what this is about.

This will be linked to wrapper SDKs to help test native Android SDK changes with wrapper SDKs.

### DIFF
--- a/docs/dev-notes/DEVELOPMENT.md
+++ b/docs/dev-notes/DEVELOPMENT.md
@@ -59,6 +59,21 @@ Create GitHub Actions secrets:
 
 This project is setup with GPG signing to sign the Maven artifacts before they are uploaded to Sonatype's servers to sync to Maven Central. GPG keys have been generated already but if you need to generate them again, [see this guide](https://gist.github.com/levibostian/ed2edcaa1ce1722d70683ce83fc429e2#sign) to do so. 
 
+### Publish SDK locally
+
+This section explains how to can publish the SDK locally to Maven Local. This could be useful in couple of scenarios:
+
+* If you want to verify that the SDK binary publishing works without errors locally without actually publishing any versions
+* Testing your changes in one of your wrapper SDKs (ReactNative, Expo or Flutter) with local changes
+
+To publish the SDK locally, you can run the following command:
+
+```bash
+IS_DEVELOPMENT=true ./gradlew publishToMavenLocal
+```
+
+This will publish a local version to Maven Local called 'local', that you can use for wrapper SDKs.
+
 # Troubleshooting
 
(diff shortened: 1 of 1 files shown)


Ahmed-Ali (approved):

docs/dev-notes/DEVELOPMENT.md
```diff
@@ -59,6 +59,21 @@ Create GitHub Actions secrets:
 
 This project is setup with GPG signing to sign the Maven artifacts before they are uploaded to Sonatype's servers to sync to Maven Central. GPG keys have been generated already but if you need to generate them again, [see this guide](https://gist.github.com/levibostian/ed2edcaa1ce1722d70683ce83fc429e2#sign) to do so. 
 
+### Publish SDK locally
+
+This section explains how to can publish the SDK locally to Maven Local. This could be useful in couple of scenarios:
```
mrehan27, line 64:
```suggestion
This section explains how to publish the SDK locally to Maven Local. This could be useful in couple of scenarios:
```

docs/dev-notes/DEVELOPMENT.md
```diff
@@ -59,6 +59,21 @@ Create GitHub Actions secrets:
...
+
+* If you want to verify that the SDK binary publishing works without errors locally without actually publishing any versions
+* Testing your changes in one of your wrapper SDKs (ReactNative, Expo or Flutter) with local changes
+
+To publish the SDK locally, you can run the following command:
```
mrehan27, line 69: Or we can select and run `install local` configuration from Android Studio as well.

mrehan27 (approved):

docs/dev-notes/DEVELOPMENT.md
mrehan27: Yup, you can select it from Run / Debug Configurations dropdown

Screenshot 2025-05-06 at 1 36 28 PM

mahmoud-elmorabea: I'm hesitant to add this option since you need to have `IS_DEVELOPMENT=true` for the local publishing to work, so you probably need to set this environment variable separately before running this, right?

I think the command should be simple enough since the vast majority of our dev docs are very command oriented.

But I can add if that's something you use regularly or think it's easier.

mrehan27: It [also sets `IS_DEVELOPMENT=true`](https://github.com/customerio/customerio-android/blob/3ebaf1a3ad903b2d21d86d41b25cca2243fca707/scripts/androidStudioRunConfigurations/install%20local.run.xml#L7) and has worked fine for me. I've always used this from UI and haven't run into any issues so far.

PR #533, opened 2025-05-07
chore: added view callback for custom action

Shahroz16:
closes: https://linear.app/customerio/issue/MBL-1020/actions-metrics-tracking-and-events-callback

changes:
- Added `InlineMessageActionListener` at the view level, so users can listen to custom events
- Logic changes, similar to iOS, if you register for `InlineMessageActionListener` you should receive callback for `messageActionTaken` and at global level listener


messaginginapp/src/main/java/io/customer/messaginginapp/ui/InAppMessageBaseView.kt
```diff
@@ -41,6 +43,12 @@ abstract class InAppMessageBaseView @JvmOverloads constructor(
     protected var currentMessage: Message? = null
     protected var currentRoute: String? = null
 
+    /**
+     * Listener to handle action clicks from inline in-app messages.
+     * Set this property to receive callbacks when actions are triggered.
+     */
+    var actionListener: InlineMessageActionListener? = null
```
mahmoud-elmorabea, line 50: Can we make this a setter only please? To avoid adding setter/getter in the public API

messaginginapp/src/main/java/io/customer/messaginginapp/type/InlineMessageActionListener.kt
```diff
@@ -0,0 +1,16 @@
...
+ */
+interface InlineMessageActionListener {
+    /**
+     * Called when a custom button is tapped in an inline message.
+     *
+     * @param message The in-app message that triggered the action
+     * @param actionValue The value associated with the action (typically a URL or identifier)
+     * @param actionName The name of the action that was triggered
+     */
+    fun onActionClick(message: InAppMessage, actionValue: String, actionName: String)
```
mahmoud-elmorabea, line 15: Can we make sure this is aligned with iOS? [iOS](https://github.com/customerio/customerio-ios/blob/main/Sources/MessagingInApp/Gist/Managers/InlineMessageManager.swift#L20) seems to be also passing the route?

mahmoud-elmorabea (approved):

mrehan27 (approved):

PR #536, opened 2025-05-09
chore: refactor in-app views to controller

mrehan27:
part of: [MBL-1091](https://linear.app/customerio/issue/MBL-1091/unified-action-handling-for-in-app-messages)

### Changes

- Moved business logic out of in-app views into dedicated controller classes for better separation and testability.
- Views now only manage UI and forward interactions to their respective controllers.
- Introduced the following structure:
  - `InAppMessageViewController`: Base controller for in-app message
  - `ModalInAppMessageViewController`: Final controller for modal in-app messages
  - `InlineInAppMessageViewController`: Final controller for inline in-app messages
  - `ModalInAppMessageView`: Final view for modal (no common parent)
  - `InlineInAppMessageView`: Final view for inline XML based usage (no common parent)

### PRs Stack:

- #535
- #536 👈🏻
- #537

### Notes

- Tests are not yet included to keep changes small and reviewable. Tests will be added once refactoring is complete.
- There are no intentional logic changes unless explicitly noted in comments. If you spot any, please flag them.

PR #535, opened 2025-05-09
chore: refactor os specific calls to delegate

mrehan27:
part of: [MBL-1091](https://linear.app/customerio/issue/MBL-1091/unified-action-handling-for-in-app-messages)

### Changes

- Introduced `InAppPlatformDelegate` to decouple all direct Android OS interactions from views, making the code more testable.
- Moved platform dependent logic into the delegate, including:
  - URI parsing
  - URL and Base64 decoding
  - Opening URLs and activities
  - View animations

### PRs Stack:

- #535 👈🏻
- #536
- #537

### Notes

- Tests are not yet included to keep changes small and reviewable. Tests will be added once refactoring is complete.
- There are no intentional logic changes unless explicitly noted in comments. If you spot any, please flag them.

PR #537, opened 2025-05-09
chore: refactor in-app view calls to delegates

mrehan27:
part of: [MBL-1091](https://linear.app/customerio/issue/MBL-1091/unified-action-handling-for-in-app-messages)

### Changes

- Added `EngineWebViewDelegate` to decouple controllers from `EngineWebView`
- Added `InAppHostViewDelegate` to decouple controllers from host in-app views
- Renamed `InAppMessageViewListener` to `InAppMessageViewCallback` (and similarly for child classes) for clearer naming

### PRs Stack:

- #535
- #536
- #537 👈🏻

### Notes

- Tests are not yet included to keep changes small and reviewable. Tests will be added once refactoring is complete.
- There are no intentional logic changes unless explicitly noted in comments. If you spot any, please flag them.

### DIFF
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/ui/bridge/EngineWebViewDelegate.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/ui/bridge/EngineWebViewDelegate.kt
@@ -1,0 +1,23 @@
+package io.customer.messaginginapp.ui.bridge
+
+import android.view.View
+import io.customer.messaginginapp.gist.data.model.engine.EngineWebConfiguration
+import io.customer.messaginginapp.gist.presentation.engine.EngineWebView
+import io.customer.messaginginapp.gist.presentation.engine.EngineWebViewListener
+
+/**
+ * Delegate interface to decouple [EngineWebView] from its consumers.
+ * Provides a minimal set of operations to control and configure the view without directly
+ * depending on its implementation.
+ * Useful for testing, mocking, or swapping the view behind the interface.
+ */
+internal interface EngineWebViewDelegate {
+    var listener: EngineWebViewListener?
+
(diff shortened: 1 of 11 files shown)


messaginginapp/src/main/java/io/customer/messaginginapp/ui/bridge/EngineWebViewDelegate.kt
```diff
@@ -0,0 +1,23 @@
...
+ */
+internal interface EngineWebViewDelegate {
+    var listener: EngineWebViewListener?
+
+    fun setup(configuration: EngineWebConfiguration)
+    fun stopLoading()
+    fun releaseResources()
+    fun getView(): View
+    fun setAlpha(alpha: Float)
+    fun bringToFront()
```
mrehan27, line 22: I reused same names from `View` class in this delegate to avoid reimplementation and keep the overall implementation simpler.

messaginginapp/src/main/java/io/customer/messaginginapp/ui/bridge/InAppHostViewDelegate.kt
```diff
@@ -0,0 +1,50 @@
...
+ * and posting actions on UI thread without directly depending on Android framework.
+ * Designed for easier testing and mocking.
+ */
+internal interface InAppHostViewDelegate {
+    var isVisible: Boolean
+
+    fun addView(delegate: EngineWebViewDelegate)
+    fun removeView(delegate: EngineWebViewDelegate)
+    fun createEngineWebViewInstance(): EngineWebViewDelegate
+    fun post(action: () -> Unit)
```
mrehan27, line 19: Same as above, reused same names from `View` class in this delegate to avoid reimplementation and keep the overall implementation simpler.

PR #541, opened 2025-05-12
chore: added compose screen in sample app

Shahroz16:
closes: https://linear.app/customerio/issue/MBL-1018/sample-app-changes-compose

changes:

- Added Jetpack Compose dependencies in the Java sample app
- Added the same UI view and navigation as the XML view

Screenshot_1747050365
Screenshot_1747050374


samples/compose-common.gradle
```diff
@@ -0,0 +1,52 @@
...
+ * This file contains shared Compose dependencies and configurations.
+ */
+
+// Common Compose version variables
+def compose_version = "1.4.1"
+def compose_bom_version = "2023.10.01"
+def compose_compiler_version = "1.4.7"
+def kotlin_bom_version = "1.8.20"
+def activity_compose_version = "1.7.1"
+def core_ktx_version = "1.10.1"
```
mrehan27, line 12: Since both apps now use Compose, could we use the existing common gradle file for this? Also, can we update sample apps to use newer versions here?

mrehan27 (approved):
Looks good

PR #544, opened 2025-05-13
chore: add tests for in-app base controller

mrehan27:
part of: [MBL-1128](https://linear.app/customerio/issue/MBL-1128/utilize-in-app-view-updates-to-increase-test-coverage)

### Changes

- Added unit tests to validate core functionality of `InAppMessageViewController`
- Included supporting extensions and test helpers required for test setup


messaginginapp/src/test/java/io/customer/messaginginapp/testutils/extension/GistExtensions.kt
```diff
@@ -38,3 +39,10 @@ fun createInAppMessage(
         )
     }
 )
+
+fun createGistAction(action: String): String = "gist://$action"
+
+internal fun InAppMessageViewController<*>.setMessageAndRouteForTest(message: Message, route: String) {
```
uros-mil, line 45: When I looked at the class being extended [InAppMessageViewController.kt](https://github.com/customerio/customerio-android/blob/8329d890c3f700576a43d8fa9359dd2adbef550f/messaginginapp/src/main/java/io/customer/messaginginapp/ui/controller/InAppMessageViewController.kt), I noticed that all properties are internal and none is private.

What is the reason for this?

If this is for testing purposes, it should not be the case.
The visibility of properties/methods should be based on what is/isn't public interface of the class.
The effect on private properties/methods should be evaluated through mocking dependencies and the return value of the public interface.

uros-mil (changes requested):

messaginginapp/src/test/java/io/customer/messaginginapp/ui/controller/InAppMessageViewControllerTest.kt
```diff
@@ -0,0 +1,490 @@
...
+        assertNoInteractions(inAppMessagingManager)
+    }
+
+    @Test
+    fun engineListener_givenSizeChanged_expectViewListenerReceivesCorrectSizeUpdates() {
+        val screenDensity = Double.random(1.0, 3.0)
+        val givenWidthInDp = Double.random(100.0, 500.0)
+        val givenHeightInDp = Double.random(100.0, 500.0)
+        val givenWidthInPx = (givenWidthInDp * screenDensity).roundToInt()
+        val givenHeightInPx = (givenWidthInDp * screenDensity).roundToInt()
```
Shahroz16, line 478: this doesnt look correct? shouldnt it be

```
val givenHeightInPx = (givenHeightInDp * screenDensity).roundToInt()
```

messaginginapp/src/test/java/io/customer/messaginginapp/ui/controller/InAppMessageViewControllerTest.kt
```diff
@@ -0,0 +1,490 @@
...
+import io.mockk.verify
+import io.mockk.verifyOrder
+import java.net.URI
+import kotlin.math.roundToInt
+import org.amshove.kluent.shouldBeEqualTo
+import org.amshove.kluent.shouldBeNull
+import org.amshove.kluent.shouldNotBeNull
+import org.junit.jupiter.api.Test
+
+class InAppMessageViewControllerTest : JUnitTest() {
```
Shahroz16, line 39: think we should also add tests for edge cases?

- Empty actions
- Null actions
- Consecutive tap handling
- Invalid URL handling

Shahroz16 (changes requested):

messaginginapp/src/test/java/io/customer/messaginginapp/testutils/extension/GistExtensions.kt
mrehan27: I've updated visibility and added annotations to internal properties that can't be made private or protected. Let me know if that helps.

messaginginapp/src/test/java/io/customer/messaginginapp/ui/controller/InAppMessageViewControllerTest.kt
mrehan27: Good catch. I've updated the test to fail in case of such typos.

messaginginapp/src/test/java/io/customer/messaginginapp/ui/controller/InAppMessageViewControllerTest.kt
mrehan27: - Empty actions

I believe `tap_givenAnyAction_expectDispatchTapActionToManager` test already covers this using random action. Is there anything specific you'd like to see added?

- Null actions

The [action is non-nullable](https://github.com/customerio/customerio-android/blob/a1c78468f9e377d804256d77811899e9980affbc/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebViewListener.kt#L5), so we shouldn't be getting `null` in the controller.

- Consecutive tap handling

We don't currently prevent multiple taps on SDK side, so I don't think there's much to verify here as it likely won't help much.

- Invalid URL handling

We're only routing to Android OS platform delegate from controller. I'll check if there's anything we can test in `PlatformDelegate`, but controller tests might not need to cover this. If you have ideas, please share.

Shahroz16 (approved):
looks good thanks for making the changes

PR #547, opened 2025-05-14
chore: added compose specific module

Shahroz16:
closes: MBL:991

changes: 
- we simply just add another module that is compose specific to hold the compose view


samples/sample-app.gradle
```diff
@@ -130,11 +130,13 @@ dependencies {
         implementation(project(":datapipelines"))
         implementation(project(":messagingpush"))
         implementation(project(":messaginginapp"))
+        implementation(project(":messaginginapp-compose"))
     } else {
         // Stable releases dependency, use published versions directly
         implementation "io.customer.android:datapipelines:$cioSDKVersion"
         implementation "io.customer.android:messaging-push-fcm:$cioSDKVersion"
         implementation "io.customer.android:messaging-in-app:$cioSDKVersion"
+        implementation "io.customer.android:messaging-in-app-compose:$cioSDKVersion"
```
mrehan27, line 139: If this is just for demonstration purposes then it's fine. Otherwise, only one of `messaginginapp` or `messaginginapp-compose` should be included on the app side.

settings.gradle
```diff
@@ -3,6 +3,7 @@ include ':base'
 include ':messagingpush'
 include ':common-test'
 include ':messaginginapp'
+include ':messaginginapp-compose'
```
mrehan27, line 6: Not related to this file, but can we please include `messaginginapp-compose` module in SDK binary size task as well?

mrehan27 (approved):

buildSrc/src/main/kotlin/io.customer/android/Versions.kt
```diff
@@ -36,4 +36,8 @@ object Versions {
     internal const val ROBOLECTRIC = "4.9"
     internal const val OKHTTP = "4.11.0"
     internal const val RETROFIT = "2.9.0"
+
+    // Compose versions
+    internal const val COMPOSE_BOM = "2023.08.00"
```
mahmoud-elmorabea, line 41: I can see that the latest is `2025.05.00`, is there a reason we are not using that?

buildSrc/src/main/kotlin/io.customer/android/Dependencies.kt
```diff
@@ -60,4 +60,14 @@ object Dependencies {
     const val okhttpLoggingInterceptor =
         "com.squareup.okhttp3:logging-interceptor:${Versions.OKHTTP}"
     const val segment = "com.segment.analytics.kotlin:android:${Versions.SEGMENT}"
+
+    // Compose dependencies
+    const val composeBom = "androidx.compose:compose-bom:${Versions.COMPOSE_BOM}"
```
mahmoud-elmorabea, line 65: BOMs FTW 👍

mahmoud-elmorabea (approved):
Looks good, but can we please address the BOM version comment 😇

buildSrc/src/main/kotlin/io.customer/android/Versions.kt
Shahroz16: `2025.05.00` has dependencies that requires our SDK to compiles with android API level `35`. We have not yet fully seen the changes/restrictions brought in that version, so just updating it would be a risky. 

I'll create a ticket to update compile with version of SDK

samples/sample-app.gradle
Shahroz16: yup for demonstration, but let me add a comment

settings.gradle
Shahroz16: good call out

PR #550, opened 2025-05-16
chore: add tests for modal in-app view controller

mrehan27:
part of: [MBL-1128](https://linear.app/customerio/issue/MBL-1128/utilize-in-app-view-updates-to-increase-test-coverage)

### Changes

- Added unit tests for `ModalInAppMessageViewController`
- Added tests for the default implementation of `InAppHostViewDelegate`
- Renamed `isMessageDisplayed` to `shouldDispatchDisplayEvent` in `ModalInAppMessageViewController` for improved clarity


messaginginapp/src/main/java/io/customer/messaginginapp/ui/controller/ModalInAppMessageViewController.kt
```diff
@@ -13,7 +14,8 @@ internal class ModalInAppMessageViewController(
     viewDelegate = viewDelegate,
     platformDelegate = platformDelegate
 ) {
-    private var isMessageDisplayed: Boolean = true
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal var shouldDispatchDisplayEvent: Boolean = true
```
Shahroz16, line 18: the class is `internal`, making the properties internal, if its a style thing thats alright but its not needed.

mrehan27: Yup, not needed in already internal class, removed it.

messaginginapp/src/main/java/io/customer/messaginginapp/ui/controller/ModalInAppMessageViewController.kt
```diff
@@ -13,7 +14,8 @@ internal class ModalInAppMessageViewController(
     viewDelegate = viewDelegate,
     platformDelegate = platformDelegate
 ) {
-    private var isMessageDisplayed: Boolean = true
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    var shouldDispatchDisplayEvent: Boolean = true
```
mahmoud-elmorabea, line 18: I suggest we don't expose internal state like that for tests, tests should depend on behavior.

For setup, let's just invoke the behavior that makes this true or false, then run our test code.
For verification, let's verify behavior changes instead of verifying the boolean itself

mrehan27: We are already doing that actually, updating this in next PR back to private.

PR #553, opened 2025-05-20
chore: fix display event handling and error recovery for in-app messages

mrehan27:
closes: [MBL-1155](https://linear.app/customerio/issue/MBL-1155/fix-inline-message-view-not-dismissed-on-error)

### Changes

- Updated `InAppMessageViewController` to dispatch the display event only once per message using flag
- In `InlineInAppMessageViewController`:
  - Reset display event flag on dismiss to allow proper handling for the next message
  - Called `onLoadingStarted` at start to improve loading UI experience
- In `ModalInAppMessageViewController`, ensured `EngineWebView` is updated only on first `routeLoaded`
- Updated `InAppMessageReducer` to dismiss inline messages on loading failure
- Moved inline message dismissal logic to an extension and extracted `InAppMessagingState` extensions to a dedicated file
- Moved min height setting from Compose view to base view for a consistent experience across Compose and XML
- Added and updated tests to validate these changes

### How to Reproduce

**Issue 1: Display Event Dispatched Multiple Times**

- Open Inline Examples
- Observe logs and display message event should never be dispatched multiple times even if `routeLoaded` is called multiple times for same message

**Issue 2: Inline View Not Dismissed on Message Loading Failure**

- Open Inline Examples
- Simulate message load failure
- Notice that the view remains partially visible and state isn't properly cleared before the fix

**Issue 3: Extra Space in Compose Inline View**

- Open Inline Compose Examples
- Before this fix, the first message loads, an empty space is shown and is fixed after the first message appears

**Issue 4: Progress Bar Not Visible in Inline XML View**

- Open Inline XML Examples
- Set `InlineInAppMessageView` height to `0dp` instead of `wrap_content`
- Before this fix, the progress bar might not show up properly due to missing `minHeight`

### DIFF
--- a/messaginginapp-compose/src/main/java/io/customer/messaginginapp/compose/InlineInAppMessageView.kt
+++ b/messaginginapp-compose/src/main/java/io/customer/messaginginapp/compose/InlineInAppMessageView.kt
@@ -46,9 +46,6 @@ fun InlineInAppMessage(
                 ViewGroup.LayoutParams.WRAP_CONTENT
             )
 
-            // Set a minimum height to ensure visibility
-            minimumHeight = (context.resources.displayMetrics.density * 48).toInt()
-
             // Set custom progress tint if provided
             progressTint?.let { color ->
                 setProgressTint(color.toArgb())
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessagingStateExtensions.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessagingStateExtensions.kt
@@ -1,0 +1,76 @@
+package io.customer.messaginginapp.state
+
+import io.customer.messaginginapp.gist.data.model.Message
+
+/**
+ * Updates [InAppMessagingState] to reflect the new state of an embedded messages.
+ *
+ * @param queueId The queueId of the embedded message to update.
+ * @param newState The new state of the embedded message.
+ * @param shownMessageQueueIds The set of queueIds that have been shown.
+ * @param messagesInQueue The set of messages in the queue.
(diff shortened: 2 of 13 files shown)


messaginginapp-compose/src/main/java/io/customer/messaginginapp/compose/InlineInAppMessageView.kt
```diff
@@ -46,9 +46,6 @@ fun InlineInAppMessage(
                 ViewGroup.LayoutParams.WRAP_CONTENT
             )
 
-            // Set a minimum height to ensure visibility
-            minimumHeight = (context.resources.displayMetrics.density * 48).toInt()
```
Shahroz16, line 50: this is now at xml view level?

messaginginapp/src/main/java/io/customer/messaginginapp/state/InAppMessagingStateExtensions.kt
```diff
@@ -0,0 +1,76 @@
...
+                        queueId = message.queueId,
+                        newState = InlineMessageState.Dismissed(message),
+                        shownMessageQueueIds = shownMessageQueueIds
+                    )
+                }
+
+                else -> {
+                    // For embedded messages without queueId
+                    // Just return the state unchanged
+                    this
```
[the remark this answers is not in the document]
mrehan27: Added

messaginginapp-compose/src/main/java/io/customer/messaginginapp/compose/InlineInAppMessageView.kt
mrehan27: Yes, correct.

PR #556, opened 2025-05-22
chore: get next message as soon as first one goes away

Shahroz16:
changes:

- Added mechanism to fetch next message when first one dismisses 
- Added test for the change


messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistSdk.kt
```diff
@@ -15,11 +15,12 @@ import java.util.Timer
 import kotlin.concurrent.timer
 import kotlinx.coroutines.flow.filter
 
-interface GistProvider {
+internal interface GistProvider {
```
mrehan27, line 18: 👍🏻

mrehan27 (approved):
Looks good 👍🏻

PR #557, opened 2025-05-22
feat: inline in app messages

Shahroz16:
closes: https://linear.app/customerio/project/inline-in-app-messages-for-android-native-a2d64e5827d6/overview

Merging the feature branch of inline in-app in the main.


mrehan27 (approved):
🚀

PR #561, opened 2025-05-29
fix: finish modal message early if sdk is not initialized

mrehan27:
part of: [MBL-940](https://linear.app/customerio/issue/MBL-940/android-crash-on-flutter-211-sdks)

### Background

When the app is restored from background by Android OS after being remove from memory while a modal message was displayed, it may crash on restoration if the SDK is not initialized in `Application` class. This occurs more frequently in wrapper SDKs where SDK initialization doesn't happen at app startup.

### Solution

While not a long-term architectural fix, this reliably prevents crashes by:

- Finishing `GistModalActivity` early if the SDK is not initialized
- Lazily accessing SDK dependent properties in `GistModalActivity` to avoid triggering runtime exceptions

Tests have been added to enforce this behavior and prevent future regressions or accidental SDK access before initialization in `GistModalActivity`.

### Code Changes

- Updated `GistModalActivity` to:
  - Finish immediately (without animation) if the SDK is not initialized
  - Lazily access SDK dependent properties to prevent crashes
  - Break down `onCreate` into smaller, readable methods
  - Added a helper method to call `super.finish()` immediately without animations
- Added tests to prevent accidental SDK access before initialization in `GistModalActivity`

### How to Reproduce

1. Make sure SDK is not initialized in `Application` class (or use wrapper sample app)
2. Open Developer Options on Test device
3. Scroll to Background process limit
4. Set the limit to "At most 1 process" (or desired value)
5. Open the Sample app
6. Send a modal in-app message
7. Without dismissing it, put the app to background
8. Open two other apps (or however many needed to hit background process limit) to force Sample app out of memory
9. Reopen the Sample app, it should crash with previous implementation

### Behavior Changes

| Before | After |
| - | - |
| App would crash on restore if the SDK was not initialized in `Application` class (❌ Bug) | `GistModalActivity` finishes early without animations on restore if the SDK is not initialized (✅ Fix) | 
| `GistModalActivity` would finish with animations if the message was `null`/invalid | `GistModalActivity` finishes early without animations if the message is `null`/invalid |

### Tested On

- [x] Native Sample App with SDK initialized in Application
- [x] Native Sample App with SDK not initialized in Application (as shown in the video)
- [x] Flutter Sample App

#### Before

https://github.com/user-attachments/assets/fb8761a7-57f2-4621-950c-0f6ea1ff0cd6

#### After (Non-persistent)

https://github.com/user-attachments/assets/c5eebe58-6b1c-4a8c-bb1e-a6cc0a1c2328

#### After (Persistent)

https://github.com/user-attachments/assets/751f3c42-7740-49b1-a579-712cdb40482c

---

### Note

**There are no intentional logic changes other than mentioned above. If any other logic change is observed, please call it out.**

### DIFF
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistModalActivity.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistModalActivity.kt
@@ -19,6 +19,7 @@ import io.customer.messaginginapp.gist.utilities.ElapsedTimer
 import io.customer.messaginginapp.gist.utilities.MessageOverlayColorParser
 import io.customer.messaginginapp.gist.utilities.ModalAnimationUtil
 import io.customer.messaginginapp.state.InAppMessagingAction
+import io.customer.messaginginapp.state.InAppMessagingManager
 import io.customer.messaginginapp.state.InAppMessagingState
 import io.customer.messaginginapp.state.ModalMessageState
 import io.customer.messaginginapp.ui.bridge.ModalInAppMessageViewCallback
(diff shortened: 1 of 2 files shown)


mahmoud-elmorabea (approved):
Looks great! 

Thanks for the thorough tests and test plan! 💯

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistModalActivity.kt
```diff
@@ -32,17 +33,22 @@ const val GIST_MODAL_POSITION_INTENT: String = "GIST_MODAL_POSITION"
...
+    private val attributesListenerJob: MutableList<Job> = mutableListOf()
+    private val elapsedTimer: ElapsedTimer = ElapsedTimer()
+    private val logger = SDKComponent.logger
+
+    // Dependencies requiring ModuleMessagingInApp to be initialized must be accessed lazily,
+    // only after confirming the SDK has been initialized.
+    private val inAppMessagingManager: InAppMessagingManager?
+        get() = kotlin.runCatching {
+            SDKComponent.inAppMessagingManager
+        }.getOrNull()
```
Shahroz16, line 47: I wonder can we consider initializing it in `onCreate`? lazy works but brings the risk on data races. 

```
    private var inAppMessagingManager: InAppMessagingManager? = null
    
    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)
        }
 ```

File: messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistModalActivity.kt [hunk 2]
```diff
@@ -135,23 +173,30 @@ class GistModalActivity : AppCompatActivity(), ModalInAppMessageViewCallback, Tr
             animationSet.start()
             animationSet.doOnEnd {
                 logger.debug("GistModelActivity finish animation completed")
-                super.finish()
+                completeFinish()
```
Shahroz16, line 176: Can the animation potentially be `cancelled` if the activity is destroyed before completion? 
Should we add animation cancellation in `onDestroy()`?

Shahroz16 (approved):
thank you for making the changes, added some considerations and suggestions.

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/GistModalActivity.kt
mrehan27: `inAppMessagingManager` is already a thread safe singleton, so I believe this shouldn't be an issue?

PR #570, opened 2025-06-03
chore: added thread safety in the inapp message controller

Shahroz16:
## Summary
  Fix thread safety issues in InApp message controllers by implementing `@Volatile` properties and `ThreadSafeController`       base class.

  ## Problem
  Controller properties like `currentMessage`, `elementId`, `contentWidthInDp`, and `contentHeightInDp` could be
  accessed from both UI thread (writes) and background threads causing potential
  race conditions and memory visibility issues.

  ## Solution
  - Created `ThreadSafeController` base class with property delegates (`threadSafe()`,
  `threadSafeWithNotification()`, `atomicState()`)
  - Updated `InAppMessageViewController` to extend `ThreadSafeController` and use `@Volatile` properties with
  `@ThreadSafeProperty` annotations
  - Updated `InlineInAppMessageViewController` to use thread-safe property delegates for `elementId`,
  `contentWidthInDp`, and `contentHeightInDp`
  - Added tests for thread safety implementation with concurrent access patterns

  ## Files Changed
  - **Modified:** `InAppMessageViewController.kt` - Extended `ThreadSafeController`, converted properties to
  use `by threadSafe()` delegates
  - **Modified:** `InlineInAppMessageViewController.kt` - Converted properties to use thread-safe delegates,
  replaced custom setter with `threadSafeWithNotification()`
  - **Added:** `ThreadSafeController.kt` - Base class with property delegates (internally uses `@Volatile`)
  - **Added:** `ThreadSafeControllerTest.kt` - Tests for property delegates and concurrent access


messaginginapp/src/main/java/io/customer/messaginginapp/ui/controller/InAppMessageViewController.kt
```diff
@@ -26,21 +26,23 @@ internal abstract class InAppMessageViewController<ViewCallback : InAppMessageVi
...
 
     @VisibleForTesting(otherwise = VisibleForTesting.PROTECTED)
-    var currentMessage: Message? = null
+    @ThreadSafeProperty("Accessed from UI thread writes and background WebView callbacks")
+    var currentMessage: Message? by threadSafe()
 
-    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
-    var currentRoute: String? = null
+    @ThreadSafeProperty("Accessed from UI and background threads during route changes")
+    internal var currentRoute: String? by threadSafe()
```
mrehan27, line 42: Since the class is already `internal`, we don't need to add it here again.

messaginginapp/src/main/java/io/customer/messaginginapp/ui/controller/InAppMessageViewController.kt
```diff
@@ -26,21 +26,23 @@ internal abstract class InAppMessageViewController<ViewCallback : InAppMessageVi
...
 
-    var viewCallback: ViewCallback? = null
+    @VisibleForTesting(otherwise = VisibleForTesting.PROTECTED)
+    var viewCallback: ViewCallback? by threadSafe()
```
mrehan27, line 45: Since it is used in `InlineInAppMessageView` and `ModalInAppMessageView`, it has to be `internal` and doesn't need `@VisibleForTesting` annotation I believe

messaginginapp/src/main/java/io/customer/messaginginapp/ui/controller/ThreadSafeController.kt
```diff
@@ -0,0 +1,99 @@
+package io.customer.messaginginapp.ui.controller
+
+import kotlin.properties.ReadWriteProperty
+import kotlin.reflect.KProperty
+
+/**
+ * Base class that enforces thread-safe property patterns for controller classes.
+ * All controller properties that might be accessed cross-thread should use these delegates.
+ */
+internal abstract class ThreadSafeController {
```
mrehan27, line 10: I think we don't need this class explicitly and can expose methods like `threadSafe` directly as a util methods instead?

messaginginapp/src/main/java/io/customer/messaginginapp/ui/controller/ThreadSafeController.kt
```diff
@@ -0,0 +1,99 @@
...
+        onChange: (old: T?, new: T?) -> Unit
+    ): ReadWriteProperty<Any?, T?> {
+        return VolatilePropertyWithNotification(initialValue, onChange)
+    }
+
+    /**
+     * For complex state requiring atomic updates of multiple properties.
+     * Use sparingly - prefer individual threadSafe() properties when possible.
+     */
+    protected fun <T> atomicState(initialValue: T): ReadWriteProperty<Any?, T> {
```
mrehan27, line 43: This probably is unused?

mrehan27 (approved):
Looks good 👍🏻

messaginginapp/src/main/java/io/customer/messaginginapp/ui/controller/ThreadSafeController.kt
Shahroz16: I thought about this one, let me draft the experience 


  Current approach (with ThreadSafeController):
```
  // Line 35: Property declaration is automatically thread-safe
  var engineWebViewDelegate: EngineWebViewDelegate? by threadSafe()
```
```
  // Lines 65, 70, 93: Usage is simple - just read/write normally
  if (engineWebViewDelegate != null) { ... }
  engineWebViewDelegate = viewDelegate.createEngineWebViewInstance()
  engineWebViewDelegate = null
```

  If using standalone utility methods instead:
```
  // You'd need a utility class like:
  object ThreadSafeUtils {
      fun <T> getVolatile(ref: AtomicReference<T>): T = ref.get()
      fun <T> setVolatile(ref: AtomicReference<T>, value: T) = ref.set(value)
  }
```
  // Then in your controller:
  ```
  private val engineWebViewDelegateRef = AtomicReference<EngineWebViewDelegate?>()

  // Every single access needs manual synchronization:
  if (ThreadSafeUtils.getVolatile(engineWebViewDelegateRef) != null) { ... }
  ThreadSafeUtils.setVolatile(engineWebViewDelegateRef,
  viewDelegate.createEngineWebViewInstance())
  ThreadSafeUtils.setVolatile(engineWebViewDelegateRef, null)
```

  The key problems with manual synchronization:

  1. Every read/write requires calling the utility method
  2. Easy to forget - developers might accidentally use direct access
  3. Verbose and error-prone - more code means more bugs
  4. No compile-time enforcement - nothing prevents direct access to the backing field

messaginginapp/src/main/java/io/customer/messaginginapp/ui/controller/ThreadSafeController.kt
Shahroz16: yup, added as utility one, will remove to not confuse

PR #571, opened 2025-06-04
chore: make device token thread safe

mrehan27:
closes: [MBL-1198](https://linear.app/customerio/issue/MBL-1198/android-non-volatile-shared-property-in-contextplugin)

### Changes

- Improved thread safety and testability for `ContextPlugin`:
  - Marked `deviceToken` as `@Volatile` to ensure proper visibility across threads
  - Extracted `ContextPluginEventProcessor` to isolate event processing logic and support test time behavior (e.g. timing assertions)
  - Added tests to validate correct propagation and visibility of `deviceToken`, following the same path as in production

### Note

The PR is created as `chore` because the SDK currently works fine without this change, but this improves reliability and helps surface potential bugs earlier.


Shahroz16 (approved):
lgtm

PR #578, opened 2025-06-15
chore: make prefs operation async 

Shahroz16:
closes: https://linear.app/customerio/issue/MBL-1231/refactor-sdk-initialization-component-to-avoid-main-thread-usage

## Summary

  Migrates device token management from `SharedPreferences` to `DataStore` to eliminate main thread violations discovered via StrictMode and synchronous access dealing with saving/retrieving data from storage. Introduces `DeviceTokenManager` as single source of truth with thread-safe access patterns.

  ## Context & Problem

  StrictMode violations revealed that SharedPreferences calls in `GlobalPreferenceStore` were blocking the main thread during SDK initialization. Device token access during event processing also had possible race condition risks with concurrent operations.

  ## Key Changes

  - DeviceTokenManager: New interface providing immediate synchronous access with background persistence via DataStore
  - GlobalPreferenceStore: Migrated from SharedPreferences to DataStore for async operations
  - ContextPlugin: Updated to delegate device token access to DeviceTokenManager
  - CustomerIO: Simplified device token operations using replaceToken() method for cleaner transition handling
  - StrictMode enabled: Added to sample app to catch future main thread violations

  ## Performance Improvements

  - Eliminates main thread blocking: DataStore operations happen asynchronously in background
  - Faster SDK startup: Device token loaded from memory immediately, storage sync happens in background
  - Improved thread safety: MutableStateFlow provides stronger memory consistency vs @Volatile fields
  - Reduced race conditions: Single source of truth pattern prevents token inconsistencies during concurrent operations
  - Cleaner token transitions: replaceToken() method handles refresh scenarios without confusing parameters

  ## Test Coverage

  - All existing tests pass (157 tests, 100% success rate)
  - Thread safety verified via ContextPluginBehaviorTest multi-threaded scenarios
  - Device token flows (new/refresh/delete/migration) remain functionally identical
  - Manual testing

### DIFF
--- a/core/src/test/java/io/customer/sdk/core/extensions/DataStoreExtensionsTest.kt
+++ b/core/src/test/java/io/customer/sdk/core/extensions/DataStoreExtensionsTest.kt
@@ -1,0 +1,161 @@
+package io.customer.sdk.core.extensions
+
+import androidx.datastore.core.DataStore
+import androidx.datastore.preferences.core.Preferences
+import androidx.datastore.preferences.core.stringPreferencesKey
+import io.customer.commontest.core.RobolectricTest
+import io.mockk.*
+import java.io.IOException
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.flowOf
+import kotlinx.coroutines.test.runTest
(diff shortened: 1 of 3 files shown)


Ahmed-Ali (changes requested):
Let's hold on for this one till we get better clarity on what the StrictMode violations were, and how are we approaching them holistically. Specially with the incomplete clarity on how the device token storage will eventually look like

PR #579, opened 2025-06-17
fix: expose inline classes for wrapper support

mrehan27:
part of: [MBL-1178](https://linear.app/customerio/issue/MBL-1178/add-android-native-view-implementation)

### Changes

- Extracted `BaseInlineInAppMessageView` from `InlineInAppMessageView` to allow easier extension and customization by wrappers
- Updated size callbacks to emit values in `DP`, converting to pixels only where necessary so wrappers receive consistent units without needing to handle conversion themselves
- Exposed selected internal classes as public with `InternalCustomerIOApi` annotation to allow controlled extension by wrapper SDKs

### Notes

There are no logic changes. This is purely refactoring to support wrapper specific needs. Please flag if any unintended logic behavior is observed.


messaginginapp/src/main/java/io/customer/messaginginapp/ui/core/BaseInlineInAppMessageView.kt
```diff
@@ -0,0 +1,65 @@
...
+    protected abstract val platformDelegate: P
+    internal val controller: InlineInAppMessageViewController by lazy {
+        InlineInAppMessageViewController(
+            viewDelegate = InAppHostViewDelegateImpl(view = this),
+            platformDelegate = platformDelegate
+        )
+    }
+
+    @InternalCustomerIOApi
+    protected fun configureView() {
```
Shahroz16, line 40: we made it manual instead of automatically initializing it in `init` due to lifecycle difference in wrapper?

Shahroz16 (approved):

messaginginapp/src/main/java/io/customer/messaginginapp/ui/core/BaseInlineInAppMessageView.kt
mrehan27: Yes, and to access `this` safely in `init` of final view

PR #584, opened 2025-07-21
fix: crash when build attributes are nullable

Shahroz16:
closes: https://linear.app/customerio/issue/MBL-984/android-sdk-running-into-anr-during-initialization-cont-gh-470

This supports a case, where `nullable` inbuilt device attributes are sent by Device. 

Changes:
This includes updating the `BuildStore` and `DeviceStore` interfaces and their implementations to make fields like `brand, model, and OS version` optional. JSON serialization logic was updated accordingly, and tests were added to verify the changes.

This is a potential fix for the Kotlin reflection crash. 

However, the more robust solution would be to update the `Kotlin serialization` library to `1.9.0`, which includes proper handling for unexpected values (introduced in 1.8.1 – PR #[2911](https://github.com/Kotlin/kotlinx.serialization/pull/2911)). But that upgrade would require bumping the `Kotlin` version to `2.2.0`.

### DIFF
--- a/datapipelines/src/test/java/io/customer/datapipelines/DataPipelinesCompatibilityTests.kt
+++ b/datapipelines/src/test/java/io/customer/datapipelines/DataPipelinesCompatibilityTests.kt
@@ -28,6 +28,7 @@ import kotlinx.coroutines.test.runTest
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.JsonArray
 import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.JsonNull
 import kotlinx.serialization.json.JsonObject
 import kotlinx.serialization.json.buildJsonObject
 import kotlinx.serialization.json.jsonArray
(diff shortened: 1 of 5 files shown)


mahmoud-elmorabea (approved):

datapipelines/src/test/java/io/customer/datapipelines/DataPipelinesCompatibilityTests.kt
```diff
@@ -432,6 +433,37 @@ class DataPipelinesCompatibilityTests : JUnitTest() {
...
 
+    @Test
+    fun device_givenTokenRegisteredWithNullableDeviceAttributes_expectDeviceAttributesAreSent() = runTest {
+        val givenToken = String.random
+        val givenAttributes = mapOf(
+            "cio_sdk_version" to "2.0.0-alpha.10",
+            "app_release" to "1.0",
+            "app_build" to "100",
+            "app_namespace" to "io.customer.android",
+            "device_os_version" to "13",
```
mrehan27, line 444: Should be int to match with `DeviceStoreImpl`?

```suggestion
            "device_os_version" to 13,
```

mrehan27 (approved):

datapipelines/src/test/java/io/customer/datapipelines/DataPipelinesCompatibilityTests.kt
Shahroz16: `device_os_version ` is just a random key :D we take that value in `device_os` i'll update

PR #586, opened 2025-07-24
chore: move setting storage to background

Shahroz16:
changes:

- Moved strict mode enabling before SDK startup rather App to avoid false positives of app
- We were storing `settings` on main thread, while this method doesn't need to be blocking nor it needs to be done in main thread. 

testing:
- Verified manually by running app and ensuring setting are being saved and running app without internet to ensure we are able to fetch loaded settings.

### DIFF
--- a/core/src/main/kotlin/io/customer/sdk/data/store/GlobalPreferenceStore.kt
+++ b/core/src/main/kotlin/io/customer/sdk/data/store/GlobalPreferenceStore.kt
@@ -26,7 +26,7 @@ internal class GlobalPreferenceStoreImpl(
     context: Context
 ) : PreferenceStore(context), GlobalPreferenceStore {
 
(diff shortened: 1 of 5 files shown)


core/src/main/kotlin/io/customer/sdk/data/store/GlobalPreferenceStore.kt
```diff
@@ -23,6 +26,8 @@ internal class GlobalPreferenceStoreImpl(
     context: Context
 ) : PreferenceStore(context), GlobalPreferenceStore {
 
+    val scope: CoroutineScope = SDKComponent.scopeProvider.globalPreferenceStoreScope
```
mrehan27, line 29: Could have been `private`?

```suggestion
    private val scope: CoroutineScope = SDKComponent.scopeProvider.globalPreferenceStoreScope
```

mrehan27 (approved):

PR #587, opened 2025-07-25
chore: moved analytics initialization to blocking background thread

Shahroz16:
closes: [MBL-1233: Changes in Segment Analytics SDK](https://linear.app/customerio/issue/MBL-1233/changes-in-segment-analytics-sdk)

Since we need to keep it synchronous, kept it blocking but we don't have to perform it on main thread. So moving it to background removes the main thread violations.

### DIFF
--- a/datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
+++ b/datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
@@ -96,6 +96,8 @@ class CustomerIO private constructor(
         }
     }
 
+    // Still synchronously create Analytics instance to ensure it is created before any other module uses it
+    // but we initialize it on background thread to avoid any storage access on main thread by Analytics.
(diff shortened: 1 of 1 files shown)


datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
```diff
@@ -94,14 +96,18 @@ class CustomerIO private constructor(
...
     }
 
-    internal val analytics: Analytics = overrideAnalytics ?: Analytics(
-        writeKey = moduleConfig.cdpApiKey,
-        context = androidSDKComponent.applicationContext,
-        configs = updateAnalyticsConfig(
-            moduleConfig = moduleConfig,
-            errorHandler = errorLogger
+    // Still synchronously create Analytics instance to ensure it is created before any other module uses it
+    // but we initialize it on background thread to avoid any storage access on main thread by Analytics.
```
mrehan27, line 100: 👍🏻

mrehan27 (approved):

PR #589, opened 2025-08-05
chore: Turn profileAttributes into setProfileAttributes

mahmoud-elmorabea:
Closes: [MBL-1299](https://linear.app/customerio/issue/MBL-1299/device-and-user-enrichment-api-alignment)

Tweaks public API `profileAttributes` to be `setProfileAttributes` for clarify and consistency with other SDKs


Shahroz16 (approved):

Shahroz16 (changes requested):
just realized we are going forward with the deprecation approach in iOS but android seems like a breaking change?

mahmoud-elmorabea:
@Shahroz16 unfortunately because `setProfileAttributes` has the same JVM signature as `profileAttributes` property, they cannot co-exist.

I'm open to suggestions though 😄

Shahroz16 (approved):

mrehan27 (commented):
I think I have an idea for deprecation approach, will continue the discussion on Slack.

mrehan27 (approved):

PR #590, opened 2025-08-05
chore: Turn deviceAttributes into setDeviceAttributes

mahmoud-elmorabea:
Closes: [MBL-1299](https://linear.app/customerio/issue/MBL-1299/device-and-user-enrichment-api-alignment)

Tweaks public API `deviceAttributes` to be `setDeviceAttributes` for clarify and consistency with other SDKs


Shahroz16 (approved):

Shahroz16 (changes requested):
just realized we are going forward with the deprecation approach in iOS but android seems like a breaking change?

mahmoud-elmorabea:
@Shahroz16 unfortunately because `setDeviceAttributes` has the same JVM signature as `deviceAttributes` property, they cannot co-exist.

I'm open to suggestions though 😄

mrehan27 (approved):

PR #604, opened 2025-08-25
chore: Add WorkManager dependency and create safer provider

mahmoud-elmorabea:
Close: [MBL-1350](https://linear.app/customerio/issue/MBL-1350/android-improve-push-metrics-background-deliverability)

Implementation is split across multiple PRs, here's a list to navigate:
- Adding `WorkManager` to push module -> **This PR**
- Move push messages delivery metrics to background -> https://github.com/customerio/customerio-android/pull/605
- Add tests for implementation -> https://github.com/customerio/customerio-android/pull/606

## Problem Statement
Android's power management features (Doze mode and App Standby) severely restrict background processing, preventing immediate delivery metric tracking when push notifications are received. Doze mode suspends network access and background services when devices are stationary and unplugged, while App Standby blocks background network access for unused apps. These restrictions cause delivery metric HTTP requests to fail or be indefinitely deferred, resulting in incomplete analytics data.

- Doze Mode & App Standby: Android suspends network access and background services when device is idle or apps are unused, preventing immediate delivery metric tracking
- Android 15 Enhanced Restrictions: Apps starting network requests outside valid process lifecycle receive exceptions (UnknownHostException or IOException) as documented in Android 15 behavior changes
- Failed Push Metrics: Traditional immediate HTTP requests to track push notification delivery fail when app is backgrounded, causing incomplete analytics data

## Solution
`PushDeliveryMetricsBackgroundScheduler` uses `WorkManager` to defer delivery metric tracking until system constraints are satisfied. The scheduler queues metric tracking tasks with network connectivity requirements, allowing WorkManager to execute them during Doze mode maintenance windows or when the device exits power-saving states.

## Implementation Details
- Constraint-based scheduling: Tasks require `NetworkType.CONNECTED` before execution
- Duplicate prevention: Uses `ExistingWorkPolicy.KEEP` with delivery ID as unique work identifier
- Graceful degradation: Handles `WorkManager` initialization failures without app crashes
- Persistent execution: Tasks survive app process death and device reboots
- Background worker: `PushDeliveryMetricsWorker` executes the actual HTTP request to track delivery metrics

## Technical Benefits
- Ensures delivery metrics are eventually tracked despite Android background restrictions
- Prevents immediate network failures during Doze mode
- Reduces battery consumption by aligning with system power management policies
- Provides reliable metric delivery without blocking the main notification processing flow

### DIFF
--- a/messagingpush/src/main/java/io/customer/messagingpush/util/WorkManagerProvider.kt
+++ b/messagingpush/src/main/java/io/customer/messagingpush/util/WorkManagerProvider.kt
@@ -7,7 +7,23 @@ import io.customer.messagingpush.logger.PushNotificationLogger
 
 /**
  * Provider for WorkManager instances with safe initialization.
- * Handles cases where WorkManager may not be initialized yet.
+ * Handles various WorkManager initialization scenarios based on hosting app configuration:
+ *
+ * 1. If the hosting app doesn't use WorkManager:
+ *    - WorkManager should be initialized automatically through androidx.work.WorkManagerInitializer
+ *    - The first WorkManager.getInstance() should return a valid instance
+ *
+ * 2. If hosting app does use WorkManager:
+ *    a) If host app depends on auto initialization - same as scenario 1
+ *    b) If host app doesn't depend on auto initialization (removed androidx.work.WorkManagerInitializer):
+ *       - If host app always initializes WorkManager early before push handling - no problems
+ *       - If host app conditionally initializes or doesn't initialize by the time we need it:
+ *         * We attempt to initialize WorkManager ourselves
+ *         * We check if host app implements Configuration.Provider to use their config
+ *         * If host app uses manual WorkManager.initialize(), we can't determine their config
+ *           so we initialize with default configuration
+ *
+ * These scenarios are handled based on documented exceptions from WorkManager's getInstance() and initialize() methods.
  */
 internal class WorkManagerProvider(
     private val context: Context,
@@ -15,7 +31,14 @@ internal class WorkManagerProvider(
 ) {
 
     /**
-     * Gets a WorkManager instance, initializing it if necessary.
+     * Gets a WorkManager instance with fallback initialization strategy.
+     *
+     * First attempts to get an existing WorkManager instance. If that fails (indicating
+     * WorkManager hasn't been initialized yet), attempts to initialize it ourselves:
+     * 1. Checks if the host app implements Configuration.Provider to respect their config
+     * 2. Falls back to default configuration if no custom config is available
+     * 3. Attempts to get the instance again after initialization
+     *
      * This method is synchronized to prevent race conditions during initialization.
      * @return WorkManager instance or null if initialization fails
      */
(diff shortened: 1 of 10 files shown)


messagingpush/src/main/java/io/customer/messagingpush/util/WorkManagerProvider.kt
```diff
@@ -0,0 +1,41 @@
...
+    fun getWorkManager(): WorkManager? {
+        return try {
+            // Try to get existing instance first
+            WorkManager.getInstance(context)
+        } catch (e: Exception) {
+            pushLogger.logWorkManagerInitializationAttempt(e)
+            try {
+                // Attempt to obtain app WorkManager config if available, otherwise use default config
+                val config = (context.applicationContext as? Configuration.Provider)?.workManagerConfiguration ?: Configuration.Builder().build()
+                WorkManager.initialize(context, config)
```
mrehan27, line 32: Why do we need to provide configuration? Isn't default initialization enough?

Also, where do we provide these configurations?

messagingpush/src/main/java/io/customer/messagingpush/util/WorkManagerProvider.kt
```diff
@@ -0,0 +1,41 @@
...
+    private val pushLogger: PushNotificationLogger
+) {
+
+    /**
+     * Gets a WorkManager instance, initializing it if necessary.
+     * This method is synchronized to prevent race conditions during initialization.
+     * @return WorkManager instance or null if initialization fails
+     */
+    @Synchronized
+    fun getWorkManager(): WorkManager? {
```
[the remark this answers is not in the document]
Shahroz16: lazy might take away the part where we are checking the instance if it fails, then initializing and then accessing the instance again. I mean its okay, i would have preferred a cleaner approach, but not a blocker.

Rehans point about config is a good one.

messagingpush/src/main/java/io/customer/messagingpush/util/WorkManagerProvider.kt
mahmoud-elmorabea: Alright, let me try to explain the whole thing, there are two main scenarios

### If the hosting app doesn't use `WorkManager`

In that case, we don't expect any problems, `WorkManager` should be initialized automatically through `androidx.work.WorkManagerInitializer` which is added internally automatically through `WorkManager` dependency. 

In this case, the first `WorkManager.getInstance()` should just return a valid instance and we are off to the races

### If hosting app does use `WorkManager`

Here's where it becomes tricky:
- If host app depends on auto initialization, that's the same as above
- If host app doesn't depend on auto initialization (Removed `androidx.work.WorkManagerInitializer`)
  - If the host app always initializes `WorkManager` very early before we have to handle a push message, again, no problems
  - If for some reason, the host app conditionally initializes `WorkManager` or it not initialized by the time we need it .. then we have to try to initialize it ourselves
    - When we try to initialize it ourselves we make a best effort to check if the host app has implemented `Configuration.Provider` in their `Application` class to avoid initializing `WorkManager` with config that is different from the host app
    - If the host app uses manual `WorkManager.initialize()` then we can't really know what config they use, so we just initialize `WorkManager` with default config
    
   

These cases are mostly driven by the documented exceptions thrown by `WorkManager`'s `getInstance()` and `initialize`

mrehan27 (approved):
My only concern was around complex and repeated attempts to initialize `WorkManager`. But based on the explanation, it seems necessary. Rest looks good to me.

PR #611, opened 2025-09-22
fix: optimize inline in-app reloading

mrehan27:
part of: [MBL-1329](https://linear.app/customerio/issue/MBL-1329/inline-in-app-message-flicker-on-return-navigation)

### Summary

Improves performance and lifecycle handling for inline in-app messages by avoiding unnecessary reloads and ensuring proper teardown when the view is destroyed. Also improves lifecycle usage to match current best practices.

### Changes

- Replaced deprecated `@OnLifecycleEvent` with `DefaultLifecycleObserver`
- Updated `InlineInAppMessageViewController` to dismiss message when view lifecycle owner is destroyed, instead of just detaching
- Skipped reloading if current message is already in `ReadyToEmbed` state and unchanged
- Cleared subscription job when the lifecycle owner is destroyed to prevent memory leaks
- Fixed relevant tests to reflect new behavior

### Testing

Can be tested on branch [mbl-1329-inline-with-tabs](https://github.com/customerio/customerio-android/tree/mbl-1329-inline-with-tabs)

#### Before

https://github.com/user-attachments/assets/520c62ff-69e4-4cda-b58a-611a2f58d8c4

#### After

https://github.com/user-attachments/assets/22fd38be-1309-4198-975b-29b31eeffa58

### Notes

- The message is not dismissed during configuration changes (e.g. screen rotation) and just recreated

### DIFF
--- a/messaginginapp/api/messaginginapp.api
+++ b/messaginginapp/api/messaginginapp.api
@@ -420,9 +420,12 @@ public abstract interface annotation class io/customer/messaginginapp/ui/control
 	public abstract fun reason ()Ljava/lang/String;
 }
 
-public abstract class io/customer/messaginginapp/ui/core/BaseInlineInAppMessageView : android/widget/FrameLayout, io/customer/messaginginapp/ui/bridge/InlineInAppMessageViewCallback {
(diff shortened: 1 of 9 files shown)


Shahroz16 (commented):
Cursor seems to have raised good points, please review

PR #613, opened 2025-09-23
chore: Add tabbed views to facilitate testing inline views

mahmoud-elmorabea:
### Context
This PR adds screens with tabs to test a bug for inline views that causes message to be prematurely dismissed when views are detached temporarily.

### Showcase of issue

[Screen_recording_20250923_150500.webm](https://github.com/user-attachments/assets/e049741d-364b-4a79-b0ad-4eb41dfb3b9e)

### Next
- Following PRs will fix this issue and these new screens will make it easier to test these changes


mrehan27 (commented):
Changes look good overall. But since Java sample app is used for internal testing as well, I think it'd be helpful to add some text to mention the element ids used because they don't match other primary sample apps now, and it'd be nearly impossible for anyone from TS to guess the `elementId` when testing an inline issue.

mrehan27 (approved):

PR #664, opened 2026-02-25
chore: make location module self-contained with generic core interfaces

Shahroz16:
Summary

  - Replace location-specific code in datapipelines (LocationPlugin, LocationSyncFilter, LocationSyncStore, TrackLocationEvent subscriber) with two generic core
   interfaces (DataPipeline, ProfileEnrichmentProvider) that any module can use
  - Move all location business logic (sync filter, userId gate, profile switch handling, track event sending) into the location module's LocationTracker, making it fully self-contained
  - Eliminate cross-module EventBus round-trips for location track events — LocationTracker now calls DataPipeline.track() directly

  What changed

  Core module — new generic interfaces:
  - DataPipeline interface (track + userId) — modules retrieve via SDKComponent.getOrNull<DataPipeline>()
  - ProfileEnrichmentProvider / ProfileEnrichmentRegistry — modules register to enrich identify events
  - Removed LocationCache, Event.TrackLocationEvent, Event.LocationData (location-specific contracts)

  Datapipelines module — zero location awareness:
  - CustomerIO implements DataPipeline, registers itself for module access
  - ProfileEnrichmentPlugin replaces LocationPlugin — generic enrichment that queries all registered providers
  - Removed LocationSyncFilter, LocationSyncStore, TrackLocationEvent subscriber, sendLocationTrack(), and clearSyncedData() calls from identifyImpl()/clearIdentifyImpl()

  Location module — fully self-contained:
  - LocationTracker implements ProfileEnrichmentProvider for identify enrichment
  - LocationTracker applies userId gate + sync filter + DataPipeline.track() directly (no EventBus)
  - LocationSyncFilter and LocationSyncStore moved from datapipelines (package change only, logic identical)
  - Profile switch detection via lastKnownUserId with sync filter reset
  - onReset() clears all state (in-memory cache, persisted location, sync filter)

---

> [!NOTE]
> **Medium Risk**
> Changes the location-identify/track flow and moves filtering/state-reset responsibilities across modules; regressions could drop or duplicate location enrichment/track events despite being covered by updated unit tests.
> 
> **Overview**
> Decouples location from `datapipelines` by removing location-specific `Event` types (`Event.LocationData`, `Event.TrackLocationEvent`) and the `LocationCache` contract, and introducing core internal interfaces for cross-module interaction: `DataPipeline` (direct `track` + `isUserIdentified`) and `IdentifyHook`/`IdentifyHookRegistry` (identify context enrichment + synchronous reset hook).
> 
> `CustomerIO` now implements and registers `DataPipeline`, replaces `LocationPlugin` with a generic `IdentifyContextPlugin` that enriches identify events from all registered hooks, and removes the EventBus-driven location track subscriber and sync-filter reset calls. The location module takes ownership of location business logic: `LocationTracker` implements `IdentifyHook`, persists/caches coordinates, applies the 24h/1km `LocationSyncFilter` (moved into the location module), and sends `LOCATION_UPDATE` via `DataPipeline.track` instead of publishing through EventBus; tests are updated accordingly and location/datapipelines builds opt into `@InternalCustomerIOApi`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5c2a122929169dbb38dceb74330a9695d8e0f546. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


core/src/main/kotlin/io/customer/sdk/core/pipeline/DataPipeline.kt
```diff
@@ -0,0 +1,17 @@
...
+ * Abstraction for sending track events to the data pipeline.
+ *
+ * Modules retrieve an implementation via `SDKComponent.getOrNull<DataPipeline>()`
+ * to send events directly without going through EventBus.
+ *
+ * This is an internal SDK contract — not intended for use by host app developers.
+ */
+@InternalCustomerIOApi
+interface DataPipeline {
+    val userId: String?
```
mahmoud-elmorabea: Are we exposing `userId` to check if the user is identified? Should we expose `isIdentified` instead?

core/src/main/kotlin/io/customer/sdk/core/pipeline/ProfileEnrichmentProvider.kt
```diff
@@ -0,0 +1,16 @@
...
+ * Interface for modules that contribute attributes to identify events.
+ *
+ * Implementations return a map of primitive-valued attributes (String, Number, Boolean)
+ * that will be added to the identify event context. Return null to skip enrichment.
+ *
+ * This is an internal SDK contract — not intended for use by host app developers.
+ */
+@InternalCustomerIOApi
+interface ProfileEnrichmentProvider {
+    fun getProfileEnrichmentAttributes(): Map<String, Any>?
```
mahmoud-elmorabea: Let's make this non-null. If there is nothing to enrich, we can return empty map.

core/src/main/kotlin/io/customer/sdk/core/pipeline/IdentifyContextRegistry.kt
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * Cleared automatically when [SDKComponent.reset] clears singletons.
+ *
+ * This is an internal SDK contract — not intended for use by host app developers.
+ */
+@InternalCustomerIOApi
+class ProfileEnrichmentRegistry {
+    private val providers = mutableListOf<ProfileEnrichmentProvider>()
+
+    @Synchronized
```
mahmoud-elmorabea: Are we sure that `@Synchronized` is the right tool here? Do plugins use threads or coroutines?

datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
```diff
@@ -135,11 +130,11 @@ class CustomerIO private constructor(
...
         // Add plugin to filter events based on SDK configuration
         analytics.add(ScreenFilterPlugin(moduleConfig.screenViewUse))
-        analytics.add(locationPlugin)
+        analytics.add(ProfileEnrichmentPlugin(SDKComponent.profileEnrichmentRegistry, logger))
         analytics.add(ApplicationLifecyclePlugin())
 
-        // Register LocationPlugin as LocationCache so the location module can update it
-        SDKComponent.registerDependency<LocationCache> { locationPlugin }
+        // Register this instance as DataPipeline so modules can send track events directly
+        SDKComponent.registerDependency<DataPipeline> { this }
```
mahmoud-elmorabea, line 137: Is this the best way to make this registration? This has to be done before the location module is inialized, correct?

location/src/main/kotlin/io/customer/location/ModuleLocation.kt
mahmoud-elmorabea, line 64: Is this access thread safe?

core/src/main/kotlin/io/customer/sdk/core/pipeline/DataPipeline.kt
Shahroz16: `userId` is exposed if we wanted to also track user being changed, but i can remove it right now and rely on `UserChangedEvent`

core/src/main/kotlin/io/customer/sdk/core/pipeline/IdentifyContextRegistry.kt
Shahroz16: `Plugins` run synchronously on the caller's thread, `EventPlugin.identify()` is blocking and executes on whatever thread called `analytics.identify()`. No coroutines involved at the plugin level.

`@Synchronized` is the right tool here because `register()`, `getAll()`, and `clear()` are fast, non-suspending operations (list add/copy/clear). A coroutine Mutex would require these to be suspend functions, which doesn't fit the EventPlugin interface.

datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
Shahroz16: Yes, it must happen before. The `init` order is guaranteed by `CustomerIO.initialize()` : `createInstance()` runs the constructor (which triggers init → registers DataPipeline), then `module.initialize()` is called in the loop afterwards. So `SDKComponent.getOrNull<DataPipeline>()` in `ModuleLocation.initialize()` always finds the registered instance.

location/src/main/kotlin/io/customer/location/ModuleLocation.kt
Shahroz16: It's safe in practice, `_locationServices` is written once during `initialize()` (main thread, during SDK startup) before any host app code can access the getter. But it's not formally guaranteed across threads. Will add `@Volatile` for extra hardening.

datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
```diff
@@ -258,7 +238,6 @@ class CustomerIO private constructor(
 
         if (isChangingIdentifiedProfile) {
             logger.info("changing profile from id $currentlyIdentifiedProfile to $userId")
-            locationSyncFilter.clearSyncedData()
```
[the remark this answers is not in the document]
Shahroz16: The sync filter not clearing on direct profile switch is intentional.                                
                                                                                                                                            
  When `user B` is `identified`, the `IdentifyContextPlugin` calls `LocationTracker.getIdentifyContext()` and attaches the current coordinates to request. This path is unfiltered, user B's profile receives the device's location immediately as part of the identify call itself.

The sync filter only gates the supplementary "Location Update" track event. If it suppresses that event because the same coordinates were recently sent, that's correct behavior — the backend already has the location from the identify event. No data is lost.

core/src/main/kotlin/io/customer/sdk/core/pipeline/DataPipeline.kt
mahmoud-elmorabea: Good point, that covers two identify call different ids with no `clearIdentify`, I will port this to iOS

core/src/main/kotlin/io/customer/sdk/core/pipeline/IdentifyContextRegistry.kt
mahmoud-elmorabea: The main idea is that the registry is called from other modules to register themselves and this could expand to more modules in the future. So we can't be sure that this invariant remains true

mahmoud-elmorabea (approved):
Looks good! 💯 

I do think we should go back to the approach where we exposed userId to detect a change. But do feel free to address this in this or later PR if we decide to go that way.

PR #665, opened 2026-03-02
chore: update Location tracking module config

Shahroz16:
## Summary

  - Replace `enableLocationTracking: Boolean` with a 3-state `LocationTrackingMode` enum (`OFF`, `MANUAL`, `ON_APP_START`) 
  - `ON_APP_START` auto-captures location on cold start, going through the normal sync filter path (cache + 24h/1km filter + track event)
  - `OFF` mode early-returns in `initialize()`, skipping all background machinery (no restoration, no enrichment, no event subscriptions)
  - Remove `stopLocationUpdates()` from public API — not needed for one-shot request pattern
  - Use first-wins request deduplication: if a request is in flight, new calls are ignored (matches iOS behavior)
  - Simplify `LocationServicesImpl` by removing lock and stop logic

---

> [!NOTE]
> **Medium Risk**
> Changes the public location module API/config and alters lifecycle-driven behavior (auto-capture on startup and cancellation on background), which could affect when/if location is collected or jobs are dropped. Also introduces new dependency on process lifecycle and changes in-flight request semantics from cancel-and-replace to first-wins.
> 
> **Overview**
> **Location module config is redesigned** from a boolean `enableLocationTracking` to a 3-state `LocationTrackingMode` (`OFF`, `MANUAL`, `ON_APP_START`), with `OFF` now short-circuiting `ModuleLocation.initialize()` to skip restoration/enrichment/subscriptions.
> 
> **Location request behavior is updated**: `LocationServices.stopLocationUpdates()` is removed, `LocationServicesImpl` switches to *first-wins* in-flight dedup (ignore new calls while a request is active), and in-flight requests are cancelled automatically when the app backgrounds via a new `ProcessLifecycleOwner` observer.
> 
> **Shared main-thread posting utility** is moved into `core` as `MainThreadPoster`/`HandlerMainThreadPoster` and reused by in-app messaging; samples/tests are updated accordingly, including removing the sample UI’s “stop updates” button and adding unit tests for `OFF` mode no-ops.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d21a04a259940b4168f4d91d06d3a280509fe106. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


location/src/main/kotlin/io/customer/location/LocationServicesImpl.kt
```diff
@@ -21,11 +19,11 @@ internal class LocationServicesImpl(
     private val scope: CoroutineScope
 ) : LocationServices {
 
-    private val lock = ReentrantLock()
```
mahmoud-elmorabea, line 24: Is there a specific reason why we removed that lock?

location/src/main/kotlin/io/customer/location/ModuleLocation.kt
```diff
@@ -92,21 +120,22 @@ class ModuleLocation @JvmOverloads constructor(
...
-            orchestrator = orchestrator,
-            scope = SDKComponent.scopeProvider.locationScope
-        )
+        // ON_APP_START: auto-capture location on cold start.
+        // Routed through LocationServicesImpl so it shares the same job tracking
+        // and first-wins dedup as user-initiated requests.
+        // If permissions are denied or services disabled, orchestrator silently no-ops.
+        if (moduleConfig.trackingMode == LocationTrackingMode.ON_APP_START) {
+            services.requestLocationUpdate()
+        }
```
mahmoud-elmorabea: I think we should actually do this in onStart of the lifecycle observer to be completely sure we are not doing this in the bacgkround

mahmoud-elmorabea (changes requested):

location/src/main/kotlin/io/customer/location/LocationServicesImpl.kt
Shahroz16: The lock was originally added when we used `last-wins` dedup (cancel old → start new). That pattern creates a concurrent jobs scenario where the lock was needed to atomically cancel-then-relaunch.

We switched to first-wins dedup (if a request is in flight, ignore the new call) what iOS is doing. With first-wins, there's only ever one job at a time — no  concurrent job scenario, so the lock's original purpose is gone.

The only theoretical race is two threads calling requestLocationUpdate() simultaneously when no job is active, both could launch a job. The consequence is one redundant GPS request

location/src/main/kotlin/io/customer/location/ModuleLocation.kt
Shahroz16: In practice, `CustomerIO.initialize()` runs during Application.onCreate(), and if the process started for a background reason, Android's background location restrictions (API 29+) silently fail the GPS request. Our `onStop` observer would also cancel it. So the current approach is safe without the extra indirection. 

Can still add it to have parity with iOS

mahmoud-elmorabea (approved):

PR #669, opened 2026-04-21
chore: prep infrastructure for live notifications

mahmoud-elmorabea:
Closes: [MBL-1643: Create path for detecting Live Activities notification](https://linear.app/customerio/issue/MBL-1643/create-path-for-detecting-live-activities-notification)

## PR Series — Live Notifications (Android Live Activities)

| # | Branch | PR |
|---|--------|----|
| **1** | `live-notifications-1-prep` | **← you are here** |
| 2 | `live-notifications-2-rendering` | https://github.com/customerio/customerio-android/pull/670 |
| 3 | `live-notifications-3-sample` | https://github.com/customerio/customerio-android/pull/671 |

---

## Context

This series introduces live notifications to the Android SDK — the Android counterpart of iOS Live Activities. Live notifications are ongoing notifications that update in-place via a stable ID, allowing the server to push real-time state changes (delivery tracking, sports scores, countdown timers) without creating duplicate notification entries.

The payload schema mirrors the iOS Live Activity contract:
- `activity_id` — stable identifier used to update the same notification slot
- `event` — lifecycle: `start` / `update` / `end`
- `activity_type` — rendering mode: `progress` / `countdown` / `text`
- `attributes` — JSON string of structural/static data (set once, e.g. `progress_max`, icons)
- `content_state` — JSON string of dynamic data that changes on every update (e.g. `title`, `body`, `progress`)

On Android 16 (API 36+) live notifications request promoted ongoing status via `POST_PROMOTED_NOTIFICATIONS`, giving them elevated treatment in the notification shade. On earlier versions they render as standard ongoing notifications using `NotificationCompat`.

---

## This PR — Infrastructure & foundations

Lays the groundwork consumed by the rendering PR. All changes compile against `main` with no dependency on the new live notification classes.

### `BitmapDownloader` utility
Extracts the repeated `URL.openStream() → BitmapFactory.decodeStream()` pattern into a shared `BitmapDownloader.download()` utility. Fixes an existing resource leak in standard push — the `InputStream` was never closed, risking file descriptor exhaustion under repeated rich push loads. Both `CustomerIOPushNotificationHandler.addImage` and the upcoming `LiveNotificationHandler` use this utility.

### `CustomerIOParsedPushPayload` — `activityId` field
Adds a nullable `activityId` field (replaces the POC's `liveNotificationId`) to carry the stable live notification ID through the click routing pipeline to `NotificationClickReceiverActivity`.

### `NotificationChannelCreator` — live notification channel
[...]

PR #670, opened 2026-04-21
chore: live notification rendering and routing

mahmoud-elmorabea:
Closes: [MBL-1644: Render templates on Android 16](https://linear.app/customerio/issue/MBL-1644/render-templates-on-android-16)

## PR Series — Live Notifications (Android Live Activities)

| # | Branch | PR |
|---|--------|----|
| 1 | `live-notifications-1-prep` | https://github.com/customerio/customerio-android/pull/669 |
| **2** | `live-notifications-2-rendering` | **← you are here** |
| 3 | `live-notifications-3-sample` | https://github.com/customerio/customerio-android/pull/671 |

---

## Context

This series introduces live notifications to the Android SDK — the Android counterpart of iOS Live Activities. Live notifications are ongoing notifications that update in-place via a stable ID, allowing the server to push real-time state changes (delivery tracking, sports scores, countdown timers) without creating duplicate notification entries.

The payload schema mirrors the iOS Live Activity contract:
- `activity_id` — stable identifier used to update the same notification slot
- `event` — lifecycle: `start` / `update` / `end`
- `activity_type` — rendering mode: `progress` / `countdown` / `text`
- `attributes` — JSON string of structural/static data (set once, e.g. `progress_max`, icons)
- `content_state` — JSON string of dynamic data that changes on every update (e.g. `title`, `body`, `progress`)

On Android 16 (API 36+) live notifications request promoted ongoing status via `POST_PROMOTED_NOTIFICATIONS`, giving them elevated treatment in the notification shade. On earlier versions they render as standard ongoing notifications using `NotificationCompat`.

---

## This PR — Rendering & routing

Implements the full live notification feature on top of the infrastructure from PR 1. Introduces the notification builders, the central handler, and the routing logic that detects a live notification payload and dispatches it away from the standard push path.

### `ActionData`
Simple data class representing a single action button (label, deep link URI, `PendingIntent`). Up to 3 actions are supported per Android's limit.

### `BasicNotificationBuilder` (pre-API 36)
Builds live notifications on API 26–35 using `NotificationCompat`. Supports a linear determinate progress bar, countdown timer via `Chronometer`, colorized background, large icon, subtext, and up to 3 action buttons. Uses a single `accentColor` parameter (matching the API 36 path) so callers make color precedence explicit rather than relying on sequential `setColor()` overwrites. Segmented progress bars and promoted ongoing status are not available on this tier.

### `Api36LiveNotificationBuilder` (API 36+)
Builds promoted ongoing notifications on Android 16+ using the native `Notification.Builder` with `Notification.ProgressStyle`. Supports segmented progress bars (with per-segment colors), progress point markers, start/end/tracker icons, and promoted ongoing status. Before requesting promoted ongoing, performs a runtime permission check for `POST_PROMOTED_NOTIFICATIONS` — if not granted, falls back silently to standard ongoing and logs a debug message.

### `LiveNotificationHandler`
Central handler that owns the full live notification lifecycle:
- Reads `activity_id`, `event`, `activity_type` from the top-level bundle
- Parses `attributes` and `content_state` as `JSONObject` — malformed JSON is caught, logged as an error, and replaced with an empty object so the notification still posts with safe defaults rather than crashing
- Dispatches to `Api36LiveNotificationBuilder` on API 36+ or `BasicNotificationBuilder` on earlier versions
- Posts the notification with a stable ID derived from `activity_id` so successive pushes update in-place
- On `event = "end"`, schedules cancellation after `dismiss_delay` ms (default 3000)
- On API 33+, logs a clear warning if `POST_NOTIFICATIONS` is not granted so developers can diagnose why no notification appeared. Behaviour then matches the standard push path — the system silently drops the notification at `notify()` time, so live and regular notifications degrade identically when the permission is missing.

### `CustomerIOPushNotificationHandler` — routing
Detects `activity_id` in the incoming bundle and routes to `LiveNotificationHandler`. The `onNotificationComposed` callback is intentionally not invoked for live notifications — their layout is SDK-controlled and cannot be safely modified by host apps (and the API 36 path uses the native `Notification.Builder`, not `NotificationCompat.Builder`, making the existing callback signature inapplicable anyway).

### Notable design decisions
- **Delivery tracking**: the `delivered` metric fires before `handleNotification()`, so it is tracked regardless of whether the notification is shown. This matches existing standard push behaviour and the iOS convention where "delivered" means "received by device".
- **`onNotificationClicked` / deep link routing**: live notifications go through the same `NotificationClickReceiverActivity` path as standard push, so click tracking and deep link handling work identically.

### DIFF
--- a/messagingpush/src/main/java/io/customer/messagingpush/LiveNotificationHandler.kt
+++ b/messagingpush/src/main/java/io/customer/messagingpush/LiveNotificationHandler.kt
@@ -1,0 +1,258 @@
+package io.customer.messagingpush
+
+import android.app.NotificationManager
+import android.app.PendingIntent
+import android.content.Context
+import android.content.Intent
+import android.content.pm.PackageManager
+import android.graphics.Bitmap
+import android.os.Build
+import android.os.Bundle
+import android.os.Handler
+import android.os.Looper
+import androidx.annotation.ColorInt
+import androidx.annotation.DrawableRes
+import androidx.core.content.ContextCompat
+import io.customer.messagingpush.activity.NotificationClickReceiverActivity
+import io.customer.messagingpush.data.model.CustomerIOParsedPushPayload
+import io.customer.messagingpush.extensions.getDrawableByName
+import io.customer.messagingpush.extensions.toColorOrNull
+import io.customer.messagingpush.util.BitmapDownloader
+import io.customer.sdk.core.di.SDKComponent
+import org.json.JSONArray
+import org.json.JSONException
+import org.json.JSONObject
+
+/**
+ * Handles live notification creation, updates, and dismissal.
+ *
+ * Live notifications are ongoing notifications that can be updated in-place
+ * (the Android counterpart of iOS Live Activities). They use a stable
+ * notification ID derived from [ACTIVITY_ID_KEY] so successive pushes
+ * replace the previous notification rather than creating new ones.
+ *
+ * Payload schema (iOS-style split):
+ * - [ACTIVITY_ID_KEY] — stable id
+ * - [EVENT_KEY] — "start" | "update" | "end"
+ * - [ACTIVITY_TYPE_KEY] — "progress" | "countdown" | "text"
+ * - [ATTRIBUTES_KEY] — JSON string: structural/static data (progress_max, icons, etc.)
+ * - [CONTENT_STATE_KEY] — JSON string: dynamic data (title, body, progress, subtext, etc.)
+ */
+internal class LiveNotificationHandler(
+    private val bundle: Bundle
+) {
+
+    companion object {
+        const val ACTIVITY_ID_KEY = "activity_id"
+        const val EVENT_KEY = "event"
+        const val ACTIVITY_TYPE_KEY = "activity_type"
+        const val CONTENT_STATE_KEY = "content_state"
+        const val ATTRIBUTES_KEY = "attributes"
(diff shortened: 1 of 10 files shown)


messagingpush/src/main/java/io/customer/messagingpush/LiveNotificationHandler.kt
```diff
@@ -0,0 +1,258 @@
...
+            ContextCompat.checkSelfPermission(context, android.Manifest.permission.POST_NOTIFICATIONS)
+            != PackageManager.PERMISSION_GRANTED
+        ) {
+            SDKComponent.logger.error(
+                "POST_NOTIFICATIONS permission not granted; live notification will be dropped by the system. " +
+                    "The host app must request this permission on Android 13+."
+            )
+        }
+
+        val activityId = bundle.getString(ACTIVITY_ID_KEY) ?: return
```
Shahroz16, line 82: can it be empty, should we gate check it?

Shahroz16 (commented):
There are alot of behavioral changes, think we should add some unit tests here

mahmoud-elmorabea:
@Shahroz16 

Will absolutely cover everything with tests, this just provides idea how things will be structured. Once we agree on templates structure, I will add tests with expected payloads to make sure it closely simulates real push payloads

Shahroz16 (approved):

PR #672, opened 2026-04-24
chore: move HttpClient to core as CustomerIOHttpClient

mrehan27:
part of: [MBL-1627](https://linear.app/customerio/issue/MBL-1627/android-send-geofence-transition-events)

> **Note:** This PR targets `main` directly since it moves shared files from `:messagingpush` to `:core`. Merging to `main` first ensures other feature branches working in parallel don't run into conflicts with the relocated files.

## Summary

Move shared infrastructure from `:messagingpush` to `:core` so all SDK modules can use them without depending on the push module.

### chore: move `HttpClient` to core as `CustomerIOHttpClient`

- Rename `HttpClient` → `CustomerIOHttpClient`, `HttpClientImpl` → `CustomerIOHttpClientImpl`
- Move from `messagingpush/network/` to `core/network/`
- Change visibility from `internal` to `public` with `@InternalCustomerIOApi`
- Move DI extension `SDKComponent.httpClient` to `SDKComponentExt.kt` in core
- Update push module imports

### chore: move `WorkManagerProvider` to core

- Move from `messagingpush/util/` to `core/util/`
- Move DI extension `SDKComponent.workManagerProvider` to `SDKComponentExt.kt` in core
- Change visibility from `internal` to `public` with `@InternalCustomerIOApi`
- Replace `PushNotificationLogger` dependency with generic `Logger`
- Remove unused `logWorkManagerInitializationAttempt` and `logWorkManagerInitializationFailed` from `PushNotificationLogger` and their tests
- Add WorkManager dependency to core
- Update push module imports

### chore: rename `WorkManagerProvider` to `CustomerIOWorkManagerProvider`

- Rename for consistency with `CustomerIOHttpClient` since it is now a public class in core
- Only import and type parameter changes in tests, no variable/function renames

### Why

The geofence module needs both direct HTTP and WorkManager for event delivery (same patterns as push delivery metrics). Without these moves, location module would need to depend on the push module or duplicate the implementations.

---

> [!NOTE]
> **Medium Risk**
> Touches shared networking/authenticated request plumbing and WorkManager initialization used by background delivery, so miswiring DI or package/visibility changes could break runtime behavior across modules.
> 
> **Overview**
> Moves shared infrastructure out of `:messagingpush` into `:core` by introducing `CustomerIOHttpClient` (renamed from `HttpClient`) and `CustomerIOWorkManagerProvider` (renamed from `WorkManagerProvider`) as `@InternalCustomerIOApi` types.
> 
> Updates DI so `SDKComponent.httpClient` and `SDKComponent.workManagerProvider` are provided from `core` (`SDKComponentExt.kt`), adds the WorkManager dependency to `core`, and switches push code/tests to the new packages/types. Also removes push-specific WorkManager init logging helpers/tests, with `CustomerIOWorkManagerProvider` now logging directly via the generic `Logger`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5773e0c85ed6fdc6be99d2ebd4616043af62b725. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -38,6 +38,7 @@ dependencies {
     api project(":base")
     api Dependencies.androidxCoreKtx
     implementation Dependencies.coroutinesAndroid
+    implementation Dependencies.workManager
     // Use this as API so customers can provide objects serializations without
     // needing to add it as a dependency to their app
     api(Dependencies.kotlinxSerializationJson)
--- a/core/src/main/kotlin/io/customer/sdk/core/util/CustomerIOWorkManagerProvider.kt
+++ b/core/src/main/kotlin/io/customer/sdk/core/util/CustomerIOWorkManagerProvider.kt
@@ -26,7 +26,7 @@ import io.customer.base.internal.InternalCustomerIOApi
  * These scenarios are handled based on documented exceptions from WorkManager's getInstance() and initialize() methods.
  */
 @InternalCustomerIOApi
-class WorkManagerProvider(
+class CustomerIOWorkManagerProvider(
     private val context: Context,
     private val logger: Logger
 ) {
--- a/core/src/main/kotlin/io/customer/sdk/core/util/CustomerIOWorkManagerProvider.kt
+++ b/core/src/main/kotlin/io/customer/sdk/core/util/CustomerIOWorkManagerProvider.kt
@@ -49,7 +49,7 @@ class CustomerIOWorkManagerProvider(
             // Try to get existing instance first
             WorkManager.getInstance(context)
         } catch (e: Exception) {
-            logger.error("WorkManager not initialized, attempting to initialize", throwable = e)
+            logger.error("WorkManager not initialized, attempting to initialize", tag = TAG, throwable = e)
             try {
                 // Check if host app implements Configuration.Provider, otherwise use default config
                 val config = (context.applicationContext as? Configuration.Provider)?.workManagerConfiguration ?: Configuration.Builder().build()
@@ -57,9 +57,13 @@ class CustomerIOWorkManagerProvider(
                 // Now try to get the instance again
                 WorkManager.getInstance(context)
             } catch (initException: Exception) {
-                logger.error("Failed to initialize WorkManager", throwable = initException)
+                logger.error("Failed to initialize WorkManager", tag = TAG, throwable = initException)
                 null
             }
         }
+    }
+
+    internal companion object {
+        private const val TAG = "WorkManager"
     }
 }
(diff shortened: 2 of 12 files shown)


core/src/main/kotlin/io/customer/sdk/core/util/CustomerIOWorkManagerProvider.kt
```diff
@@ -48,15 +49,15 @@ internal class WorkManagerProvider(
             // Try to get existing instance first
             WorkManager.getInstance(context)
         } catch (e: Exception) {
-            pushLogger.logWorkManagerInitializationAttempt(e)
+            logger.error("WorkManager not initialized, attempting to initialize", throwable = e)
```
mahmoud-elmorabea: These logs lose the tag used to be there, maybe we can create a `CioWorkManager` tag or smth shorter 😄 ?

core/build.gradle
```diff
@@ -38,6 +38,7 @@ dependencies {
     api project(":base")
     api Dependencies.androidxCoreKtx
     implementation Dependencies.coroutinesAndroid
+    implementation Dependencies.workManager
```
mahmoud-elmorabea, line 41: Should this be `api`? For now this works because push module has it as well, but do we want to keep adding it to all modules that use it?

mahmoud-elmorabea (approved):

core/src/main/kotlin/io/customer/sdk/core/util/CustomerIOWorkManagerProvider.kt
mrehan27: Added `WorkManager` tag to it, since it is using `Logger` class, `[CIO]` will be used automatically for log tag

core/build.gradle
mrehan27: Keeping `implementation` as changing to `api` would expose `WorkManager` classes in public classpath, making them visible to customers' apps. Each module that needs `WorkManager` can declare its own implementation dependency instead for now.

mahmoud-elmorabea: I don't think that's true, `api` would expose the types only to the direct consumers not our customers. You can try it out and see if that exposes the types to our sample apps.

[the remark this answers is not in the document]
mrehan27: Each module that uses `CustomerIOWorkManagerProvider` also uses WorkManager types directly so they already have their own `implementation Dependencies.workManager`. The redundant dependency is intentional as keeping `implementation` in `core` prevents leaking `WorkManager` to modules that don't need it (`datapipelines`, `messaginginapp`) and avoids exposing `WorkManager` classes to customers' apps.

mrehan27: I believe `api` is transitive through the dependency chain. Also, I tested and `WorkManager` types do become available in the sample app's compile classpath when `core` uses `api`. Keeping `implementation` in core and each consuming module avoids this exposure in sample app.

PR #673, opened 2026-04-24
chore: add geofence OS listen layer

mrehan27:
part of: [MBL-1625](https://linear.app/customerio/issue/MBL-1625/android-handle-os-geofence-monitoring)

## Summary

Foundation for geofence monitoring in `location` module. This PR introduces the OS integration layer, the classes needed to register geofences with Android's GeofencingClient, receive transition callbacks, and re-register after device reboot.

### What's included

- **GeofenceRegion**: Data model with `distanceTo()` and `toGmsTransitionTypes()` extensions
- **GeofenceManager**: Wraps `GeofencingClient` with suspend API for add/remove/clear operations, proper permission checks (`ACCESS_FINE_LOCATION` + `ACCESS_BACKGROUND_LOCATION` on Q+), and movement trigger initial-trigger handling
- **GeofenceReceiverToggle**: Injectable wrapper for `PackageManager.setComponentEnabledSetting`, enabling/disabling receivers dynamically when geofences are registered or cleared
- **GeofenceBroadcastReceiver**: Receives OS geofence transitions (stub, event delivery wired in next PR)
- **GeofenceBootReceiver**: Re-registers geofences after reboot (stub, restore logic wired in later PR)
- **GeofenceLogger**: Structured log methods following `PushNotificationLogger` pattern
- **GeofenceConstants**: All thresholds and identifiers
- **AndroidManifest**: Receiver declarations with `android:enabled="false"` (dynamically enabled by GeofenceManager). Geofence permissions not declared by SDK, host app must request them
- **DI wiring**: `GeofencingClient`, `GeofenceReceiverToggle`, `GeofenceManager`, `GeofenceLogger` on appropriate DI graphs (`AndroidSDKComponent` for Android-dependent, `SDKComponent` for pure)

### What's not included (coming in next PRs)

- Event delivery via WorkManager
- Persistence, sync, and movement trigger loop
- Module integration and lifecycle wiring

---

> [!NOTE]
> **Medium Risk**
> Adds new Android manifest receivers and permission-dependent geofencing registration logic, which can impact app behavior if enabled/misconfigured. Functionality is mostly additive and guarded (receivers disabled by default) with test coverage, reducing rollout risk.
> 
> **Overview**
> Adds the *Android OS integration layer* for geofencing in the `location` module, including a `GeofenceManager` wrapper around `GeofencingClient` (suspending add/remove/clear APIs, permission gating for fine + background location, request batching/initial-trigger handling, and receiver enable/disable).
> 
> Introduces new broadcast receivers for geofence transitions and reboot handling (`GeofenceBroadcastReceiver`, `GeofenceBootReceiver`), wires them into DI, and declares them in the manifest as **disabled by default** (host apps must supply required permissions). Adds supporting types/constants (`GeofenceRegion`, `GeofenceReceiverToggle`, `GeofenceLogger`, `GeofenceConstants`) plus Robolectric tests covering manager behavior and transition bitmask conversion.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 33c454244020eacb5cf0f6bcc907a5c9e9d1f24e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/location/src/main/kotlin/io/customer/location/geofence/GeofenceBroadcastReceiver.kt
+++ b/location/src/main/kotlin/io/customer/location/geofence/GeofenceBroadcastReceiver.kt
@@ -1,0 +1,44 @@
+package io.customer.location.geofence
+
+import android.content.BroadcastReceiver
+import android.content.Context
+import android.content.Intent
+import com.google.android.gms.location.GeofencingEvent
+import io.customer.location.geofence.di.geofenceLogger
+import io.customer.sdk.core.di.SDKComponent
+import io.customer.sdk.core.di.setupAndroidComponent
+
+/** Receives OS geofence transition callbacks and dispatches them to the SDK. */
+class GeofenceBroadcastReceiver : BroadcastReceiver() {
+
+    override fun onReceive(context: Context, intent: Intent) {
+        SDKComponent.setupAndroidComponent(context = context)
+
+        try {
+            handleGeofenceEvent(intent)
+        } catch (e: Exception) {
+            SDKComponent.geofenceLogger.logSyncFailed("BroadcastReceiver error: ${e.message}")
+        }
+    }
+
+    private fun handleGeofenceEvent(intent: Intent) {
+        val geofencingEvent = GeofencingEvent.fromIntent(intent) ?: return
+        val logger = SDKComponent.geofenceLogger
+
+        if (geofencingEvent.hasError()) {
+            logger.logGeofencingError(geofencingEvent.errorCode)
+            return
+        }
+
+        val transitionType = geofencingEvent.geofenceTransition
+        val triggeringGeofences = geofencingEvent.triggeringGeofences ?: return
+        val location = geofencingEvent.triggeringLocation
+
+        triggeringGeofences.forEach { geofence ->
+            logger.logTransitionReceived(
+                geofence.requestId,
+                "type=$transitionType at (${location?.latitude}, ${location?.longitude})"
+            )
+        }
+    }
+}
--- a/location/src/main/kotlin/io/customer/location/geofence/GeofenceLogger.kt
+++ b/location/src/main/kotlin/io/customer/location/geofence/GeofenceLogger.kt
@@ -1,0 +1,51 @@
+package io.customer.location.geofence
+
+import io.customer.sdk.core.util.Logger
+
+/** Structured logger for geofence operations, tagged for logcat filtering. */
+internal class GeofenceLogger(private val logger: Logger) {
+
+    fun logGeofencesRegistered(count: Int) {
+        logger.debug("Registered $count geofences with OS", tag = TAG)
+    }
+
+    fun logGeofencesRemoved(count: Int) {
+        logger.debug("Removed $count geofences from OS", tag = TAG)
+    }
+
+    fun logGeofencesCleared() {
+        logger.debug("Cleared all geofences from OS", tag = TAG)
+    }
+
+    fun logRegistrationFailed(message: String?) {
+        logger.error("Failed to register geofences: $message", tag = TAG)
+    }
+
+    fun logRemovalFailed(message: String?) {
+        logger.error("Failed to remove geofences: $message", tag = TAG)
+    }
+
+    fun logMissingPermission(permission: String) {
+        logger.error("Cannot register geofences: $permission not granted. Host app must request this permission.", tag = TAG)
+    }
+
+    fun logTransitionReceived(geofenceId: String, transitionName: String) {
+        logger.debug("Transition $transitionName for geofence $geofenceId", tag = TAG)
+    }
+
+    fun logGeofencingError(errorCode: Int) {
+        logger.error("Geofencing error code: $errorCode", tag = TAG)
+    }
+
+    fun logSyncFailed(message: String?) {
+        logger.error("Geofence sync failed: $message", tag = TAG)
+    }
+
+    fun logReceiverSkipped(reason: String) {
+        logger.debug("Geofence receiver skipped: $reason", tag = TAG)
+    }
+
+    companion object {
+        private const val TAG = "Geofence"
+    }
+}
--- a/location/src/main/kotlin/io/customer/location/geofence/GeofenceManager.kt
+++ b/location/src/main/kotlin/io/customer/location/geofence/GeofenceManager.kt
@@ -1,0 +1,171 @@
+package io.customer.location.geofence
+
+import android.Manifest
+import android.app.PendingIntent
+import android.content.Context
+import android.content.Intent
+import android.content.pm.PackageManager
+import android.os.Build
+import androidx.annotation.RequiresPermission
+import androidx.core.content.ContextCompat
+import com.google.android.gms.location.Geofence
+import com.google.android.gms.location.GeofencingClient
+import com.google.android.gms.location.GeofencingRequest
+import kotlin.coroutines.resume
+import kotlinx.coroutines.suspendCancellableCoroutine
+
+/** Wraps GeofencingClient to register/remove geofences with the OS. */
+internal class GeofenceManager(
+    private val context: Context,
+    private val client: GeofencingClient,
+    private val receiverToggle: GeofenceReceiverToggle,
+    private val logger: GeofenceLogger
+) {
+
+    private val pendingIntent: PendingIntent by lazy {
+        val intent = Intent(context, GeofenceBroadcastReceiver::class.java)
+        val flags = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
+            PendingIntent.FLAG_UPDATE_CURRENT or PendingIntent.FLAG_MUTABLE
+        } else {
+            PendingIntent.FLAG_UPDATE_CURRENT
+        }
+        PendingIntent.getBroadcast(
+            context,
+            GeofenceConstants.PENDING_INTENT_REQUEST_CODE,
+            intent,
+            flags
+        )
+    }
+
+    @RequiresPermission(allOf = [Manifest.permission.ACCESS_FINE_LOCATION, Manifest.permission.ACCESS_BACKGROUND_LOCATION])
+    suspend fun addGeofences(regions: List<GeofenceRegion>): Result<Unit> {
+        if (regions.isEmpty()) return Result.success(Unit)
+        if (!hasRequiredPermissions()) {
+            return Result.failure(SecurityException("Required location permissions not granted"))
+        }
+
+        // Split into separate requests: GMS only supports one initial-trigger per request.
+        // Movement trigger uses no initial trigger to avoid spurious EXIT on re-registration.
+        // Business geofences use INITIAL_TRIGGER_ENTER for immediate detection if already inside.
+        val movementTrigger = regions.filter { it.id == GeofenceConstants.MOVEMENT_TRIGGER_ID }
+        val businessGeofences = regions.filter { it.id != GeofenceConstants.MOVEMENT_TRIGGER_ID }
+
+        if (movementTrigger.isNotEmpty()) {
+            val result = registerBatch(movementTrigger, initialTrigger = 0)
+            if (result.isFailure) return result
+        }
+
+        if (businessGeofences.isNotEmpty()) {
+            val result = registerBatch(businessGeofences, initialTrigger = GeofencingRequest.INITIAL_TRIGGER_ENTER)
+            if (result.isFailure) {
+                // Clean up movement trigger if business batch fails to avoid orphaned geofences
+                if (movementTrigger.isNotEmpty()) {
+                    removeGeofencesByIds(movementTrigger.map { it.id })
+                }
+                return result
+            }
+        }
+
+        receiverToggle.setEnabled(true)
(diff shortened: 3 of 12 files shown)


location/src/main/kotlin/io/customer/location/geofence/GeofenceReceiverToggle.kt
```diff
@@ -0,0 +1,31 @@
+package io.customer.location.geofence
+
+import android.content.ComponentName
+import android.content.Context
+import android.content.pm.PackageManager
+
+/** Enables/disables geofence broadcast receivers via PackageManager. */
+internal interface GeofenceReceiverToggle {
```
mahmoud-elmorabea: I think this can just be a class, we can make it an interface if we ever need polymorphic behavior

location/src/main/kotlin/io/customer/location/geofence/GeofenceReceiverToggle.kt
```diff
@@ -0,0 +1,31 @@
...
+    fun setEnabled(enabled: Boolean)
+}
+
+internal class GeofenceReceiverToggleImpl(
+    private val context: Context
+) : GeofenceReceiverToggle {
+
+    override fun setEnabled(enabled: Boolean) {
```
mahmoud-elmorabea: Cool idea for folks not using Geofences 👍

location/src/main/kotlin/io/customer/location/geofence/GeofenceBroadcastReceiver.kt
```diff
@@ -0,0 +1,44 @@
...
+class GeofenceBroadcastReceiver : BroadcastReceiver() {
+
+    override fun onReceive(context: Context, intent: Intent) {
+        SDKComponent.setupAndroidComponent(context = context)
+
+        try {
+            handleGeofenceEvent(intent)
+        } catch (e: Exception) {
+            SDKComponent.geofenceLogger.logSyncFailed("BroadcastReceiver error: ${e.message}")
+        }
```
mahmoud-elmorabea, line 21: In future PRs consider using [goAsync()](https://developer.android.com/reference/android/content/BroadcastReceiver#goAsync()) to tell the OS you are doing async stuff

location/src/main/kotlin/io/customer/location/geofence/GeofenceManager.kt
```diff
@@ -0,0 +1,171 @@
...
+import androidx.annotation.RequiresPermission
+import androidx.core.content.ContextCompat
+import com.google.android.gms.location.Geofence
+import com.google.android.gms.location.GeofencingClient
+import com.google.android.gms.location.GeofencingRequest
+import kotlin.coroutines.resume
+import kotlinx.coroutines.suspendCancellableCoroutine
+
+/** Wraps GeofencingClient to register/remove geofences with the OS. */
+internal class GeofenceManager(
```
mahmoud-elmorabea, line 18: I think this manager might be doing a tad too much. Maybe we can create a `GeofencingClient` wrapper that makes unit testing easier and have it know how to register a group of fences.

Then we can create an orchestration layer on top of it that understands movements vs business, that doesn't care about pending intents or GSM APIs.

location/src/main/kotlin/io/customer/location/geofence/GeofenceManager.kt
```diff
@@ -0,0 +1,171 @@
...
+            context,
+            GeofenceConstants.PENDING_INTENT_REQUEST_CODE,
+            intent,
+            flags
+        )
+    }
+
+    @RequiresPermission(allOf = [Manifest.permission.ACCESS_FINE_LOCATION, Manifest.permission.ACCESS_BACKGROUND_LOCATION])
+    suspend fun addGeofences(regions: List<GeofenceRegion>): Result<Unit> {
+        if (regions.isEmpty()) return Result.success(Unit)
```
mahmoud-elmorabea: Maybe this should log something, it doesn't seem like a natural success case, this will probably be masking an issue somewhere upstream of this call

location/src/main/kotlin/io/customer/location/geofence/GeofenceManager.kt
```diff
@@ -0,0 +1,171 @@
...
+
+        // Split into separate requests: GMS only supports one initial-trigger per request.
+        // Movement trigger uses no initial trigger to avoid spurious EXIT on re-registration.
+        // Business geofences use INITIAL_TRIGGER_ENTER for immediate detection if already inside.
+        val movementTrigger = regions.filter { it.id == GeofenceConstants.MOVEMENT_TRIGGER_ID }
+        val businessGeofences = regions.filter { it.id != GeofenceConstants.MOVEMENT_TRIGGER_ID }
+
+        if (movementTrigger.isNotEmpty()) {
+            val result = registerBatch(movementTrigger, initialTrigger = 0)
+            if (result.isFailure) return result
```
mahmoud-elmorabea, line 58: If registration of the movement trigger fails, we will ignore the rest of the regions?

location/src/main/kotlin/io/customer/location/geofence/GeofenceManager.kt
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        // Split into separate requests: GMS only supports one initial-trigger per request.
+        // Movement trigger uses no initial trigger to avoid spurious EXIT on re-registration.
+        // Business geofences use INITIAL_TRIGGER_ENTER for immediate detection if already inside.
+        val movementTrigger = regions.filter { it.id == GeofenceConstants.MOVEMENT_TRIGGER_ID }
+        val businessGeofences = regions.filter { it.id != GeofenceConstants.MOVEMENT_TRIGGER_ID }
+
+        if (movementTrigger.isNotEmpty()) {
+            val result = registerBatch(movementTrigger, initialTrigger = 0)
```
mahmoud-elmorabea: Maybe we can create a constant for 0 to mean no triggers

location/src/main/kotlin/io/customer/location/geofence/GeofenceManager.kt
mahmoud-elmorabea, line 18: It's difficult to assess the functions this class exposes when it's not used in this PR

location/src/main/kotlin/io/customer/location/geofence/GeofenceRegion.kt
```diff
@@ -0,0 +1,43 @@
...
+import android.location.Location
+import com.google.android.gms.location.Geofence
+
+/** A geographic region to monitor for enter/exit transitions. */
+internal data class GeofenceRegion(
+    val id: String,
+    val latitude: Double,
+    val longitude: Double,
+    val radiusMeters: Float,
+    val name: String = "",
```
mahmoud-elmorabea, line 12: If `name` is optional maybe a nullable property better communicates that

location/src/main/kotlin/io/customer/location/geofence/GeofenceRegion.kt
```diff
@@ -0,0 +1,43 @@
...
+    val transitionTypes: List<GeofenceTransitionType> = listOf(
```
mahmoud-elmorabea, line 13: Does this need to be a property on the data class, seems like a static list to me?

location/src/main/kotlin/io/customer/location/geofence/GeofenceRegion.kt
```diff
@@ -0,0 +1,43 @@
...
+    ENTER(Geofence.GEOFENCE_TRANSITION_ENTER),
+    EXIT(Geofence.GEOFENCE_TRANSITION_EXIT)
+}
+
+/**
+ * Straight-line distance in meters from this region's center to the given coordinates.
+ */
+internal fun GeofenceRegion.distanceTo(lat: Double, lng: Double): Float {
+    val result = FloatArray(1)
+    Location.distanceBetween(latitude, longitude, lat, lng, result)
```
mahmoud-elmorabea, line 35: I believe this [throws](https://developer.android.com/reference/android/location/Location#distanceBetween(double,%20double,%20double,%20double,%20float[])) if lat or lng are not valid coordinates . Let's either make sure that the caller validates or it does

location/src/main/kotlin/io/customer/location/geofence/GeofenceReceiverToggle.kt
mrehan27: Converted to a class. We can add an interface if polymorphic behavior is needed later.

location/src/main/kotlin/io/customer/location/geofence/GeofenceBroadcastReceiver.kt
mrehan27: Good call 👍🏻 Will add `goAsync()` in upcoming PRs when the receiver does async work (WorkManager scheduling + EventBus publishing). In this PR it's still a stub that only logs.

location/src/main/kotlin/io/customer/location/geofence/GeofenceManager.kt
mrehan27: Valid point. For now `GeofenceManager` is the only GMS touchpoint and `registerBatch` already separates the low level registration. Will consider splitting with `GeofenceRepository` in PR 3 as at that point the boundary between GMS wrapper and business logic will be clearer.

location/src/main/kotlin/io/customer/location/geofence/GeofenceManager.kt
mrehan27: Added a debug log so empty calls are visible in logs.

location/src/main/kotlin/io/customer/location/geofence/GeofenceManager.kt
mrehan27: Intentional. If movement trigger registration fails (permissions, GMS unavailable), business geofences will fail for the same reason. And business geofences without a movement trigger won't get re-evaluated when the user moves as they'd go stale. Failing early avoids a broken state.

location/src/main/kotlin/io/customer/location/geofence/GeofenceManager.kt
mrehan27: Added `NO_INITIAL_TRIGGER` constant in `GeofenceConstants`.

location/src/main/kotlin/io/customer/location/geofence/GeofenceManager.kt
mrehan27: This is used in PR 3 (`GeofenceRepository`) which uses operations like add/remove/clear. This PR establishes the GMS integration layer that the repository builds on.

location/src/main/kotlin/io/customer/location/geofence/GeofenceRegion.kt
mrehan27: The API model isn't fully finalized yet unfortunately. will adjust name and other fields when the server response shape is defined. Keeping `String=""` for now to avoid churn.

location/src/main/kotlin/io/customer/location/geofence/GeofenceRegion.kt
mrehan27: Not static, the server is expected to configure transition types per geofence (`ENTER` only, `EXIT` only, or both). The default is both, but each geofence from the API can have different values. It needs to be per-instance.

location/src/main/kotlin/io/customer/location/geofence/GeofenceRegion.kt
mrehan27: Agreed. Added `@throws` doc on `distanceTo()` noting the coordinate range requirement. Actual validation will be in `GeofenceApiRegion.toDomain()` (PR 3) at the API boundary where server data enters the SDK.

mahmoud-elmorabea (approved):

location/src/main/kotlin/io/customer/location/geofence/GeofenceLogger.kt
```diff
@@ -0,0 +1,51 @@
+package io.customer.location.geofence
+
+import io.customer.sdk.core.util.Logger
+
+/** Structured logger for geofence operations, tagged for logcat filtering. */
+internal class GeofenceLogger(private val logger: Logger) {
```
hollyschilling-cio, line 6: I know we use this pattern elsewhere in both SDKs, but this seems like a strange pattern. Most, if not all, of these logging messages are only used once. Why wrap the logger and implement a module specific logger? 

The output log text is odd too. For example, with `logSyncFailed` the only use case will always output, `"Geofence sync failed: BroadcastReceiver error: ${e.message}"`.

hollyschilling-cio (approved):

location/src/main/kotlin/io/customer/location/geofence/GeofenceLogger.kt
mrehan27: I'll improve the log messages in upcoming PRs. For your other point about dedicated logger, I followed the current project convention, but we can discuss improving that later some time if needed.

PR #674, opened 2026-04-27
chore: add store for encrypted user identity caching

mrehan27:
part of: [MBL-1627](https://linear.app/customerio/issue/MBL-1627/android-send-geofence-transition-events)

## Summary

Add `SecureUserStore` — an encrypted, user-scoped preference store in `:core` for persisting sensitive data that must survive app restarts and be accessible without full SDK initialization.

### Why

Geofence events need to include `userId` in direct API calls made from WorkManager workers and BroadcastReceivers. In killed state (especially RN/Flutter wrappers), the SDK isn't initialized so`analytics.userId()` is unavailable. This store caches the identity encrypted so workers can read it independently.

### What

- **`SecureUserStore`** interface + `SecureUserStoreImpl` in `core/data/store/`
  - Encrypted via `PreferenceCrypto` (AES-256-GCM, Android Keystore)
  - Separate prefs file (`io.customer.sdk.secure_user.<package>`) — no migration risk with existing stores
  - User-scoped: cleared on `clearIdentify`/`reset` (unlike `GlobalPreferenceStore` which is device-scoped)
- **`AndroidSDKComponent.secureUserStore`** — singleton, accessible via `SDKComponent.android().secureUserStore`
- **`CustomerIO`** wiring:
  - `UserChangedEvent` subscription saves `userId` on identify
  - `ResetEvent` subscription calls `clearAll()` on reset

### Design

| Store | Scope | Encrypted | Survives reset |
|---|---|---|---|
| `GlobalPreferenceStore` | Device | No | Yes |
| `SecureUserStore` | User | Yes | No |

---

> [!NOTE]
> **Medium Risk**
> Introduces new persistence for user identity and Keystore-backed encryption; mistakes could leak/retain IDs across sessions or break background flows if encryption/decryption fails.
> 
> **Overview**
> Adds a new `SecureUserStore` (and `SecureUserStoreImpl`) in `:core` to persist the current `userId` in a dedicated SharedPreferences file, encrypting values at rest via `PreferenceCrypto`.
> 
> Exposes the store through `AndroidSDKComponent` as a singleton dependency, and updates `CustomerIO` to keep it in sync by saving the `userId` on `Event.UserChangedEvent` and clearing it on `Event.ResetEvent` for use by background workers/receivers without full SDK initialization.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d0a6098aa878593622f909de8d50250b3096a59f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/core/src/main/kotlin/io/customer/sdk/data/store/SecureUserStore.kt
+++ b/core/src/main/kotlin/io/customer/sdk/data/store/SecureUserStore.kt
@@ -2,6 +2,7 @@ package io.customer.sdk.data.store
 
 import android.content.Context
 import androidx.core.content.edit
+import io.customer.base.internal.InternalCustomerIOApi
 import io.customer.sdk.core.util.Logger
 
 /**
@@ -16,6 +17,7 @@ import io.customer.sdk.core.util.Logger
  * Currently stores user identity for direct API calls from WorkManager
  * workers and BroadcastReceivers (e.g., geofence events, push metrics).
  */
+@InternalCustomerIOApi
 interface SecureUserStore {
     fun saveUserId(userId: String?)
     fun getUserId(): String?
--- a/core/src/main/kotlin/io/customer/sdk/data/store/SecureUserStore.kt
+++ b/core/src/main/kotlin/io/customer/sdk/data/store/SecureUserStore.kt
@@ -1,0 +1,54 @@
+package io.customer.sdk.data.store
+
+import android.content.Context
+import androidx.core.content.edit
+import io.customer.sdk.core.util.Logger
+
+/**
+ * Encrypted, user-scoped preference store for sensitive data that must
+ * persist across app restarts and be accessible without full SDK init.
+ *
+ * Unlike [GlobalPreferenceStore] (device-scoped, survives reset), this
+ * store is cleared on clearIdentify/reset — all data is tied to the
+ * current user session.
+ *
+ * All values are encrypted at rest via [PreferenceCrypto] (AES-256-GCM).
+ * Currently stores user identity for direct API calls from WorkManager
+ * workers and BroadcastReceivers (e.g., geofence events, push metrics).
+ */
+interface SecureUserStore {
+    fun saveUserId(userId: String?)
+    fun getUserId(): String?
+    fun clearAll()
+}
+
+internal class SecureUserStoreImpl(
+    context: Context,
+    logger: Logger
+) : PreferenceStore(context), SecureUserStore {
+
+    private val crypto = PreferenceCrypto(KEY_ALIAS, logger)
+
+    override val prefsName: String by lazy {
+        "io.customer.sdk.secure_user.${context.packageName}"
+    }
+
+    override fun saveUserId(userId: String?) = prefs.edit {
+        if (userId != null) {
+            putString(KEY_USER_ID, crypto.encrypt(userId))
+        } else {
+            remove(KEY_USER_ID)
+        }
+    }
+
+    override fun getUserId(): String? = prefs.read {
+        getString(KEY_USER_ID, null)?.let { crypto.decrypt(it) }
+    }
+
+    override fun clearAll() = prefs.edit { clear() }
+
+    companion object {
+        private const val KEY_ALIAS = "cio_secure_user_key"
+        private const val KEY_USER_ID = "cio_secure_user_id"
+    }
+}
--- a/datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
+++ b/datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
@@ -71,6 +71,7 @@ class CustomerIO private constructor(
     private val logger: Logger = SDKComponent.logger
     private val dataPipelinesLogger: DataPipelinesLogger = SDKComponent.dataPipelinesLogger
     private val globalPreferenceStore = androidSDKComponent.globalPreferenceStore
+    private val secureUserStore = androidSDKComponent.secureUserStore
     private val deviceStore = androidSDKComponent.deviceStore
     private val eventBus = SDKComponent.eventBus
     internal var migrationProcessor: MigrationProcessor? = null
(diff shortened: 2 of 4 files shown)


core/src/main/kotlin/io/customer/sdk/data/store/SecureUserStore.kt
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * Unlike [GlobalPreferenceStore] (device-scoped, survives reset), this
+ * store is cleared on clearIdentify/reset — all data is tied to the
+ * current user session.
+ *
+ * All values are encrypted at rest via [PreferenceCrypto] (AES-256-GCM).
+ * Currently stores user identity for direct API calls from WorkManager
+ * workers and BroadcastReceivers (e.g., geofence events, push metrics).
+ */
+interface SecureUserStore {
```
mahmoud-elmorabea, line 21: Does this need to be an interface?

datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
```diff
@@ -158,6 +159,14 @@ class CustomerIO private constructor(
...
             registerDeviceToken(deviceToken = it.token)
         }
+        // Cache user identity encrypted for direct API calls (WorkManager workers,
+        // BroadcastReceivers) that run without full SDK initialization.
+        eventBus.subscribe<Event.UserChangedEvent> {
+            secureUserStore.saveUserId(it.userId)
+        }
+        eventBus.subscribe<Event.ResetEvent> {
+            secureUserStore.clearAll()
+        }
```
mahmoud-elmorabea, line 169: When I attempted to do this previously, I found out that `UserChangedEvent` is triggered as null when logout happens so this will probably double clear

mahmoud-elmorabea: Any reason why not directly call the storage at `identify` and `clearIdentity` instead of using the event bus??

core/src/main/kotlin/io/customer/sdk/data/store/SecureUserStore.kt
mahmoud-elmorabea, line 21: Should this be internal CIO API?

core/src/main/kotlin/io/customer/sdk/data/store/SecureUserStore.kt
```diff
@@ -0,0 +1,54 @@
...
+    logger: Logger
+) : PreferenceStore(context), SecureUserStore {
+
+    private val crypto = PreferenceCrypto(KEY_ALIAS, logger)
+
+    override val prefsName: String by lazy {
+        "io.customer.sdk.secure_user.${context.packageName}"
+    }
+
+    override fun saveUserId(userId: String?) = prefs.edit {
```
mahmoud-elmorabea, line 38: Maybe add tests for this class as well

core/src/main/kotlin/io/customer/sdk/data/store/SecureUserStore.kt
```diff
@@ -0,0 +1,54 @@
...
+ * Currently stores user identity for direct API calls from WorkManager
+ * workers and BroadcastReceivers (e.g., geofence events, push metrics).
+ */
+interface SecureUserStore {
+    fun saveUserId(userId: String?)
+    fun getUserId(): String?
+    fun clearAll()
+}
+
+internal class SecureUserStoreImpl(
```
mahmoud-elmorabea, line 27: will we directly instantiate this for broadcasts and workers?

core/src/main/kotlin/io/customer/sdk/data/store/SecureUserStore.kt
mrehan27: Yes, it helps with mocking when writing tests for classes that depend on this store.

datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
mrehan27: I would keep `EventBus` as it covers all paths that change identity automatically. Direct calls would require adding `secureUserStore` calls to every new identify path.

> UserChangedEvent is triggered as null when logout

Yes, it does get called. The `ResetEvent` subscription with `clearAll()` is kept as a safety measure as it clears the entire prefs file. This helps future proof things when more fields are added to the store without needing to handle each one individually in `UserChangedEvent`.

core/src/main/kotlin/io/customer/sdk/data/store/SecureUserStore.kt
mrehan27: Updated

core/src/main/kotlin/io/customer/sdk/data/store/SecureUserStore.kt
mrehan27: The store is a thin wrapper over `SharedPreferences` + `PreferenceCrypto`, so testing it directly would mostly validate Android APIs rather than our logic. It will be tested indirectly through geofence event tracker tests later, which verify that `userId` is included in API calls.

core/src/main/kotlin/io/customer/sdk/data/store/SecureUserStore.kt
mrehan27: It will be accessed via DI (`SDKComponent.android().secureUserStore`). Broadcasts/workers call `setupAndroidComponent` first, then access the store through the DI graph, the same pattern as other stores.

datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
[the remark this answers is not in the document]
mrehan27: [Replied above](https://github.com/customerio/customerio-android/pull/674#discussion_r3167200726). I think we should also improve this to avoid confusion with `ResetEvent`, but will address it in a separate PR.

mahmoud-elmorabea (approved):

hollyschilling-cio (approved):
Looks good to merge.

PR #677, opened 2026-05-04
chore: skip periodic flush while backgrounded on Android 15+

mrehan27:
## Summary

Linear: [MBL-1499 — Noisy error logs for background HTTP calls on Android 15](https://linear.app/customerio/issue/MBL-1499/noisy-error-logs-for-background-http-calls-on-android-15)

On Android 15+ the OS restricts background network access for non-foreground apps. The bundled Segment `analytics-kotlin` library's `FrequencyFlushPolicy` keeps polling every 30s regardless of app state, so each backgrounded flush attempt fails with `UnknownHostException` and produces three repeating `[CIO]` log lines until the next foreground:

```
D Customer.io Data Pipelines performing flush
E Unable to resolve host "cdp.customer.io": No address associated with hostname
E Error uploading events from batch file fileUrl=… msg=Unable to resolve host …
```

iOS already does the equivalent gating in [`IntervalBasedFlushPolicy.enterBackground/enterForeground`](https://github.com/customerio/cdp-analytics-swift/blob/main/Sources/Segment/Utilities/Policies/IntervalBasedFlushPolicy.swift) — this PR brings Android to parity.

## What changed

- **`BackgroundAwareFrequencyFlushPolicy`** — a `FlushPolicy` whose periodic loop calls `foregroundState.isInForeground()` on each tick and only flushes when the app is in foreground.
- **`AppForegroundState`** interface + **`ProcessLifecycleForegroundState`** impl — exposes a `suspend fun isInForeground()` that reads `Lifecycle.currentState` inside `withContext(dispatchers.main) { … }`, honoring the `@MainThread` contract regardless of caller's dispatcher.
- **`CustomerIODestination`** picks the new policy on API 35+ and keeps Segment's vanilla `FrequencyFlushPolicy` on older Android (no behavior change for pre-Android-15 devices).
- Default builder behavior, no opt-in flag.

## What didn't change

- The one-shot `analytics.flush()` triggered by `Application Backgrounded` still runs (it fires before OS network restrictions kick in, so it usually drains the queue successfully).
- Count-based and manual `flush()` paths are untouched.
- Foreground behavior — including airplane-mode error logging — is unchanged. Real network outages still log at ERROR.
- API < 35 behavior is bit-for-bit identical.

## Why this approach

Considered alternatives:
- **Suppress the noisy logs** — simpler in code, but still requires lifecycle plumbing for the gate, and string-matches Segment's log strings (brittle across upstream bumps).
- **Bump Segment to 1.25.0** — verified upstream's `EventPipeline` and `FrequencyFlushPolicy` are unchanged for this scenario. The fix would still be required.
- **Stop the entire `EventPipeline` on background** — drops events queued while backgrounded; unsafe.

This PR fixes the root cause at the smallest correct scope.

### Automated

- [x] `./gradlew :datapipelines:assembleDebug` — passes
- [x] `./gradlew :datapipelines:testDebugUnitTest` — 199 tests, 0 failures (13 new across 2 new test classes)
- [x] `make lint-no-format` — clean

#### Android 15 (API 35) or higher — primary scenario

- [x] Build sample app: `./gradlew :samples:java_layout:installDebug` (or `:samples:kotlin_compose:installDebug`)
- [x] Open the app, identify a user, trigger a few events
- [x] Attach logcat: `adb logcat | grep '\[CIO\]'`
- [x] Background the app (Home button or Recents → switch to another app)
- [x] **Expected:** no recurring `[CIO]` log lines while backgrounded — specifically, no `performing flush` / `Unable to resolve host` / `Error uploading events from batch file` repeating every ~30 seconds
- [x] Wait at least 60 seconds in background to confirm
- [x] Bring the app back to foreground
- [x] **Expected:** within ~30s the periodic flush resumes; queued events upload on the next tick (or sooner if a count-based flush triggers)

#### Android 14 (API 34) — regression check

- [x] Repeat steps above on an API 34 device
- [x] **Expected:** behavior unchanged from before the PR — periodic flush continues in background (we use the upstream `FrequencyFlushPolicy` on this path)

#### minSdk 24 (Android 7) — boundary smoke test

- [x] Same flow on an API 24 device
- [x] **Expected:** no crash; behavior identical to API 34 (same code path)

#### Foreground network outage — sanity check (optional)

- [x] Disable network on an Android 15 device while in **foreground**, generate events, watch logs
- [x] **Expected:** real `Unable to resolve host` ERROR logs still appear (we only gate on app foreground state, not network availability — legitimate foreground errors still surface for debugging)

---

> [!NOTE]
> **Medium Risk**
> Changes flush scheduling behavior on Android 15+ by skipping periodic flushes while backgrounded, which could delay event uploads if foreground detection is wrong. Adds lifecycle/main-thread checks and coroutine scheduling that could introduce edge-case timing/cancellation issues.
> 
> **Overview**
> Prevents recurring background network flush attempts on Android 15+ by switching `CustomerIODestination`’s default interval-based flush policy to a new `BackgroundAwareFrequencyFlushPolicy` when `SDK_INT` is API 35+.
> 
> Adds `AppForegroundState`/`ProcessLifecycleForegroundState` to detect foreground via `ProcessLifecycleOwner` (read on the main dispatcher), and includes unit tests validating flush is skipped in background, resumes in foreground, and scheduling/unscheduling doesn’t spawn duplicate coroutines.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aa8170e13db3f58bcc645651df833ef4e1949e40. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/datapipelines/src/main/kotlin/io/customer/datapipelines/util/AppForegroundState.kt
+++ b/datapipelines/src/main/kotlin/io/customer/datapipelines/util/AppForegroundState.kt
@@ -1,0 +1,21 @@
+package io.customer.datapipelines.util
+
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.LifecycleOwner
+import androidx.lifecycle.ProcessLifecycleOwner
+
+/** Whether the app process is currently in foreground. */
+internal interface AppForegroundState {
+    val isInForeground: Boolean
+}
+
+/** [AppForegroundState] using [ProcessLifecycleOwner]; owner is lazy so the class is safe to construct in pure JVM unit tests. */
+internal class ProcessLifecycleForegroundState(
+    processLifecycleOwnerProvider: () -> LifecycleOwner = { ProcessLifecycleOwner.get() }
+) : AppForegroundState {
+
+    private val processLifecycleOwner: LifecycleOwner by lazy(processLifecycleOwnerProvider)
+
+    override val isInForeground: Boolean
+        get() = processLifecycleOwner.lifecycle.currentState.isAtLeast(Lifecycle.State.STARTED)
+}
--- a/datapipelines/src/main/kotlin/io/customer/datapipelines/util/AppForegroundState.kt
+++ b/datapipelines/src/main/kotlin/io/customer/datapipelines/util/AppForegroundState.kt
@@ -13,14 +13,18 @@ internal interface AppForegroundState {
 }
 
 /**
- * [AppForegroundState] backed by [ProcessLifecycleOwner]. The `@MainThread`
- * `currentState` getter is invoked via `withContext(dispatchers.main)` so the
- * read happens on the main thread regardless of where the caller is suspended.
+ * [AppForegroundState] backed by [ProcessLifecycleOwner]. The `LifecycleOwner`
+ * is resolved lazily on first read of [isInForeground], which always happens
+ * inside `withContext(dispatchers.main)` — so `ProcessLifecycleOwner.get()`
+ * and the `@MainThread` `currentState` getter both run on the main thread
+ * regardless of where the SDK was initialized.
  */
 internal class ProcessLifecycleForegroundState(
-    private val processLifecycleOwner: LifecycleOwner = ProcessLifecycleOwner.get(),
+    processLifecycleOwnerProvider: () -> LifecycleOwner = { ProcessLifecycleOwner.get() },
     private val dispatchersProvider: DispatchersProvider = SDKComponent.dispatchersProvider
 ) : AppForegroundState {
+
+    private val processLifecycleOwner: LifecycleOwner by lazy(processLifecycleOwnerProvider)
 
     override suspend fun isInForeground(): Boolean = withContext(dispatchersProvider.main) {
         processLifecycleOwner.lifecycle.currentState.isAtLeast(Lifecycle.State.STARTED)
--- a/datapipelines/src/main/kotlin/io/customer/datapipelines/util/AppForegroundState.kt
+++ b/datapipelines/src/main/kotlin/io/customer/datapipelines/util/AppForegroundState.kt
@@ -3,19 +3,26 @@ package io.customer.datapipelines.util
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.LifecycleOwner
 import androidx.lifecycle.ProcessLifecycleOwner
+import io.customer.sdk.core.di.SDKComponent
+import io.customer.sdk.core.util.DispatchersProvider
+import kotlinx.coroutines.withContext
 
 /** Whether the app process is currently in foreground. */
 internal interface AppForegroundState {
-    val isInForeground: Boolean
+    suspend fun isInForeground(): Boolean
 }
 
-/** [AppForegroundState] using [ProcessLifecycleOwner]; owner is lazy so the class is safe to construct in pure JVM unit tests. */
+/**
+ * [AppForegroundState] backed by [ProcessLifecycleOwner]. The `@MainThread`
+ * `currentState` getter is invoked via `withContext(dispatchers.main)` so the
+ * read happens on the main thread regardless of where the caller is suspended.
+ */
 internal class ProcessLifecycleForegroundState(
-    processLifecycleOwnerProvider: () -> LifecycleOwner = { ProcessLifecycleOwner.get() }
+    private val processLifecycleOwner: LifecycleOwner = ProcessLifecycleOwner.get(),
+    private val dispatchersProvider: DispatchersProvider = SDKComponent.dispatchersProvider
 ) : AppForegroundState {
 
-    private val processLifecycleOwner: LifecycleOwner by lazy(processLifecycleOwnerProvider)
-
-    override val isInForeground: Boolean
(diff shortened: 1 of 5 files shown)


datapipelines/src/main/kotlin/io/customer/datapipelines/util/AppForegroundState.kt
```diff
@@ -0,0 +1,21 @@
...
+internal interface AppForegroundState {
+    val isInForeground: Boolean
+}
+
+/** [AppForegroundState] using [ProcessLifecycleOwner]; owner is lazy so the class is safe to construct in pure JVM unit tests. */
+internal class ProcessLifecycleForegroundState(
+    processLifecycleOwnerProvider: () -> LifecycleOwner = { ProcessLifecycleOwner.get() }
+) : AppForegroundState {
+
+    private val processLifecycleOwner: LifecycleOwner by lazy(processLifecycleOwnerProvider)
```
mahmoud-elmorabea: I remember this not being happy when it's initialized/registered on background threads. I know it's tricky having this done on background thread but it's worth double checking.

datapipelines/src/main/kotlin/io/customer/datapipelines/util/AppForegroundState.kt
```diff
@@ -0,0 +1,21 @@
...
+
+    override val isInForeground: Boolean
+        get() = processLifecycleOwner.lifecycle.currentState.isAtLeast(Lifecycle.State.STARTED)
```
mahmoud-elmorabea: Can this be done on background thread?

datapipelines/src/test/java/io/customer/datapipelines/util/ProcessLifecycleForegroundStateTest.kt
```diff
@@ -0,0 +1,58 @@
...
+                mockk()
+            }
+        )
+
+        providerCalled shouldBeEqualTo false
+    }
+
+    @Test
+    fun isInForeground_givenStartedState_expectsTrue() {
+        ProcessLifecycleForegroundState(processLifecycleOwnerProvider = ownerInState(Lifecycle.State.STARTED))
```
mahmoud-elmorabea: Maybe add a test for INITIALIZED for completeness 😇

mahmoud-elmorabea (approved):
Good solution! Much better than just skipping logs. Do we know if iOS does this as well?

I suggest we test OS versions boundaries with this, min SDK 24, 34 and 35 to make sure they all work as expected 💯

mrehan27:
Thanks!

iOS does the same — [`IntervalBasedFlushPolicy.enterBackground`](https://github.com/customerio/cdp-analytics-swift/blob/main/Sources/Segment/Utilities/Policies/IntervalBasedFlushPolicy.swift) suspends the timer on background and resumes on foreground. So this brings Android to parity.

Added API 35 / 34 / 24 to the test plan in the description.

Also reshaped per the inline feedback — `suspend fun isInForeground()` with `withContext(dispatchers.main)` (details on each thread).

datapipelines/src/main/kotlin/io/customer/datapipelines/util/AppForegroundState.kt
mrehan27: Good catch — ditched lazy + observer entirely. Now:

```kotlin
override suspend fun isInForeground(): Boolean = withContext(dispatchersProvider.main) {
    processLifecycleOwner.lifecycle.currentState.isAtLeast(Lifecycle.State.STARTED)
}
```

`ProcessLifecycleOwner.get()` is bg-safe (static singleton, no `@MainThread`), so constructing from a worker thread is fine. The `@MainThread` ops fire later inside the `withContext`.

datapipelines/src/main/kotlin/io/customer/datapipelines/util/AppForegroundState.kt
mrehan27: Fixed — `currentState` is read inside `withContext(dispatchersProvider.main) { … }`, so it runs on main regardless of caller.

datapipelines/src/test/java/io/customer/datapipelines/util/ProcessLifecycleForegroundStateTest.kt
mrehan27: Added: `isInForeground_givenInitializedState_expectsFalse`. Tests are now state-driven (`STARTED`/`RESUMED` → true, `INITIALIZED`/`CREATED`/`DESTROYED` → false).

datapipelines/src/main/kotlin/io/customer/datapipelines/util/AppForegroundState.kt
mrehan27: Quick follow-up: after a bit more digging, decided to be doubly defensive and re-add the lazy — `processLifecycleOwnerProvider` is captured at construction but `get()` only fires on first read of `isInForeground()`, which always happens inside `withContext(dispatchers.main)`. So `get()` itself lands on main thread, in case AndroidX tightens enforcement on it later. Pushed in `aa8170e1`.

PR #683, opened 2026-05-11
chore: send geofence transition events via EventBus (MBL-1627) [1/3]

mrehan27:
## Linear ticket

[MBL-1627: Android — Send Geofence Transition Events](https://linear.app/customerio/issue/MBL-1627/android-send-geofence-transition-events)

## Stack

This is **PR 1 of 3** stacked PRs under MBL-1627. All three target `feature/geofence-on-device` and will be merged in order.

1. **This PR** — Event types, EventNames, EventBus subscription, receiver publishes to EventBus
2. **PR 2 (next)** — WorkManager-based direct HTTP delivery for events that survive process death
3. **PR 3** — 1-hour cooldown filter to suppress duplicate transitions

Each PR is self-contained but builds on the previous one.

## Summary

When the OS fires a geofence transition, the SDK now publishes a typed `Event.GeofenceTransitionEvent` to the in-process `EventBus`, and the DataPipelines module subscribes and forwards it to `track()` so the event lands in the customer's Customer.io dashboard.

This PR establishes the **EventBus delivery path**. The redundant WorkManager-based delivery (so events survive process death) lands in PR 2, mirroring the `PushMessageProcessorImpl.trackDeliveredMetrics` pattern (parallel WM + EventBus delivery).

### Changes

**Core**
- `Event.GeofenceTransition` enum (`ENTER`, `EXIT`) + `Event.GeofenceTransitionEvent` data class inside the sealed `Event`
- `EventNames.GEOFENCE_ENTERED = "GeoFence Entered"`, `GEOFENCE_EXITED = "GeoFence Exited"` (note the capital `F` — matches the platform's event naming convention)
- `Clock` interface + `SystemClock` impl + DI singleton — injectable time source, used here for event timestamp; used by the cooldown filter in PR 3

**Location module**
- `GeofenceBroadcastReceiver` refactored into three testable methods: `onReceive(context, intent)` → `handleGeofencingEvent(event: GeofencingEvent?)` → `dispatchTransition(...)`. The `@VisibleForTesting` split keeps the parsing layer separately verifiable from the dispatch logic and avoids needing to static-mock `GeofencingEvent.fromIntent` in tests.
- Receiver behavior:
  - Skips the movement-trigger geofence (`TODO(MBL-1623)` — full handling lands when the sync layer ships)
  - Maps GMS transition int → `Event.GeofenceTransition`; only `ENTER` and `EXIT` are tracked (other types like `DWELL` are logged and skipped)
[...]

PR #684, opened 2026-05-11
chore: deliver geofence transition events via WorkManager (MBL-1627) [2/3]

mrehan27:
## Linear ticket

[MBL-1627: Android — Send Geofence Transition Events](https://linear.app/customerio/issue/MBL-1627/android-send-geofence-transition-events)

## Stack

This is **PR 2 of 3** stacked PRs under MBL-1627. All three target `feature/geofence-on-device`.

1. [PR 1 (#683)](https://github.com/customerio/customerio-android/pull/683) — Event types, EventNames, EventBus subscription, receiver publishes to EventBus
2. **This PR** — WorkManager-based direct HTTP delivery so events survive process death, plus `goAsync()` hardening on the receiver
3. **PR 3 (next)** — 1-hour cooldown filter to suppress duplicate transitions

## Summary

PR 1 established the EventBus delivery path — but EventBus only delivers if the SDK is initialized in the current process. Geofence transitions often fire when the OS wakes a cold app process for the broadcast, which means events can land before (or fail entirely if) the SDK init completes.

This PR adds a **redundant WorkManager-based direct HTTP delivery** in parallel with EventBus, modeled on the existing push-delivery infrastructure (`PushDeliveryTracker` + `PushDeliveryMetricsBackgroundScheduler`). The receiver now fires both paths on every transition. WorkManager persists the work to its own DB so delivery survives process death; if WorkManager itself is unavailable, the scheduler falls back to an in-process async tracker.

### Changes

**Core module — modified**
- `ScopeProvider` gains a `geofenceScope` so the receiver can launch its async work through the SDK's scope provider rather than constructing a `CoroutineScope` inline.

**Location module — new files**
- `worker/GeofenceEventTracker.kt` — interface + `Impl` posts `/track` JSON via `CustomerIOHttpClient`, using a cached `userId` from `SecureUserStore` so delivery works without full SDK init. When no user is identified, the tracker skips the request and returns `Result.success()` with a debug log (no retry, no error-level noise). Also defines `AsyncGeofenceEventTracker` — a fire-and-forget wrapper used when WorkManager is unavailable.
- `worker/GeofenceEventWorker.kt` — `GeofenceEventScheduler` builds `Data`, enqueues a `OneTimeWorkRequest` with `NetworkType.CONNECTED` and unique key `${id}_${transition}_${timestamp}`, then `await()`s the WM `Operation` so the BroadcastReceiver doesn't `finish()` before WM has persisted the work spec. `GeofenceEventWorker` (`CoroutineWorker`) calls `setupAndroidComponent` defensively at the top of `doWork()` (for cold-process restarts), reads `inputData`, resolves the tracker via DI, returns `Result.retry()` on `IOException` / `Result.failure()` on other errors / `Result.success()` on success.

**Location module — modified**
- `GeofenceBroadcastReceiver.onReceive()` now calls `goAsync()` and runs the work in a coroutine on `scopeProvider.geofenceScope`, ensuring the process stays alive until WM commits the work spec. `dispatchTransition()` calls `scheduler.schedule(...)` (suspend) immediately before `eventBus.publish(...)` — parallel delivery, both paths fire on every transition.
- `GeofenceLogger` gains `logEventDeliveryRetryable` (debug; includes network-error message), `logEventDeliveryFailed` (error; includes failure message), `logEventInvalidInput` (error; includes which field was missing), `logEventDeliverySkippedNoUser` (debug; skip reason for anonymous users).
- `DIGraphGeofence` adds three singletons on `AndroidSDKComponent`: `geofenceEventTracker`, `asyncGeofenceEventTracker`, `geofenceEventScheduler`. Placed on `AndroidSDKComponent` (not `SDKComponent`) since they depend on `secureUserStore` which requires Android context.

**Build**
- `location/build.gradle` adds `implementation Dependencies.workManager` and `testImplementation Dependencies.workManagerTesting`.

### Why parallel delivery (not "WorkManager only")

Matches the existing pattern from `PushMessageProcessorImpl.trackDeliveredMetrics`:
- **WorkManager path** — primary, persisted to WM database, survives process death, retries on network failures (`Result.retry()` on `IOException`).
- **EventBus path** — secondary, in-process, runs through the analytics pipeline. Useful when the SDK *is* initialized in the same process (host app foreground use cases).

Backend should tolerate / dedupe by `(geofence_id, transition_type, timestamp)`.

### Notable design notes for reviewers

- **`goAsync()` + `Operation.await()`** — `WorkManager.enqueueUniqueWork()` returns an `Operation` that completes after the work spec is committed to WM's internal DB on a background executor. Without `goAsync()`, the OS can kill our process between `enqueue` returning and that commit, dropping the event entirely. The receiver wraps its work in `goAsync()` + coroutine + `finish()`-in-`finally`, and the scheduler `await()`s the `Operation` before returning so the receiver doesn't `finish()` early.
- **Unique key includes second-precision timestamp** — `${id}_${transition}_$timestamp` so same-second burst broadcasts collide on the key (`KEEP` dedupes them at the WM layer), while later transitions get distinct keys so an offline-queued worker can't block legitimate re-entries.
- **`Result.success()` for the no-userId case** — there's no point retrying; the userId state won't change just because WorkManager runs us again. Returning success drops the work item cleanly and avoids logging a misleading "HTTP delivery failed" at error level for what is actually a deliberate skip.
- **Worker calls `setupAndroidComponent` defensively** — WM can restart workers in cold processes where the host app's SDK init hasn't run yet. The receiver does this for the same reason.
- **Constructor-injected deps** — `GeofenceEventTrackerImpl(httpClient, secureUserStore, logger)` and `AsyncGeofenceEventTracker(tracker, dispatcher)` take their deps via constructor; they don't reach back into `SDKComponent`. The DI singletons wire them up.

## Verification

Automated tests added in this PR:
- [x] `GeofenceEventTrackerTest` — 6 tests covering ENTER body shape, EXIT event name, anonymous-user skip (no HTTP call, success result), null lat/lng omission, HTTP failure result, Content-Type header.
- [x] `AsyncGeofenceEventTrackerTest` — 2 tests covering pass-through to underlying tracker (with `UnconfinedTestDispatcher` so the `launch` block runs synchronously inside `runTest`).
- [x] `GeofenceEventSchedulerTest` — 3 tests covering WM-available enqueue (with constraints, timestamped unique key, input data), null lat/lng input data, WM-unavailable fallback to async tracker.
- [x] `GeofenceEventWorkerTest` — 8 tests covering valid ENTER/EXIT inputs, missing/invalid input failure paths, IOException → retry, non-IOException → failure.
- [x] `GeofenceBroadcastReceiverTest` (updated) — scheduler is verified called on ENTER, EXIT, and null-lat/lng paths; verified NOT called on movement-trigger and unknown-transition paths. Symmetry with EventBus assertions on every branch.

End-to-end manual verification deferred to when MBL-1623 (sync layer) + MBL-1629 (module integration) land — same constraint as PR 1. This PR is verified by unit tests only.

## Out of scope

- 1-hour cooldown filter to dedupe rapid repeated transitions — lands in PR 3 (next).

---

> [!NOTE]
> **Medium Risk**
> Adds a new WorkManager-based delivery path that performs direct HTTP `/track` calls from background workers, plus changes BroadcastReceiver lifecycle handling; failures could impact event delivery/duplication and background execution reliability.
> 
> **Overview**
> Geofence transition handling now **schedules a WorkManager job for direct HTTP delivery** (surviving process death) *in parallel* with the existing `EventBus` publish path.
> 
> `GeofenceBroadcastReceiver` is hardened with `goAsync()` and coroutine-based handling on a new `ScopeProvider.geofenceScope`, and transition dispatch isolates WorkManager scheduling errors so the EventBus path still runs.
> 
> Adds a small geofence worker stack (`GeofenceEventScheduler` + `GeofenceEventWorker` + `GeofenceEventTracker`) with network constraints, unique-work dedupe keys, retry-on-`IOException`, and an async in-process fallback when WorkManager is unavailable; updates DI wiring, logging, Gradle deps, and introduces focused unit tests for the new delivery path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 040a526712b708579d8f1fc46ce9ac4e17ca34bf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/location/src/main/kotlin/io/customer/location/geofence/GeofenceBroadcastReceiver.kt
+++ b/location/src/main/kotlin/io/customer/location/geofence/GeofenceBroadcastReceiver.kt
@@ -12,8 +12,6 @@ import io.customer.sdk.communication.Event
 import io.customer.sdk.core.di.SDKComponent
 import io.customer.sdk.core.di.clock
 import io.customer.sdk.core.di.setupAndroidComponent
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.launch
 
@@ -25,7 +23,7 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
         // without it the OS may kill us between enqueue and persist.
         val pendingResult = goAsync()
         SDKComponent.setupAndroidComponent(context = context)
-        val scope = CoroutineScope(SDKComponent.dispatchersProvider.background + SupervisorJob())
+        val scope = SDKComponent.scopeProvider.geofenceScope
 
         scope.launch {
             try {
--- a/location/src/main/kotlin/io/customer/location/geofence/GeofenceBroadcastReceiver.kt
+++ b/location/src/main/kotlin/io/customer/location/geofence/GeofenceBroadcastReceiver.kt
@@ -6,6 +6,7 @@ import android.content.Intent
 import androidx.annotation.VisibleForTesting
 import com.google.android.gms.location.Geofence
 import com.google.android.gms.location.GeofencingEvent
+import io.customer.location.geofence.di.geofenceEventScheduler
 import io.customer.location.geofence.di.geofenceLogger
 import io.customer.sdk.communication.Event
 import io.customer.sdk.core.di.SDKComponent
@@ -58,6 +59,8 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
     ) {
         val logger = SDKComponent.geofenceLogger
         val timestamp = SDKComponent.clock.currentTimeSeconds()
+        val androidComponent = SDKComponent.android()
+        val scheduler = androidComponent.geofenceEventScheduler
         val eventBus = SDKComponent.eventBus
 
         triggeringGeofenceIds.forEach { geofenceId ->
@@ -77,6 +80,15 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
             }
 
             logger.logTransitionEmitting(geofenceId, transition.name)
+
+            // Parallel delivery: WorkManager (survives process death) + EventBus (analytics pipeline).
+            scheduler.schedule(
+                geofenceId = geofenceId,
+                transition = transition,
+                latitude = latitude,
+                longitude = longitude,
+                timestamp = timestamp
+            )
 
             val properties = buildMap<String, Any> {
                 put("geofence_id", geofenceId)
--- a/location/src/main/kotlin/io/customer/location/geofence/GeofenceBroadcastReceiver.kt
+++ b/location/src/main/kotlin/io/customer/location/geofence/GeofenceBroadcastReceiver.kt
@@ -13,6 +13,7 @@ import io.customer.sdk.core.di.SDKComponent
 import io.customer.sdk.core.di.clock
 import io.customer.sdk.core.di.setupAndroidComponent
 import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.launch
 
@@ -23,9 +24,22 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
         // goAsync keeps the process alive until WorkManager has committed the work spec;
         // without it the OS may kill us between enqueue and persist.
         val pendingResult = goAsync()
-        SDKComponent.setupAndroidComponent(context = context)
-        val scope = SDKComponent.scopeProvider.geofenceScope
+        try {
+            SDKComponent.setupAndroidComponent(context = context)
+            val scope = SDKComponent.scopeProvider.geofenceScope
+            launchTransitionHandler(scope, intent, pendingResult)
+        } catch (e: Throwable) {
+            // Setup threw before the coroutine could register its finally — release the PendingResult here.
+            SDKComponent.geofenceLogger.logSyncFailed("BroadcastReceiver setup failed: ${e.message}")
+            pendingResult.finish()
+        }
+    }
 
+    private fun launchTransitionHandler(
+        scope: CoroutineScope,
+        intent: Intent,
+        pendingResult: PendingResult
+    ) {
         scope.launch {
             try {
                 handleGeofencingEvent(GeofencingEvent.fromIntent(intent))
--- a/location/src/main/kotlin/io/customer/location/geofence/GeofenceBroadcastReceiver.kt
+++ b/location/src/main/kotlin/io/customer/location/geofence/GeofenceBroadcastReceiver.kt
@@ -12,6 +12,7 @@ import io.customer.sdk.communication.Event
 import io.customer.sdk.core.di.SDKComponent
 import io.customer.sdk.core.di.clock
 import io.customer.sdk.core.di.setupAndroidComponent
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.launch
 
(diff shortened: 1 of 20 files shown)


location/src/main/kotlin/io/customer/location/geofence/worker/GeofenceEventTracker.kt
```diff
@@ -0,0 +1,85 @@
...
+            put("transition_type", transition.name.lowercase())
+            latitude?.let { put("latitude", it) }
+            longitude?.let { put("longitude", it) }
+            put("timestamp", timestamp)
+        }
+        val bodyJson = JSONObject().apply {
+            put("properties", propertiesJson)
+            put("event", eventName)
+            // Cached userId so direct-HTTP delivery works without full SDK init.
+            secureUserStore.getUserId()?.let { put("userId", it) }
```
mahmoud-elmorabea: If `userId` is nil, do we want to make this request?

location/src/main/kotlin/io/customer/location/geofence/GeofenceBroadcastReceiver.kt
```diff
@@ -78,6 +81,15 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
...
 
+            // Parallel delivery: WorkManager (survives process death) + EventBus (analytics pipeline).
+            scheduler.schedule(
+                geofenceId = geofenceId,
+                transition = transition,
+                latitude = latitude,
+                longitude = longitude,
+                timestamp = timestamp
+            )
+
```
mahmoud-elmorabea, line 127: I don't know if you plan to do this in a future PR but I think we should add `goAsync()` call here somewhere

location/src/main/kotlin/io/customer/location/geofence/worker/GeofenceEventWorker.kt
```diff
@@ -0,0 +1,119 @@
...
+                Constraints.Builder()
+                    .setRequiredNetworkType(NetworkType.CONNECTED)
+                    .build()
+            )
+            .setInputData(input)
+            .addTag(WORK_MANAGER_TAG_CIO)
+            .addTag(WORK_MANAGER_TAG_GEOFENCE)
+            .build()
+
+        val uniqueKey = "${geofenceId}_${transition.name}_$timestamp"
```
mahmoud-elmorabea, line 64: Will this generate the same key uniquely? 

The case I think of, is that for some reason, we get two broadcasts for entering/exiting the same geofence. The timestamp is generated when we get the receiver, this won't generate the same key for the same geofence triggering multiple times.

mahmoud-elmorabea (changes requested):

location/src/main/kotlin/io/customer/location/geofence/worker/GeofenceEventTracker.kt
mrehan27: Good catch, updated

location/src/main/kotlin/io/customer/location/geofence/GeofenceBroadcastReceiver.kt
mrehan27: The plan was to add it in later PRs, but added to this to avoid confusion

location/src/main/kotlin/io/customer/location/geofence/worker/GeofenceEventWorker.kt
mrehan27: This is intentional. The timestamp is at second level resolution, so if the OS fires twice within the same second, they collide on the same key and get deduplicated at the WorkManager level. Events across different seconds generate distinct keys by design, so an offline queued worker doesn't block a legitimate later re-entry. For broader deduplication, upcoming PR introduces a 1 hour cooldown filter at the receiver level, which applies to both delivery paths (WorkManager and EventBus), not just WorkManager.

mahmoud-elmorabea (approved):

PR #686, opened 2026-05-14
chore: suppress duplicate geofence transitions via cooldown filter (MBL-1627)

mrehan27:
## Linear ticket

[MBL-1627: Android — Send Geofence Transition Events](https://linear.app/customerio/issue/MBL-1627/android-send-geofence-transition-events)

## Stack

This is **PR 3 of 3** stacked PRs under MBL-1627, targeting `feature/geofence-on-device`.

1. [PR 1 (#683)](https://github.com/customerio/customerio-android/pull/683) — Event types, EventNames, EventBus subscription
2. [PR 2 (#684)](https://github.com/customerio/customerio-android/pull/684) — WorkManager-based direct HTTP delivery
3. **This PR** — 1-hour cooldown filter to suppress duplicate transitions at the receiver

## Summary

PR 2's WorkManager-key dedup catches OS double-fire bursts within the same second (timestamp at second resolution). This PR adds an *application-level* cooldown filter at the receiver that suppresses identical transitions across a wider 1-hour window, applying to **both** delivery paths (WorkManager + EventBus) before either is invoked.

The filter exposes a single atomic `tryAcquire(geofenceId, transition): Boolean` method — check-and-record happens inside one `@Synchronized` block so two concurrent broadcasts can't both pass the cooldown check and emit duplicates.

### Changes

**Location module — new files**
- `GeofenceCooldownFilter.kt` — `tryAcquire(...)` reads the last-emit timestamp from the store, compares against `GeofenceConstants.DEDUPE_COOLDOWN_MS` (1h), records the new timestamp if allowed, returns true/false. Both `tryAcquire` and `clearAll` are `@Synchronized` so they share the filter monitor — `clearAll` waits for any in-flight `tryAcquire` to complete before wiping the store.
- `store/GeofenceCooldownStore.kt` — interface + `PreferenceStore`-backed impl. Keys are `cio_cooldown_${geofenceId}_${transition.name}`. Prefs file is namespaced per app (`io.customer.sdk.geofence_cooldown.${packageName}`).

**Location module — modified**
- `GeofenceBroadcastReceiver.dispatchTransition()` — between transition validation and the delivery paths, calls `cooldownFilter.tryAcquire(...)`; if it returns false, logs a `suppressed` event and skips both the scheduler and EventBus for that geofence.
- `ModuleLocation.initialize()` — subscribes to `Event.ResetEvent` and calls `geofenceCooldownFilter.clearAll()` so a new user isn't suppressed by transitions emitted under the previous identity.
- `GeofenceLogger` gains `logTransitionSuppressed` (debug). Wording is generic (`"within the cooldown window"`) so it doesn't lie if the constant changes.
- `DIGraphGeofence` adds `geofenceCooldownStore` and `geofenceCooldownFilter` singletons on `AndroidSDKComponent` (the store needs Android context).

### Notable design notes for reviewers

- **Atomic `tryAcquire`** — single `@Synchronized` method instead of separate `shouldEmit`/`recordEmit`. Prevents a TOCTOU race where two concurrent broadcasts could both pass the check and emit. Caller decides (`if (!tryAcquire(...)) return`).
- **Record-before-deliver** — `tryAcquire` records the timestamp before the receiver invokes the scheduler/EventBus. If delivery fails downstream (HTTP error, etc.), the WM worker retries; the cooldown protects against repeat *receiver invocations*, not against delivery failures.
- **`@Synchronized clearAll`** — closes the narrow window where `clearAll` could run between an in-flight `tryAcquire`'s read and write. A residual race (a new `tryAcquire` arriving immediately after `clearAll`) is acceptable: cooldown is a device-level anti-burst mechanism, and any stale entry self-heals within 1h. Backend dedup on `(geofence_id, transition_type, timestamp)` remains the strict-correctness layer.
- **Cooldown is shared across delivery paths** — the filter gates at the receiver, so a suppressed transition skips *both* WM and EventBus. This is by design — the two paths are redundant, not independent.

## Verification

Automated tests added in this PR:
- [x] `GeofenceCooldownFilterTest` — 6 tests covering `tryAcquire` boundary cases (null / within / exactly at / outside cooldown), per-key independence (ENTER vs EXIT not collapsed), and `clearAll` pass-through. Each `tryAcquire` test asserts both the return value and whether `store.recordEmit` was called.
- [x] `GeofenceCooldownStoreTest` — 5 tests covering null read, write+read round trip, multi-key independence, same-key overwrite, `clearAll`. Uses real SharedPreferences via Robolectric.
- [x] `GeofenceBroadcastReceiverTest` (updated) — 2 new tests covering the suppress branch (`tryAcquire` returns false → nothing scheduled or published) and the allow branch with `coVerifyOrder` proving `tryAcquire → scheduler.schedule → eventBus.publish` ordering.

End-to-end manual verification deferred to when MBL-1623 (sync layer) + MBL-1629 (module integration) land — same constraint as PR 1 and PR 2.

## Out of scope

- Sign-out wiring beyond `clearAll` — the `ResetEvent` subscription is wired in this PR, but full sign-out lifecycle integration belongs to MBL-1629.
- E2E manual verification — depends on MBL-1623 (sync) and MBL-1629 (module config flag).

---

> [!NOTE]
> **Medium Risk**
> Changes geofence transition dispatch to suppress events for up to 1 hour using persisted state, which could inadvertently drop legitimate transitions if the cooldown logic/keying is wrong. Touches runtime event delivery (WorkManager + EventBus) and adds SharedPreferences persistence, but scope is limited and covered by new tests.
> 
> **Overview**
> Adds an application-level cooldown dedupe for geofence transitions: `GeofenceBroadcastReceiver.dispatchTransition()` now gates each geofenceId+transition with `GeofenceCooldownFilter.tryAcquire()` (backed by a new `SharedPreferences` store) and skips both WorkManager scheduling and EventBus publishing when suppressed.
> 
> Wires the new `geofenceCooldownStore`/`geofenceCooldownFilter` into the geofence DI graph, adds a suppression log line, and clears cooldown state on `Event.ResetEvent` to avoid a new identity inheriting the previous user’s suppression window. Includes new unit tests for the filter/store and updates receiver tests to cover suppression and call ordering.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5c5fcd232007a3317cf7eb13c6f44e896e3716a5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/location/src/main/kotlin/io/customer/location/ModuleLocation.kt
+++ b/location/src/main/kotlin/io/customer/location/ModuleLocation.kt
@@ -2,6 +2,7 @@ package io.customer.location
 
 import android.location.Location
 import androidx.lifecycle.ProcessLifecycleOwner
+import io.customer.location.geofence.di.geofenceCooldownFilter
 import io.customer.location.provider.FusedLocationProvider
 import io.customer.location.store.LocationPreferenceStoreImpl
 import io.customer.location.sync.LocationSyncFilter
@@ -118,6 +119,12 @@ class ModuleLocation @JvmOverloads constructor(
             if (!it.userId.isNullOrEmpty()) {
                 locationTracker.onUserIdentified()
             }
+        }
+
+        // Clear geofence cooldown state on user reset so a new user isn't suppressed
+        // by transitions emitted under the previous identity.
+        eventBus.subscribe<Event.ResetEvent> {
+            SDKComponent.android().geofenceCooldownFilter.clearAll()
         }
 
         // Register lifecycle observer for background cancellation and ON_APP_START.
--- a/location/src/main/kotlin/io/customer/location/geofence/GeofenceBroadcastReceiver.kt
+++ b/location/src/main/kotlin/io/customer/location/geofence/GeofenceBroadcastReceiver.kt
@@ -6,6 +6,7 @@ import android.content.Intent
 import androidx.annotation.VisibleForTesting
 import com.google.android.gms.location.Geofence
 import com.google.android.gms.location.GeofencingEvent
+import io.customer.location.geofence.di.geofenceCooldownFilter
 import io.customer.location.geofence.di.geofenceEventScheduler
 import io.customer.location.geofence.di.geofenceLogger
 import io.customer.sdk.communication.Event
@@ -89,6 +90,7 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
         val timestamp = SDKComponent.clock.currentTimeSeconds()
         val androidComponent = SDKComponent.android()
         val scheduler = androidComponent.geofenceEventScheduler
+        val cooldownFilter = androidComponent.geofenceCooldownFilter
         val eventBus = SDKComponent.eventBus
 
         triggeringGeofenceIds.forEach { geofenceId ->
@@ -107,6 +109,10 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
                 }
             }
 
+            if (!cooldownFilter.tryAcquire(geofenceId, transition)) {
+                logger.logTransitionSuppressed(geofenceId, transition.name)
+                return@forEach
+            }
             logger.logTransitionEmitting(geofenceId, transition.name)
 
             // Parallel delivery: WorkManager (survives process death) + EventBus (analytics pipeline).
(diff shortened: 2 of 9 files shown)


mahmoud-elmorabea (approved):
I think we should also consider using `GEOFENCE_TRANSITION_DWELL` for transition type that might prevent some enter/exit jitter on lower accuracy location provider sources.

mrehan27:
@mahmoud-elmorabea Dwell isn't part of V1 and is planned for later phase. But I'll keep an eye on it and include it if it adds value.

PR #689, opened 2026-05-14
chore: per-install installation id on every event

mahmoud-elmorabea:
[MBL-1700: Implement lifecycle API calls to backend](https://linear.app/customerio/issue/MBL-1700/implement-lifecycle-api-calls-to-backend)

## Summary
- Adds a stable per-install `installationId` (UUID) generated once on first SDK launch and persisted in `GlobalPreferenceStore`. Survives identify/clearIdentify, SDK re-init, and process death; only wiped by an OS app-data clear.
- Exposes the id lazily on `AndroidSDKComponent` (no public getter on `CustomerIO`) so the first access reads SharedPreferences, generates+persists a UUID if absent, and caches it for the rest of the process.
- `ContextPlugin` now attaches `installationId` under `context.device.installationId` (camelCase, alongside `device.token`) on every outgoing event so the backend can correlate events to a single install.

### Automated (done in this PR)
- [x] `./gradlew :core:test :datapipelines:test` green (one unrelated pre-existing flake in `ContextPluginBehaviorTest.execute_whenDeviceTokenIsSetFromAnotherThread...` passes on rerun; not introduced by this change)
- [x] `./gradlew :core:apiCheck :datapipelines:apiCheck` green (only intentional addition: `AndroidSDKComponent.installationId`, matching the existing pattern of other DI-graph properties)
- [x] `ktlint --android` clean
- [x] New unit tests cover persistence (`GlobalPreferenceStoreImplTest`), ContextPlugin attachment, generate-and-persist-on-first-access, and survival across `clearIdentify` (`ContextPluginTest`)

### Manual (TODO — reviewer to run on a device/emulator)
- [ ] Fresh install → first event includes `context.device.installationId` as a UUID
- [ ] `identify(...)` then `clearIdentify()` → same UUID on next event
- [ ] Kill + relaunch app → UUID unchanged
- [ ] Android Settings → Clear data → new UUID generated on next launch
- [ ] (optional) `adb shell run-as <pkg> cat shared_prefs/io.customer.sdk.<pkg>.xml` shows an `installation_id` entry

---

> [!NOTE]
> **Medium Risk**
> Adds a new persisted identifier that is generated and sent on every event, which can affect backend correlation and analytics data semantics. Risk is moderate due to new SharedPreferences storage and context payload changes but limited surface area.
> 
> **Overview**
> Adds a stable per-install `installationId` persisted in `GlobalPreferenceStore` (generate-once UUID when absent).
> 
> Updates `CustomerIO`/`ContextPlugin` to pass this id through and attach it to every event under `context.device.installationId`, and extends tests/utilities to validate persistence, generation, and behavior across `clearIdentify()`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d78d340a79ec56de5b3e1845ede45c666bf09f09. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


core/src/main/kotlin/io/customer/sdk/data/store/GlobalPreferenceStore.kt
```diff
@@ -48,10 +54,23 @@ internal class GlobalPreferenceStoreImpl(
...
+    }
+
     override fun removeDeviceToken() = clear(KEY_DEVICE_TOKEN)
 
+    // installation_id is a per-install identifier whose lifecycle is owned by the OS
+    // (wiped only on app uninstall or "Clear data"). Bulk-clear paths must preserve it.
+    override fun clearAll() {
+        val preservedId = prefs.read { getString(KEY_INSTALLATION_ID, null) }
+        super.clearAll()
+        preservedId?.let(::saveInstallationId)
```
mrehan27: Are we actually calling `clearAll` for `GlobalPreferenceStore` anywhere? I think we should  also not remove the device token.

If we have to keep this. wouldn't it be better to implement this the other way around? Instead of clearing everything and preserving specific values, we could just remove the keys we want, like only `KEY_CONFIG_SETTINGS` here?

core/src/test/java/io/customer/sdk/data/store/GlobalPreferenceStoreImplTest.kt
```diff
@@ -0,0 +1,85 @@
...
+        val secondId = "00000000-0000-0000-0000-000000000004"
+
+        store.saveInstallationId(firstId)
+        store.saveInstallationId(secondId)
+
+        store.getInstallationId() shouldBeEqualTo secondId
+    }
+
+    @Test
+    fun getInstallationId_givenClearAllInvoked_expectIdPreserved() {
```
mrehan27: This is a good test to have, but the rest of the tests don't seem to add much value and could probably be reduced/removed since most of them appear to only verify OS calls?

datapipelines/src/main/kotlin/io/customer/datapipelines/plugins/ContextPlugin.kt
```diff
@@ -55,6 +66,10 @@ internal class DefaultContextPluginEventProcessor : ContextPluginEventProcessor
             event.putInContextUnderKey("device", "token", token)
         }
 
+        // Attach installation id to device in context so backend can correlate
+        // events to a single install across identify/clearIdentify cycles.
+        event.putInContextUnderKey("device", "installationId", installationId)
```
mrehan27, line 71: Unlike the device token, we attach this every time and override it if it exists? (Though I think it may never already exist). Just mentioning it to make sure expectations are aligned here.

datapipelines/src/test/java/io/customer/datapipelines/plugins/ContextPluginTest.kt
```diff
@@ -41,6 +46,17 @@ class ContextPluginTest : JUnitTest() {
...
+    private fun setupWithInstallationId(installationId: String?): DataPipelinesTestConfig {
+        return testConfiguration {
+            diGraph {
+                android {
+                    // Stub before CustomerIO init so the init-block resolution reads this value.
+                    every { globalPreferenceStore.getInstallationId() } returns installationId
+                }
+            }
+        }
+    }
```
mrehan27: I believe `diGraph` block is mainly for overriding dependencies, and this statement doesn't necessarily need to be in it. It could simply be something like:

```suggestion
    private fun setupWithInstallationId(installationId: String?) {
        // Stub before CustomerIO init so the init-block resolution reads this value.
        every { globalPreferenceStore.getInstallationId() } returns installationId
    }
```

mrehan27 (approved):
Just a few comments, looks good overall

datapipelines/src/main/kotlin/io/customer/datapipelines/plugins/ContextPlugin.kt
mahmoud-elmorabea: Intentional: `installationId` is `String` (non-null) on `ContextPlugin` and is SDK-owned — no upstream plugin or migration path sets `device.installationId`, so there's nothing in the event context to preserve. We always write it for the same reason we always write `userAgent`: the SDK is the source of truth for the value on every event.

PR #691, opened 2026-05-14
chore: add geofence distance filter for nearest-N selection

mrehan27:
## Linear ticket

[MBL-1623: Android — Fetch and Sync Geofences](https://linear.app/customerio/issue/MBL-1623/android-fetch-and-sync-geofences)

## Stack

This is **PR 2 of 5** stacked PRs under MBL-1623.

1. [PR 1 (#687)](https://github.com/customerio/customerio-android/pull/687) — Region persistence store
2. **This PR** — Distance filter for selecting the nearest N regions
3. PR 3 (next) — API service + DTO mapper (stub backend)
4. PR 4 — Repository: sync orchestration
5. PR 5 — Services entry point + movement-trigger EXIT wiring

This PR targets `mbl-1623-geofence-store` (PR 1's branch); GitHub will auto-retarget to `feature/geofence-on-device` once PR 1 merges.

## Summary

Adds `GeofenceDistanceFilter` — picks the geographically closest N regions from a candidate pool. Used by the repository (PR 4) when the candidate set exceeds the OS's per-app geofence limit (`MAX_BUSINESS_GEOFENCES = 19`).

The filter is stateless, takes `max` explicitly (no default — caller owns the limit decision so a server-provided override can be passed in cleanly), and delegates the actual distance computation to `Location.distanceBetween` via the existing `GeofenceRegion.distanceTo` extension.

## Changes

**New files**
- `GeofenceDistanceFilter.kt` (22 LOC) — class with one method `nearest(regions, latitude, longitude, max): List<GeofenceRegion>`. Returns `emptyList()` for empty input or `max <= 0`; otherwise sorts by distance and takes `max`.
- `GeofenceDistanceFilterTest.kt` (75 LOC) — 5 tests covering empty input, `max=0`, fewer-than-max (returns all sorted), more-than-max (returns nearest), and stable order for ties.

**Modified**
- `DIGraphGeofence.kt` — registers `geofenceDistanceFilter` as a `newInstance<>` on `SDKComponent` (not `AndroidSDKComponent` — the filter is pure Kotlin, no Context required).

## Notable design notes for reviewers

- **`SDKComponent`, not `AndroidSDKComponent`** — the filter only uses `Location.distanceBetween` which is a static utility (no Context). Putting it on `SDKComponent` accurately reflects the dependency surface; the repository (PR 4) will access it via `SDKComponent.geofenceDistanceFilter` from its `AndroidSDKComponent` extension.
[...]

PR #692, opened 2026-05-14
chore: add geofence API service and DTOs

mrehan27:
## Summary
- Adds `GeofenceApiService` + DTOs for the upcoming nearby-geofence fetch flow.
- Tolerant JSON parsing (`ignoreUnknownKeys = true`) shared by parser + future consumers via `GeofenceApiJson`.
- Service is **inert by default**: `ENDPOINT_PATH` is empty until backend exposes the real path, so production short-circuits to `Result.failure` with no stray network calls. TODOs in `GeofenceApiService.kt` point reviewers at what changes when the endpoint lands (path + delete the guard + add real service tests).
- Stateless, registered as `newInstance` in DI to match the precedent set by `GeofenceDistanceFilter`.

## Stack
Targets `mbl-1623-geofence-distance-filter` (PR #691). Will auto-retarget to `main` once #691 merges.

## Linear

## Test plan
- [x] `GeofenceApiResponseTest` covers parser + DTO→domain mapping: full sample, unknown top-level field (tolerance), mixed valid/invalid transition types, all-invalid transition types (region dropped).
- [x] No service-layer test in this PR — see TODO on `ENDPOINT_PATH`; real tests (success decode, body shape, HTTP failure, malformed response) come with the wire-up PR when backend lands.
- [x] `./gradlew :location:testDebugUnitTest --tests "io.customer.location.geofence.api.*"` — green.

---

> [!NOTE]
> **Low Risk**
> Adds new internal DTO/parsing and a DI-registered geofence fetch service, but it is intentionally inert (blank endpoint) and does not change existing runtime behavior unless later wired to a real path.
> 
> **Overview**
> Introduces a new server-driven geofence API layer: `GeofenceApiResponse` DTOs plus mapping to domain (`GeofenceConfig` and `GeofenceRegion`), including seconds→milliseconds normalization and filtering of unsupported `transition_types`.
> 
> Adds `GeofenceApiResponseParser` with `ignoreUnknownKeys` for forward-compatible decoding, and an internal `GeofenceApiService` that builds a JSON request and parses the response; it currently short-circuits to `Result.failure` while `ENDPOINT_PATH` is blank.
> 
> Registers the parser and service in the geofence DI graph, and adds Robolectric tests covering decode/mapping, unknown-field tolerance, and transition-type handling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1bd143aabaeccfa442a4f90519104356bb6319ab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/location/src/main/kotlin/io/customer/location/geofence/api/GeofenceApiResponse.kt
+++ b/location/src/main/kotlin/io/customer/location/geofence/api/GeofenceApiResponse.kt
@@ -1,0 +1,87 @@
+package io.customer.location.geofence.api
+
+import io.customer.location.geofence.GeofenceConfig
+import io.customer.location.geofence.GeofenceRegion
+import io.customer.location.geofence.GeofenceTransitionType
+import java.util.concurrent.TimeUnit
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.Json
+
+/** Shared parser config: tolerant to unknown keys so backend additions don't break decoding. */
+internal val GeofenceApiJson: Json = Json { ignoreUnknownKeys = true }
+
+@Serializable
+internal data class GeofenceApiResponse(
+    val config: GeofenceApiConfig,
+    val geofences: List<GeofenceApiRegion>
+)
+
+@Serializable
+internal data class GeofenceApiConfig(
+    @SerialName("movement_trigger_radius")
+    val movementTriggerRadius: Float,
+    @SerialName("local_refresh_trigger_radius")
+    val localRefreshTriggerRadius: Float,
+    @SerialName("remote_fetch_refresh_trigger_radius")
+    val remoteFetchRefreshTriggerRadius: Float,
+    @SerialName("remote_fetch_refresh_expiry_time")
+    val remoteFetchRefreshExpiryTime: Long,
+    @SerialName("duplicate_events_expiry_time")
+    val duplicateEventsExpiryTime: Long,
+    val android: GeofenceApiPlatformConfig
+)
+
+@Serializable
+internal data class GeofenceApiPlatformConfig(
+    @SerialName("max_business_geofence")
+    val maxBusinessGeofence: Int
+)
+
+@Serializable
+internal data class GeofenceApiRegion(
+    val id: String,
+    val name: String,
+    val latitude: Double,
+    val longitude: Double,
+    val radius: Float,
+    @SerialName("transition_types")
+    val transitionTypes: List<String>,
+    @SerialName("last_updated")
+    val lastUpdated: Long
+)
+
+internal fun GeofenceApiResponse.toDomainConfig(): GeofenceConfig = config.toDomain()
+
+internal fun GeofenceApiResponse.toDomainRegions(): List<GeofenceRegion> =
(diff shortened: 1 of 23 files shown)


location/src/main/kotlin/io/customer/location/geofence/api/GeofenceApiResponse.kt
```diff
@@ -0,0 +1,87 @@
...
+import io.customer.location.geofence.GeofenceConfig
+import io.customer.location.geofence.GeofenceRegion
+import io.customer.location.geofence.GeofenceTransitionType
+import java.util.concurrent.TimeUnit
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.Json
+
+/** Shared parser config: tolerant to unknown keys so backend additions don't break decoding. */
+internal val GeofenceApiJson: Json = Json { ignoreUnknownKeys = true }
```
mahmoud-elmorabea: Can we avoid this being global like that, let's create a parser type that can be injected into the service.

location/src/main/kotlin/io/customer/location/geofence/api/GeofenceApiResponse.kt
```diff
@@ -0,0 +1,87 @@
...
+    val remoteFetchRefreshExpiryTime: Long,
+    @SerialName("duplicate_events_expiry_time")
+    val duplicateEventsExpiryTime: Long,
+    val android: GeofenceApiPlatformConfig
+)
+
+@Serializable
+internal data class GeofenceApiPlatformConfig(
+    @SerialName("max_business_geofence")
+    val maxBusinessGeofence: Int
```
mahmoud-elmorabea, line 34: Nice 👍

location/src/main/kotlin/io/customer/location/geofence/api/GeofenceApiResponse.kt
```diff
@@ -0,0 +1,87 @@
...
+        transitionTypes = types,
+        lastUpdated = lastUpdated
+    )
+}
+
+private fun parseTransitionType(value: String): GeofenceTransitionType? =
+    when (value.lowercase()) {
+        "enter" -> GeofenceTransitionType.ENTER
+        "exit" -> GeofenceTransitionType.EXIT
+        else -> null
```
mahmoud-elmorabea, line 82: Do we want to allow that? I think we have logic that depends on this being required

mahmoud-elmorabea (approved):

location/src/main/kotlin/io/customer/location/geofence/api/GeofenceApiResponse.kt
mrehan27: Good point. Already safe from a crash standpoint — if every transition type on a region is unknown, we drop the whole region (line 70), so we never end up with an empty list downstream.

The real risk is silent data loss: if backend ever adds a new type (`"dwell"` etc.), we'd just stop registering those geofences with no warning. I'll add a log at the service level so we catch it. Folding into the service-tests follow-up when the live endpoint lands.

PR #693, opened 2026-05-14
chore(in-app): dedupe onMessageShown metric by deliveryID per session

mahmoud-elmorabea:
Dedupes the in-app `onMessageShown` `opened` metric publish by `deliveryID` within the SDK process lifetime. First show of a given delivery publishes the metric; subsequent shows of the same delivery skip the publish.

The host `eventListener.messageShown` callback still fires on every call, so app-level UI side-effects are unaffected — only the analytics metric publish is deduped.

Backend already dedupes `(deliveryID, opened)` pairs server-side, so this is bandwidth and analytics-queue noise reduction, not a metric-correctness change.

---

> [!NOTE]
> **Low Risk**
> Scoped analytics deduplication with explicit reset clearing; host callbacks unchanged and backend already dedupes opened metrics.
> 
> **Overview**
> **In-app `Opened` metrics** are now deduplicated per `deliveryID` for the lifetime of the SDK process. `ModuleMessagingInApp` keeps a thread-safe in-memory set; the first `onMessageShown` for a campaign publishes `TrackInAppMetricEvent` with `Metric.Opened`, and later shows of the same delivery skip that publish.
> 
> **Host behavior is unchanged:** `eventListener.messageShown` still runs on every show. Only the analytics bus publish is suppressed on repeats. **`ResetEvent`** clears the set so logout or reset allows a new session to emit `Opened` again for the same delivery. Messages without a `campaignId` still skip metric publish and do not touch the dedup set.
> 
> Unit tests cover duplicate delivery IDs, reset clearing, distinct campaigns, and null delivery ID.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4d1c00c686cd44e5c074a3fbc21f788b341b2072. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
@@ -13,8 +13,8 @@ import io.customer.sdk.communication.subscribe
 import io.customer.sdk.core.di.SDKComponent
 import io.customer.sdk.core.module.CustomerIOModule
 import io.customer.sdk.events.Metric
-import java.util.concurrent.locks.ReentrantLock
-import kotlin.concurrent.withLock
+import java.util.Collections
+import java.util.concurrent.ConcurrentHashMap
 
 class ModuleMessagingInApp(
     config: MessagingInAppModuleConfig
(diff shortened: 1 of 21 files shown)


messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
```diff
@@ -25,6 +27,14 @@ class ModuleMessagingInApp(
         get() = SDKComponent.gistProvider
     private val logger = SDKComponent.logger
 
+    // Per-session set of deliveryIDs whose `Opened` metric has already been published.
+    // Used to dedupe `onMessageShown` emissions so the same in-app view does not produce
+    // multiple `TrackInAppMetricEvent`s within a single SDK process lifetime.
+    // Lifetime: scoped to this `ModuleMessagingInApp` instance (per-process). No persistence.
+    // Mirrors iOS `GistDelegateImpl` behavior — see Phase 1 PR 2.
+    private val shownDeliveryIdsLock = ReentrantLock()
```
mrehan27: Wouldn't `ConcurrentHashMap` be a simpler choice instead?

messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
```diff
@@ -25,6 +27,14 @@ class ModuleMessagingInApp(
...
+    private val shownDeliveryIds = mutableSetOf<String>()
```
mrehan27: The ids are never cleared anywhere. I believe delivery ids are unique per message, but still think we should clear them on logout?

messaginginapp/src/main/java/io/customer/messaginginapp/ModuleMessagingInApp.kt
```diff
@@ -87,6 +97,17 @@ class ModuleMessagingInApp(
         moduleConfig.eventListener?.messageShown(InAppMessage.getFromGistMessage(message))
 
         message.gistProperties.campaignId?.let { deliveryID ->
+            // Dedupe `Opened` metric emission per-deliveryID for this session. Host UI
+            // side-effects (eventListener.messageShown above) still fire on every call —
+            // only the analytics publish is suppressed on duplicates. See PR 6 in
+            // binary-tinkering-treasure-phase1.md.
+            val isFirstShown = shownDeliveryIdsLock.withLock {
+                shownDeliveryIds.add(deliveryID)
```
mrehan27: For persistent messages, it looks like they'll continue sending opened API calls every time they're shown, but not analytics events, correct?

Also, have we communicated this change to in-app squad to make sure expectations are aligned? I assume they mostly rely on the API side, so this likely isn't a major concern. But it still feels worth informing them in case this changes any expected behavior.

mahmoud-elmorabea:
I was unable to convince myself that this could happen in real production scenario but rather only a technical possibility, dropping this for now

PR #694, opened 2026-05-14
chore: dedup consecutive auto-tracked screen views by name

mahmoud-elmorabea:
Dedupes consecutive identical screen events emitted by `AutomaticActivityScreenTrackingPlugin`. Name-only, case-sensitive comparison; in-memory per plugin instance.

Brings Android to parity with iOS's existing in-memory `lastScreenTracked` dedup. Pure additive guard before the screen emit — no API changes.

---

> [!NOTE]
> **Medium Risk**
> Changes screen-view emission behavior by suppressing consecutive same-name auto-tracked screens, which could impact downstream analytics counts (though scoped in-memory per plugin instance). Logic is simple and covered by new unit tests, keeping overall risk moderate.
> 
> **Overview**
> **Dedupes consecutive auto-tracked screen views in Android.** `AutomaticActivityScreenTrackingPlugin` now keeps an in-memory `lastScreenTracked` and skips emitting when `onActivityStarted` resolves to the same (case-sensitive) screen name as the last one.
> 
> The plugin only updates the dedup state after a successful `CustomerIO.instance().screen(...)` call or its `analytics.screen(...)` fallback, and adds a new test suite validating same/different-name behavior plus basic thread-safety of the synchronized state.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit af91671f6e8464e87b87e6252bf265a68025fbff. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/datapipelines/src/main/kotlin/io/customer/datapipelines/plugins/AutomaticActivityScreenTrackingPlugin.kt
+++ b/datapipelines/src/main/kotlin/io/customer/datapipelines/plugins/AutomaticActivityScreenTrackingPlugin.kt
@@ -61,10 +61,16 @@ class AutomaticActivityScreenTrackingPlugin : Plugin, AndroidLifecycle {
             if (!screenName.isNullOrBlank()) {
                 // Dedup by name only (case-sensitive), parity with iOS.
                 if (isDuplicateScreen(screenName)) return
-                runCatching { CustomerIO.instance().screen(screenName) }.onFailure { analytics.screen(screenName) }
-                // Mark only after the emission attempt so a thrown emit doesn't silently swallow
-                // the next same-name screen.
-                markScreenTracked(screenName)
+                // Mark only when the emit actually succeeds — if both the SDK call and the
+                // analytics fallback throw, the screen wasn't delivered and the next
+                // same-name screen must not be silently deduped.
+                runCatching {
+                    CustomerIO.instance().screen(screenName)
+                    markScreenTracked(screenName)
+                }.onFailure {
+                    analytics.screen(screenName)
+                    markScreenTracked(screenName)
+                }
             }
         } catch (e: PackageManager.NameNotFoundException) {
             logger.error(e.message ?: "Unable to activity screen NameNotFoundException, $activity")
--- a/datapipelines/src/main/kotlin/io/customer/datapipelines/plugins/AutomaticActivityScreenTrackingPlugin.kt
+++ b/datapipelines/src/main/kotlin/io/customer/datapipelines/plugins/AutomaticActivityScreenTrackingPlugin.kt
@@ -25,16 +25,17 @@ class AutomaticActivityScreenTrackingPlugin : Plugin, AndroidLifecycle {
     private val logger: Logger = SDKComponent.logger
 
     // Last screen name emitted by this plugin instance. Guarded by `@Synchronized` on
-    // the dedup helper below — the codebase idiom (see CustomerIO.createInstance,
+    // the dedup helpers below — the codebase idiom (see CustomerIO.createInstance,
     // IdentifyHookRegistry, PreferenceCrypto) is `@Synchronized` rather than ReentrantLock.
     // Scope: per-plugin-instance, lifetime of the process; not persisted across launches.
     private var lastScreenTracked: String? = null
 
     @Synchronized
-    private fun shouldSkipDuplicate(screenName: String): Boolean {
-        if (lastScreenTracked == screenName) return true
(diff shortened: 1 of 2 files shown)


datapipelines/src/main/kotlin/io/customer/datapipelines/plugins/AutomaticActivityScreenTrackingPlugin.kt
```diff
@@ -40,6 +58,8 @@ class AutomaticActivityScreenTrackingPlugin : Plugin, AndroidLifecycle {
             }
             // If screen name is null or blank, we do not track the screen
             if (!screenName.isNullOrBlank()) {
+                // Dedup by name only (case-sensitive), parity with iOS.
+                if (shouldSkipDuplicate(screenName)) return
```
mrehan27: So are we deliberately doing this only for automatic screen tracking? For apps using Jetpack Compose and probably have a higher chance of tracking the same screen multiple times, shouldn't we be handling it for them as well?

datapipelines/src/test/java/io/customer/datapipelines/plugins/AutomaticActivityScreenTrackingPluginTest.kt
```diff
@@ -0,0 +1,165 @@
...
+        plugin.onActivityStarted(activityForScreen("Home"))
+        plugin.onActivityStarted(activityForScreen("Profile"))
+        plugin.onActivityStarted(activityForScreen("Home"))
+
+        // Home re-emits because the last tracked screen was Profile.
+        assertEquals(listOf("Home", "Profile", "Home"), publishedScreenNames())
+    }
+
+    @Test
+    fun onActivityStarted_concurrent_isThreadSafe() {
```
mrehan27: Copying Claude's feedback on this just in case you want to have another look:

> The assertions here pass even if shouldSkipDuplicate always returns false — each thread uses a unique screen name, so the only same-name pairs are back-to-back calls within a single thread (sequential, not concurrent), and the [N, 2N] range covers both the perfect-dedup and no-dedup outcomes. This is exercising @Synchronized/MockK plumbing, not the dedup decision.
> 
> If we want a real concurrent test, have every thread emit the same name and assert exactly one emission — that's the property that would actually fail if @Synchronized were removed. Otherwise I'd drop this and rely on the single-threaded tests.

datapipelines/src/main/kotlin/io/customer/datapipelines/plugins/AutomaticActivityScreenTrackingPlugin.kt
mahmoud-elmorabea: I am intentionally keeping the scope for now to automatic only to match iOS behavior, when we do some data analysis and deem the deduplication on the manual API level as well safe, I will extend it to be at the `Customer.screen()` level

datapipelines/src/test/java/io/customer/datapipelines/plugins/AutomaticActivityScreenTrackingPluginTest.kt
mahmoud-elmorabea: 👍

datapipelines/src/main/kotlin/io/customer/datapipelines/plugins/AutomaticActivityScreenTrackingPlugin.kt
```diff
@@ -40,7 +59,12 @@ class AutomaticActivityScreenTrackingPlugin : Plugin, AndroidLifecycle {
             }
             // If screen name is null or blank, we do not track the screen
             if (!screenName.isNullOrBlank()) {
+                // Dedup by name only (case-sensitive), parity with iOS.
+                if (isDuplicateScreen(screenName)) return
                 runCatching { CustomerIO.instance().screen(screenName) }.onFailure { analytics.screen(screenName) }
+                // Mark only after the emission attempt so a thrown emit doesn't silently swallow
+                // the next same-name screen.
+                markScreenTracked(screenName)
```
mrehan27: Should this be in `runCatching` instead to avoid updating in case of failure?

mrehan27 (approved):

datapipelines/src/main/kotlin/io/customer/datapipelines/plugins/AutomaticActivityScreenTrackingPlugin.kt
mahmoud-elmorabea: Good point!

PR #696, opened 2026-05-14
chore: dedup identify(userId) within a session when no traits supplied

mahmoud-elmorabea:
First `identify(userId)` per session always fires fully — preserves device-token re-registration and Device Created or Updated cadence.

Subsequent `identify(userId)` calls with the same userId and empty traits short-circuit (no analytics call, no `UserChangedEvent` publish, no device-token re-registration). `clearIdentify()` resets the session marker. Calls with non-empty traits always pass through.

Targets the misuse pattern of identify-on-every-event within a session. Per-session scope ensures token rotation still propagates on cold start.

Internal-only change — no public API surface modification.

---

> [!NOTE]
> **Medium Risk**
> Changes core `identify` behavior by skipping repeated same-user, empty-traits calls, which can affect downstream `UserChangedEvent` publications and device-token re-registration if the dedup conditions are incorrect.
> 
> **Overview**
> Adds **per-session `identify` deduping** in `CustomerIO.identifyImpl`: repeated `identify(userId)` calls with the *same* `userId` and *empty Map-shaped traits* now short-circuit (no analytics identify, no `UserChangedEvent`, no device re-registration), while profile changes or non-empty traits still flow through.
> 
> `clearIdentify()` now resets the session dedup marker. Tests were updated/added to cover the new dedup behavior and to regression-guard that the *first* identify still publishes `UserChangedEvent` and re-registers the device token.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4a85c3cf46be7c6c522d6165675c6bf0b0be3116. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
+++ b/datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
@@ -111,6 +111,11 @@ class CustomerIO private constructor(
 
     private val contextPlugin: ContextPlugin = ContextPlugin(deviceStore)
 
+    // Tracks the last userId successfully identified in this SDK session. Used to dedup
+    // back-to-back identify(userId) calls with no traits, which are no-ops server-side.
+    // Reset on clearIdentify so a subsequent identify of the same userId is not deduped.
+    private var lastIdentifiedUserIdThisSession: String? = null
+
     init {
         // Set analytics logger and debug logs based on SDK logger configuration
         Analytics.debugLogsEnabled = logger.logLevel == CioLogLevel.DEBUG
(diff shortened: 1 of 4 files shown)


datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
```diff
@@ -111,6 +111,11 @@ class CustomerIO private constructor(
 
     private val contextPlugin: ContextPlugin = ContextPlugin(deviceStore)
 
+    // Tracks the last userId successfully identified in this SDK session. Used to dedup
+    // back-to-back identify(userId) calls with no traits, which are no-ops server-side.
+    // Reset on clearIdentify so a subsequent identify of the same userId is not deduped.
+    private var lastIdentifiedUserIdThisSession: String? = null
```
mrehan27, line 117: Did we deliberately add this new variable `lastIdentifiedUserIdThisSession` instead of reusing `analytics.userId`? One difference I can think of is that the old one persists across app relaunches, while the new one is cleared and starts fresh after a kill/relaunch. For the problem we're solving here, I think either could have worked? Or did you want to continue with existing approach if the app was relaunched? Or was there another reason for introducing a new one?

mahmoud-elmorabea: Yes, this is deliberate. Using `analytics.userId()` / `registeredUserId` would persist the dedup across cold starts and an app that only does it once per session ends up doing it once per its lifetime. 

I am trying to prevent abuse or buggy code that calls identify way too much in a single session. But once per app launch is okay for now.

Once we finish our data investigation we might decide to restrict this even more, but for now. This prevents abuse and limits the blast radius of the change.

mrehan27 (approved):

PR #697, opened 2026-05-14
chore: add geofence sync repository

mrehan27:
## Summary
- Adds `GeofenceRepository.refresh(lat, lng)` — single entry point for the geofence sync pipeline: userId lookup → API fetch → persist regions + timestamp → distance filter (`config.maxBusinessGeofences`) → register top-N with `GeofenceManager`.
- No-userId / blank-userId → log skipped + `Result.success(Unit)` (deliberate skip, matches the EventBus tracker precedent for "no user, no work").
- API failure → log + propagate `Result.failure`. **No cached-data fallback in this PR** — backend is unavailable for months, so the failure path is already the dominant code path; lifecycle/retry semantics belong in PR 5.
- Manager failure → propagate, but regions stay cached for the next attempt to use.
- **Concurrency:** `Mutex` wraps `refresh()`. Two parallel triggers (e.g., movement-trigger fires during a manual refresh) would otherwise race `GeofencingClient.addGeofences` and leave the OS with inconsistent state. Repo is registered as `singleton` so the mutex actually serializes across consumers.
- Adds `GeofenceLogger.logSyncSucceeded` / `logSyncSkipped`.
- DI: registers `geofenceRegionStore` (missing from PR 1, first consumer adds it) + `geofenceRepository`.

## Stack
Targets `mbl-1623-geofence-api` (PR #692). Will auto-retarget on merge.

## Linear

## Test plan
- [x] 6 unit tests in `GeofenceRepositoryTest`:
  - skip on null userId (no API call, no register)
  - skip on blank userId
  - API failure → propagate + no persist + no register
  - success → persist + setLastSyncTimestamp + filter by `config.maxBusinessGeofences` + register + `logSyncSucceeded`
  - manager failure → propagate but regions still persisted
  - **concurrent refreshes → serialized through Mutex** (atomic counter inside the manager mock asserts max-in-flight = 1; the test fails without `withLock`)
- [x] `./gradlew :location:testDebugUnitTest` — green
- [ ] End-to-end smoke testing waits on PR 5 (trigger wiring) + a debug-only sample-app fake `GeofenceApiService` (follow-up); backend endpoint is not coming soon.

---

> [!NOTE]
> **Medium Risk**
> Introduces a new geofence sync entry point that coordinates network fetch, persistence, and OS registration with concurrency/user-change guards; issues here could lead to missed or stale geofence registrations.
> 
> **Overview**
> Adds a new `GeofenceRepository.refresh(lat, lng)` as the **single entry point** for geofence syncing: it looks up the current `userId`, fetches regions from the API, persists regions + last-sync timestamp, filters to the nearest `config.maxBusinessGeofences`, then registers them via `GeofenceManager`.
> 
> Adds **concurrency and state-safety** protections: an in-flight refresh is deduped (skipped with success), and a mutex-protected write block rechecks `userId` to avoid persisting/registering geofences if the user changes/signs out mid-refresh.
> 
> Wires the repository + missing `GeofenceRegionStore` into DI, extends `GeofenceLogger` with `logSyncSucceeded`/`logSyncSkipped`, and adds comprehensive unit tests covering skip/failure/success paths and refresh deduping.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bceaacc013f1cad4412a53e51a4ad30e8b6961ad. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/location/src/main/kotlin/io/customer/location/geofence/GeofenceRepository.kt
+++ b/location/src/main/kotlin/io/customer/location/geofence/GeofenceRepository.kt
@@ -7,6 +7,7 @@ import io.customer.location.geofence.api.toDomainConfig
 import io.customer.location.geofence.api.toDomainRegions
 import io.customer.location.geofence.store.GeofenceRegionStore
 import io.customer.sdk.data.store.SecureUserStore
+import java.util.concurrent.atomic.AtomicBoolean
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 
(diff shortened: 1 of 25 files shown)


location/src/main/kotlin/io/customer/location/geofence/GeofenceRepository.kt
```diff
@@ -0,0 +1,66 @@
...
+    private val secureUserStore: SecureUserStore,
+    private val logger: GeofenceLogger
+) : GeofenceRepository {
+
+    // Serializes refresh() so concurrent triggers can't race
+    // GeofencingClient.addGeofences and leave inconsistent OS state.
+    private val refreshMutex = Mutex()
+
+    @RequiresPermission(allOf = [Manifest.permission.ACCESS_FINE_LOCATION, Manifest.permission.ACCESS_BACKGROUND_LOCATION])
+    override suspend fun refresh(latitude: Double, longitude: Double): Result<Unit> = refreshMutex.withLock {
```
mahmoud-elmorabea: That lock is held for way too long, I think we only want to avoid double requesting, a simple in-flight status check and early exist should be good, no? The lock only for state checking/update

mrehan27: Good call — refactored. The mutex now only wraps the state writes (API call runs lock-free), and an `AtomicBoolean` gate drops duplicate triggers up front instead of queueing them. userId is rechecked inside the lock to handle sign-out during the API call. Same fix covers your #698 comment.

location/src/main/kotlin/io/customer/location/geofence/GeofenceRepository.kt
```diff
@@ -0,0 +1,90 @@
...
+                        val currentUserId = secureUserStore.getUserId()
+                        if (currentUserId != userId) {
+                            logger.logSyncSkipped("user changed during refresh")
+                            return@withLock Result.success(Unit)
+                        }
+                        store.saveAll(regions)
+                        store.setLastSyncTimestamp(System.currentTimeMillis())
+                        manager.addGeofences(nearest).also { result ->
+                            if (result.isSuccess) logger.logSyncSucceeded(nearest.size)
+                        }
```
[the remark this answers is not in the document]
mrehan27: Correct — `addGeofences` is additive and successive syncs would accumulate stale OS registrations otherwise. This is addressed in the next PR in the stack (#699): it diffs the new set against the persisted store and removes any IDs that fell out before registering the new set.

mahmoud-elmorabea (approved):

PR #698, opened 2026-05-14
chore: wire geofence sync triggers and services facade

mrehan27:
## Summary
Closes the MBL-1623 stack by wiring `GeofenceRepository` into two trigger sources and giving the OS-side machinery something to actually fire.

- **`GeofenceServices` facade** — owns the coroutine scope, null-location gate, and permission gate. Two entry points: `onMovementTriggerExit(lat?, lng?)` and `onUserIdentified(lat?, lng?)`. Fire-and-forget delegation to `repository.refresh(...)`.
- **`GeofencePermissionChecker`** — small Context-scoped class for `ACCESS_FINE_LOCATION` + `ACCESS_BACKGROUND_LOCATION` (mirrors the `GeofenceReceiverToggle` pattern). Short-circuits sync before wasted network calls when permissions are revoked.
- **Movement-trigger creation in `GeofenceRepository`** — when business geofences are non-empty, prepend a synthetic movement trigger (`cio_movement_trigger`, EXIT-only, radius from `config.movementTriggerRadius`) centered on the request location. When business is empty, register nothing — **zero OS-side cost for accounts without configured geofences**.
- **Broadcast receiver** — replaced the `TODO(MBL-1623)` "do nothing" path with a dispatch to `services.onMovementTriggerExit`. Defensive guard: non-EXIT transitions on the movement-trigger ID are logged and ignored (movement trigger is registered `NO_INITIAL_TRIGGER`, so ENTER should never fire).
- **`ModuleLocation.initialize`** — after the existing `locationTracker.onUserIdentified()`, hand the cached location to `services.onUserIdentified(...)` so an identify event seeds the initial sync.
- **`LocationTracker.lastLocation`** — promoted to `internal var ... private set` so `ModuleLocation` can hand it to services without a wrapper getter.
- **Logger** — removed stale `logMovementTriggerSkipped` ("lands in MBL-1623" — now landed). Added `logSyncTriggered`, `logSyncSkippedNoLocation`, `logSyncSkippedNoPermission`, `logMovementTriggerIgnoredNonExit`.

## Threading
No new races introduced:
- `GeofenceServicesImpl.triggerSync` has no shared mutable state — args are local, deps are thread-safe, `scope.launch` is safe from any thread.
- The critical section (network + persist + register) remains protected by the `Mutex` in `GeofenceRepositoryImpl.refresh()` (PR #697).
- Receiver's scope is fresh-and-cancelled-per-invocation; services holds its own long-lived scope — receiver dying does not kill in-flight refresh.
- `LocationTracker.lastLocation` is `@Volatile` — safe cross-thread read.

## Stack
Targets `mbl-1623-geofence-repository` (PR #697). Will auto-retarget on merge.

## Linear

## Test plan
- [x] `GeofenceServicesTest` (4 tests): launches refresh on valid location+permissions; skips on null location; skips on permission denied; reason-string differs per trigger source.
- [x] `GeofenceRepositoryTest`: success path now asserts movement trigger is prepended with the right ID/coords/radius/transitionTypes; new zero-business-geofences test pins the "0 impact" contract (nothing registered, no movement trigger).
- [x] `GeofenceBroadcastReceiverTest`: movement-trigger EXIT routes to services with the triggering lat/lng; non-EXIT defensive guard ignores stray ENTER/DWELL on movement-trigger ID.
- [x] `./gradlew :location:testDebugUnitTest` — green.

## Known follow-ups (not in this PR)
- **Boot receiver restore** — `GeofenceBootReceiver` is still a stub. After reboot, the OS wipes all geofences; we need to re-register from `GeofenceRegionStore`. Has its own location-source question (no live trigger, no live user) so deferred.
- **Stale-geofence cleanup** — `manager.addGeofences` doesn't remove old IDs that aren't in the new list; geofences accumulate across refreshes. Needs diff-or-clear logic.
- **Movement-trigger rollback** — `GeofenceManager` removes the movement trigger if the business batch fails to "avoid orphans," which works against the "safety net always there" intent. Worth inverting.
- **Backend endpoint** — still `ENDPOINT_PATH = ""` from PR #692; sync returns `Result.failure` cleanly until backend ships.

---

> [!NOTE]
> **Medium Risk**
> Introduces new geofence sync trigger paths (identify + movement-trigger EXIT) and new permission gating, which can change when/if geofences are registered and refreshed in production environments.
> 
> **Overview**
> Wires geofence refreshes to real trigger sources by adding a `GeofenceServices` facade that gates sync on *location availability* and *runtime permissions*, then launches `GeofenceRepository.refresh(...)` on a dedicated scope.
> 
> Movement-trigger geofence callbacks are now handled in `GeofenceBroadcastReceiver` (EXIT-only, with defensive logging for non-EXIT), and `ModuleLocation` kicks off an initial sync after `identify` using `LocationTracker.lastLocation` (now `internal` with a private setter).
> 
> Geofence registration is updated to (1) prepend a synthetic movement-trigger region when business geofences exist, (2) register nothing when there are zero business geofences (no OS-side cost), and (3) centralize permission checks via a new `GeofencePermissionChecker`, with expanded logging and new tests covering these paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b874da991fca5a9a175a0cc25f2f22f5c6d5fad1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/location/src/main/kotlin/io/customer/location/geofence/GeofenceRepository.kt
+++ b/location/src/main/kotlin/io/customer/location/geofence/GeofenceRepository.kt
@@ -7,6 +7,7 @@ import io.customer.location.geofence.api.toDomainConfig
 import io.customer.location.geofence.api.toDomainRegions
 import io.customer.location.geofence.store.GeofenceRegionStore
 import io.customer.sdk.data.store.SecureUserStore
+import java.util.concurrent.atomic.AtomicBoolean
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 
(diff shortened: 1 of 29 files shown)


location/src/main/kotlin/io/customer/location/geofence/GeofenceRepository.kt
```diff
@@ -53,7 +53,14 @@ internal class GeofenceRepositoryImpl(
                     longitude = longitude,
                     max = config.maxBusinessGeofences
                 )
-                manager.addGeofences(nearest).also { result ->
+                // Zero business geofences => register nothing (no movement trigger either).
```
mahmoud-elmorabea: I just noticed can we narrow down `refreshMutex.withLock` to only state and maybe early exit if something already is in-flight?

This will hold the lock till the entire request is done? Could be few seconds

location/src/main/kotlin/io/customer/location/geofence/GeofenceServices.kt
```diff
@@ -0,0 +1,55 @@
...
+            logger.logSyncSkippedNoPermission(reason)
+            return
+        }
+        logger.logSyncTriggered(reason)
+        // Guarded by permissionChecker above; Android kills the process when
+        // permissions are revoked, so no mid-flight revocation to handle.
+        @SuppressLint("MissingPermission")
+        scope.launch {
+            repository.refresh(latitude, longitude)
+        }
```
mahmoud-elmorabea, line 48: Is this going to make a request every time the user moves out of the movement trigger? If that's 1k, this is probably too aggressive.

Also, does this fire in the background? This will probably fail then

location/src/main/kotlin/io/customer/location/geofence/GeofenceRepository.kt
mrehan27: Same fix as discussed on #697 — `refreshMutex` replaced with an `AtomicBoolean` in-flight gate (drops duplicate triggers up front) plus a small mutex held only around the state writes. API call runs lock-free. userId is rechecked inside the lock to handle sign-out racing with refresh.

location/src/main/kotlin/io/customer/location/geofence/GeofenceServices.kt
mrehan27: Both valid — current behavior is temporary. Next PR introduces tiered refresh (local cache re-sort at 5km, remote API fetch at 10km), so 1km EXITs stop hitting the network. Background reliability for the remote tier handled there too.

mahmoud-elmorabea (approved):

PR #700, opened 2026-05-16
chore: buffer track/identify/screen calls made before initialize completes

mahmoud-elmorabea:
## Summary
- Pre-init event calls on `CustomerIO.instance()` are now absorbed into a bounded FIFO (capacity 100, drop-most-recent on overflow) and replayed in order once `CustomerIO.initialize()` completes.
- Behavior matches the iOS sibling (customerio-ios#1054): **no new API, no migration**. Existing callers — `CustomerIO.instance().track(...)`, wrapper SDK bridges, host-app cold-start code — benefit automatically. The legacy `IllegalStateException` thrown by pre-init `instance()` is gone.

This is item 2 of a cross-SDK plan to port targeted improvements from the proposed iOS SDK rewrite back to the current native SDKs.

## Public API — unchanged
| Symbol | Behavior |
|---|---|
| `CustomerIO.instance(): CustomerIO` | **Return type unchanged.** Now lazy: returns the singleton whether or not `initialize()` has run. Never throws. |
| `CustomerIO.initialize(config)` | Unchanged signature. Now wires real components into the existing singleton instead of constructing a new one. |

`api/datapipelines.api` diff: removes one `public synthetic` constructor (the compiler artifact for the now-no-arg primary constructor — the constructor was always `private`, so no external caller could have invoked it).

## How it works
- `CustomerIO`'s private primary constructor is now no-arg. The heavyweight setup (Analytics instance, plugins, event-bus subscriptions, identity broadcast) moved from `init {}` into a `synchronized initializeComponents()` method called from `Companion.createInstance()` exactly once.
- Runtime state (`analytics`, `moduleConfig`, `globalPreferenceStore`, `deviceStore`, `contextPlugin`) is `lateinit` and gated by an `isReady`/`componentsReady` predicate.
- Each event-shaped `*Impl` method checks `isReady`. When ready, calls execute against `analytics` directly. When not ready, calls enqueue onto a `PreInitEventBuffer` (ReentrantLock-backed FIFO).
- On `initialize()` the buffer drains synchronously against the now-ready singleton **before** any later caller observes the ready state — so the first post-init call cannot race ahead of buffered ones.
- Read-side properties (`userId`, `anonymousId`, `registeredDeviceToken`, `isUserIdentified`, `profileAttributes`/`deviceAttributes` getters) return safe defaults (`null` / empty string / empty map) pre-init.

## Scope of buffering
All event-shaped methods on `DataPipelineInstance` buffer when called pre-init:
- `track`, `identify`, `screen` (incl. deprecated typed overloads)
- `clearIdentify`, `setProfileAttributes`, `setDeviceAttributes`
- `registerDeviceToken`, `deleteDeviceToken`, `trackMetric`

## Decisions worth flagging
- **Buffer cap: 100.** Matches the iOS sibling.
- **Overflow policy: drop most recent.** Deliberate divergence from the iOS rewrite (which drops oldest). Preserves the earliest events — install attribution, initial identify — which tend to be the highest-value signals.
- **No `initialize()` ever called → buffer sits at cap.** No TTL, no cleanup. Memory is bounded.
- **Order preserved through identify boundaries.** A sequence like `[identify(A), track(x), clearIdentify, identify(B), track(y)]` drains in that exact order.
- **Debug-level logs on overflow drops and drain.** Counter tracked across pre-init; logged once on drain.

## Test plan
- [x] 9 new unit tests in `PreInitEventBufferTest.kt`: state transitions, overflow, drain order, reentrancy, concurrent enqueue, post-init direct execution, idempotent transition.
- [x] All existing `datapipelines` unit tests pass.
- [x] `messaginginapp`, `messagingpush`, `location` unit tests pass.
- [x] API snapshot regenerated; the only diff is removal of the compiler-synthetic constructor.
- [ ] CI green on this PR.
- [ ] Local wrapper validation (RN + Flutter sample apps with deliberate pre-`initialize()` `track()` call) — done locally, no PRs.

---

> [!NOTE]
> **Medium Risk**
> Changes the core `CustomerIO` singleton initialization and dispatch path, which can affect event ordering, concurrency, and identity/device-token handling during app startup.
> 
> **Overview**
> **Adds pre-initialization buffering for event-shaped SDK calls.** `CustomerIO.instance()` is now lazily created and no longer throws pre-init; calls like `track`/`identify`/`screen` (and related setters/token/metric APIs) enqueue into a bounded FIFO (`PreInitEventBuffer`, cap 100, drop-most-recent) and replay in order once initialization completes.
> 
> **Refactors initialization lifecycle and ordering.** Heavy setup moves into `initializeComponents()`, `initialize()` is synchronized and idempotent, and `finishInitialization()` now publishes the initial `UserChangedEvent` *after* all modules have initialized and then drains the buffer; read-side accessors return safe defaults (and log debug) until ready. Adds unit tests covering buffer behavior, init idempotency, and module-subscription/event ordering; API snapshot updates by removing a synthetic constructor.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cdd1c8debfe4085a4eb4611d4f7f79276dd525d1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
```diff
@@ -200,15 +275,27 @@ class CustomerIO private constructor(
...
     override var profileAttributes: CustomAttributes
-        get() = analytics.traits() ?: emptyMap()
+        get() = if (isReady) analytics.traits() ?: emptyMap() else emptyMap()
         set(value) {
             setProfileAttributes(value)
         }
 
     override fun setProfileAttributes(attributes: CustomAttributes) {
-        val identifier = this.userId
+        if (!isReady) {
```
mahmoud-elmorabea: Let's create a `dispatch` helper like in https://github.com/customerio/customerio-ios/pull/1054 to avoid repeating that check everywhere

datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
```diff
@@ -317,16 +440,16 @@ class CustomerIO private constructor(
...
-        get() = analytics.anonymousId()
+        get() = if (isReady) analytics.anonymousId() else ""
 
     override val userId: String?
-        get() = analytics.userId()
+        get() = if (isReady) analytics.userId() else null
 
     override val isUserIdentified: Boolean
-        get() = !analytics.userId().isNullOrEmpty()
+        get() = if (isReady) !analytics.userId().isNullOrEmpty() else false
```
mahmoud-elmorabea: These property access calls shouldn't silently return default values

datapipelines/src/main/kotlin/io/customer/sdk/PreInitEventBuffer.kt
```diff
@@ -0,0 +1,177 @@
...
+        internal const val LOG_TAG = "PreInitEventBuffer"
+    }
+
+    private sealed class State {
+        data class Buffering(val blocks: MutableList<Block>) : State()
+        data class Draining(val impl: DataPipelineInstance, val pending: MutableList<Block>) : State()
+        data class Ready(val impl: DataPipelineInstance) : State()
+    }
+
+    private val lock = ReentrantLock()
```
mahmoud-elmorabea: Consider using kotlin's synchronized on this object instead

hollyschilling-cio (changes requested):
This doesn't serve the same purpose as it does on iOS. Specifically, on Android the initialization is entirely synchronous. Given the default installation suggested on the [Quick Start Guide](https://docs.customer.io/integrations/sdk/android/quick-start-guide/), this only collects events that occur before the SDK initialization begins. No Activity can be created before that point. 

This also introduces potential issues. If the hosting app spawns a thread before init and that thread creates events rapidly, the draining process will hang indefinitely causing the hosting app to trigger ANR on startup. Additionally, this suppresses the previous `IllegalStateException` that was triggered if an event was submitted without the SDK being initialized. That means that if the SDK initialization gets removed from the app's startup sequence, there is no error to indicate the problem to the developer and events simply don't make it to the backend. 

On iOS, disk operations are handled asynchronously, which necessitates an async startup. That is not the case on Android. Adding this does not solve any existing pain points on Android.

PR #701, opened 2026-05-18
fix: drain pre-init buffer after modules subscribe; preserve setter ordering

mahmoud-elmorabea:
## Summary
Addresses both P2 review comments on #700.

- **Deferred drain (P2 #2)** — A new `CustomerIO.finishInitialization()` publishes the initial `UserChangedEvent` and drains the pre-init buffer **after** every module has run its `initialize()`. Late-subscribing modules (e.g. `MessagingInApp`) attach before the identity event is published, so a screen-heavy buffer can no longer evict it from the SharedFlow replay window.
- **Setter ordering (P2 #1)** — `isReady` now sources from `preInitBuffer.isReady` instead of a separate `componentsReady` flag. The readiness signal flips atomically with the drain completing, so `setProfileAttributes` / `setDeviceAttributes` (which don't synchronize via `DataPipelineInstance`) route through the buffer for the entire pre-init and drain window, preserving FIFO order with earlier buffered calls.
- **Recursion-free replay** — Buffered closures now invoke private `*Internal` methods that contain the bodies without the readiness gate, so replay against the real impl doesn't re-enter `enqueue`. The companion-level double-init guard moved to a separate `initStarted` flag latched at the start of `initializeComponents()`.
- **Capacity during drain** — `PreInitEventBuffer.enqueue` now enforces capacity + drop accounting in the `Draining` state as well as `Buffering`, mirroring the iOS sister fix.

Aligned with `customerio-ios-tng` (no deliberate divergence): module wiring, identity broadcast, and event-stream drain occur in that order.

## Test plan
- [x] `./gradlew :datapipelines:testDebugUnitTest` — full suite green, including:
  - New `PreInitEventBufferTest.drainingStateEnforcesCapacity` — reentrant enqueues during drain hit the cap.
  - New `FinishInitializationOrderTest.givenScreenHeavyPreInitBuffer_*` — late-subscribing module receives the initial `UserChangedEvent` even when 101 pre-init screens are queued; buffered screens publish after subscription.
- [x] `./gradlew :messaginginapp:testDebugUnitTest` — green.
- [x] `./gradlew :datapipelines:lintDebug` — green.
- [x] `./ktlint --android` on changed files — clean.
- [ ] CI lint + tests (this PR).

### DIFF
--- a/datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
+++ b/datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
@@ -87,9 +87,18 @@ class CustomerIO private constructor() :
     // initializeComponents() runs and replays them in order on init.
     private val preInitBuffer = PreInitEventBuffer()
 
+    // Single source of truth for "the SDK has finished initializing AND the
+    // pre-init buffer has been fully drained." Reading from the buffer means
+    // public setters that arrive during the drain window route through
+    // preInitBuffer.enqueue (which puts them on `pending`) instead of racing
+    // ahead of earlier buffered calls.
+    private val isReady: Boolean get() = preInitBuffer.isReady
+
+    // Latched at the start of [initializeComponents] to make the companion
+    // [initialize] idempotent. Distinct from [isReady] so the double-init
+    // guard fires even before the drain completes.
     @Volatile
-    private var componentsReady: Boolean = false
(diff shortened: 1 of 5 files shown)


mahmoud-elmorabea:
Rolled into the original PR #700 as commit `bdc4e1f4` (pushed directly onto `pre-init-event-buffer`). Closing this duplicate.

PR #702, opened 2026-05-18
chore: harden geofence sync lifecycle

mrehan27:
## Summary

Lifecycle hardening for the geofence sync pipeline so a refresh behaves correctly across user-state and app-state transitions, and so the SDK keeps working through gradual backend rollout of the geofence config API.

- **Receiver disabled when there are no geofences.** `GeofenceManager.addGeofences(emptyList)` now disables the broadcast receiver. Covers fresh-zero accounts and has→0 transitions so we don't burn resources listening for events that can't fire.
- **Sign-out cleanup via `ResetEvent`.** `clearIdentify()` publishes `ResetEvent` before `UserChangedEvent(null)`, so it's the explicit "wipe user state" signal — analogous to `analytics.reset()`. One subscription clears both `geofenceServices.onUserSignedOut()` (store + OS-side registrations) and `geofenceCooldownFilter.clearAll()`.
- **24h freshness throttle (`STALE_THRESHOLD_MS`).** Repeated identify/app-launch within the window is a cheap no-op. Movement-trigger EXIT passes `force = true` to bypass the throttle so the trigger's center can be updated.
- **Defensive app-launch trigger.** If a user identified in a previous session is still persisted, `ModuleLocation.initialize()` kicks a refresh now. The 24h threshold makes it a no-op when identify also fires shortly after init (the common case), so it only does work when the host doesn't call identify on this launch.
- **`setLastSyncTimestamp` only on full success.** Moved inside the `if (result.isSuccess)` block so partial failures (manager add fails) leave the timestamp stale and the next threshold-based retry actually fires.
- **Stale-cleanup reordered to AFTER successful add (7g).** `removeGeofencesByIds` now runs only when `addGeofences` succeeded. Original "remove-then-add" destroyed previous OS state on add failure, leaving the device with no geofences at all until the next refresh.
- **Per-field fallbacks in `GeofenceApiConfig.toDomain()`.** API DTO fields now nullable with `= null` defaults; `toDomain()` applies `GeofenceConstants.FALLBACK_*` per-field when missing or non-positive. Lets the SDK keep working through gradual backend rollout / bogus values.
  - Exception: `maxBusinessGeofences` passes through as 0 (no `toDomain` fallback) so the repository can log via `logInvalidMaxBusinessGeofences(max, fallback)` then apply `FALLBACK_MAX_BUSINESS_GEOFENCES`. Logging belongs at the boundary that has the logger; `toDomain` stays pure.
- **Brief inline comments on every `GeofenceConstants` entry.** Numeric literals use `_` separators (1_000f, 5_000f, 10_000f).

## Linear

## Stack

- #702 👈🏻 
- #703
- #704 
- #705 
- #706 
- #707 
- #708  

## Test plan

- [x] Unit tests pass: `./gradlew :location:testDebugUnitTest`
- [x] Existing refresh-path tests updated for new `force` param
- [x] New tests:
  - Repository: threshold skip / stale / forced / no-prior-sync, invalid `maxBusinessGeofences` fallback + log, `coVerifyOrder { add; remove }` for 7g, `removeGeofencesByIds` NOT called on add failure, `reset()` clears store + manager
  - Services: `force = true` for movement EXIT, `force = false` for identify and app-launch, `onUserSignedOut` hand-off
  - Manager: empty-list path asserts receiver disabled; mixed-batch failure no longer enables receiver
  - `GeofenceApiConfig.toDomain()`: all config fields missing / zero / negative both fall back to SDK defaults
- [x] Manual verification deferred — possible once the live backend endpoint lands

---

> [!NOTE]
> **Medium Risk**
> Touches geofence refresh/registration lifecycle, including new sign-out reset behavior and freshness throttling; mistakes could suppress expected geofence updates or leave stale OS registrations. Changes are well-covered by updated/new unit tests but still affect runtime behavior in a device/OS-integrated area.
> 
> **Overview**
> Improves geofence sync robustness across app/user lifecycle events by adding a **freshness throttle** to `GeofenceRepository.refresh()` (with a `force` bypass for movement-trigger EXIT) and triggering a defensive `onAppLaunch()` refresh when a persisted user exists.
> 
> Adds explicit **sign-out cleanup**: `ModuleLocation` listens for `Event.ResetEvent` to call `geofenceServices.onUserSignedOut()`, which now clears OS registrations and persisted state via a new `GeofenceRepository.reset()`.
> 
> Hardens OS registration behavior by disabling the receiver when registering an empty geofence set, reordering stale geofence removal to run **only after** a successful add, and making API config parsing tolerant of missing/invalid fields via nullable DTOs plus `GeofenceConstants.FALLBACK_*` defaults (including bounds on `maxBusinessGeofences`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit be3b73a0181223a6ab73b0558d62d1da1fbb4b6c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/location/src/main/kotlin/io/customer/location/geofence/GeofenceManager.kt
+++ b/location/src/main/kotlin/io/customer/location/geofence/GeofenceManager.kt
@@ -39,6 +39,11 @@ internal class GeofenceManager(
     @RequiresPermission(allOf = [Manifest.permission.ACCESS_FINE_LOCATION, Manifest.permission.ACCESS_BACKGROUND_LOCATION])
     suspend fun addGeofences(regions: List<GeofenceRegion>): Result<Unit> {
         if (regions.isEmpty()) {
+            // No geofences to register => disable the receiver so we don't burn
+            // resources listening for events that can't fire. Covers both the
+            // fresh-account-with-no-geofences case and the account-transitioned-to-0
+            // case (where stale cleanup just removed the previous registrations).
+            receiverToggle.setEnabled(false)
             logger.logGeofencesRegistered(0)
             return Result.success(Unit)
         }
@@ -61,7 +66,11 @@ internal class GeofenceManager(
         if (businessGeofences.isNotEmpty()) {
             val result = registerBatch(businessGeofences, initialTrigger = GeofencingRequest.INITIAL_TRIGGER_ENTER)
             if (result.isFailure) {
-                // Clean up movement trigger if business batch fails to avoid orphaned geofences
+                // Roll back the movement trigger so we don't leave an unattended
+                // safety-net geofence in the OS without any business geofences to
+                // recover. Business-batch failure is a rare edge case (transient OS
+                // / GMS issue); recovery happens on the next identify or app-launch
+                // trigger past the freshness threshold.
                 if (movementTrigger.isNotEmpty()) {
                     removeGeofencesByIds(movementTrigger.map { it.id })
                 }
--- a/location/src/main/kotlin/io/customer/location/geofence/GeofenceRepository.kt
+++ b/location/src/main/kotlin/io/customer/location/geofence/GeofenceRepository.kt
@@ -16,8 +16,21 @@ import kotlinx.coroutines.sync.withLock
  * current user/location, persist them, then register the closest N with the OS.
  */
 internal interface GeofenceRepository {
+    /**
+     * @param force when true, bypasses the [GeofenceConstants.STALE_THRESHOLD_MS]
+     *   freshness check. Use for triggers that need a guaranteed re-fetch (e.g.
+     *   movement-trigger EXIT, where the trigger's center must be updated).
+     *   When false, a recent successful sync short-circuits the API call.
+     */
     @RequiresPermission(allOf = [Manifest.permission.ACCESS_FINE_LOCATION, Manifest.permission.ACCESS_BACKGROUND_LOCATION])
-    suspend fun refresh(latitude: Double, longitude: Double): Result<Unit>
+    suspend fun refresh(latitude: Double, longitude: Double, force: Boolean = false): Result<Unit>
+
+    /**
+     * Clears all geofence state: persisted regions in the store and OS-side
+     * registrations via the manager. Called on user sign-out so the next user
+     * (or anonymous session) doesn't inherit the previous user's geofences.
+     */
+    suspend fun reset(): Result<Unit>
 }
 
 internal class GeofenceRepositoryImpl(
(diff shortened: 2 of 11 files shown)


location/src/main/kotlin/io/customer/location/geofence/api/GeofenceApiResponse.kt
```diff
@@ -13,25 +14,28 @@ internal data class GeofenceApiResponse(
...
-    val remoteFetchRefreshTriggerRadius: Float,
+    val remoteFetchRefreshTriggerRadius: Float? = null,
     @SerialName("remote_fetch_refresh_expiry_time")
-    val remoteFetchRefreshExpiryTime: Long,
+    val remoteFetchRefreshExpiryTime: Long? = null,
     @SerialName("duplicate_events_expiry_time")
-    val duplicateEventsExpiryTime: Long,
-    val android: GeofenceApiPlatformConfig
+    val duplicateEventsExpiryTime: Long? = null,
+    val android: GeofenceApiPlatformConfig? = null
```
mahmoud-elmorabea, line 32: I like the idea of resilience if the backend returns non-expected output. But I'm wondering if we will just fallback to defaults that might be not great for us or rather just no-op the feature if we don't get valid response temporarily from server

location/src/main/kotlin/io/customer/location/geofence/GeofenceManager.kt
```diff
@@ -39,6 +39,11 @@ internal class GeofenceManager(
     @RequiresPermission(allOf = [Manifest.permission.ACCESS_FINE_LOCATION, Manifest.permission.ACCESS_BACKGROUND_LOCATION])
     suspend fun addGeofences(regions: List<GeofenceRegion>): Result<Unit> {
```
mahmoud-elmorabea, line 40: The semantics of this method is to always replace any existing fences with the new list, correct? In that case it would make sense that an empty list means disable the receiver. But maybe we can consider a rename? or maybe docs? `addGeofences` doesn't signal that behavioral

location/src/main/kotlin/io/customer/location/geofence/GeofenceRepository.kt
```diff
@@ -51,11 +63,21 @@ internal class GeofenceRepositoryImpl(
                 logger.logSyncSkipped("no identified user")
                 return Result.success(Unit)
             }
+
+            if (!force) {
+                val lastSync = store.getLastSyncTimestamp()
+                if (lastSync != null && System.currentTimeMillis() - lastSync < GeofenceConstants.STALE_THRESHOLD_MS) {
```
mahmoud-elmorabea, line 69: Can we use our time abstraction here instead of the static call to `currentTimeMillis()`?

location/src/main/kotlin/io/customer/location/geofence/GeofenceRepository.kt
```diff
@@ -122,6 +147,23 @@ internal class GeofenceRepositoryImpl(
...
+        // Order matters: clear OS-side FIRST, then the persisted record. If
+        // manager.clearAll fails (transient GMS error), preserving the store
+        // lets the next refresh's stale-cleanup diff see the previous regions
+        // and retry their removal — without it, OS state orphans with no
+        // record to drive the cleanup.
+        manager.clearAll().also { result ->
+            if (result.isSuccess) {
+                store.clearAll()
+            }
+        }
```
mahmoud-elmorabea, line 164: Should this also clear the refresh last sync timestamp?

mahmoud-elmorabea (approved):

location/src/main/kotlin/io/customer/location/geofence/api/GeofenceApiResponse.kt
mrehan27: The backend doesn't expose config options for this yet, so I'm designing it in a way that the SDK behaves safely without them. Once the backend adds support, the SDK should automatically start leveraging it. This will improve further in upcoming PRs.

location/src/main/kotlin/io/customer/location/geofence/GeofenceManager.kt
mrehan27: Good point, will rename it to `replaceGeofences` in the follow up PR addressing review feedback.

location/src/main/kotlin/io/customer/location/geofence/GeofenceRepository.kt
mrehan27: Yes, currently it does. But good call, this makes me think we should evaluate whether we actually want to persist this across logout and re-trigger it on identify. I'll take another look and update if needed in follow up PR.

PR #703, opened 2026-05-18
chore: prepare storage for tiered geofence refresh

mrehan27:
## Summary

Foundation for the upcoming tiered-refresh PR — caches everything the future tier dispatch needs without yet wiring the new dispatch path. **No behavior change at the movement-EXIT path**; that lands in the follow-up. This PR is intentionally narrow and reviewable in isolation.

## Stack

- #702 
- #703 👈🏻 
- #704 
- #705 
- #706 
- #707 
- #708  

## Background on the tiered-refresh flow

For reviewer context — the next PR will implement two tiers per movement EXIT:

| Tier | Condition | Action |
|---|---|---|
| **A** (local) | distance(current, `lastApiFetchLocation`) < 5km | Re-rank cached regions by current location, register new top-N + new movement trigger. **No API call.** |
| **B** (remote) | distance(current, `lastApiFetchLocation`) ≥ 5km | Fetch from API, replace cache, update anchor, then re-rank + register. |

Both tiers always re-center the 1km movement trigger on the user's current location so the next EXIT fires from there. Today every 1km EXIT hits the API → tier A will eliminate ~most of those calls.

To make that work, the sync pipeline needs three new pieces of persisted state, which this PR adds and populates on every successful `refresh()`.

## Field / constant simplification

The old API config had three radii (`movement_trigger_radius`, `local_refresh_trigger_radius`, `remote_fetch_refresh_trigger_radius`) but only two distinct concepts on the client. Resolved by dropping the redundant one:

| Before | After |
|---|---|
| `movementTriggerRadius` (1km — geofence circle) | **dropped** — same concept as below |
| `localRefreshTriggerRadius` (5km — old "local re-sort threshold") | **repurposed** to 1km, the OS geofence circle radius for the movement trigger |
| `remoteFetchRefreshTriggerRadius` (10km — old "API threshold") | **repurposed** to 5km, the anchor-distance threshold for tier B |

The 10km value was only relevant server-side; nothing on the client needs it. Constants updated:
- `FALLBACK_MOVEMENT_TRIGGER_RADIUS_METERS` → **dropped**
- `FALLBACK_LOCAL_REFRESH_RADIUS_METERS`: 5_000f → **1_000f**
- `FALLBACK_REMOTE_FETCH_RADIUS_METERS`: 10_000f → **5_000f**

`remoteFetchRefreshTriggerRadius` is still parsed from the API config (server-driven); the constant is just the fallback.

## Store refactor

`GeofenceRegionStore` was previously a single `List<GeofenceRegion>` ("registered") + timestamp. The tiered path needs more shape; renamed and expanded:

| Old | New | Purpose |
|---|---|---|
| `saveAll(List<GeofenceRegion>) / getAll()` | `saveRegisteredIds(Set<String>) / getRegisteredIds()` | Subset live in OS — drives the stale-cleanup diff. Stores just IDs (only thing the diff uses). |
| — | `saveCachedRegions(List<GeofenceRegion>) / getCachedRegions()` | Full backend response — source for the future tier-A re-rank. |
| — | `saveCachedConfig(GeofenceConfig) / getCachedConfig()` | Last server-driven thresholds between API fetches. |
| — | `saveLastApiFetchLocation(GeofenceLocation) / getLastApiFetchLocation()` | Single anchor for the tier-B distance check. |
| `getLastSyncTimestamp / setLastSyncTimestamp` | unchanged | 24h freshness throttle for identify / app-launch. |
| `clearAll()` | unchanged | Sign-out cleanup wipes everything. |

`refresh()` success path now populates all four. Failure paths leave them stale (symmetric with the existing `lastSyncTimestamp` behavior — partial-failure retries instead of skipping as "fresh").

## Schema-drift safe deserialization

All cache decodes go through `Json { ignoreUnknownKeys = true }` + try/catch with **key wipe on parse failure**. A future SDK adding fields to `GeofenceConfig` / `GeofenceRegion` won't crash older clients reading older payloads. A corrupted payload returns null/empty (caller falls back to "no cached value") and the bad key is wiped so subsequent reads don't keep failing.

## Json wrapper consolidation

`GeofenceApiResponseParser` and the temporary cache codec collapsed into one `GeofenceJsonSerializer`:

- `decode(serializer, raw)` (strict, throws) — for the API service, where a malformed payload should propagate as `Result.failure`.
- `decodeOrNull(serializer, raw)` (lenient, returns null) — for the store, where schema drift / corruption should degrade gracefully.
- `encode(serializer, value)` — for both.

One DI binding (`SDKComponent.geofenceJsonSerializer`) replaces the two old ones.

## `@SerialName` on every persisted field

`GeofenceConfig`, `GeofenceRegion`, `GeofenceLocation`, and the `GeofenceTransitionType` enum values are now `@SerialName`-annotated. JSON keys are pinned independently of:

- Kotlin field renames in future SDK versions
- R8 / obfuscation behavior
- Any tooling that pokes the bytecode

Three new tests assert the exact JSON keys are present so the contract is enforced — accidentally removing `@SerialName` would now fail a test instead of silently changing the persisted JSON shape.

## Test coverage added

- **`GeofenceJsonSerializerTest`** — encode/decode roundtrip; strict decode throws on invalid JSON; `decodeOrNull` returns null on invalid; forward-compat (unknown fields ignored).
- **`GeofenceRegionStoreTest`** — rewritten for the new API. Roundtrips per accessor; `clearAll` wipes everything; corrupted JSON → null/empty + key wipe; schema-drift forward-compat; `@SerialName` key pinning for config / region / location.
- **`GeofenceRepositoryTest`** — new `refresh_givenSuccess_expectCacheAndConfigAndAnchorPersisted` and `refresh_givenManagerFails_expectCacheAndAnchorNotPersisted` tests; existing tests adapted to `Set<String>` IDs.
- **`GeofenceApiResponseTest`** — sample JSON updated; existing fallback / max-business edge cases all still covered.

## Linear

## Diff stat

15 files, +563 / -178. ~205 production LOC net, ~360 test LOC.

## Test plan

- [x] `./gradlew :location:testDebugUnitTest` passes
- [x] Manual code review of the storage rename — no `saveAll/getAll` left on the region store
- [x] CI green
- [x] Manual verification deferred until the tier-logic PR lands (no behavior change here)

---

> [!NOTE]
> **Medium Risk**
> Changes geofence sync persistence and registration bookkeeping (new cached fields + registered-ID diff) and alters fallback radius constants/field mapping, which could affect refresh/cleanup behavior if assumptions are wrong.
> 
> **Overview**
> Prepares the geofence module for tiered refresh by expanding persisted state from a single stored region list to **separate cached regions/config/anchor location plus a `Set<String>` of OS-registered IDs**, and updates `GeofenceRepository.refresh()` to populate these on successful API+OS registration.
> 
> Replaces `GeofenceApiResponseParser` with a shared `GeofenceJsonSerializer` (strict decode for API, lenient `decodeOrNull` for cache with key-wipe on parse failure), wires it through DI, and pins persisted JSON keys via `@Serializable`/`@SerialName` on `GeofenceConfig`, `GeofenceRegion`, `GeofenceTransitionType`, and new `GeofenceLocation`.
> 
> Simplifies config semantics by **dropping** `movementTriggerRadius` and repurposing `localRefreshTriggerRadius` for the movement trigger radius, while adjusting fallback radius constants (1km local refresh, 5km remote-fetch threshold). Tests are updated/added to cover the new store API, serializer behavior, and persistence guarantees.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6b592f8168fb8d632aef48ee8487af23b9a4bdb9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


location/src/main/kotlin/io/customer/location/geofence/GeofenceJsonSerializer.kt
```diff
@@ -0,0 +1,31 @@
...
+
+    fun <T> encode(serializer: KSerializer<T>, value: T): String =
+        json.encodeToString(serializer, value)
+
+    fun <T> decode(serializer: KSerializer<T>, raw: String): T =
+        json.decodeFromString(serializer, raw)
+
+    fun <T> decodeOrNull(serializer: KSerializer<T>, raw: String): T? = try {
+        decode(serializer, raw)
+    } catch (_: Throwable) {
```
mahmoud-elmorabea, line 28: Can this swallow `CancellationException`?

location/src/main/kotlin/io/customer/location/geofence/store/GeofenceRegionStore.kt
```diff
@@ -48,9 +85,34 @@ internal class GeofenceRegionStoreImpl(
...
+
+    /**
+     * Returns the decoded value, or `null` if the key is absent OR the stored
+     * payload can't be parsed (schema drift, corruption). On parse failure the
+     * key is wiped so a stale, unparseable value won't keep failing on every read.
+     */
+    private fun <T> readJson(key: String, serializer: KSerializer<T>): T? = prefs.read {
+        val raw = getString(key, null) ?: return@read null
+        jsonSerializer.decodeOrNull(serializer, raw) ?: run {
+            prefs.edit { remove(key) }
```
mahmoud-elmorabea, line 104: I am not confident what `prefs.read` does internally but I'd feel better if we don't nest a `prefs.edit` inside of it. Maybe we can extract a read helper func that retuns boolean to indicate failure and key needing to be cleaned up and do the removal serially after the read is complete

mahmoud-elmorabea (approved):

location/src/main/kotlin/io/customer/location/geofence/GeofenceJsonSerializer.kt
mrehan27: Good catch, will address in follow up PR

location/src/main/kotlin/io/customer/location/geofence/store/GeofenceRegionStore.kt
mrehan27: Agreed, will address in follow up PR

location/src/main/kotlin/io/customer/location/geofence/GeofenceJsonSerializer.kt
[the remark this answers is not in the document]
mrehan27: As discussed above, will improve in follow up PR

location/src/main/kotlin/io/customer/location/geofence/store/GeofenceRegionStore.kt
```diff
@@ -48,9 +85,34 @@ internal class GeofenceRegionStoreImpl(
...
+            null
+        }
+    }
```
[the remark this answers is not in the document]
mrehan27: As discussed above, will improve in follow up PR

Issue #10, opened 2021-05-07
Initialization of SDK with CIO configuration


levibostian:
In order for our customers to use the SDK, they must give us some information in order for us to be able to send data to their CIO workspace. This task is to get this information and save it on the device. With this task, we are going to cater the authentication process along with the storage of `authentication token`, also create an interface for the customer to interact and configure the SDK accordingly. 

# Why are we doing this for our customers? 

This task is required for our customers to use the SDK. The SDK cannot do any work successfully without this task. 

# How do we test that this change is successful? 

Not much to test here. If future tasks such as [identify a customer](https://github.com/customerio/customerio-android/issues/3) works after this task, we have tested it successfully. 

# How are we building this? 

- [x] Create a public interfacing class, ideally singleton to configure the SDK with the required options.
```kt
CIO.configure(siteId: String, apiKey: String)
```

Save the `siteId` and `apiKey` to shared preferences to be reused and add the `region` entity to be configured. 

- [x] Create a queue task to [test credentials valid](https://customer.io/docs/api/#operation/testTrackAuth). In development mode, throw an `IllegalArgumentException()`. If not in development, ignore the request so no siteid or apikey get saved. 

- [x] Create a queue task to [find account region](https://customer.io/docs/api/#operation/getRegion). Once response comes back from that, save the country code.

PR #704, opened 2026-05-18
chore: add tiered refresh dispatch on movement EXIT

mrehan27:
## Summary

Final piece of MBL-1629: wires the storage prep from #703 into a real behavior change. Movement-trigger EXIT now goes through a two-tier dispatch instead of always hitting the API — eliminates the ~10× over-call problem where every 1km EXIT hits the backend.

## Stack

- #702 
- #703 
- #704 👈🏻 
- #705 
- #706 
- #707 
- #708  

## The two-tier flow

Every movement EXIT computes straight-line distance from the user's current location to `lastApiFetchLocation` (the cached anchor from the last API fetch):

| Tier | Condition | Action |
|---|---|---|
| **B** (remote) | no anchor OR distance ≥ 5km (`remoteFetchRefreshTriggerRadius`) | Fetch from API → replace cache → update anchor → re-rank → register top-N |
| **A** (local) | within 5km of anchor | Re-rank `cachedRegions` by current location → register top-N. **No API call.** Anchor stays put. |

Both tiers always re-center the 1km movement trigger on the user's current location so the next EXIT fires from there.

Example walk (5km threshold):
- t0: API call, anchor=A. Trigger at A.
- t1: 1.2km away → EXIT. dist(t1, A)=1.2km < 5km → Tier A. Trigger at t1.
- t2: 2.5km from A → EXIT. dist=2.5km < 5km → Tier A. Trigger at t2.
- t3: 5.3km from A → EXIT. dist=5.3km ≥ 5km → **Tier B**. New anchor=t3. Trigger at t3.
- t4: 1.1km from t3 → EXIT. dist(t4, t3)=1.1km < 5km → Tier A. Trigger at t4.

## Surface changes

**Repository interface**
- New: `suspend fun handleMovement(latitude, longitude): Result<Unit>` — for movement EXIT
- Existing `refresh(latitude, longitude)` keeps its identify / app-launch role with the 24h freshness threshold. The `force` parameter is dropped — movement EXIT no longer goes through `refresh()`, so the only caller of `force = true` is gone.

**Services**
- `onMovementTriggerExit` → `repository.handleMovement(...)`
- `onUserIdentified` / `onAppLaunch` → `repository.refresh(...)` (unchanged behaviour)
- `triggerSync` helper generalized to take the suspending action as a lambda so each entry point is explicit about which repository method it invokes.

**Internal refactor**
- `performRemoteRefresh` / `performLocalRefresh` / `registerNearestAndPersist` extracted from the old monolithic `refresh()` body.
- `registerNearestAndPersist` takes `onRegistered: () -> Unit = {}` so Tier B can pass the cache + anchor + timestamp writes at the call site instead of branching on a flag inside the helper.

**Shared in-flight gate**
- `refreshInProgress: AtomicBoolean` now gates both `refresh()` and `handleMovement()`. A movement EXIT during an in-flight identify (or vice versa) drops fast — same dedup guarantee `refresh()` had before, extended across both entry points.

## Thread safety / known limitation

The userId-recheck-inside-mutex pattern from #702 carries through. One race I noticed during review (**pre-existing, not introduced here**): sign-out followed by immediate sign-in as a different user can interleave `reset()` and a subsequent `refresh()` on the shared `geofenceScope`, since `Dispatchers.Default` doesn't guarantee launch-order execution and `reset()` doesn't acquire the in-flight gate. If `refresh()` wins the race, its writes get clobbered by the later-arriving `reset()` and the new user ends up with empty state until the next trigger.

Mitigation in this PR: **none** — it's pre-existing and out of scope. Tracked for a follow-up (likely folds into #10 defensive failure cleanup). The race window only matters when sign-out + sign-in happen within milliseconds of each other.

## Test coverage

New `handleMovement_*` tests:
- `givenNoUserId_expectSkipAndNoOp`
- `givenNoAnchor_expectRemoteFetch` (first EXIT after install / sign-out / clearAll)
- `givenNoCachedConfig_expectRemoteFetch` (defensive)
- `givenAnchorWithinThreshold_expectLocalRerankAndNoApiCall` (~111m from anchor)
- `givenAnchorBeyondThreshold_expectRemoteFetch` (~11km from anchor)
- `givenAnchorBeyondThreshold_expectCacheAndAnchorPersisted` — confirms Tier B via `handleMovement` writes cache/config/anchor/timestamp with the new anchor at the current location
- `givenLocalRerankSucceeds_expectCacheAndAnchorAndTimestampNotUpdated` — Tier A invariant
- `givenSecondCallWhileFirstInFlight_expectSecondDropped` — shared gate

Coverage not duplicated:
- Stale cleanup, userId recheck, manager-add-fails state preservation — all exercised via the shared `registerNearestAndPersist` helper through existing `refresh_*` tests. Adding `handleMovement_*` versions would test the same code under different entry-point names.

Services tests simplified: each entry-point test now verifies which repository method was called (`handleMovement` vs `refresh`) instead of the old `force` param assertion.

Dropped: `refresh_givenRecentSyncButForced_expectApiCalled` — the `force` param it was testing no longer exists.

## Linear

## Diff stat

4 files, +367 / -160. ~75 production LOC net, ~130 test LOC net.

## Test plan

- [x] `./gradlew :location:testDebugUnitTest` passes
- [x] CI green
- [x] Manual verification deferred — backend endpoint not yet live; once live, Tier B will exercise the real API and Tier A will exercise the cache.

---

> [!NOTE]
> **Medium Risk**
> Changes the geofence sync decision logic for movement-trigger EXIT to sometimes skip the API and rely on cached data, which could affect geofence registration accuracy if cache/anchor state is incorrect. Refactors core refresh flow and persistence timing, so regressions could impact when geofences are registered/cleaned up.
> 
> **Overview**
> **Movement-trigger EXIT is no longer a forced API refresh.** The repository adds `handleMovement()` and routes EXIT events through a two-tier dispatch: if the device is within `remoteFetchRefreshTriggerRadius` of the last API fetch anchor, it re-ranks and re-registers using cached regions/config; otherwise it performs a remote fetch and updates the cache/anchor/timestamp.
> 
> `refresh()` is simplified to the identify/app-launch path only (drops the `force` parameter and always honors the freshness window), and the shared implementation is refactored into `performRemoteRefresh`/`performLocalRefresh` plus a common `registerNearestAndPersist` helper with remote-only persistence hooks. `GeofenceServices` is updated to call `handleMovement()` for EXIT and uses a generalized `triggerSync` lambda; tests are updated accordingly with new tier-dispatch coverage and removed `force` assertions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 99d90363c77ac4ec4787bce75605d74f7f4ff597. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


location/src/main/kotlin/io/customer/location/geofence/GeofenceRepository.kt
```diff
@@ -63,89 +65,145 @@ internal class GeofenceRepositoryImpl(
                 logger.logSyncSkipped("no identified user")
                 return Result.success(Unit)
             }
+            val lastSync = store.getLastSyncTimestamp()
+            if (lastSync != null && System.currentTimeMillis() - lastSync < GeofenceConstants.STALE_THRESHOLD_MS) {
```
mahmoud-elmorabea, line 69: Can we use an abstraction over `System.currentTimeMillis`?

location/src/main/kotlin/io/customer/location/geofence/GeofenceRepository.kt
```diff
@@ -12,18 +12,20 @@ import kotlinx.coroutines.sync.Mutex
...
- * current user/location, persist them, then register the closest N with the OS.
+ * Geofence sync pipeline. Two public entry points:
+ *
+ * - [refresh] — for identify / app-launch. Honors the 24h freshness threshold; a recent
+ *   successful sync short-circuits the API call.
+ * - [handleMovement] — for movement-trigger EXIT. Two-tier dispatch: re-rank cached regions
+ *   when within `remoteFetchRefreshTriggerRadius` of the last API anchor, otherwise hit
+ *   the API for fresh data.
  */
 internal interface GeofenceRepository {
```
mahmoud-elmorabea, line 23: This class is getting a bit out of hand, can we please extract two types, one that does server refresh and another that does local refresh. This repo can still do the book-keeping of is there refresh in progress and such, but it delegates the actual refresh to another smaller type

mahmoud-elmorabea (approved):

location/src/main/kotlin/io/customer/location/geofence/GeofenceRepository.kt
mrehan27: It is little hard to breakdown the complexity yet, but will see if something we can extract without breaking things and will update.

PR #706, opened 2026-05-18
chore: drop & remove orphan transitions, defer reset on quick re-login

mrehan27:
## Ticket

https://linear.app/customerio/issue/MBL-1629/smart-geofence-updates

## Stack

- #702 
- #703 
- #704 
- #705 
- #706 👈🏻 
- #707 
- #708 

## Summary

Two defensive fixes for the geofence sync lifecycle.

### 1. Sign-out + immediate-sign-in race

`ResetEvent` (sign-out) and `UserChangedEvent` (sign-in) both `scope.launch { ... }` on `geofenceScope`, which runs on `Dispatchers.Default`. The dispatcher doesn't guarantee execution order across coroutine launches — so a refresh for a new user can complete (writing state) **before** the reset for the previous user runs. The late reset would then wipe the new user's freshly-committed geofences.

The locks didn't cover this:
- `stateMutex` only makes writes atomic — refresh's write and reset's wipe can't interleave mid-operation, but nothing prevents a later reset from running after refresh has already released the mutex.
- `refreshInProgress` is only acquired by refresh / handleMovement; reset never held it.

Fix: `reset()` now reads `secureUserStore.getUserId()` at execution. If a user is signed in, it skips the wipe and logs `"reset superseded by signed-in user"`. The new user's refresh handles previous-user cleanup via the existing stale-cleanup diff in `registerNearestAndPersist` (any ID not in the new set gets removed).

### 2. Orphan transition events

`GeofenceBroadcastReceiver.dispatchTransition` now partitions incoming transition IDs against `store.getRegisteredIds()`. Unknown IDs are:
- **Dropped** — no scheduler / EventBus / cooldown work.
- **Batch-removed from the OS** — single GMS call regardless of how many orphans in the batch — so they stop firing on every future transition.

Catches orphans from a failed `clearAll`, OS-side state surviving an app-data wipe, or an SDK upgrade that changed ID format.

### 3. `clearAll` failure retry — no new code needed

The third item from the original plan is **already handled** by the existing patterns:
- PR #702 preserves the store when `manager.clearAll()` fails, so the next refresh's stale-cleanup diff retries removal of the previous user's IDs.
- The new orphan-event filter (above) backs this up: any orphan that never gets explicitly removed via a refresh gets silently dropped + removed at event-fire time.

So there's no new code for this — just calling it out so reviewers know it's covered.

## Change breakdown

- `GeofenceRepository.reset()`: userId check at top; skip wipe + log when a user is signed in. Pre-existing `stateMutex`-guarded clearAll/store.clearAll order unchanged.
- `GeofenceBroadcastReceiver.dispatchTransition()`: partition incoming IDs; log + batch-remove unknowns via `manager.removeGeofencesByIds(unknownIds)`; remaining known IDs go through the existing per-ID processing.
- `GeofenceLogger`: new `logTransitionDroppedUnknownId(geofenceId)`.

Production code: ~30 LOC. Tests: ~95 LOC.

## Test plan

- [x] `:location:test` passes locally (debug + release variants).
- [x] `:location:apiCheck` passes (no public-API leakage).
- Unit tests added:
  - `GeofenceRepositoryTest.reset_givenNewUserSignedInDuringWait_expectSkipWipe` — verifies the race fix.
  - `GeofenceBroadcastReceiverTest.dispatchTransition_givenIdNotInStore_expectDroppedAndRemovedFromOs` — single orphan dropped from all downstream paths and removed.
  - `GeofenceBroadcastReceiverTest.dispatchTransition_givenMovementTriggerNotInStore_expectMovementHandlerNotCalledAndRemoved` — movement trigger as orphan (separate downstream path).
  - `GeofenceBroadcastReceiverTest.dispatchTransition_givenMixedKnownAndUnknownIds_expectOnlyKnownProcessedAndUnknownsRemovedAsBatch` — per-ID filter + batched removal (one GMS call, not N).
  - `GeofenceBroadcastReceiverTest.dispatchTransition_givenAllIdsKnown_expectNoRemoveCall` — normal path pins "no orphans → no removal call".
- Existing reset tests (`reset_givenManagerSucceeds_*`, `reset_givenManagerFails_*`) continue to cover the no-signed-in-user paths via the default `secureUserStore.getUserId() returns null`.
- E2E manual verification is deferred — the API endpoint isn't wired yet (`ENDPOINT_PATH = ""`).

## Deferred follow-ups (MBL-1629 stack)

- **Boot `INITIAL_TRIGGER_EXIT`** (`mbl-1629-boot-initial-trigger-exit`, committed locally) — registers the movement trigger with `INITIAL_TRIGGER_EXIT` during boot restore so the OS fires EXIT immediately if the user moved beyond the cached anchor while the device was off. Held back pending design re-think on whether to also cache the last movement-trigger center.
- **No-geofences memo** — TTL cache for "user has 0 geofences" to skip redundant API calls.
- **Doc / log follow-ups** — region store KDoc (#R3), unknown-transition-type warning log (#R5).

---

> [!NOTE]
> **Medium Risk**
> Touches geofence lifecycle/state cleanup and OS registration removal logic; mistakes could drop valid transitions or leave stale geofences registered, though changes are defensive and covered by new unit tests.
> 
> **Overview**
> **Hardens geofence sync against two edge cases.** `GeofenceRepository.reset()` now checks `secureUserStore.getUserId()` at execution time and *skips clearing OS/store state* if a user is already signed in, preventing a sign-out reset from wiping a freshly-registered new-user state.
> 
> `GeofenceBroadcastReceiver.dispatchTransition()` now filters incoming transition IDs against `geofenceRegionStore.getRegisteredIds()`: *unknown/orphan IDs are logged, dropped (no cooldown/scheduler/EventBus work), and batch-removed from the OS via `geofenceManager.removeGeofencesByIds`* so they stop firing. Adds a new logger entry and unit tests covering orphan dropping/removal and the reset race.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 04a9abbde58b53cb99e9af68dbd3e714ccb3a136. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


location/src/main/kotlin/io/customer/location/geofence/GeofenceRepository.kt
```diff
@@ -244,10 +244,16 @@ internal class GeofenceRepositoryImpl(
...
+        // Skip if a new user is signed in by the time this reset runs.
+        // geofenceScope runs on Dispatchers.Default, which doesn't order coroutines —
+        // a refresh queued after this reset may have already written the new user's
+        // state. Wiping now would clobber it; the new user's refresh handles previous-
+        // user cleanup via the stale-diff in registerNearestAndPersist.
+        val currentUserId = secureUserStore.getUserId()
+        if (!currentUserId.isNullOrBlank()) {
+            logger.logSyncSkipped("reset superseded by signed-in user")
+            return@withLock Result.success(Unit)
+        }
```
mahmoud-elmorabea, line 256: Are we sure that this always runs after the userId is cleared? Can we add a test for wherever this happens?

mahmoud-elmorabea (approved):

location/src/main/kotlin/io/customer/location/geofence/GeofenceRepository.kt
mrehan27: It should be covered by `reset_givenNewUserSignedInDuringWait_expectSkipWipe`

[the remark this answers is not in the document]
mrehan27: Will take care in follow up PR

PR #707, opened 2026-05-19
chore: respect server-configured freshness expiry; inject Clock; log unknown transition types

mrehan27:
## Ticket

https://linear.app/customerio/issue/MBL-1629/smart-geofence-updates

## Stack

- #702 
- #703 
- #704 
- #705 
- #706 
- #707 👈🏻 
- #708 

## Summary

Four related cleanups bundled into one PR.

### 1. Respect server-configured freshness expiry

`GeofenceConfig.remoteFetchRefreshExpiryMs` has been parsed from the API response for a while but `refresh()` never read it — the staleness check was hardcoded to `GeofenceConstants.STALE_THRESHOLD_MS` (24h). `refresh()` now reads the cached config's value with a `takeIf { it > 0 }` defense, falling back to the constant when no cache exists or the value is non-positive.

Why this matters: backend can now tune cadence **per-account** without an SDK release — e.g., longer expiry for accounts with no geofences to reduce wasted API calls, shorter expiry for tenants needing fresher data. Also drops the now-stale "24h" reference from the `GeofenceRegionStore` KDoc.

### 2. Inject `Clock` instead of `System.currentTimeMillis()`

Pre-existing testability gap in `GeofenceRepositoryImpl` — both timestamp call sites used the system clock directly while `GeofenceCooldownFilter` already used the project's `Clock` abstraction. Now consistent: `GeofenceRepositoryImpl` takes `Clock` via constructor (wired from `SDKComponent.clock`), and the two call sites (`refresh()` freshness check + `setLastSyncTimestamp` in `performRemoteRefresh`) go through it. Tests can override with `every { clock.currentTimeMillis() } returns X` for deterministic timing.

### 3. Log unknown API transition types

`parseTransitionType` returned null for unknown strings and `mapNotNull` silently dropped them, so a backend rolling out a new transition type (or any SDK-vs-backend version drift) would invisibly break the SDK's view of affected regions. `GeofenceApiRegion.toDomain()` now logs each unknown string via `SDKComponent.geofenceLogger.logUnknownApiTransitionType(value)` — distinct error log per unknown string, naming the actual value so it's debuggable.

### 4. Lowercase `@SerialName` on `GeofenceTransitionType`

The enum carried `@SerialName("ENTER")` / `@SerialName("EXIT")` (uppercase) while the API wire format is lowercase (`"enter"` / `"exit"`). Cache and API spoke different dialects for the same value — stylistically inconsistent. Both are now lowercase.

**Upgrade impact:** existing caches with uppercase values fail to decode after the SDK upgrade. The schema-drift-safe deserialization in `GeofenceRegionStoreImpl.readJson` handles this — wipes the key and degrades to "no cache", which the next refresh repopulates. Net effect: one extra API call per already-installed user on first launch after upgrade. Acceptable.

## Change breakdown

- `GeofenceRepository.kt`: Clock constructor param; `refresh()` reads `cachedConfig.remoteFetchRefreshExpiryMs?.takeIf { it > 0 } ?: STALE_THRESHOLD_MS`; both `System.currentTimeMillis()` call sites → `clock.currentTimeMillis()`. KDoc trimmed.
- `DIGraphGeofence.kt`: passes `SDKComponent.clock` to repository.
- `GeofenceApiResponse.kt`: `toDomain()` uses `?: run { logger.logUnknownApiTransitionType(value); null }` to log + skip unknowns.
- `GeofenceLogger.kt`: new `logUnknownApiTransitionType(value: String)`.
- `GeofenceRegion.kt`: `@SerialName("ENTER")` → `"enter"`, `@SerialName("EXIT")` → `"exit"`.
- `GeofenceRegionStore.kt`: KDoc drops "24h" reference.

Production code: ~30 LOC net. Tests: ~95 LOC.

## Test plan

- [x] `:location:test` passes locally (debug + release).
- [x] `:location:apiCheck` passes (no public-API leakage).
- Unit tests added/updated:
  - `GeofenceRepositoryTest`:
    - `refresh_givenCachedConfigWithLongerExpiry_expectSkipWhenWithinIt` — server expiry overrides constant.
    - `refresh_givenCachedConfigWithShorterExpiry_expectApiCallSooner` — symmetric: shorter window trips earlier than constant.
    - `refresh_givenCachedConfigWithNonPositiveExpiry_expectFallbackToConstant` — `takeIf { it > 0 }` defense.
    - `Clock` mock injected; defaults to mirroring real time so existing tests using relative timestamps work without churn.
  - `GeofenceApiResponseTest`:
    - `parseAndMap_givenMixedValidAndInvalidTransitionTypes_*` and `parseAndMap_givenAllInvalidTransitionTypes_*` updated to verify the unknown-type log.
    - `parseAndMap_givenOnlyValidTransitionTypes_expectNoUnknownLogged` — happy path pinned, no spurious logs.
    - DI override for `GeofenceLogger`.
  - `GeofenceRegionStoreTest`:
    - Pinned JSON-key test updated from `"ENTER"` → `"enter"`.
- E2E manual verification is deferred — the API endpoint isn't wired yet (`ENDPOINT_PATH = ""`).

## Deferred follow-ups (still queued)

- **Region store storage rationale KDoc** (#R3) — author wants to revisit independently.
- **Boot `INITIAL_TRIGGER_EXIT`** — `mbl-1629-boot-initial-trigger-exit` committed locally, pending design rethink on whether to also cache the last movement-trigger center.

---

> [!NOTE]
> **Medium Risk**
> Changes geofence refresh throttling and persisted enum serialization, which can affect sync cadence and forces cache invalidation on upgrade; behavior is covered by added unit tests but impacts runtime syncing.
> 
> **Overview**
> Geofence `refresh()` now uses the cached server config’s `remoteFetchRefreshExpiryMs` (with defensive fallback) instead of a hardcoded 24h staleness window, and `GeofenceRepositoryImpl` is updated to depend on an injected `Clock` for all timestamp reads/writes.
> 
> API-to-domain mapping now **logs and drops** unknown `transition_types` values (new `GeofenceLogger.logUnknownApiTransitionType`), and `GeofenceTransitionType` serial names are changed to lowercase (`enter`/`exit`) to match the API wire format (which invalidates existing cached region entries and relies on schema-drift-safe store reads to self-heal). Tests are expanded to cover the new expiry behavior, clock injection, and unknown-transition logging, and updated for the lowercase enum persistence.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 122d97c44ea410fdac18a6f1dde56cfd2350487a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mahmoud-elmorabea (approved):
💯

PR #710, opened 2026-05-19
chore: wire geofence nearby endpoint and encrypt location snapshots

mrehan27:
## Ticket

[MBL-1629](https://linear.app/customerio/issue/MBL-1629/smart-geofence-updates)

## Stack

Final piece of the MBL-1629 SDK work. Sits alongside the open #709 (review feedback) — they touch overlapping files (`GeofenceRepository.kt`, `GeofenceRegionStore.kt`, repo tests) but are independent in intent. Whichever lands first, the other will need a small rebase.

⚠️ **Blocked on backend** — the matching `GET /v1/geofences/nearby` endpoint isn't merged or deployed yet. E2E verification runs once backend lands. Unit tests + apiCheck pass locally.

## Summary

Aligns the Android SDK with the backend `GET /v1/geofences/nearby` contract. The endpoint is workspace-scoped (no `userId` on the wire — auth is the SDK writeKey), returns a flat `geofences` array, and currently does **not** ship a `config` block or per-region `transition_types` / `last_updated`. Those last three are modeled as forward-compat slots so the SDK adopts them automatically when backend ships them — no release required.

Also adds `PreferenceCrypto` (AES-256-GCM via Android Keystore) for the two user-location snapshots only. Workspace config + region cache stay in plaintext under UID isolation.

## Wire contract mapping

```
GET /v1/geofences/nearby?latitude={lat}&longitude={lng}
Authorization: Basic <base64("writeKey:")>

200 OK:
{
  "geofences": [
    {
      "id": 123,                  // int (Customer.io internal)
      "name": "NYC Store",
      "latitude": 40.7128,
      "longitude": -74.0060,
      "radius": 500,              // int meters, [100..10000]
      "external_id": "store-001"  // string, customer-supplied, nullable
    }
  ]
}
```

Backend defaults: `radius_km=10`, `limit=100`. SDK omits both.

## Forward-compat slots

| Field | Today | Future |
|---|---|---|
| `config` block | absent → null on the DTO → repo uses cached config or `GeofenceConfig.fallback()` | once shipped → SDK parses + persists + uses next refresh |
| `transition_types` per region | absent → SDK defaults to `[ENTER, EXIT]` | once shipped → SDK honors values, falls back on all-unknown, logs each unknown |
| `last_updated` per region | absent → `0L` sentinel | once shipped → preserved (currently unused but cached) |
| `external_id` | nullable, customer-supplied | preserved end-to-end so host apps can correlate |

`id` (numeric) is the OS request ID and matches the `geofence_id` key on the event-processing path (planned in the next backend PR).

### Core
- `HttpRequestParams` gains `method: HttpMethod` (enum `{ GET, POST }`, default POST) and `queryParams: Map<String, String>`.
- `CustomerIOHttpClientImpl` builds URLs via `android.net.Uri.Builder` (`appendQueryParameter` percent-encodes both key and value automatically). Method is sourced from `params.method.name`.

### Location module
- `GeofenceApiResponse.kt`: rewritten DTO matching the wire shape; `@SerialName` on every field for R8 safety; `config` / `transitionTypes` / `lastUpdated` / `externalId` are nullable with defaults; `toDomainConfig()` returns `GeofenceConfig?` (gates cache save); transition-type resolution is three-tier (null/empty/all-unknown → defaults; mixed → valid subset; unknowns logged).
- `GeofenceApiService.kt`: switched to `GET`, no `userId` parameter, query params via the new `Map`.
- `GeofenceConfig.kt`: new `fallback()` companion factory wrapping `GeofenceConstants.FALLBACK_*`.
- `GeofenceRepository.kt`: 3-tier config preference `parsedConfig ?: cached ?: fallback`; `saveCachedConfig` only fires when parsedConfig non-null (null parse must not clobber prior cached value); `handleMovement` + `restoreFromCache` consume `cachedConfig ?: fallback`.
- `GeofenceRegion.kt`: added `externalId: String?` (preserves null vs empty distinction); KDoc explains `id` vs `externalId` split.
- `GeofenceRegionStore.kt`: `PreferenceCrypto` instance with its own key alias `cio_geofence_location_key`; `writeEncryptedJson` / `readEncryptedJson` helpers wrap the two location-snapshot writes/reads only; rest stays plaintext.

### Tests
- `GeofenceApiResponseTest.kt` rewritten for the new contract: full sample, minimal/defaults, external_id null/empty/value, transition_types five-rule fallback, last_updated, config field-level fallbacks, max-business 0/100/500 edge cases.
- `GeofenceRepositoryTest.kt`: updated for the new API signature; new test for parsed-config-wins-over-cached path; new test for store-cross-instance encryption survival in `GeofenceRegionStoreTest.kt`.

Production: ~250 net LOC (most of it DTO + helpers). Tests: ~250 net LOC.

## Encryption rationale

Used the existing `PreferenceCrypto` (same one backing `SecureUserStore`), **not** `androidx.security:security-crypto` which is deprecated. Migration via `decrypt`'s catch-all → returns input on Keystore failure → JSON parser decides → next write rewrites encrypted. Threat model: defends against on-device forensics (file copied off rooted device). Doesn't defend against runtime memory access — but that's a different attack surface.

Only the two location snapshots (`lastApiFetchLocation`, `lastMovementTriggerLocation`) are encrypted. Region cache + config + timestamps stay plaintext — workspace config, not user PII.

## Test plan

- [x] `:location:test` passes (debug + release).
- [x] `:location:apiCheck` clean.
- [x] `:core:apiCheck` clean.
- [ ] E2E — blocked on the backend endpoint merging + deploying.

## Open caveats

- `GeofenceApiService.fetchGeofences` will return `Result.failure` until backend deploys. The SDK degrades to "no geofences" gracefully (no crashes, no events).
- Wire-up assumes the backend's response shape stays as drafted. If reviewers there ask for changes (e.g., camelCase, different field names, different `id` type), this PR will need a small DTO patch.

---

> [!NOTE]
> **Medium Risk**
> Moderate risk: changes request-building to support GET+query params and updates the geofence sync pipeline/DTOs to match a new backend contract, plus introduces encrypted persistence for location snapshots which could affect cache restore/migration behavior.
> 
> **Overview**
> **Wires the geofence sync to the new `GET /geofences/nearby` contract** by adding HTTP method + query-param support to `CustomerIOHttpClient` and switching `GeofenceApiService` from a POST body (with `userId`) to a GET with `latitude`/`longitude` query params.
> 
> **Reworks geofence API decoding and config handling for forward compatibility**: `GeofenceApiResponse` now models nullable `config` and optional per-region fields (`external_id`, `transition_types`, `last_updated`), applies defaults/logging for unknown transition types, and makes `toDomainConfig()` nullable so the repo only overwrites cached config when the server actually sends one.
> 
> **Hardens persistence and restore paths** by adding `GeofenceConfig.fallback()` and using it throughout refresh/movement/boot-restore when cached config is missing, and by encrypting the two stored user location snapshots in `GeofenceRegionStore` via `PreferenceCrypto` (workspace-level region/config cache remains plaintext). Tests are updated/expanded to cover the new wire shape, fallback behavior, and cross-instance persistence/encryption.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 271a305ba8e7c39ce2b0e55ba12da518a078b762. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/location/src/main/kotlin/io/customer/location/geofence/api/GeofenceApiResponse.kt
+++ b/location/src/main/kotlin/io/customer/location/geofence/api/GeofenceApiResponse.kt
@@ -6,19 +6,24 @@ import io.customer.location.geofence.GeofenceRegion
 import io.customer.location.geofence.GeofenceTransitionType
 import io.customer.location.geofence.di.geofenceLogger
 import io.customer.sdk.core.di.SDKComponent
-import java.util.concurrent.TimeUnit
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 
+/**
+ * Wire shape of `GET /v1/geofences/nearby`. Backend ships only `geofences`
+ * today; `config` and per-region `transition_types` / `last_updated` are
+ * nullable forward-compat slots the SDK will honor when backend adds them.
+ */
 @Serializable
 internal data class GeofenceApiResponse(
-    val config: GeofenceApiConfig,
+    @SerialName("config")
+    val config: GeofenceApiConfig? = null,
+    @SerialName("geofences")
     val geofences: List<GeofenceApiRegion>
 )
 
-// All config fields are nullable with defaults so the SDK keeps working if the
-// backend rolls out fields gradually or temporarily returns invalid values.
-// Per-field fallbacks are applied in [toDomain].
+// Every field nullable so backend can roll fields out gradually; per-field
+// fallbacks live in [toDomain].
 @Serializable
 internal data class GeofenceApiConfig(
     @SerialName("local_refresh_trigger_radius")
@@ -29,6 +34,7 @@ internal data class GeofenceApiConfig(
     val remoteFetchRefreshExpiryTime: Long? = null,
     @SerialName("duplicate_events_expiry_time")
     val duplicateEventsExpiryTime: Long? = null,
+    @SerialName("android")
     val android: GeofenceApiPlatformConfig? = null
 )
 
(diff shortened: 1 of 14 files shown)


location/src/main/kotlin/io/customer/location/geofence/api/GeofenceApiResponse.kt
```diff
@@ -40,57 +46,72 @@ internal data class GeofenceApiPlatformConfig(
 
 @Serializable
 internal data class GeofenceApiRegion(
-    val id: String,
-    val name: String,
+    // Used as the OS request ID and the `geofence_id` key on transition events.
+    @SerialName("id")
+    val id: Int,
```
mahmoud-elmorabea: I understand that maybe on DB level this might be an Int, but do we need to treat it as such? Safer to send it from server as string so in case if changes in the future, we shouldn't care.

mahmoud-elmorabea (approved):

location/src/main/kotlin/io/customer/location/geofence/api/GeofenceApiResponse.kt
mrehan27: Updated with lenient parser

PR #711, opened 2026-05-20
chore: geofence followups — permissions, first-run, cooldown config

mrehan27:
part of: [MBL-1735](https://linear.app/customerio/issue/MBL-1735/android-fix-geofencing-reliability-and-lifecycle-issues-identified)

## Summary

Six independent follow-up items surfaced while testing MBL-1629 end-to-end. Bundled here for review economy; each could split into a separate PR if preferred.

- Relax `GeofencePermissionChecker` so FINE-only registration works (foreground-only delivery) instead of silently no-opping when BACKGROUND is missing
- Recover from a first-run race where `identify()` lands before the first GPS fix: retry the geofence sync once a fix arrives
- Make `GeofenceCooldownFilter` honor the server-configured `duplicateEventsExpiry` — it was hardcoded to a constant despite the field being parsed and persisted
- Add the location permissions geofence registration requires to the sample app manifests
- Avoid spurious EXIT events by skipping re-registration of business geofences whose IDs are already in GMS — `GeofencingClient.addGeofences()` treats same-ID re-upserts as fresh state evaluations and can fire reconciliation EXITs
- Repair post-reboot GMS state when boot-restore races the app-launch refresh — boot-restore now bypasses the in-flight gate so business geofences aren't left out of GMS after a reboot

## Stack

```
main
 └─ feature/geofence-on-device   (open, target of this PR)
     └─ geofence-followups       ← THIS PR
```

Note: `feature/geofence-on-device` has not yet merged to main. All current geofence work targets that feature branch.

### 1. Permission gate over-demanded BACKGROUND for registration

**Before:** `hasRequiredLocationPermissions()` required both `ACCESS_FINE_LOCATION` AND (on API 29+) `ACCESS_BACKGROUND_LOCATION`. If a host granted only FINE, registration silently no-op'd — zero geofences registered, no clear signal to the host.

**GMS contract:** `GeofencingClient.addGeofences()` only requires FINE. `ACCESS_BACKGROUND_LOCATION` only controls whether transitions are delivered while the app is in the background. Foreground-only registration is a perfectly viable degraded mode.

**After:**
- `GeofencePermissionChecker.hasRequiredLocationPermissions()` returns `true` on FINE alone
- New `isBackgroundDeliveryAvailable()` separately reports whether transitions will survive backgrounding
- `GeofenceServicesImpl.triggerSync` proceeds with registration on FINE alone, but emits a new info log via `GeofenceLogger.logBackgroundDeliveryUnavailable()` when BACKGROUND is missing so hosts know transitions won't fire while backgrounded
- `GeofenceManager.replaceGeofencesInternal` log message updated to reflect that FINE alone gates registration

**Sample UX:** `samples/java_layout/LocationTestActivity` adds a "Grant background location" button. Walks the user through the FINE-first chain, hides on pre-Q, and routes to Settings on API 30+ where the runtime dialog no longer fires.

### 2. First-run no-location reliability hole

**Before:** On fresh install, `identify()` typically fires within milliseconds of app start. The GPS hardware takes 2–10 seconds to produce the first fix. If identify lands before the first fix:
- `LocationTracker.lastLocation` is still null
- `UserChangedEvent` subscriber calls `onUserIdentified(null, null)` → `triggerSync` skips with `logSyncSkippedNoLocation`
- The GPS fix eventually arrives but `LocationTracker.onLocationReceived` doesn't trigger any retry

The skip self-heals only on (a) next identify, (b) cold launch with persisted location, or (c) movement-trigger EXIT (chicken-and-egg — requires registration that requires this chain to have worked). For apps whose primary value is "welcome on first arrival", the first store visit on install day likely never fires.

**After:** Three pieces wired together:

- `LocationTracker.onLocationReceivedListener: ((Double, Double) -> Unit)?` — new internal callback fired after every successful `onLocationReceived` (NOT on `restorePersistedLocation()`, which is state recovery, not a new event)
- `GeofenceServicesImpl.lastSkippedForNoLocation: AtomicBoolean` — set on no-location skip, cleared on any successful trigger
- `onLocationAcquired(lat, lng)` — new interface method, CAS-protected so it only acts on the rising edge of skip (no refresh storm for hosts that stream locations)
- `ModuleLocation.initialize()` wires the listener at module init

The defensive `onAppLaunch` block in `ModuleLocation.initialize()` is preserved — it covers a different case (cached location from previous session + no new fix this session). The new listener handles only "new fix arrived after a skip"; the defensive read handles "we already have a fix on disk". Both are needed.

### 3. Cooldown filter ignored server-pushed config

**Before:** `GeofenceConfig.duplicateEventsExpiry` is parsed in `GeofenceApiResponse.toDomain()`, persisted via `GeofenceRegionStore.saveCachedConfig`, and readable via `getCachedConfig()`. But `GeofenceCooldownFilter.tryAcquire` used `GeofenceConstants.DEDUPE_COOLDOWN_MS` directly (60 min hardcoded fallback). The server-configured value was silently ignored.

**After:** `GeofenceCooldownFilter` consults `regionStore.getCachedConfig()?.duplicateEventsExpiry ?: GeofenceConstants.DEDUPE_COOLDOWN_MS`. Server can now tune the cooldown window without an SDK release.

**Audit:** Every `GeofenceConstants.X` reference in `location/src/main` was checked against `GeofenceConfig` field equivalents. Only `GeofenceCooldownFilter` was using a constant directly where a config field exists. All other constants are either sentinels (`MOVEMENT_TRIGGER_ID`), Android API flags (`NO_INITIAL_TRIGGER`, `GEOFENCE_EXPIRATION_NEVER`, `PENDING_INTENT_REQUEST_CODE`), or already use the `?:` fallback pattern.

### 4. Sample manifest gaps

- `samples/java_layout/src/main/AndroidManifest.xml`: added `ACCESS_BACKGROUND_LOCATION` (FINE was already there)
- `samples/kotlin_compose/src/main/AndroidManifest.xml`: added both `ACCESS_FINE_LOCATION` and `ACCESS_BACKGROUND_LOCATION` (had neither)

### 5. Spurious EXIT events from GMS state reconciliation

**Before:** Every Tier A / Tier B refresh sent the full nearest-set to `GeofencingClient.addGeofences` — including business IDs already registered with the same parameters. Observed on emulator during repeated Tier A refreshes: the OS fired EXIT for all 4 cached regions on every refresh despite no actual transition. The cooldown filter dedupes within the window, but every refresh outside the window flushes the same spurious events again.

**Root cause:** `GeofencingClient.addGeofences()` treats a same-ID re-upsert as a fresh state evaluation. GMS recomputes the region's inside/outside state from the current location; if the cached internal state has drifted from reality (e.g. a "should be inside" region that the OS now sees as "outside"), an EXIT fires immediately.

**After:**
- `GeofenceManager.replaceGeofences` accepts `existingBusinessIds: Set<String> = emptySet()` — caller tells the manager what's already in GMS
- Manager `partition`s incoming business regions into `businessToAdd` (new − existing) and `businessKept` (overlap); only `businessToAdd` is sent to GMS. Overlapping IDs are left untouched, so no reconciliation event fires for them.
- Movement trigger is always re-registered because its center moves with the user.
- `GeofenceRepository.registerWithBusinessDiff` reads the persisted registered IDs (minus movement trigger) and forwards them on the default Tier A / Tier B refresh path.
- Boot restore bypasses the diff via `replaceGeofencesForBootRestore` — after a reboot OS state is empty regardless of what the store says, so any "existing" set would leave us with no registrations in GMS.
- The Repository's existing stale-cleanup pass (existing − new for removes) is the symmetric inverse and is unchanged; KDoc now notes the symmetry.

**Logging:** New `GeofenceLogger.logBusinessGeofencesKept(count)` fires when overlap is skipped, so the kept-vs-added split is visible in logs.

**Param-drift handling (commit d2297e8):** Cursor Bugbot flagged that the initial diff would skip re-registering business geofences whose backend-side parameters had changed — the cache would update via `saveCachedRegions(regions)` but GMS would retain the stale values until boot, sign-out, or the geofence dropping out of the nearest set. Fixed by tightening `registerWithBusinessDiff` to treat an ID as skip-safe only when `cachedRegion == incomingRegion` (data-class equality on `GeofenceRegion`). Any param drift — radius, coords, transition types, name, or `lastUpdated` — falls out of `existingBusinessIds` and forces re-registration. Tier A (cache as input) is unaffected by construction; only Tier B (API refresh) can introduce drift.

**Tests (commit 26a84b3):** `GeofenceRepositoryTest` adds two Tier-B remote-refresh cases that pin the equality-based skip filter end-to-end — one where cached equals incoming (overlap ID forwarded as existing), one where the radius drifted from 100m to 200m (overlap ID excluded, forcing re-register). Removing the `cachedById[it.id] == it` check would fail the second assertion immediately. Existing 240 cases still pass against the updated 2-arg `replaceGeofences` matchers.

### 6. Post-reboot GMS state lost when boot-restore races app-launch

**Before:** After a device reboot GMS wipes all geofence registrations, but our persisted `registeredIds` still reflects pre-boot state. If the app launches around the same time the OS dispatches `ACTION_BOOT_COMPLETED` (common on emulator restarts, sometimes seen on real devices), the app-launch `refresh()` grabs the `refreshInProgress` gate first. The boot receiver's `restoreFromCache()` then fails its own `compareAndSet` and bails with `"refresh already in progress"`. Meanwhile `registerWithBusinessDiff` (the path refresh takes) treats every business region as "unchanged" via the equality diff (cached == incoming) and skips them — leaving GMS with only the movement trigger and zero business geofences until a sign-out, reinstall, or `remote_fetch_refresh_expiry` tick. End-to-end testing surfaced this as "no geofence ENTER/EXIT events fire after reboot even though movement-trigger still does."

**After:** `GeofenceRepositoryImpl.restoreFromCache` no longer takes the `refreshInProgress` gate. It still uses `replaceGeofencesForBootRestore` (which bypasses the equality diff) and goes through `registerNearestAndPersist → stateMutex.withLock { ... }` so writes against `GeofencingClient` and the persisted `registeredIds` remain serialized. The two orderings now converge on the correct GMS state:

- *Refresh wins the mutex first:* registers movement, skips business via diff. Boot-restore then acquires the mutex and `replaceGeofencesForBootRestore` adds the 4 business + re-upserts movement. End state: movement + 4 business.
- *Boot-restore wins the mutex first:* registers movement + 4 business. Refresh then acquires the mutex, sees cached == incoming, skips business, re-upserts movement only. End state: movement + 4 business.

**Tests (commit c94b910):** New `restoreFromCache_givenRefreshInFlight_expectStillRunsViaBootRestoreVariant` launches refresh + restoreFromCache concurrently — refresh holds the path 100ms via a mocked `replaceGeofences` delay, restoreFromCache enters at +20ms — and asserts `manager.replaceGeofencesForBootRestore(any())` was actually called. Under the old gated behavior that call would never have happened.

**Verified end-to-end on emulator:** post-reboot log showed boot-restore firing, both refresh paths writing through `stateMutex` in sequence, and a full Eiffel→Shahkam→Saleem→Thokar journey producing 4 ENTER + 4 EXIT events with no spurious EXIT bursts.

## Test plan

Automated:
- [x] `:location:testDebugUnitTest` passes
- [x] `:location:apiCheck` passes (no public API changes)
- [x] `:samples:java_layout:assembleDebug` and `:samples:kotlin_compose:assembleDebug` build

New test coverage:
- `GeofencePermissionCheckerTest` (new) — FINE/BACKGROUND combinations across SDK 28 and 29
- `GeofenceServicesTest` — `onLocationAcquired` happy path, no-prior-skip no-op, no-userId no-op, no-retrigger-after-successful-sync, and BACKGROUND-missing-warn behavior
- `LocationTrackerTest` — listener invoked with coordinates, null listener no-exception
- `GeofenceCooldownFilterTest` — server-configured cooldown is actually applied (new test), all existing constant-fallback tests preserved
- `GeofenceManagerTest` — FINE-only registration now succeeds (was: expected failure)

Manual (java sample):
- [ ] Grant only FINE (not BACKGROUND) → geofences register, `[Geofence] Geofence sync (...): ACCESS_BACKGROUND_LOCATION not granted` info log appears
- [ ] Tap "Grant background location" → API 29: runtime dialog; API 30+: Settings deep link
- [ ] Fresh install, identify before any GPS fix → confirm refresh fires when the first fix lands (visible in logs via `Geofence sync triggered: user-identified` shortly after `Location update received` for the first fix)
- [ ] Repeated Tier A refresh within the freshness window → confirm `Kept N business geofences unchanged in OS` appears and no spurious EXIT events fire for cached regions
- [ ] Reboot device, do NOT open the app, then cross a business geofence → confirm `sync triggered: boot-restore` fires, `Registered N geofences with OS` appears (N = movement + business count), and subsequent ENTER/EXIT events log for each business region crossed

---

> [!NOTE]
> **Medium Risk**
> Touches geofence registration and refresh logic (OS/GMS interaction and permission gating), which can change when/if transitions fire and when network refreshes occur. Changes are well-scoped and backed by added/updated unit tests, but regressions would impact geofence reliability.
> 
> **Overview**
> Improves geofence reliability and configurability by **allowing registration with only `ACCESS_FINE_LOCATION`** (foreground-only delivery) while separately warning when background delivery is unavailable.
> 
> Adds a one-time **retry path when `identify()` happens before the first GPS fix**, by wiring a `LocationTracker` location-received callback into `GeofenceServices.onLocationAcquired()`.
> 
> Reduces spurious GMS reconciliation transitions by **skipping re-upsert of unchanged business geofences** (`existingBusinessIds` diff) while still re-registering the movement trigger; boot restore bypasses the diff to handle wiped OS state. Also updates duplicate-event suppression to **honor server-configured `duplicateEventsExpiry`**.
> 
> Updates sample apps to declare background location and adds sample UI flow to request it, and expands tests around permissions, cooldown config, diff-based registration, and the new location-acquired retry hook.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c94b9103f7e30ae4db878225a99b7ef7aa684f15. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/location/src/main/kotlin/io/customer/location/geofence/GeofenceRepository.kt
+++ b/location/src/main/kotlin/io/customer/location/geofence/GeofenceRepository.kt
@@ -141,39 +141,36 @@ internal class GeofenceRepositoryImpl(
 
     @RequiresPermission(allOf = [Manifest.permission.ACCESS_FINE_LOCATION, Manifest.permission.ACCESS_BACKGROUND_LOCATION])
     override suspend fun restoreFromCache(): Result<Unit> {
-        if (!refreshInProgress.compareAndSet(false, true)) {
-            logger.logSyncSkipped("refresh already in progress")
+        // Bypasses the in-flight gate: after a reboot, app-launch refresh's
+        // registerWithBusinessDiff would see persisted registeredIds matching
+        // the incoming set and skip business as "unchanged" — but GMS was wiped
+        // by the reboot. Boot-restore must still run via
+        // replaceGeofencesForBootRestore (no diff); stateMutex serializes the
+        // concurrent writes.
+        val userId = secureUserStore.getUserId()
+        if (userId.isNullOrBlank()) {
+            logger.logSyncSkipped("no identified user")
             return Result.success(Unit)
         }
-        try {
-            val userId = secureUserStore.getUserId()
-            if (userId.isNullOrBlank()) {
-                logger.logSyncSkipped("no identified user")
-                return Result.success(Unit)
-            }
-            // Prefer the most recent movement-trigger center as the effective
-            // location — it tracks Tier A drift and is much closer to the user's
-            // real position than the anchor (only updated on Tier B fetches).
-            // Fall back to the anchor if there's no movement-trigger location yet
-            // (older cache / first-ever boot restore).
-            val effectiveLocation = store.getLastMovementTriggerLocation()
-                ?: store.getLastApiFetchLocation()
-            if (effectiveLocation == null) {
-                logger.logSyncSkipped("no cached state to restore")
-                return Result.success(Unit)
-            }
-            val cachedConfig = store.getCachedConfig() ?: GeofenceConfig.fallback()
-            // Boot-restore variant — see [GeofenceManager.replaceGeofencesForBootRestore].
-            return performLocalRefresh(
-                userId = userId,
-                latitude = effectiveLocation.latitude,
-                longitude = effectiveLocation.longitude,
-                cachedConfig = cachedConfig,
-                register = manager::replaceGeofencesForBootRestore
-            )
-        } finally {
-            refreshInProgress.set(false)
+        // Prefer the most recent movement-trigger center as the effective
+        // location — it tracks Tier A drift and is much closer to the user's
+        // real position than the anchor (only updated on Tier B fetches).
+        // Fall back to the anchor if there's no movement-trigger location yet
+        // (older cache / first-ever boot restore).
+        val effectiveLocation = store.getLastMovementTriggerLocation()
+            ?: store.getLastApiFetchLocation()
+        if (effectiveLocation == null) {
+            logger.logSyncSkipped("no cached state to restore")
+            return Result.success(Unit)
         }
+        val cachedConfig = store.getCachedConfig() ?: GeofenceConfig.fallback()
+        return performLocalRefresh(
+            userId = userId,
+            latitude = effectiveLocation.latitude,
+            longitude = effectiveLocation.longitude,
+            cachedConfig = cachedConfig,
+            register = manager::replaceGeofencesForBootRestore
+        )
     }
 
     @RequiresPermission(allOf = [Manifest.permission.ACCESS_FINE_LOCATION, Manifest.permission.ACCESS_BACKGROUND_LOCATION])
--- a/location/src/main/kotlin/io/customer/location/geofence/GeofenceRepository.kt
+++ b/location/src/main/kotlin/io/customer/location/geofence/GeofenceRepository.kt
@@ -230,15 +230,19 @@ internal class GeofenceRepositoryImpl(
     )
 
     /**
-     * Default register path for Tier A / Tier B refreshes — forwards the
-     * persisted business IDs as `existingBusinessIds` so the manager skips
-     * re-registering the overlap (see [GeofenceManager.replaceGeofences]).
+     * Default register path for Tier A / Tier B refreshes. An ID is treated
+     * as skip-safe only when the cached region equals the incoming one — any
+     * param drift forces a re-register so GMS doesn't keep stale values.
      * Boot restore bypasses this; OS state is empty after reboot.
      */
     @RequiresPermission(allOf = [Manifest.permission.ACCESS_FINE_LOCATION, Manifest.permission.ACCESS_BACKGROUND_LOCATION])
     private suspend fun registerWithBusinessDiff(regions: List<GeofenceRegion>): Result<Unit> {
-        val existingBusinessIds =
-            store.getRegisteredIds() - GeofenceConstants.MOVEMENT_TRIGGER_ID
+        val registeredBusinessIds = store.getRegisteredIds() - GeofenceConstants.MOVEMENT_TRIGGER_ID
+        val cachedById = store.getCachedRegions().associateBy { it.id }
+        val existingBusinessIds = regions
+            .filter { it.id in registeredBusinessIds && cachedById[it.id] == it }
+            .map { it.id }
+            .toSet()
         return manager.replaceGeofences(
             regions = regions,
             existingBusinessIds = existingBusinessIds
--- a/location/src/main/kotlin/io/customer/location/geofence/GeofenceRepository.kt
+++ b/location/src/main/kotlin/io/customer/location/geofence/GeofenceRepository.kt
@@ -219,7 +219,7 @@ internal class GeofenceRepositoryImpl(
         latitude: Double,
         longitude: Double,
         cachedConfig: GeofenceConfig,
-        register: suspend (List<GeofenceRegion>) -> Result<Unit> = manager::replaceGeofences
+        register: suspend (List<GeofenceRegion>) -> Result<Unit> = ::registerWithBusinessDiff
     ): Result<Unit> = registerNearestAndPersist(
         userId = userId,
         latitude = latitude,
@@ -229,6 +229,22 @@ internal class GeofenceRepositoryImpl(
         register = register
     )
 
+    /**
+     * Default register path for Tier A / Tier B refreshes — forwards the
+     * persisted business IDs as `existingBusinessIds` so the manager skips
+     * re-registering the overlap (see [GeofenceManager.replaceGeofences]).
+     * Boot restore bypasses this; OS state is empty after reboot.
+     */
+    @RequiresPermission(allOf = [Manifest.permission.ACCESS_FINE_LOCATION, Manifest.permission.ACCESS_BACKGROUND_LOCATION])
+    private suspend fun registerWithBusinessDiff(regions: List<GeofenceRegion>): Result<Unit> {
+        val existingBusinessIds =
+            store.getRegisteredIds() - GeofenceConstants.MOVEMENT_TRIGGER_ID
+        return manager.replaceGeofences(
+            regions = regions,
+            existingBusinessIds = existingBusinessIds
+        )
+    }
+
     @RequiresPermission(allOf = [Manifest.permission.ACCESS_FINE_LOCATION, Manifest.permission.ACCESS_BACKGROUND_LOCATION])
     private suspend fun registerNearestAndPersist(
         userId: String,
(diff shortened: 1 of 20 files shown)


location/src/main/kotlin/io/customer/location/geofence/GeofenceRepository.kt
```diff
@@ -229,14 +229,30 @@ internal class GeofenceRepositoryImpl(
...
+     */
+    @RequiresPermission(allOf = [Manifest.permission.ACCESS_FINE_LOCATION, Manifest.permission.ACCESS_BACKGROUND_LOCATION])
+    private suspend fun registerWithBusinessDiff(regions: List<GeofenceRegion>): Result<Unit> {
+        val existingBusinessIds =
+            store.getRegisteredIds() - GeofenceConstants.MOVEMENT_TRIGGER_ID
+        return manager.replaceGeofences(
+            regions = regions,
+            existingBusinessIds = existingBusinessIds
+        )
+    }
```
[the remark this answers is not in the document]
mrehan27: Good catch — addressed in d2297e81.

`registerWithBusinessDiff` now only treats an ID as skip-safe when `cachedRegion == incomingRegion` (data-class equality on `GeofenceRegion`). Any param drift — radius, coords, transition types, name, or `lastUpdated` — falls out of `existingBusinessIds` and forces re-registration, so cache + GMS stay in sync.

Tier A (cache as input) is unaffected by construction (incoming *is* the cache). Tier B (API refresh) is where the divergence could occur, and that's exactly the path the equality check now protects.

samples/kotlin_compose/src/main/AndroidManifest.xml
```diff
@@ -2,6 +2,8 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
 
     <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
+    <uses-permission android:name="android.permission.ACCESS_BACKGROUND_LOCATION" />
```
mahmoud-elmorabea, line 6: Let's now move forward with testing but before release we should check how the feature behaves if it has a subset of these permissions

mahmoud-elmorabea (approved):
Looks good 💯 

I think it will be helpful if we create tiny internal settings UI or maybe config screen for this to adjust values so our testing is easier. Shorter/smaller threshholds and such

samples/kotlin_compose/src/main/AndroidManifest.xml
mrehan27: Absolutely. I'll note this down and make sure we test this before final release.

PR #695, opened 2026-05-14
refactor: cache push-delivered metrics and flush via analytics pipeline

mahmoud-elmorabea:
Reworks the push-`delivered` metric path so a single delivery produces a single emit on the happy path.

- Append a pending entry to a local disk-backed store before scheduling the WorkManager job (or direct-HTTP fallback when WorkManager is unavailable).
- On WorkManager / HTTP success, remove the pending entry. On failure or any other case, keep it.
- On app launch, flush any remaining pending entries through the analytics pipeline (off the main thread) and clear them.

Removes the prior dual-emit (WorkManager direct HTTP + analytics event bus). Backend already dedupes `(deliveryID, delivered)` pairs server-side, so the motivation is bandwidth, analytics-queue load, and removing the two-path confusion.

---

> [!NOTE]
> **Medium Risk**
> Changes the push `Delivered` metric delivery flow by adding disk-backed persistence and altering when metrics are emitted, which could affect analytics correctness and duplicate/missed delivery behavior if edge cases are wrong.
> 
> **Overview**
> Refactors push `Delivered` metric tracking to **avoid dual-emitting** on the happy path by introducing a disk-backed `PendingPushDeliveryStore` that records `(deliveryId, token, timestamp)` before scheduling the primary delivery (WorkManager or direct HTTP fallback).
> 
> On successful backend receipt, the WorkManager worker and direct-HTTP fallback now **remove** the corresponding pending entry; on failure they leave it for retry. On app launch, `ModuleMessagingPushFCM.initialize()` now flushes any remaining pending entries by publishing `Event.TrackPushMetricEvent(Metric.Delivered)` on a background dispatcher and then removes only the IDs that were loaded, so mid-flush appends survive.
> 
> Updates DI wiring and tests accordingly, and adds comprehensive store tests (capacity cap, concurrency locking, corrupted-file handling, and targeted removals).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 46cfe8d387798719710d22ce0110151ed40d95b0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #717, opened 2026-06-05
chore: geofence reliability fixes + logger tag dispatch

mrehan27:
## Stack

```
main
 └── feature/geofence-on-device
      └── mbl-1760-geofence-reliability-fixes  ← this PR
```

Linear: [MBL-1760](https://linear.app/customerio/issue/MBL-1760/android-execute-geofencing-validation-and-lifecycle-testing)

## Summary

Three related fixes surfaced during MBL-1760 validation (long-distance journey testing + log inspection):

### 1. Movement-trigger geofence stuck after first EXIT

**Symptom.** A 20km drive produced exactly one movement EXIT — the second and subsequent ones never fired. After re-registration the trigger geofence appeared centered on the user, but GMS would not emit EXIT until the user physically returned to the OLD circle.

**Root cause.** `GeofencingClient` caches per-ID INSIDE/OUTSIDE state across same-ID `addGeofences` calls. A just-fired EXIT leaves the ID stuck OUTSIDE; re-registering the same ID at a new center inherits that OUTSIDE state, so the next EXIT can only fire when the user re-enters the OLD circle.

**Fix.** Remove the prior movement-trigger registration before adding the new one — forces GMS to allocate a fresh state machine. Register with `INITIAL_TRIGGER_ENTER` so GMS evaluates state at register time (user is at the new center → INSIDE). The receiver gracefully ignores the resulting ENTER ([`GeofenceBroadcastReceiver.kt`](location/src/main/kotlin/io/customer/location/geofence/GeofenceBroadcastReceiver.kt#L111)).

`NO_INITIAL_TRIGGER` is removed; it's no longer used anywhere.

Verified end-to-end on emulator: a long-distance journey now produces one EXIT per ~radius traveled instead of stalling after the first.

### 2. App-launch refresh ignored anchor distance

**Symptom.** A user who travels far while the app is killed (e.g. flies to a new city) reopens the app and sees stale geofences from the old location — until the time-based freshness window expires (default 24h).

**Root cause.** [`GeofenceRepository.refresh()`](location/src/main/kotlin/io/customer/location/geofence/GeofenceRepository.kt#L86) only checked the time-based freshness window. Movement-trigger EXITs can't fire while the app process is dead, so the cache stays stale until the time-based threshold trips.

**Fix.** Mirror [`handleMovement()`](location/src/main/kotlin/io/customer/location/geofence/GeofenceRepository.kt#L131) — skip the API call only if **both** are true: time-fresh **and** user near the last fetch anchor (`distance < remoteFetchRefreshTriggerRadius`). Null anchor (never fetched) preserves existing behavior via a `0f` distance fallback so the time-only check still applies.

### 3. `setLogDispatcher` silently dropped the subsystem tag

**Symptom.** Custom dispatchers (used for downstream telemetry / observability) received messages without the `[Subsystem]` prefix that Logcat saw — e.g. `Geofence sync succeeded: 4 regions registered` instead of `[Geofence] Geofence sync succeeded: 4 regions registered`.

**Root cause.** [`LoggerImpl.logIfMatchesCriteria`](core/src/main/kotlin/io/customer/sdk/core/util/LoggerImpl.kt#L45) passed the raw `message` to the dispatcher, but the tagged form (via `prependTagToMessage(tag, message)`) to `actualLogger`. The two paths diverged.

**Fix.** Prepend the tag once and pass the same tagged string to both paths. No public API change.

## Test plan

- [x] `LoggerTest.givenTagAndDispatcher_shouldDeliverTaggedMessageToDispatcher` — pins ERROR/INFO/DEBUG with tag through dispatcher
- [x] `GeofenceManagerTest.replaceGeofences_givenMovementTrigger_expectInitialTriggerEnter` — pins the new flag
- [x] `GeofenceManagerTest.replaceGeofences_givenMovementTrigger_expectPriorRegistrationRemovedFirst` — pins the remove-first behavior
- [x] `GeofenceRepositoryTest.refresh_givenFreshButMovedFarFromAnchor_expectApiCalled` — pins the distance-check force-fetch
- [x] All existing geofence + Logger tests still green
- [x] Verified end-to-end on emulator: long journey now produces multiple movement EXITs

## Risk

Low — production code diff is ~46 LOC across 5 files. The movement-trigger remove-first runs inside the existing `stateMutex`-protected path; no new threading surface. The distance check has a `null` anchor fallback that preserves the pre-fix behavior for users without an established anchor.

## Follow-ups (separate tickets, intentionally not in this PR)

- Freshness self-extension catch-22: when the cached config has a 24h `remoteFetchRefreshExpiry`, you can't shorten it without `pm clear` or natural expiry. Edge case; bites testing more than production.
- Duplicate emit per transition: WorkManager + EventBus both call `track()` for each transition. Plan to mirror PR #695's `PendingPushDeliveryStore` pattern once that merges.
- Real-device verification of the movement-trigger fix (emulator-only so far).

---

> [!NOTE]
> **Low Risk**
> Small, targeted changes to geofence registration/refresh and logging formatting; movement remove-first stays on the existing mutex-protected path, and null anchor preserves prior time-only skip behavior.
> 
> **Overview**
> Fixes **geofence movement-trigger reliability**, **app-launch refresh when the user has moved**, and **logger tag parity** for custom dispatchers.
> 
> **Movement trigger:** Re-registers the movement geofence with `INITIAL_TRIGGER_ENTER` and **removes the prior OS registration first** so GMS does not keep a stale OUTSIDE state that blocks later EXIT events. `GeofenceBroadcastReceiver` still treats non-EXIT movement events as no-ops; `NO_INITIAL_TRIGGER` is dropped.
> 
> **`GeofenceRepository.refresh`:** Skips the API only when sync is **time-fresh and** the current location is within `remoteFetchRefreshTriggerRadius` of the last fetch anchor (aligned with `handleMovement`), so a long trip while the app was killed can still force a remote fetch inside the time window.
> 
> **`LoggerImpl`:** Builds the tagged message once and sends the same string to `setLogDispatcher` and Logcat.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 43a846531c23120d8edd424661f608d7a517ca54. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/core/src/main/kotlin/io/customer/sdk/core/util/LoggerImpl.kt
+++ b/core/src/main/kotlin/io/customer/sdk/core/util/LoggerImpl.kt
@@ -45,13 +45,13 @@ internal open class LoggerImpl(
     private fun logIfMatchesCriteria(levelForMessage: CioLogLevel, message: String, tag: String?, throwable: Throwable?) {
         if (!shouldLog(levelForMessage)) return
 
-        // Dispatch log event to log dispatcher only if the log level is met and the dispatcher is set
-        // Otherwise, log to Logcat
-        logDispatcher?.invoke(levelForMessage, message) ?: when (levelForMessage) {
+        // Tag prepended once so dispatcher and Logcat see the same final message.
+        val taggedMessage = prependTagToMessage(tag, message)
+        logDispatcher?.invoke(levelForMessage, taggedMessage) ?: when (levelForMessage) {
             CioLogLevel.NONE -> {}
-            CioLogLevel.ERROR -> actualLogger.error(TAG, prependTagToMessage(tag, message), throwable)
-            CioLogLevel.INFO -> actualLogger.info(TAG, prependTagToMessage(tag, message))
-            CioLogLevel.DEBUG -> actualLogger.debug(TAG, prependTagToMessage(tag, message))
+            CioLogLevel.ERROR -> actualLogger.error(TAG, taggedMessage, throwable)
+            CioLogLevel.INFO -> actualLogger.info(TAG, taggedMessage)
+            CioLogLevel.DEBUG -> actualLogger.debug(TAG, taggedMessage)
         }
     }
 
--- a/core/src/test/java/io/customer/sdk/core/util/LoggerTest.kt
+++ b/core/src/test/java/io/customer/sdk/core/util/LoggerTest.kt
@@ -200,4 +200,19 @@ class LoggerTest : JUnit5Test() {
         assertCalledOnce { mockLogger.info(LoggerImpl.TAG, "[Tag?] Test info message") }
         assertCalledOnce { mockLogger.debug(LoggerImpl.TAG, "[Tag?] Test debug message") }
     }
+
+    @Test
+    fun givenTagAndDispatcher_shouldDeliverTaggedMessageToDispatcher() {
+        logger.logLevel = CioLogLevel.DEBUG
+        val logEventListenerMock = mockk<(CioLogLevel, String) -> Unit>(relaxed = true)
+        logger.setLogDispatcher(logEventListenerMock)
+
+        logger.error("error msg", "Subsystem")
+        logger.info("info msg", "Subsystem")
+        logger.debug("debug msg", "Subsystem")
+
+        assertCalledOnce { logEventListenerMock(CioLogLevel.ERROR, "[Subsystem] error msg") }
+        assertCalledOnce { logEventListenerMock(CioLogLevel.INFO, "[Subsystem] info msg") }
+        assertCalledOnce { logEventListenerMock(CioLogLevel.DEBUG, "[Subsystem] debug msg") }
+    }
 }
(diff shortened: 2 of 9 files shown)


core/src/main/kotlin/io/customer/sdk/core/util/LoggerImpl.kt
```diff
@@ -45,13 +45,13 @@ internal open class LoggerImpl(
     private fun logIfMatchesCriteria(levelForMessage: CioLogLevel, message: String, tag: String?, throwable: Throwable?) {
         if (!shouldLog(levelForMessage)) return
 
-        // Dispatch log event to log dispatcher only if the log level is met and the dispatcher is set
-        // Otherwise, log to Logcat
-        logDispatcher?.invoke(levelForMessage, message) ?: when (levelForMessage) {
+        // Tag prepended once so dispatcher and Logcat see the same final message.
+        val taggedMessage = prependTagToMessage(tag, message)
+        logDispatcher?.invoke(levelForMessage, taggedMessage) ?: when (levelForMessage) {
```
mahmoud-elmorabea, line 50: 💯

mahmoud-elmorabea (approved):

PR #720, opened 2026-06-08
chore: snapshot userId on pending geofence entries to fix cross-user attribution

mrehan27:
## Stack

```
main
 └── feature/geofence-on-device
      └── mbl-1760-pending-delivery-userid-snapshot  ← this PR
```

Linear: [MBL-1760](https://linear.app/customerio/issue/MBL-1760/android-execute-geofencing-validation-and-lifecycle-testing)

### The bug

A geofence transition queued while User A was identified could be delivered as **User B** if A signed out and B signed in before delivery completed. Both delivery channels resolved the userId at *delivery time*, not at *queue time*:

- The WorkManager worker called `secureUserStore.getUserId()` inside `GeofenceEventTrackerImpl.trackEvent` (always returns the current user).
- The foreground flush published `Event.GeofenceTransitionEvent` to the analytics pipeline, which called `track()` — also resolving to the current identity.

Also: events queued during an anonymous session were silently dropped by the worker (`Result.success()` with no HTTP) instead of being deferred to the pipeline path.

### The fix

Snapshot the userId on [`PendingGeofenceDelivery`](location/src/main/kotlin/io/customer/location/geofence/store/PendingGeofenceDelivery.kt) at queue time, then route per-entry:

| Snapshot userId | Channel | Why |
|---|---|---|
| Non-null | Worker — direct HTTP using the snapshot | The worker can carry an explicit userId; this preserves attribution across sign-out/sign-in. |
| Null (anonymous at queue time) | Flush — analytics pipeline using `anonymousId` | Direct HTTP requires a userId; the pipeline handles anonymous-to-identified merging naturally on later identify. |

To save a no-op WorkManager schedule, the broadcast receiver skips `scheduler.schedule()` for null-userId entries — they can only be delivered via the foreground flush anyway.

### Refactor along the way

[`GeofenceEventTracker.trackEvent`](location/src/main/kotlin/io/customer/location/geofence/worker/GeofenceEventTracker.kt) now takes just the entry — the HTTP body's `properties` come from `entry.toEventProperties()` and the userId is read from the snapshot. The impl guards `entry.userId.isNullOrEmpty()` as an explicit precondition; anonymous entries belong on the foreground-flush path and must not be passed in.

## Test plan

- [x] `dispatchTransition_givenIdentifiedUser_expectUserIdSnapshottedOnEntry` — pins snapshot at queue time
- [x] `dispatchTransition_givenAnonymousSession_expectEntryQueuedAndNoWorkerScheduled` — pins null snapshot + WM skip
- [x] `serialization_givenUserIdSnapshot_expectRoundTripPreservesIt` — pins persistence
- [x] `serialization_givenNullUserId_expectRoundTripPreservesNull` — pins anonymous-entry persistence round-trip
- [x] `schedule_givenWorkManagerAvailable_expectUniqueWorkEnqueued` — pins userId is written into WM input data
- [x] `doWork_givenNullUserId_expectDeferredWithoutClaimingOrTracking` — pins worker defers without claiming
- [x] `trackEvent_givenNullUserId_expectNoClaimAndNoSend` — same for async fallback
- [x] `trackEvent_givenNullUserIdEntry_expectFailureWithoutHttpCall` — pins the precondition guard inside the tracker
- [x] All existing geofence + `:core` + `:messagingpush` tests still green

## Risk

Low — production code diff is ~50 net lines across 7 files in the location module. No `:core` API change.

- **Threading**: no new state — `userId` is `val` on a data class. `secureUserStore.getUserId()` reads from thread-safe storage. The async tracker early-returns *before* launching its coroutine on null userId.
- **Anonymous events**: previously silently dropped by the worker; now correctly deferred to the pipeline flush. Behavior change is intentional and desirable.

## Follow-up (out of scope)

The `if (inputData.hasKeyWithValueOfType(KEY_LATITUDE, Double::class.javaObjectType)) inputData.getDouble(...) else null` dance in `GeofenceEventWorker.doWork` (lines 93-94) is pre-existing and verbose. A small `Data.getDoubleOrNull` extension would tidy it — happy to do that in a separate PR if anyone bites.

---

> [!NOTE]
> **Medium Risk**
> Touches core `GeofenceTransitionEvent` API, analytics user attribution, and geofence delivery routing (HTTP worker vs pipeline flush), including intentional behavior change for anonymous-at-queue-time events.
> 
> **Overview**
> Fixes **cross-user geofence attribution** by snapshotting `userId` when a transition is queued, instead of resolving identity at delivery time.
> 
> **`PendingGeofenceDelivery`** and **`Event.GeofenceTransitionEvent`** now carry an optional **`userId`** from queue time through persistence, WorkManager input, foreground flush, and the datapipelines subscriber. The **broadcast receiver** reads `secureUserStore` when appending an entry; **identified** transitions still schedule WorkManager/direct HTTP using the snapshot, while **anonymous** entries skip scheduling and stay in the pending store for the **foreground flush** (pipeline `anonymousId` path). **`GeofenceEventTracker`** takes the full entry and POSTs the snapshotted `userId`; null/empty `userId` is a hard precondition failure rather than a silent skip.
> 
> On the datapipelines side, the geofence handler **`track()`s** with optional **enrichment** so a non-null snapshot overrides the current SDK user for that event only. Logging is updated for deferred anonymous delivery; tests cover snapshotting, routing, WM payload, worker deferral, and pinned vs current identity.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d6dc4c22e0a8ff5503b808236f483021797a1e50. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/location/src/main/kotlin/io/customer/location/geofence/store/PendingGeofenceDelivery.kt
+++ b/location/src/main/kotlin/io/customer/location/geofence/store/PendingGeofenceDelivery.kt
@@ -20,7 +20,8 @@ internal data class PendingGeofenceDelivery(
     val transition: Event.GeofenceTransition,
     val latitude: Double?,
     val longitude: Double?,
-    val timestamp: Long
+    val timestamp: Long,
+    val userId: String?
 ) : PendingDeliveryStore.PendingDeliveryEntry {
     override val key: String get() = "${geofenceId}_${transition.name}_$timestamp"
 
--- a/location/src/main/kotlin/io/customer/location/geofence/worker/GeofenceEventTracker.kt
+++ b/location/src/main/kotlin/io/customer/location/geofence/worker/GeofenceEventTracker.kt
@@ -7,7 +7,6 @@ import io.customer.sdk.core.network.CustomerIOHttpClient
 import io.customer.sdk.core.network.HttpRequestParams
 import io.customer.sdk.core.util.DispatchersProvider
 import io.customer.sdk.data.store.PendingDeliveryStore
-import io.customer.sdk.data.store.SecureUserStore
 import io.customer.sdk.data.store.claimSendRestore
 import io.customer.sdk.util.EventNames
 import kotlinx.coroutines.CoroutineScope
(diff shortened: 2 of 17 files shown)


location/src/main/kotlin/io/customer/location/geofence/store/PendingGeofenceDelivery.kt
```diff
@@ -20,7 +20,8 @@ internal data class PendingGeofenceDelivery(
     val transition: Event.GeofenceTransition,
     val latitude: Double?,
     val longitude: Double?,
-    val timestamp: Long
+    val timestamp: Long,
+    val userId: String?
```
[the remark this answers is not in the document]
mrehan27: Deliberate drop, not an oversight.

`PendingGeofenceDelivery` was introduced in #716 (commit `4bb0e767`) and exists only on `feature/geofence-on-device` and its descendants — never on `main`. Since no shipped SDK persists this type, there's no legacy JSON on disk to deserialize, and the `= null` default isn't earning its keep.

An earlier iteration of this PR did have `= null` plus a `deserialization_givenOldJsonWithoutUserId_expectDefaultsToNull` test; both were dropped during review for this reason. When `feature/geofence-on-device` eventually merges to `main`, the whole delivery store ships for the first time — future field additions on top of *that* baseline will need defaults, but that's the first cutover, not the upgrade path this analysis assumes.

mahmoud-elmorabea, line 24: Since `userId` is not included in `toEventProperties` and is added at top level key in the event, are we missing that value from the app-launch flush operation?

location/src/main/kotlin/io/customer/location/geofence/worker/GeofenceEventTracker.kt
```diff
@@ -18,53 +17,35 @@ import org.json.JSONObject
...
-            put("transition_type", transition.name.lowercase())
-            latitude?.let { put("latitude", it) }
-            longitude?.let { put("longitude", it) }
-            put("timestamp", timestamp)
-        }
         val bodyJson = JSONObject().apply {
-            put("properties", propertiesJson)
+            put("properties", JSONObject(entry.toEventProperties()))
             put("event", eventName)
             put("userId", userId)
```
mahmoud-elmorabea, line 50: Here the `userId` is pulled from the pending entry but I think on the flush path it doesn't pull the entry pinned `userId` rather the current `userId` at flush time

mahmoud-elmorabea (changes requested):
Looks good! I think we just need to make sure that the app launch flush path, ensures that the current `userId` matches the one saved in the pending entry. Either drop it or attribute it to the pinned `userId` in the pending entry

mrehan27:
@mahmoud-elmorabea Good catch. Updated to handle this properly.

mahmoud-elmorabea (approved):
Looks good, thanks for the update! 💯

PR #723, opened 2026-06-09
chore(live-notifications): backend lifecycle reporting + out-of-order/dismissal handling

mahmoud-elmorabea:
## What

Adds the Android analogue of the iOS Live Activity **backend lifecycle**, plus the **`timestamp` / `dismissal_date`** envelope handling Android needs because it renders FCM data directly (there's no APNs/ActivityKit ordering layer to lean on).

Stacked on #722 (set base to `feature/live-notifications` once #722 merges).

### Backend lifecycle
- `HttpClient` gains a per-request `HttpMethod` (PUT/DELETE) and a status-aware `HttpRequestException`, so callers can retry 5xx but not 4xx.
- **`LiveNotificationLifecycleClient`**
  - per-template registration: `PUT /v1/live_activities/registration/{activityType}` with `{token, os:"android", transport:"fcm", userId}` (Android's equivalent of iOS push-to-start — no special token, registers the FCM token per type)
  - user dismissal: `DELETE /v1/live_activities/{activityId}`
  - retries 5xx up to 3×, skips 4xx (mirrors the iOS client)
- **`LiveNotificationRegistrar`** registers all built-in types on device-token rotation and on user change, deduped per type via `LiveNotificationStore`; clears registrations on reset / token deletion.
- **User dismissal** is detected via the notification's delete intent (`LiveNotificationDismissReceiver`). Programmatic `NotificationManager.cancel()` does **not** fire a delete intent, so a server-driven `end` is never mis-reported as a user dismissal.

### Envelope metadata
- **`timestamp`** — out-of-order / duplicate guard: a push whose timestamp is `<= ` the last one rendered for that `activity_id` is dropped; the guard is skipped when `timestamp` is absent (arrival order), and the entry is cleared on `end`. Persisted in `LiveNotificationStore` with a launch-time trim of stale entries.
- **`dismissal_date`** — honored on `event:"end"` only; schedules removal at that time (epoch ms), or removes immediately when absent (no invented delay; replaces the old hardcoded 3s).

### Decisions baked in (per earlier alignment discussion)
- Token-based device identity (no `installationId`), even though it is now available on the base — can revisit.
- No per-instance content-state echo (`update` PUT) — not applicable on Android.

## Tests
`LiveNotificationStore` (dedup get/set/clear + stale trim), `LiveNotificationLifecycleClient` (request shape + 5xx-retry / 4xx-no-retry policy), and handler `timestamp` dedup + `dismissal_date` scheduling. Full `:messagingpush` unit suite, ktlint, and Android Lint pass locally.

## Notes for reviewers (cross-team)
- `timestamp` / `dismissal_date` are treated as **epoch milliseconds** (consistent with the other Android template time fields), unlike iOS APNs seconds — backend must send Android-flavored values.
- Backend must emit the per-type registration / accept the DELETE with token-only identity.

---

> [!NOTE]
> **Medium Risk**
> New authenticated backend calls and push-handling logic affect live notification delivery and teardown; mis-timed timestamps or dismissal_date could drop updates or remove UI early, but scope is limited to the live-notification path.
> 
> **Overview**
> Adds **Android live-notification backend lifecycle** (FCM token registration per built-in activity type, user-dismissal reporting) and **envelope handling** for out-of-order FCM updates and scheduled teardown on `end`.
> 
> **Lifecycle:** `HttpClient` now supports configurable methods and surfaces `HttpRequestException` with status codes for retry policy. `LiveNotificationLifecycleClient` performs `PUT /v1/live_activities/registration/{activityType}` and `DELETE /v1/live_activities/{activityId}` with 5xx retries. `LiveNotificationRegistrar` starts during push module init, re-registers on token/user changes with SharedPreferences dedup, and clears on reset/token deletion. User swipes are reported via notification **delete intents** and `LiveNotificationDismissReceiver` (programmatic `cancel` does not fire delete, avoiding false dismissals on server `end`).
> 
> **Handler:** Drops pushes when `timestamp` ≤ last rendered for the same `activity_id` (persisted in `LiveNotificationStore`); on `end`, honors `dismissal_date` (epoch ms) or cancels immediately. Both notification builders wire `setDeleteIntent`.
> 
> **API surface:** Public `LiveNotificationDismissReceiver` is declared in the API dump and manifest.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 252fc3ea91bdf6aeed043e7bc5c10348e6be5ce0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/messagingpush/src/main/java/io/customer/messagingpush/LiveNotificationHandler.kt
+++ b/messagingpush/src/main/java/io/customer/messagingpush/LiveNotificationHandler.kt
@@ -7,13 +7,17 @@ import android.content.Intent
 import android.content.pm.PackageManager
 import android.os.Build
 import android.os.Bundle
+import android.os.Handler
+import android.os.Looper
 import androidx.annotation.ColorInt
 import androidx.annotation.DrawableRes
 import androidx.core.content.ContextCompat
 import io.customer.messagingpush.activity.NotificationClickReceiverActivity
 import io.customer.messagingpush.data.model.CustomerIOParsedPushPayload
+import io.customer.messagingpush.di.liveNotificationStore
 import io.customer.messagingpush.di.pushModuleConfig
 import io.customer.messagingpush.livenotification.LiveNotificationBranding
+import io.customer.messagingpush.livenotification.LiveNotificationDismissReceiver
 import io.customer.messagingpush.livenotification.template.TemplateAssets
 import io.customer.messagingpush.livenotification.template.TemplateRegistry
 import io.customer.messagingpush.util.PushTrackingUtil
@@ -95,6 +99,21 @@ internal class LiveNotificationHandler(
             )
             return
         }
+
+        // Out-of-order / duplicate guard. Android renders FCM data directly, so unlike
+        // iOS (where APNs/ActivityKit order updates) the SDK must drop stale pushes itself.
+        val store = SDKComponent.liveNotificationStore
+        val timestamp = bundle.getString(TIMESTAMP_KEY)?.toLongOrNull()
+        if (timestamp != null) {
+            val lastSeen = store.lastTimestamp(activityId)
+            if (lastSeen != null && timestamp <= lastSeen) {
+                SDKComponent.logger.debug(
+                    "Dropping out-of-order/duplicate live notification for '$activityId' (timestamp $timestamp <= $lastSeen)."
+                )
+                return
+            }
+        }
+
         val activityType = bundle.getString(ACTIVITY_TYPE_KEY)
         val template = TemplateRegistry.find(activityType)
         if (template == null) {
@@ -102,6 +121,12 @@ internal class LiveNotificationHandler(
                 "Unknown live notification template '$activityType'; dropping push for activity '$activityId'."
             )
             return
+        }
+
+        // We have committed to acting on this push: record its timestamp so later,
+        // out-of-order pushes for the same activity are dropped. `end` clears it below.
+        if (event != EVENT_END && timestamp != null) {
+            store.setLastTimestamp(activityId, timestamp)
         }
 
         val data = extractData(bundle)
@@ -134,6 +159,7 @@ internal class LiveNotificationHandler(
             activityId = activityId
         )
         val pendingIntent = createIntentForNotificationClick(context, notifId, parsedPayload)
+        val deletePendingIntent = createDeleteIntent(context, notifId, activityId)
 
         val notification = when {
             Build.VERSION.SDK_INT >= Build.VERSION_CODES.BAKLAVA -> {
@@ -153,6 +179,7 @@ internal class LiveNotificationHandler(
                     endIconRes = result.endIconRes,
                     trackerIconRes = result.trackerIconRes,
                     pendingIntent = pendingIntent,
+                    deleteIntent = deletePendingIntent,
                     countdownUntil = result.countdownUntil,
                     largeIcon = result.largeIcon,
                     showProgress = result.showProgress
@@ -172,6 +199,7 @@ internal class LiveNotificationHandler(
                     progress = result.progress,
                     progressMax = result.progressMax,
                     pendingIntent = pendingIntent,
+                    deleteIntent = deletePendingIntent,
                     countdownUntil = result.countdownUntil,
                     largeIcon = result.largeIcon,
                     showProgress = result.showProgress
@@ -183,10 +211,48 @@ internal class LiveNotificationHandler(
         notificationManager.notify(activityId, notifId, notification)
 
         if (event == EVENT_END) {
-            // Without a dismissal_date the activity is removed immediately.
-            // Scheduled dismissal via dismissal_date is added with lifecycle reporting.
-            notificationManager.cancel(activityId, notifId)
+            store.clearTimestamp(activityId)
+            scheduleEndDismissal(bundle, notificationManager, activityId, notifId)
         }
+    }
+
+    /**
+     * On `end`, removes the notification at the server-provided `dismissal_date`
+     * (epoch ms). When absent, the activity is removed immediately — there is no
+     * invented default delay. (Best-effort: a long delay does not survive
+     * process death.)
+     */
+    private fun scheduleEndDismissal(
+        bundle: Bundle,
+        notificationManager: NotificationManager,
+        activityId: String,
+        notifId: Int
+    ) {
+        val dismissAtMs = bundle.getString(DISMISSAL_DATE_KEY)?.toLongOrNull()
+        if (dismissAtMs == null) {
+            notificationManager.cancel(activityId, notifId)
+            return
+        }
+        val delayMs = (dismissAtMs - System.currentTimeMillis()).coerceAtLeast(0L)
+        Handler(Looper.getMainLooper()).postDelayed({
+            notificationManager.cancel(activityId, notifId)
+        }, delayMs)
+    }
+
+    private fun createDeleteIntent(
+        context: Context,
+        requestCode: Int,
+        activityId: String
+    ): PendingIntent {
+        val intent = Intent(context, LiveNotificationDismissReceiver::class.java).apply {
+            putExtra(LiveNotificationDismissReceiver.EXTRA_ACTIVITY_ID, activityId)
+        }
+        val flags = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+            PendingIntent.FLAG_UPDATE_CURRENT or PendingIntent.FLAG_IMMUTABLE
+        } else {
+            PendingIntent.FLAG_UPDATE_CURRENT
+        }
+        return PendingIntent.getBroadcast(context, requestCode, intent, flags)
     }
 
     /**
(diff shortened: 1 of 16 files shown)


messagingpush/src/main/java/io/customer/messagingpush/livenotification/LiveNotificationRegistrar.kt
```diff
@@ -0,0 +1,66 @@
...
+            token = event.token
+            registerAll()
+        }
+        eventBus.subscribe(Event.UserChangedEvent::class) { event ->
+            userId = event.userId ?: event.anonymousId
+            registerAll()
+        }
+        eventBus.subscribe(Event.DeleteDeviceTokenEvent::class) {
+            token = null
+            store.clearRegistrations()
```
mrehan27, line 44: We don't need to call `unregister` on `LiveNotificationStore` here, right?

messagingpush/src/main/java/io/customer/messagingpush/livenotification/LiveNotificationRegistrar.kt
```diff
@@ -0,0 +1,66 @@
...
+        }
+        eventBus.subscribe(Event.ResetEvent::class) {
+            store.clearRegistrations()
+        }
+    }
+
+    private suspend fun registerAll() {
+        val currentToken = token ?: return
+        val signature = "$currentToken|$userId"
```
mrehan27, line 53: If `userId` is `null` on first call, either because it's received later or because the user isn't logged in yet, and then the user logs in, would this register again without unregistering the previous entry?
Same question for token refresh cases, which are rare but possible since FCM can refresh the token.

messagingpush/src/main/java/io/customer/messagingpush/livenotification/LiveNotificationStore.kt
```diff
@@ -0,0 +1,74 @@
...
+ * - **Out-of-order / dedup guard** (per `activity_id`): the last `timestamp`
+ *   seen, so a delayed or duplicate push that is older than one already
+ *   rendered is dropped. Unlike iOS (where APNs/ActivityKit order updates), the
+ *   Android SDK renders FCM data directly and must guard ordering itself.
+ *
+ * Timestamp entries are stored with their record time so stale ones (for
+ * activities that ended long ago without an explicit `end`) can be trimmed on
+ * app launch.
+ */
+internal class LiveNotificationStore(context: Context) {
```
mrehan27, line 23: Should we use `PreferenceStore` as the base here for consistency and ease?

messagingpush/src/main/java/io/customer/messagingpush/network/HTTPClient.kt
```diff
@@ -62,7 +78,7 @@ internal class HttpClientImpl : HttpClient {
             // Configure the connection
             connection.connectTimeout = connectTimeoutMs
             connection.readTimeout = readTimeoutMs
-            connection.requestMethod = "POST"
+            connection.requestMethod = params.method.name
```
mrehan27, line 81: Looks like this branch is behind `main`. I updated this class earlier to give us more flexibility. Would you prefer to update this/feature branch from `main` first, or merge this into feature branch and then merge `main` there to resolve conflicts?

messagingpush/src/main/java/io/customer/messagingpush/LiveNotificationHandler.kt
```diff
@@ -95,6 +99,21 @@ internal class LiveNotificationHandler(
...
             return
         }
+
+        // Out-of-order / duplicate guard. Android renders FCM data directly, so unlike
+        // iOS (where APNs/ActivityKit order updates) the SDK must drop stale pushes itself.
+        val store = SDKComponent.liveNotificationStore
+        val timestamp = bundle.getString(TIMESTAMP_KEY)?.toLongOrNull()
+        if (timestamp != null) {
+            val lastSeen = store.lastTimestamp(activityId)
+            if (lastSeen != null && timestamp <= lastSeen) {
```
mrehan27, line 109: Does this guard also run for end events? If a live notification end event has a timestamp that isn't strictly newer than the last update, would it get dropped and never cancel the notification?

messagingpush/src/main/java/io/customer/messagingpush/LiveNotificationHandler.kt
```diff
@@ -104,6 +123,12 @@ internal class LiveNotificationHandler(
             return
         }
 
+        // We have committed to acting on this push: record its timestamp so later,
+        // out-of-order pushes for the same activity are dropped. `end` clears it below.
+        if (event != EVENT_END && timestamp != null) {
+            store.setLastTimestamp(activityId, timestamp)
+        }
```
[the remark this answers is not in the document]
mrehan27: Can we confirm whether this is valid?

mrehan27 (commented):
Only a few questions from my side. Also, I believe the PR should use `chore` prefix instead of `feat`?

messagingpush/src/main/java/io/customer/messagingpush/LiveNotificationHandler.kt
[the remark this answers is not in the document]
mahmoud-elmorabea: Confirmed + fixed: `end` now records its timestamp as the high-water mark instead of clearing the guard, so a stale `update` arriving after `end` gets dropped. Covered by `handle_givenStaleUpdateAfterEnd_isDropped`.

messagingpush/src/main/java/io/customer/messagingpush/LiveNotificationHandler.kt
```diff
@@ -183,10 +211,48 @@ internal class LiveNotificationHandler(
...
+    ) {
+        val dismissAtMs = bundle.getString(DISMISSAL_DATE_KEY)?.toLongOrNull()
+        if (dismissAtMs == null) {
             notificationManager.cancel(activityId, notifId)
+            return
+        }
+        val delayMs = (dismissAtMs - System.currentTimeMillis()).coerceAtLeast(0L)
+        Handler(Looper.getMainLooper()).postDelayed({
+            notificationManager.cancel(activityId, notifId)
+        }, delayMs)
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Fixed — scheduled cancels are tracked in a per-`activity_id` pending map now, so a new lifecycle for the same id cancels the stale one.

messagingpush/src/main/java/io/customer/messagingpush/livenotification/LiveNotificationRegistrar.kt
mahmoud-elmorabea: Right, no client-side unregister by design — backend runs cleanup jobs for stale registrations.

messagingpush/src/main/java/io/customer/messagingpush/livenotification/LiveNotificationRegistrar.kt
mahmoud-elmorabea: Yeah, it re-registers on login or token refresh — that's intended. We don't unregister the old entry client-side; backend cleanup handles stale ones. The store dedupes per type so the same `token|userId` won't get sent twice.

messagingpush/src/main/java/io/customer/messagingpush/livenotification/LiveNotificationStore.kt
mahmoud-elmorabea: Kept raw prefs here — the dynamic per-type/per-activity keys + composite values + launch-time trim don't really fit PreferenceStore's fixed-key model. Open to revisiting if you feel strongly.

messagingpush/src/main/java/io/customer/messagingpush/network/HTTPClient.kt
mahmoud-elmorabea: Yep — you moved this to core in #672. Plan is to reconcile during the feature→main integration: merge main in, take your core client, and port the PUT/DELETE + status-aware retry onto it. Tracking as a follow-up, will coordinate with you.

messagingpush/src/main/java/io/customer/messagingpush/LiveNotificationHandler.kt
mahmoud-elmorabea: No — `end` bypasses the timestamp guard, so a stale-timestamped end still cancels. Test: `handle_givenStaleEndTimestamp_stillCancels`.

messagingpush/src/main/java/io/customer/messagingpush/LiveNotificationHandler.kt
```diff
@@ -95,6 +99,21 @@ internal class LiveNotificationHandler(
...
+        val store = SDKComponent.liveNotificationStore
+        val timestamp = bundle.getString(TIMESTAMP_KEY)?.toLongOrNull()
+        if (timestamp != null) {
+            val lastSeen = store.lastTimestamp(activityId)
+            if (lastSeen != null && timestamp <= lastSeen) {
+                SDKComponent.logger.debug(
+                    "Dropping out-of-order/duplicate live notification for '$activityId' (timestamp $timestamp <= $lastSeen)."
+                )
+                return
+            }
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Correct for this PR in isolation — the `end`-bypass and `handle_givenStaleEndTimestamp_stillCancels` come in the next PR in the stack (#724, commit `bc8b7df8`), where we also harden the high-water mark so a stale `end` still cancels without letting later stale updates resurrect it. Stacked the review-round fixes on top rather than rewriting each PR.

mrehan27 (approved):

PR #724, opened 2026-06-09
chore(live-notifications): local start API + custom types + app render callback

mahmoud-elmorabea:
## What

Adds the host-app surface for live notifications. Stacked on #723 (retarget to `feature/live-notifications` as the chain merges down).

### Start a live notification locally
- `ModuleMessagingPushFCM.startLiveNotification(data: LiveNotificationData): String` and `startLiveNotification(activityType: String, data: Map<String, Any?>): String` — the SDK generates the `activity_id` (UUID), renders the notification immediately, and registers the instance with `PUT /v1/live_activities/{activityId}/push_token` so the backend can push updates.
- Rendering goes straight through `LiveNotificationHandler` (not the FCM delivery path), so **no delivered/opened metric is fabricated** for a locally-started notification.
- Sealed **`LiveNotificationData`** with the 5 built-in templates; nested `Airport`/`Team` serialize to JSON. The `Map` overload is for customer-defined types.

### Customer-defined activity types
- `MessagingPushModuleConfig.Builder.registerLiveNotificationTypes(vararg)` (the iOS `register()` analogue). The registrar registers built-ins **+** custom types with the backend so it can push to them.

### App render override
- `CustomerIOPushNotificationCallback.createLiveNotification(payload, context): Notification?` lets the app fully render *any* live notification. A non-null return is posted **as-is** (the SDK doesn't wrap it) while the SDK still owns the posting lifecycle — keyed by `activity_id` so updates replace it, and cancelled on `end`. Returning null falls back to the SDK template. Custom types have no built-in template, so the callback is **required** for them (else the push is dropped + logged).

### Decisions (per earlier plan)
- API surface lives on `ModuleMessagingPushFCM` (no public push facade existed).
- App-rendered notifications are not wrapped, so user-dismissal DELETE reporting is the app's responsibility for those.

## Tests
`LiveNotificationData` field mapping (incl. nested JSON), `registerInstance` request shape, and handler coverage for the callback override (built-in + custom) and custom-type-without-callback drop. Full `:messagingpush` unit suite + ktlint + Android Lint pass locally; binary API dump updated (only the intended public symbols).

---

> [!NOTE]
> **Medium Risk**
> New public API and opt-in behavior change when live notifications run; push handling and lifecycle registration paths are expanded but covered by tests.
> 
> **Overview**
> Makes **live notifications opt-in** via `MessagingPushModuleConfig.Builder.setLiveNotificationTypes`: nothing registers with Customer.io and incoming pushes are ignored until at least one activity type is enabled. Backend registration follows only those types (built-in and custom), not every built-in template by default.
> 
> Adds **`ModuleMessagingPushFCM.startLiveNotification`** (`LiveNotificationData` for the five built-in templates, or `activityType` + `Map` for custom types): the SDK assigns an `activity_id`, renders immediately through `LiveNotificationHandler` (no fabricated push metrics), and registers the instance with `PUT /v1/live_activities/{activityId}/push_token`.
> 
> Introduces **`CustomerIOPushNotificationCallback.createLiveNotification`** so the host app can post a fully custom `Notification` (SDK still owns notify/cancel on `end`); custom enabled types without a callback are dropped. Handler logic is tightened: enabled-type gate, app-vs-template rendering, cancellable scheduled `end` dismissals, and timestamp ordering so stale updates cannot resurrect an activity after `end`.
> 
> Also adds shared **field-name constants**, future-only countdown chronometers in notification builders, and sample/demo wiring for the local-start API.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 60ca0e20f79b719349729094095cae723b4e86c4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/messagingpush/api/messagingpush.api
+++ b/messagingpush/api/messagingpush.api
@@ -25,7 +25,7 @@ public final class io/customer/messagingpush/MessagingPushModuleConfig : io/cust
 	public synthetic fun <init> (ZLio/customer/messagingpush/data/communication/CustomerIOPushNotificationCallback;Lio/customer/messagingpush/config/PushClickBehavior;Lio/customer/messagingpush/livenotification/LiveNotificationBranding;Ljava/util/Set;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public final fun getAutoTrackPushEvents ()Z
 	public final fun getLiveNotificationBranding ()Lio/customer/messagingpush/livenotification/LiveNotificationBranding;
-	public final fun getLiveNotificationCustomTypes ()Ljava/util/Set;
+	public final fun getLiveNotificationTypes ()Ljava/util/Set;
 	public final fun getNotificationCallback ()Lio/customer/messagingpush/data/communication/CustomerIOPushNotificationCallback;
 	public final fun getPushClickBehavior ()Lio/customer/messagingpush/config/PushClickBehavior;
 	public fun toString ()Ljava/lang/String;
@@ -35,9 +35,9 @@ public final class io/customer/messagingpush/MessagingPushModuleConfig$Builder :
 	public fun <init> ()V
 	public fun build ()Lio/customer/messagingpush/MessagingPushModuleConfig;
 	public synthetic fun build ()Lio/customer/sdk/core/module/CustomerIOModuleConfig;
-	public final fun registerLiveNotificationTypes ([Ljava/lang/String;)Lio/customer/messagingpush/MessagingPushModuleConfig$Builder;
 	public final fun setAutoTrackPushEvents (Z)Lio/customer/messagingpush/MessagingPushModuleConfig$Builder;
 	public final fun setLiveNotificationBranding (Lio/customer/messagingpush/livenotification/LiveNotificationBranding;)Lio/customer/messagingpush/MessagingPushModuleConfig$Builder;
+	public final fun setLiveNotificationTypes ([Ljava/lang/String;)Lio/customer/messagingpush/MessagingPushModuleConfig$Builder;
 	public final fun setNotificationCallback (Lio/customer/messagingpush/data/communication/CustomerIOPushNotificationCallback;)Lio/customer/messagingpush/MessagingPushModuleConfig$Builder;
 	public final fun setPushClickBehavior (Lio/customer/messagingpush/config/PushClickBehavior;)Lio/customer/messagingpush/MessagingPushModuleConfig$Builder;
 }
@@ -330,6 +330,15 @@ public final class io/customer/messagingpush/livenotification/LiveNotificationDi
 }
 
 public final class io/customer/messagingpush/livenotification/LiveNotificationDismissReceiver$Companion {
+}
+
+public final class io/customer/messagingpush/livenotification/LiveNotificationType {
+	public static final field AUCTION_BID Ljava/lang/String;
+	public static final field COUNTDOWN_TIMER Ljava/lang/String;
+	public static final field DELIVERY_TRACKING Ljava/lang/String;
+	public static final field FLIGHT_STATUS Ljava/lang/String;
+	public static final field INSTANCE Lio/customer/messagingpush/livenotification/LiveNotificationType;
+	public static final field LIVE_SCORE Ljava/lang/String;
 }
 
 public final class io/customer/messagingpush/processor/PushMessageProcessor$Companion {
--- a/messagingpush/api/messagingpush.api
+++ b/messagingpush/api/messagingpush.api
@@ -22,9 +22,10 @@ public final class io/customer/messagingpush/CustomerIOFirebaseMessagingService$
 
 public final class io/customer/messagingpush/MessagingPushModuleConfig : io/customer/sdk/core/module/CustomerIOModuleConfig {
 	public static final field Companion Lio/customer/messagingpush/MessagingPushModuleConfig$Companion;
-	public synthetic fun <init> (ZLio/customer/messagingpush/data/communication/CustomerIOPushNotificationCallback;Lio/customer/messagingpush/config/PushClickBehavior;Lio/customer/messagingpush/livenotification/LiveNotificationBranding;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public synthetic fun <init> (ZLio/customer/messagingpush/data/communication/CustomerIOPushNotificationCallback;Lio/customer/messagingpush/config/PushClickBehavior;Lio/customer/messagingpush/livenotification/LiveNotificationBranding;Ljava/util/Set;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public final fun getAutoTrackPushEvents ()Z
 	public final fun getLiveNotificationBranding ()Lio/customer/messagingpush/livenotification/LiveNotificationBranding;
+	public final fun getLiveNotificationCustomTypes ()Ljava/util/Set;
 	public final fun getNotificationCallback ()Lio/customer/messagingpush/data/communication/CustomerIOPushNotificationCallback;
 	public final fun getPushClickBehavior ()Lio/customer/messagingpush/config/PushClickBehavior;
 	public fun toString ()Ljava/lang/String;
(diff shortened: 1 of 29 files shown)


messagingpush/api/messagingpush.api
mrehan27, line 1: Do all these changes here need to be public? 😱

messagingpush/src/main/java/io/customer/messagingpush/livenotification/LiveNotificationManager.kt
```diff
@@ -0,0 +1,96 @@
...
+            deliveryToken = "",
+            smallIcon = smallIcon,
+            tintColor = tintColor,
+            channelId = channelId,
+            notificationManager = notificationManager
+        )
+    }
+
+    private fun registerInstance(activityId: String, activityType: String) {
+        fcmTokenProvider.getCurrentToken { token ->
```
mrehan27: We already call `fcmTokenProvider.getCurrentToken` in `ModuleMessagingPushFCM`. Should we hook into that instead?

messagingpush/src/main/java/io/customer/messagingpush/LiveNotificationHandler.kt
```diff
@@ -143,27 +149,58 @@ internal class LiveNotificationHandler(
...
+        val notification = appNotification ?: result?.let {
+            buildSdkNotification(context, channelId, effectiveSmallIcon, it, pendingIntent, deletePendingIntent)
+        }
+        if (notification == null) {
+            SDKComponent.logger.error(
+                "No renderer for live notification type '$activityType': no built-in template and " +
+                    "createLiveNotification returned null; dropping activity '$activityId'."
+            )
+            return
+        }
```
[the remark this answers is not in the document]
mrehan27: Can we please confirm this behavior?

messagingpush/src/main/java/io/customer/messagingpush/MessagingPushModuleConfig.kt
```diff
@@ -20,13 +20,15 @@ class MessagingPushModuleConfig private constructor(
...
-    val liveNotificationBranding: LiveNotificationBranding?
+    val liveNotificationBranding: LiveNotificationBranding?,
+    val liveNotificationCustomTypes: Set<String>
 ) : CustomerIOModuleConfig {
     class Builder : CustomerIOModuleConfig.Builder<MessagingPushModuleConfig> {
         private var autoTrackPushEvents: Boolean = true
         private var notificationCallback: CustomerIOPushNotificationCallback? = null
         private var pushClickBehavior: PushClickBehavior = ACTIVITY_PREVENT_RESTART
         private var liveNotificationBranding: LiveNotificationBranding? = null
+        private var liveNotificationCustomTypes: Set<String> = emptySet()
```
mrehan27: With this config, can live notifications be opt-in feature? Should we skip setting up Live Notification classes in `ModuleMessagingPushFCM` if the types are empty?

mrehan27 (approved):
Looks good overall, just a few comments. Also, I think this PR should also use `chore` prefix instead of `feat`?

messagingpush/src/main/java/io/customer/messagingpush/LiveNotificationHandler.kt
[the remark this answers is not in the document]
mahmoud-elmorabea: Confirmed + fixed — end-event cleanup (cancel + clear pending) now runs even when the callback returns null / no renderer, before the early return. Test: `customType_endWithoutRenderer_stillCancels`.

messagingpush/api/messagingpush.api
mahmoud-elmorabea: Most of it, yeah — the start API, the enable config, `LiveNotificationData`/types, and the dismiss receiver are all intended public surface. apiDump's updated to match; internals (handler/manager/registrar/store) stay internal.

messagingpush/src/main/java/io/customer/messagingpush/livenotification/LiveNotificationManager.kt
mahmoud-elmorabea: Switched to `registrar.currentToken()`, which reuses the token the registrar already tracks (fetched once in ModuleMessagingPushFCM) instead of re-requesting it.

messagingpush/src/main/java/io/customer/messagingpush/MessagingPushModuleConfig.kt
mahmoud-elmorabea: Done — opt-in now via `setLiveNotificationTypes(...)`. ModuleMessagingPushFCM skips starting the registrar entirely when no types are enabled.

messagingpush/src/main/java/io/customer/messagingpush/livenotification/LiveNotificationManager.kt
```diff
@@ -0,0 +1,100 @@
...
+
+    private fun buildBundle(activityId: String, activityType: String, fields: Map<String, Any?>): Bundle =
+        Bundle().apply {
+            putString(LiveNotificationHandler.ACTIVITY_ID_KEY, activityId)
+            putString(LiveNotificationHandler.EVENT_KEY, EVENT_START)
+            putString(LiveNotificationHandler.ACTIVITY_TYPE_KEY, activityType)
+            putString(LiveNotificationHandler.TIMESTAMP_KEY, System.currentTimeMillis().toString())
+            for ((key, value) in fields) {
+                if (value != null) putString(key, value.toString())
+            }
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Good catch — fixed by writing the user fields first and the envelope keys after, so reserved keys (timestamp/activity_id/etc.) always win over a colliding field.

messagingpush/src/main/java/io/customer/messagingpush/LiveNotificationHandler.kt
```diff
@@ -114,26 +133,22 @@ internal class LiveNotificationHandler(
...
+        // A new event for this activity supersedes any scheduled end-dismissal (handles
+        // activity_id reuse where the delayed cancel would otherwise kill the new notification).
+        cancelPendingDismissal(activityId)
 
-        // We have committed to acting on this push: record its timestamp so later,
-        // out-of-order pushes for the same activity are dropped. `end` clears it below.
-        if (event != EVENT_END && timestamp != null) {
+        // Record the high-water timestamp for ALL events (incl. `end`) so a later stale
+        // update — even one arriving after `end` — is dropped by the guard above.
+        if (timestamp != null) {
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Good catch — fixed in bddacbd1. The write now only advances the mark (`timestamp > lastSeen`), so a stale out-of-order `end` can't lower it and a later stale update still gets dropped. Added `handle_givenStaleEndThenStaleUpdate_doesNotResurrect`.

messagingpush/src/main/java/io/customer/messagingpush/LiveNotificationHandler.kt
```diff
@@ -143,27 +160,60 @@ internal class LiveNotificationHandler(
...
+            title = result?.title ?: bundle.getString(CustomerIOPushNotificationHandler.TITLE_KEY).orEmpty(),
+            body = result?.body ?: bundle.getString(CustomerIOPushNotificationHandler.BODY_KEY).orEmpty(),
             activityId = activityId
         )
         val pendingIntent = createIntentForNotificationClick(context, notifId, parsedPayload)
         val deletePendingIntent = createDeleteIntent(context, notifId, activityId)
 
-        val notification = when {
-            Build.VERSION.SDK_INT >= Build.VERSION_CODES.BAKLAVA -> {
-                val params = Api36LiveNotificationParams(
```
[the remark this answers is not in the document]
mahmoud-elmorabea: By design — `cancelImmediately` is a terminal signal from a built-in template (only the expired-countdown-with-no-expiredMessage case): the activity is over and should be dismissed. The render callback is for custom/template-less types, where there's no template and so no `cancelImmediately` path — it always runs for those. The "fully render" wording is about rendering content, not overriding a built-in template's terminal dismissal. Leaving behavior as-is

PR #734, opened 2026-06-15
chore: enable Live Notifications

mahmoud-elmorabea:
Live Notification feature

---

> [!NOTE]
> **High Risk**
> Large notification/FCM surface area with ordering, logout supersede, and identity-gated CDP events; incorrect behavior would strand notifications or drop backend-tracked activities.
> 
> **Overview**
> Adds **Android Live Notifications** (ongoing, in-place updates analogous to iOS Live Activities), plus a datapipelines fix so identity checks are reliable right after login.
> 
> **Live notifications (messagingpush):** Opt-in via `MessagingPushModuleConfig` (enabled types, branding, optional `CustomerIOLiveNotificationsCallback` for custom types). Public APIs on `ModuleMessagingPushFCM`: `startLiveNotification` / `updateLiveNotification` / `endLiveNotification`. Server pushes with `cioInstanceId` are routed off the standard FCM handler into a dedicated live path (separate channel, serialized render queue). Built-in **segments** and **countdown timer** templates; API 36+ uses promoted/progress styling when permitted, otherwise `NotificationCompat` fallback. Lifecycle is reported as CDP track events (`start`/`end`, `register_push_to_start`); local updates are render-only. Persistent store handles ordering, terminal `end`, dismiss reporting, and logout cleanup.
> 
> **Push / identity plumbing:** `CustomerIOParsedPushPayload` gains `activityId`; opened metrics are skipped when delivery id/token are blank (locally started). Shared `BitmapDownloader` replaces inline image fetch on the FCM thread.
> 
> **Synchronous `isUserIdentified`:** `CustomerIO` keeps a caller-thread `syncUserIdentified` mirror updated in `identify()` / `clearIdentify()`, with analytics fallback for restored sessions; `DataPipeline.isUserIdentified` is documented for modules (e.g. live-notification lifecycle gating immediately after `identify()`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fd69467ec76daaed964e3798efb1964505ccbe8d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


messagingpush/src/main/java/io/customer/messagingpush/LiveNotificationHandler.kt
```diff
@@ -0,0 +1,437 @@
...
+                return
+            }
+            advanceHighWaterMark()
+            notificationManager.cancel(activityId, notifId)
+            return
+        }
+
+        bundle.putInt(CustomerIOPushNotificationHandler.NOTIFICATION_REQUEST_CODE, notifId)
+        val parsedPayload = CustomerIOParsedPushPayload(
+            extras = Bundle(bundle),
```
Shahroz16: Could we give the custom callback the same flattened fields used by built-in templates? Services always sends customer content inside the JSON `payload` data field. `extractData()` unwraps that for built-ins, but this payload keeps the raw Bundle even though the callback KDoc promises flattened extras. As a result every server-delivered custom notification sees `payload` rather than `driverName`, `status`, etc.; only local custom starts work. A nested-payload callback test would catch this.

messagingpush/src/main/java/io/customer/messagingpush/LiveNotificationHandler.kt
```diff
@@ -0,0 +1,437 @@
...
+        )
+        val pendingIntent = createIntentForNotificationClick(context, notifId, parsedPayload)
+        // Only in-progress notifications carry the dismiss intent (user-swipe -> end).
+        // A terminal `end` notification must not, or swiping it would report a second end.
+        val deletePendingIntent = if (isEnd) null else createDeleteIntent(context, notifId, activityId, activityType)
+
+        // The host app may fully render the notification; otherwise fall back to the SDK template.
+        val appNotification = SDKComponent.pushModuleConfig.liveNotificationCallback
+            ?.createLiveNotification(parsedPayload, context)
+        val notification = appNotification ?: result?.let {
```
Shahroz16, line 279: Could we keep the SDK click and dismiss intents when an app-rendered notification doesn't provide its own? Both intents are created just above, but a notification returned by the callback bypasses `buildSdkNotification` and drops them. The native, React Native and Flutter samples all return notifications without either intent, so custom notifications cannot open, report `opened`, deep-link, or report a swipe. Since the SDK still owns posting and lifecycle, defaulting only missing intents would preserve app overrides without making every callback reconstruct internal receiver details.

messagingpush/src/main/java/io/customer/messagingpush/livenotification/LiveNotificationManager.kt
```diff
@@ -0,0 +1,336 @@
...
+    private fun reportStart(
+        activityId: String,
+        activityType: String,
+        attributes: Map<String, Any?>,
+        contentState: Map<String, Any?>
+    ) {
+        val deviceId = SDKComponent.android().globalPreferenceStore.getDeviceToken()
+        if (deviceId.isNullOrBlank()) {
+            SDKComponent.logger.debug(
+                "No FCM token available yet; skipping start event for live notification '$activityId'."
```
Shahroz16, line 314: Could we retain this start until the initial FCM token request finishes instead of dropping it permanently? Module initialization asks Firebase for the token asynchronously, so on a first install `initialize → identify → startLiveNotification` can render and return an id while this branch skips its only backend registration. When the token arrives, the registrar sends the type token but never replays this instance, so the backend can never update or end it. This is a normal immediate-start path; a small pending-start queue flushed by `RegisterDeviceTokenEvent` and cleared on logout would close it.

mahmoud-elmorabea: Thanks for the detailed trace — I dug into this one but I don't think we should queue and replay these starts.

Not reporting here is the intended contract rather than a timing gap. `startLiveNotification`'s KDoc already draws the line: the notification renders regardless of identity, but lifecycle events are only reported for an identified, registered device — "call `identify` first if you need the backend to track this activity and push updates/remote end (matches iOS Live Activities)". A start made before the device is registered falls on the not-tracked side of that line by design, so buffering it would let an unsupported call through rather than fix a bug — we deliberately don't support start/update outside an identified, registered state, and this isn't only a started-too-early scenario.

One accuracy note on the trace: the branch that fires in your example is the blank-device-token guard, not the identity check (identity gating lives further down in the lifecycle client). Same conclusion either way — device registration is part of the same prerequisite set as identity.

Leaving the behavior as-is. If the current wording reads as accidental rather than deliberate, I'm happy to make the drop louder — raising that debug log, or calling the prerequisite out more explicitly in the local-start docs — just say which you'd prefer.

Shahroz16: Thanks, this makes sense. I checked the iOS path too and it has the same identity plus device-token gate, so I am comfortable treating this as an existing low-frequency limitation rather than holding this PR. I do not think we need queue or replay here. The follow-up I would keep is making the public Android KDoc explicit that a registered device token is also required and that an early lifecycle event is not retried; a readiness API can come after release.

messagingpush/src/main/java/io/customer/messagingpush/data/communication/CustomerIOLiveNotificationsCallback.kt
```diff
@@ -0,0 +1,44 @@
...
+     * intents, or `null` to fall back to the SDK's built-in template.
+     *
+     * Required for customer-defined activity types (enabled via
+     * [io.customer.messagingpush.MessagingPushModuleConfig.Builder.enableCustomLiveNotificationTypes]),
+     * which have no built-in template; if this returns `null` for such a type, the
+     * notification is dropped.
+     *
+     * The SDK still owns the posting lifecycle: it posts the returned notification
+     * keyed by the activity id (so later updates replace it) and cancels it on
+     * `end`. It does NOT modify the returned notification, so any dismissal
```
Shahroz16: Small docs fix, not a release blocker: this paragraph no longer matches the behavior in 6058bc14. The SDK now fills missing click and dismiss intents, reports dismissal through the default delete intent, and re-invokes the callback for a terminal end rather than always cancelling. Could we update this paragraph so apps know which parts they still own?

messagingpush/src/main/java/io/customer/messagingpush/LiveNotificationHandler.kt
```diff
@@ -0,0 +1,486 @@
...
+            ?.withDefaultIntents(pendingIntent, deletePendingIntent)
+        val notification = appNotification ?: result?.let {
+            buildSdkNotification(context, channelId, effectiveSmallIcon, it, pendingIntent, deletePendingIntent, ongoing = !isEnd)
+        }
+
+        // A logout/reset may have landed while the branding logo downloaded or the app
+        // renderer ran above. If so, drop this render: don't post the notification and don't
+        // write the activity type or the high-water mark back into the store that reset
+        // just cleared.
+        if (isSuperseded()) {
```
Shahroz16, line 287: This still leaves a practical dismissal race on Android 14 and newer. The ended check runs before template rendering and the app callback, both of which can block; a remote logo download can wait up to 20 seconds. If the user swipes the existing ongoing notification during that work, the receiver marks it ended, but this path only rechecks logout before calling notify, so the notification is posted again and stays until the user dismisses it a second time. Could we repeat the non-end isEnded check here, immediately before committing the notification? A test that marks the activity ended from inside the callback would cover the real window.

Shahroz16 (approved):

messagingpush/src/main/java/io/customer/messagingpush/LiveNotificationHandler.kt
mahmoud-elmorabea: Good catch — you're right, and my earlier check was in the wrong place. It sat at the pre-render guard, which closes the enqueue-to-render window but not the render-to-post one, and that second window is the long one: it spans the template render and the app callback, where a remote logo can block for ~20s.

Fixed in d3082831 by repeating the non-end `isEnded` check immediately before committing, on both the server and local paths.

Your test suggestion was the important part. `localRender_forAlreadyEndedActivity_isDropped` marks the activity ended before invoking, so it only ever exercised the early guard — it still passes with the new check removed. The new `dismissalDuringRender_isNotReposted` marks it ended from inside the callback and does fail without the fix, so the real window is covered now.

messagingpush/src/main/java/io/customer/messagingpush/data/communication/CustomerIOLiveNotificationsCallback.kt
mahmoud-elmorabea: Updated in d3082831. The paragraph now says the SDK fills in its click and dismiss intents wherever the app left them unset — so a tap reports `opened` and follows the deep link and a swipe reports `end` without the app wiring receivers — and that setting either intent yourself keeps it, along with the corresponding reporting.

Also documented the `end` behavior, which the old text had wrong in the other direction: the callback is invoked again for the terminal non-ongoing state, and returning `null` there cancels instead. Noted that a terminal notification deliberately carries no dismiss intent so swiping it doesn't report a second `end`.

messagingpush/src/main/java/io/customer/messagingpush/livenotification/LiveNotificationManager.kt
mahmoud-elmorabea: Thanks — agreed on scope. I've taken the KDoc follow-up in d3082831: both `startLiveNotification` overloads now state that a registered device token is required alongside an identified user, and that an event dropped for a missing prerequisite is not retried, so the backend can neither update nor end that instance. I also called out that the token is fetched asynchronously at initialization, which is the window your trace hit.

Leaving a readiness API for after release, as you suggested.

messagingpush/src/main/java/io/customer/messagingpush/LiveNotificationHandler.kt
mahmoud-elmorabea: Fixed in 6058bc14. The extras handed to the callback are now flattened the same way `extractData` unwraps them for built-in templates, so a server-delivered custom type sees `driverName`, `status` etc. rather than a raw `payload` string. Envelope keys already in the bundle win on collision, so nothing the SDK owns gets overwritten.

Added the nested-payload callback test you asked for — `serverPush_customType_callbackSeesFlattenedPayloadFields`, which fails without the change.

messagingpush/src/main/java/io/customer/messagingpush/LiveNotificationHandler.kt
mahmoud-elmorabea: Fixed in 6058bc14, using the defaulting approach you suggested rather than making hosts reconstruct receiver details: missing click and dismiss intents are filled in on the returned notification, and a notification that sets either keeps its own. Covered by `appRenderedNotification_getsSdkClickAndDismissIntents` plus `appRenderedNotification_keepsItsOwnIntents` as an override guard, and the KDoc is updated to match.

messagingpush/src/main/java/io/customer/messagingpush/LiveNotificationHandler.kt
```diff
@@ -0,0 +1,446 @@
...
+                // share a second; reject only strictly-older pushes.
+                val lastSeen = store.lastTimestamp(activityId)
+                if (timestamp != null && lastSeen != null && timestamp < lastSeen) {
+                    SDKComponent.logger.debug(
+                        "Dropping out-of-order/duplicate live notification for '$activityId' (timestamp $timestamp < $lastSeen)."
+                    )
+                    return
+                }
+            }
+        }
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Valid — fixed across 6058bc14 and d3082831.

Local start/update renders bypassed the order guard entirely, so they skipped `store.isEnded`. Two windows needed closing: enqueue-to-render (a re-check at the guard, 6058bc14) and render-to-post (a re-check immediately before `notify`, d3082831 — this is the long one, spanning the template render and the app callback where a remote logo can block ~20s). `end` is exempt from both since it renders precisely because it just went terminal.

Covered by `localRender_forAlreadyEndedActivity_isDropped` and `dismissalDuringRender_isNotReposted`.

messagingpush/src/main/java/io/customer/messagingpush/LiveNotificationHandler.kt
```diff
@@ -0,0 +1,446 @@
...
+            )
+            // A server-delivered `end` claims the terminal transition in the store *before*
+            // rendering, so a failure here would leave the previous ongoing (non-dismissible)
+            // notification on screen while every retry `end` is dropped by that same claim.
+            // Cancel it so a failed render can't strand the activity visible and unclearable.
+            // Local renders are excluded: LiveNotificationManager owns their terminal flow and
+            // claims ended only after the render.
+            if (!bypassOrderGuard && bundle.getString(EVENT_KEY) == EVENT_END && activityId != null) {
+                notificationManager.cancel(activityId, notificationId(activityId))
+            }
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Valid, and self-inflicted — the cancel-on-failed-render I added for server pushes was gated on `!bypassOrderGuard` on the reasoning that `LiveNotificationManager` owned the local terminal flow. That was wrong on both counts: `end()` claims the terminal transition before its queued render, and never cancelled on failure.

Gate removed in 6058bc14, so both paths cancel. Covered by `throwingCallback_onLocalEnd_cancelsToo`.

messagingpush/src/main/java/io/customer/messagingpush/livenotification/LiveNotificationManager.kt
```diff
@@ -0,0 +1,314 @@
...
+        // high-water mark must survive so a delayed older push can't resurrect the
+        // activity, and logout can still cancel a still-visible ended notification.
+        // Reclamation happens via the store's TTL trim / logout clear.
+        if (activityType == null) {
+            SDKComponent.logger.debug(
+                "No known live notification for '$activityId'; nothing to end."
+            )
+        } else if (store.markEnded(activityId)) {
+            reportEnd(activityId, activityType)
+        }
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Reopening my own assessment here: I initially judged this not worth fixing because the stated consequence — that nothing records the dismissal — doesn't hold, since `end()` has already reported the `end` before the dismiss receiver runs. That part was right, but I stopped there and missed the actual bug, which is the resurrection: the queued terminal render still posts after the user cleared the notification.

Fixed in 6058bc14: `end()` now claims `markEnded` before enqueuing its render and skips rendering when the claim is lost, which means a dismissal already ended it and there is nothing left to post. The related non-end case is covered by the render-time check in d3082831.

messagingpush/src/main/java/io/customer/messagingpush/livenotification/LiveNotificationManager.kt
```diff
@@ -0,0 +1,293 @@
...
+        val bundle = buildBundle(activityId, activityType, attributes + contentState, EVENT_START)
+        lastBundles[activityId] = Bundle(bundle)
+        render(bundle)
+        // Lifecycle reporting is intentionally decoupled from the local render: the
+        // app called start/end, so Customer.io must track the activity (and be
+        // able to push updates / remote-end it) even when the local render posts nothing
+        // — e.g. a custom type with no built-in template, or a transient render failure.
+        // The report is not blocking (token read is an in-memory pref; track() enqueues),
+        // so it stays on the caller thread; identity gating still applies in the client.
+        reportStart(activityId, activityType, attributes, contentState)
```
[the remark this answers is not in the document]
mahmoud-elmorabea: This one is intended behavior rather than a defect, so leaving it as-is. The decoupling is deliberate and documented at the call site: the app asked for the start, so Customer.io must track the activity — and be able to push updates or remote-end it — even when the local render posts nothing, e.g. a custom type with no built-in template or a transient render failure.

The fair part of the observation is that starting a type you never enabled is a misconfiguration that currently fails silently. If we want to surface that, a warning log when `start` is called for a type absent from `liveNotificationTypes` would do it without changing the reporting contract. Happy to add that if reviewers want it.

messagingpush/src/main/java/io/customer/messagingpush/livenotification/LiveNotificationDismissReceiver.kt
```diff
@@ -0,0 +1,66 @@
...
+        val activityType = intent.getStringExtra(EXTRA_ACTIVITY_TYPE) ?: return
+        SDKComponent.setupAndroidComponent(context = context)
+
+        val deviceId = SDKComponent.android().globalPreferenceStore.getDeviceToken()
+        if (deviceId.isNullOrBlank()) {
+            SDKComponent.logger.debug(
+                "No FCM token available; skipping end event for live notification '$activityId'."
+            )
+            return
+        }
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Valid, and it defeated the guards I'd just added — those all consult the terminal marker, so a dismissal that never sets it reopens exactly the resurrection they were meant to close. Reachable through the start-before-token window raised earlier in this PR, since a local start renders regardless of token.

Fixed in fd69467e: the terminal claim now happens first, and reporting is attempted after.

That reverses a deliberate ordering — the previous code claimed only after confirming the end was reportable, to avoid losing the end and blocking a later one. That trade no longer holds: lifecycle events require a registered device token and are never retried (documented in d3082831), so a 'later end' for an already-dismissed notification isn't reportable either, while the resurrection is user-visible.

`onReceive_withoutFcmToken_doesNotMarkEnded` asserted the old behavior, so it's inverted rather than added to. Also added `updateAfterTokenlessDismissal_isNotReposted`, which drives the receiver and the render guard together — the combination is where the gap actually was.

messagingpush/src/main/java/io/customer/messagingpush/LiveNotificationHandler.kt
```diff
@@ -0,0 +1,499 @@
...
+            else -> {
+                val reason = if (template != null) {
+                    "required content fields are missing (payload not flattened, or empty)"
+                } else {
+                    "no built-in template and createLiveNotification returned null"
+                }
+                SDKComponent.logger.error(
+                    "Not posting live notification '$activityId' (type '$activityType'): $reason."
+                )
+                return
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Valid — fixed in fd69467e by moving `advanceHighWaterMark()` past the post/cancel branch.

It also contradicted the function's own stated intent, 'invoked only once this render is committed': the `else` branch that logs and returns isn't a commit, so a render that showed nothing was still recording a last-seen timestamp and letting the out-of-order guard drop a later push carrying an older timestamp. Only the notify and cancel paths advance it now.

Covered by `renderThatPostsNothing_doesNotAdvanceTheHighWaterMark`, which fails without the move.

PR #740, opened 2026-06-17
chore: share one CDP timestamp formatter across delivery paths

mrehan27:
**Linear:** [MBL-1831](https://linear.app/customerio/issue/MBL-1831/remove-precise-location-data-from-geofence-events)

### Stack
- #739 — remove precise location from geofence events
- **this PR** — send the geofence event timestamp as ISO-8601 to CDP `/track`

> Stacked on #739; base will move to `feature/geofence-on-device` once #739 merges.

### Why
The Pipelines (CDP) `/track` endpoint reads the reserved top-level `timestamp` as an **ISO-8601 string**. A bare integer is interpreted as **milliseconds** — and the SDK's geofence timestamp is in **seconds**, so sending it as an integer would have dated every worker-delivered event to ~1970. (Verified directly against `cdp.customer.io/v1/track`: ISO string ✅, integer ms ✅, integer seconds ❌.)

The geofence worker's direct-HTTP path now emits the timestamp as ISO-8601, matching the analytics pipeline so both delivery channels attribute the event to the same instant.

### What changed
- New `Iso8601TimestampFormatter` in `:core` — the single source of truth for the CDP timestamp string, reachable by both `:datapipelines` and `:geofence` (the worker can't depend on `:datapipelines`).
- Removed `SegmentInstantFormatter` (a `:datapipelines`-only class) and repointed its callers at the core type: the geofence pipeline subscriber (`fromDate`), the tracking-migration path (`fromUnixSeconds`).
- Geofence worker emits the top-level `timestamp` via the core formatter.

### Tests
- `:core` `Iso8601TimestampFormatterTest` — format correctness (literal ISO, ms preservation, null-on-failure).
- `:datapipelines` `Iso8601TimestampFormatterParityTest` (replaces `SegmentInstantFormatterTest`) — asserts output stays identical to a real Segment `TrackEvent.timestamp`; kept in `:datapipelines` because it needs the Segment types.
- Worker test asserts the top-level `timestamp` is the expected ISO string.
- `:core` / `:datapipelines` / `:geofence` unit tests, lint, and apiCheck pass.

---

> [!NOTE]
> **Medium Risk**
> Changes how event timestamps are serialized on CDP `/track` for geofence worker delivery and migration/geofence pipeline enrichment; incorrect formatting would mis-attribute events, but behavior is guarded by parity and explicit ISO tests.
> 
> **Overview**
> Introduces **`Iso8601TimestampFormatter`** in `:core` as the shared way to build the ISO-8601 string CDP `/track` expects for the top-level `timestamp` field (avoids bare integers being read as milliseconds while geofence times are in seconds). **Removes** `:datapipelines`’s `SegmentInstantFormatter` and **switches** migration replay (`fromUnixSeconds`) and geofence EventBus tracking (`fromDate`) to the core formatter.
> 
> The **geofence worker** direct HTTP `/track` body now **includes** a top-level `timestamp` via `fromUnixSeconds`, aligned with the analytics pipeline. Tests move to `:core` unit coverage, a Segment parity test in `:datapipelines`, and an updated worker assertion for the ISO string.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 15de526304cb6a2fdb39c7dff91383fd8a6c1c9b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mahmoud-elmorabea (approved):
Awesome refactor 💯

PR #741, opened 2026-06-17
chore(geofence): coarsen sync location and widen refresh radii

mrehan27:
**Linear:** [MBL-1843](https://linear.app/customerio/issue/MBL-1843/use-coarse-location-for-nearby-geofence-fetch-requests)

### Why
The `/geofences/nearby` sync call sent the device's **precise** lat/lng as GET query params, so exact location landed in access logs/proxies on every fetch. This blurs the outbound coordinates so the SDK never transmits an exact position to fetch geofences — precise location stays on-device for proximity and movement-trigger logic.

### What changed
- **`GeofenceCoordinateCoarsener`** snaps the outbound coordinates to a **~1km grid** (round to 2 decimals; 0.01° ≈ 1.1km). Applied at the `GeofenceApiServiceImpl.fetchGeofences` boundary — a single chokepoint, so no caller can send precise coords. The repository still uses precise location for its own anchor/distance math; only the wire value is coarsened.
- Snapping is **deterministic, not jittered** — repeated syncs from the same area send the same value, so the true position can't be recovered by averaging many requests.
- **Widened fallback refresh radii** to minimize server calls: `localRefreshTriggerRadius` 1km→**3km**, `remoteFetchRefreshTriggerRadius` 5km→**20km**. These are fallbacks only — server-shipped config still overrides both.

### Cross-team dependencies
- **Server fetch radius must be ≥ ~21km** (20km re-sync + 1km coarsening cell), or a geofence near a cell/anchor boundary can be missed.
- The 20km re-sync stays accurate only if the **100-geofence cap** realistically spans ~20km for a deployment; in dense areas the cap truncates to the nearest 100 around the anchor, leaving a coverage gap. Accepted trade-off (fewer server calls vs coarser coverage).
- Server should ship config matching 3/20 or the new fallbacks won't be what runs in production.

### Tests
- `GeofenceCoordinateCoarsenerTest` — rounding to grid, same-cell collapse (the anti-averaging property), negatives/zero.
- `GeofenceApiServiceTest` — wire-level: precise in → coarse out (`fetchGeofences` had no test before).
- Full `:geofence` unit tests + lint pass.

---

> [!NOTE]
> **Medium Risk**
> Changes geofence fetch location and refresh thresholds; server fetch radius and config must stay aligned or boundary geofences can be missed.
> 
> **Overview**
[...]

PR #742, opened 2026-06-17
chore(geofence): fetch all geofences without sending location

mrehan27:
**Linear:** [MBL-1843](https://linear.app/customerio/issue/MBL-1843/use-coarse-location-for-nearby-geofence-fetch-requests)

> Supersedes #741 (coarse-location-only). Backend agreed to accept no location and return the full set, so we default to fetch-all and **retain** the coarse-nearby path behind a switch for a later phase.
>
> **Base:** `feature/geofence-on-device` (after #739 + #740, both merged).

### Why
The `/geofences/nearby` sync call sent the device's location to fetch geofences. The backend now returns the full set (capped at 100) without location, so the SDK can fetch everything and filter on-device — sending **zero** location to the backend.

### What changed
- **`GeofenceSyncMode`** (`FETCH_ALL` default, `NEARBY` retained). The active mode is an **SDK-internal** decision — not a runtime or server-pushed toggle — so the shipped SDK never sends location, and re-enabling nearby is a deliberate SDK release.
- **`FETCH_ALL`**: no location sent; remote fetch is time/staleness-based; movement only **re-ranks the cached set on-device** (movement can't change a location-independent set).
- **`refresh()` decision table** — only the re-fetch question depends on the mode; everything else is mode-agnostic:
  - re-fetch when stale in time, or (NEARBY) the device outran the remote radius **from the last API fetch**;
  - re-rank locally when the device left the trigger radius **since the last registration**, or the cached set isn't registered with the OS;
  - otherwise skip.
- **Two reference points** (a bug fix): re-fetch distance is measured from the last API fetch; re-rank distance from the last registration (the movement-trigger center). The re-rank condition is exactly "a movement EXIT missed while the app was dead," so it applies to **both** modes.
- **`NEARBY`** (retained, still tested): sends coarse (~1km grid) location via `GeofenceCoordinateCoarsener`, re-fetches on movement beyond the API anchor. Widened fallback radii (local 3km, remote 20km).
- **`GeofenceApiService`** split into two overloads: `fetchGeofences()` (fetch-all, no query params) and `fetchGeofences(lat, lng)` (nearby, coarsened at the boundary).

### Tests
- `GeofenceRepositoryTest` — decision-table coverage incl. the NEARBY re-rank-on-moved-device case, its within-trigger SKIP complement, and a **regression guard** that the re-rank is measured from the last registration, not the last API fetch.
- `GeofenceApiServiceTest` — coarse coords for the location overload, none for the no-arg overload.
- `GeofenceCoordinateCoarsenerTest` — grid rounding, same-cell collapse (anti-averaging), negatives/zero.
- Full `:geofence` unit tests, ktlint, and apiCheck pass.

---

> [!NOTE]
> **Medium Risk**
> Changes geofence sync behavior and location transmission (privacy-sensitive), but FETCH_ALL is the shipped default with broad test coverage; wrong staleness logic could leave stale OS registrations.
> 
> **Overview**
> **Default geofence sync no longer sends device location** to `/geofences/nearby`. The SDK fetches the full capped set with `fetchGeofences()` (no query params), ranks nearest regions on-device, and only re-fetches on **time staleness**—not on travel distance.
> 
> Introduces **`GeofenceSyncMode`**: **`FETCH_ALL`** is the active default (SDK-internal, not runtime-togglable); **`NEARBY`** remains for a future release and still sends **coarse** lat/lng via `GeofenceCoordinateCoarsener` (~1 km grid).
> 
> **`refresh()`** is refactored into a **`RefreshAction`** decision table: remote when time-stale or (NEARBY only) beyond the remote radius from the **last API fetch**; local re-rank when beyond the local trigger from the **last registration** (movement-trigger center) or when cache exists but OS regs are empty; otherwise skip. That split fixes conflating fetch-anchor distance with ranking staleness.
> 
> **`handleMovement()`** uses the same mode rules: FETCH_ALL never remote-fetches on movement; NEARBY still escalates to API past the remote radius. Fallback radii widen to **3 km** local and **20 km** remote.
> 
> Tests cover FETCH_ALL vs NEARBY, coarsening, and the registration-vs-fetch-anchor regression.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b7387ae1ae682f2d8efb7c546b360084f2ae5cae. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceConstants.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceConstants.kt
@@ -8,11 +8,11 @@ internal object GeofenceConstants {
 
     // Fallback for the movement trigger's OS geofence radius (meters). API field:
     // `local_refresh_trigger_radius`.
-    const val FALLBACK_LOCAL_REFRESH_RADIUS_METERS = 1_000f
+    const val FALLBACK_LOCAL_REFRESH_RADIUS_METERS = 3_000f
 
     // Fallback for the distance from the last API anchor at which an EXIT escalates
     // to a fresh API fetch (meters). API field: `remote_fetch_refresh_trigger_radius`.
-    const val FALLBACK_REMOTE_FETCH_RADIUS_METERS = 5_000f
+    const val FALLBACK_REMOTE_FETCH_RADIUS_METERS = 20_000f
 
     // Fallback for `maxBusinessGeofences` from the API config. Matches the
     // historical cap so customers aren't punished by a misconfigured backend.
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
@@ -6,6 +6,7 @@ import io.customer.geofence.api.GeofenceApiService
 import io.customer.geofence.api.toDomainConfig
 import io.customer.geofence.api.toDomainRegions
 import io.customer.geofence.store.GeofenceRegionStore
+import io.customer.location.LocationCoordinates
 import io.customer.sdk.core.util.Clock
 import io.customer.sdk.data.store.SecureUserStore
 import java.util.concurrent.atomic.AtomicBoolean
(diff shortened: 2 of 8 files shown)


mrehan27:
The PR has drifted a bit from the original scope, so I'm going to close it and open a few more focused PRs to address the changes separately.

PR #746, opened 2026-06-19
fix: scope in-app polling to process lifecycle instead of per-activity

mahmoud-elmorabea:
## Summary
Scopes in-app message queue **polling** to the process foreground lifecycle (`ProcessLifecycleOwner`) instead of per-activity `ON_RESUME`/`ON_PAUSE` callbacks, mirroring how `SseLifecycleManager` already works.

Previously, because a modal in-app runs in its own `GistModalActivity`, dismissing it resumed the host activity and triggered an immediate refetch — which re-displayed a message that had failed to load, in a tight retry loop. The same coupling also restarted polling on ordinary activity navigation. With this change a single polling timer survives activity navigation and modal display, and a failed-to-load message is retried only on the normal poll cadence (matching iOS and the web SDK).

Added `[Polling]` verification logs for foreground/background/poll-tick transitions.

## Test plan
- Unit: `:messaginginapp:testDebugUnitTest` green; new `GistSDKTest` cases for foreground-starts-polling, SSE-active-skips-polling, and not-foregrounded-skips-polling.
- Manual (emulator, java_layout): verified both polling and SSE modes — single timer on foreground, **no churn** when navigating/stacking activities, polling stops on background and resumes on foreground, and polling stays off entirely while SSE is active.

---

> [!NOTE]
> **Medium Risk**
> Changes when and how often the SDK fetches in-app messages (foreground/background and modal flows), which can affect delivery timing and duplicate-fetch behavior; mitigated by dedicated unit tests and alignment with existing SSE lifecycle patterns.
> 
> **Overview**
> Moves in-app message **polling** out of `GistSdk` into a new **`PollingLifecycleManager`**, wired as a DI singleton like `SseLifecycleManager`.
> 
> Polling now follows **`ProcessLifecycleOwner`** (`onStart` / `onStop`) instead of per-activity resume/pause (and the old `GistModalActivity` filter). One timer stays active across navigation and modal display, so closing a modal no longer forces an immediate queue refetch that could re-show a failed message in a tight loop.
> 
> `GistSdk` only holds references to the lifecycle managers and delegates `fetchInAppMessages()` / `reset()`. Dismiss middleware triggers **`pollingLifecycleManager.fetchInAppMessages()`** instead of the SDK. The manager still skips polling when SSE is active, synchronizes timer start/stop to avoid orphan timers, and avoids redundant restarts on poll-interval replay and SSE/identification churn. **`[Polling]`** debug logs were added.
> 
> New **`PollingLifecycleManagerTest`**; existing tests mock the manager instead of activity lifecycle for `GistSdk`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8bcf3c92c465390ad811081064cac6c19f30d0a4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/PollingLifecycleManager.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/PollingLifecycleManager.kt
@@ -105,6 +105,12 @@ internal class PollingLifecycleManager(
         resetTimer()
     }
 
+    // @Synchronized so the cancel-then-assign is atomic. startPolling is reachable concurrently
+    // (main-thread foreground handler, redux/event-bus fetchInAppMessages, and the attribute
+    // subscription coroutines); without serialization two calls could each create a Timer while
+    // only the last is retained in `timer`, orphaning the other so it keeps polling forever and
+    // resetTimer() can never cancel it.
+    @Synchronized
     private fun startPolling(duration: Long, initialDelay: Long = 0) {
         val currentState = state
         // Only skip polling if SSE should be used (both flag enabled AND user identified)
@@ -122,6 +128,7 @@ internal class PollingLifecycleManager(
         }
     }
 
+    @Synchronized
     private fun resetTimer() {
         timer?.cancel()
         timer = null
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/PollingLifecycleManager.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/PollingLifecycleManager.kt
@@ -1,0 +1,167 @@
+package io.customer.messaginginapp.gist.presentation
+
+import androidx.lifecycle.DefaultLifecycleObserver
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.LifecycleOwner
+import io.customer.messaginginapp.gist.data.listeners.GistQueue
+import io.customer.messaginginapp.state.InAppMessagingManager
+import io.customer.messaginginapp.state.InAppMessagingState
+import io.customer.sdk.core.util.HandlerMainThreadPoster
+import io.customer.sdk.core.util.Logger
+import io.customer.sdk.core.util.MainThreadPoster
+import java.util.Timer
+import java.util.concurrent.atomic.AtomicBoolean
+import kotlin.concurrent.timer
+
+/**
+ * Manages lifecycle-aware in-app message polling.
+ *
+ * Polling is scoped to the *process* foreground lifecycle, not individual activities, mirroring
+ * [SseLifecycleManager]. A single polling timer survives activity navigation and the display of
+ * our own [GistModalActivity], so dismissing a modal (normally or after a load failure) never
+ * triggers an immediate refetch. Polling stays disabled while SSE is active.
+ */
+internal class PollingLifecycleManager(
+    private val inAppMessagingManager: InAppMessagingManager,
+    processLifecycleOwner: LifecycleOwner,
+    private val gistQueue: GistQueue,
+    private val logger: Logger,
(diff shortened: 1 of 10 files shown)


messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/PollingLifecycleManager.kt
```diff
@@ -0,0 +1,167 @@
...
+                return@subscribeToAttribute
+            }
+
+            val currentState = state
+            if (currentState.shouldUseSse) {
+                return@subscribeToAttribute
+            }
+            logger.debug("[Polling] Poll interval changed to ${interval}ms - restarting polling")
+            startPolling(duration = interval, initialDelay = interval)
+        }
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Acknowledged, but resolving as won't-fix for this PR.

This is a pre-existing race, not introduced here: the prior `GistSdk` had the same pattern — a foreground catch-up fetch (`onActivityResumed` → `fetchInAppMessages(initialDelay = 0)`) alongside this same `pollInterval` subscription that restarts with `initialDelay = interval`. The extraction moved the code verbatim; behavior is unchanged.

It's also largely masked in practice: user identification independently triggers `fetchInAppMessages()` (another `initialDelay = 0` start), and the server lowers the poll interval via response headers shortly after the first fetch, producing a genuine change emission that restarts polling. The worst case (dropped catch-up) only delays the *first* poll and only when the initial replay emission wins the race against the catch-up timer.

If we want to harden this later, the clean fix is to have the `pollInterval` subscription ignore its initial replay emission (the initial start is owned by `handleForegrounded()`/`fetchInAppMessages()`) and only react to actual interval changes. Tracking that separately rather than expanding the scope of this lifecycle-scoping change.

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/PollingLifecycleManager.kt
```diff
@@ -0,0 +1,167 @@
...
+            return
+        }
+
+        val currentState = state
+        if (currentState.shouldUseSse) {
+            logger.debug("[Polling] $reason - SSE now active, stopping polling")
+            resetTimer()
+        } else {
+            logger.debug("[Polling] $reason - SSE not active, ensuring polling is running")
+            startPolling(duration = currentState.pollInterval)
```
mrehan27, line 191: Copying feedback from Claude, would be nice to validate this:

---

This starts polling with `initialDelay=0` (immediate fetch). On an anonymous→identified transition while SSE is disabled, this fires `fetchUserMessages()` at the same time as `ModuleMessagingInApp`'s `UserChangedEvent` handler, which also calls `fetchInAppMessages()` — so we get two near-simultaneous fetches.

The old code deliberately avoided this: its `isUserIdentified`/`sseEnabled` subscriptions only stopped polling ("Starting polling is handled by … event handlers"). Same redundant restart happens on an `sseEnabled` false→true flip while the user is anonymous (`shouldUseSse` stays false → pointless timer reset + fetch).

Suggest restarting here without the immediate fetch (use `initialDelay = currentState.pollInterval`, letting the event handlers own the catch-up fetch), or guarding `startPolling` so it doesn't restart when a timer with the same interval is already running.

mrehan27 (approved):
Changes look good overall. As I mentioned in slack earlier, it would be good to call out the slight behavior change so everyone is aware of it and aligned with expectations.

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/PollingLifecycleManager.kt
mahmoud-elmorabea: Good callout, fixed!

PR #748, opened 2026-06-19
chore(geofence): remove on-device location-send and add monitoring-distance cap

mrehan27:
[MBL-1843](https://linear.app/customerio/issue/MBL-1843/use-coarse-location-for-nearby-geofence-fetch-requests)

### Why
The on-device sync used to send a coarsened device location to the backend. Removing it entirely means the SDK has **no code path that transmits device location**, and lets the backend own "return the right set." Replaces #742 (closed; scope changed from gating fetch-all to removing location-send).

> ⚠️ **Held until backend ships return-all-without-location.** Today `/geofences/nearby` requires lat/long; this PR's no-location fetch depends on that change landing first.

### What
1. **Remove on-device location-send.** `fetchGeofences()` sends no location; dropped the 2-arg overload and `GeofenceCoordinateCoarsener`. `GeofenceSyncMode` stays as a single-value (`FETCH_ALL`) documented seam so re-introducing a location mode is a localized change + deliberate release (prior impl in git history).
2. **`maxMonitoringDistance` cap** (wire `max_monitoring_distance`). Caps how far a business geofence can be from the device to be registered. Semantics: `null` → 1000 km default (don't register another continent's geofence for a local user), `0` → disabled, below the trigger radius → fall back (avoids a dead-zone).
3. **Three-way empty fix.** Distinguishes "no geofences at all" (register nothing) from "geofences exist but all beyond the cap" (register the movement trigger only, so an EXIT re-ranks them in as the device approaches). Without this the cap is self-defeating.
4. **Config self-correction** at the wire→domain boundary: non-positive values fall back, positive out-of-range values clamp (local radius 100 m–5 km, remote-fetch expiry 1 min–7 d, dedupe 1 min–24 h). `geofences` remains the only required wire field (missing → parse fails → preserve cache + retry, fail-safe).

### Testing
- `:geofence` unit suite green (incl. new cap, three-way-empty, and coercion cases).
- apiCheck + ktlint green via pre-commit.

---

> [!NOTE]
> **High Risk**
> Changes core geofence sync, privacy behavior, and when remote vs local refresh runs; depends on backend supporting locationless `/geofences/nearby` before release.
> 
> **Overview**
> Stops sending device location on geofence API fetches: `fetchGeofences()` is parameterless with empty query params, while on-device lat/lng still drive ranking and anchors. **`GeofenceSyncMode.FETCH_ALL`** means movement no longer escalates to remote fetch based on distance from the last API anchor; **`refreshAction`** now keys remote vs local vs skip on time staleness, ranking staleness from **last movement-trigger registration** (not fetch anchor), and missing OS registrations.
> 
> Adds **`maxMonitoringDistance`** (`max_monitoring_distance`) so the distance filter can drop far regions; server config is coerced/clamped at parse time. Registration keeps the **movement trigger** when geofences exist but none are within the cap (trigger-only), vs truly empty sets that register nothing. Fallback radii and new clamp bounds are updated in constants.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 430288c67cd4861769d156d5626827b8a8c92717. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceConfig.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceConfig.kt
@@ -9,16 +9,24 @@ import kotlinx.serialization.Serializable
  */
 @Serializable
 internal data class GeofenceConfig(
+    // Radius (m) of the movement-trigger geofence; its EXIT re-ranks nearby business geofences.
     @SerialName("localRefreshTriggerRadius")
     val localRefreshTriggerRadius: Float,
+    // Distance (m) from the last API anchor at which an EXIT escalates to a fresh server fetch.
     @SerialName("remoteFetchRefreshTriggerRadius")
     val remoteFetchRefreshTriggerRadius: Float,
+    // Min interval (ms) between server fetches; a successful sync within this window is reused.
     @SerialName("remoteFetchRefreshExpiry")
     val remoteFetchRefreshExpiry: Long,
+    // Window (ms) for suppressing duplicate transition events for the same geofence.
     @SerialName("duplicateEventsExpiry")
     val duplicateEventsExpiry: Long,
+    // Max business geofences registered at once (the movement trigger is extra).
     @SerialName("maxBusinessGeofences")
-    val maxBusinessGeofences: Int
+    val maxBusinessGeofences: Int,
+    // Cap (m) on how far a business geofence can be from the device to be registered.
+    @SerialName("maxMonitoringDistance")
+    val maxMonitoringDistance: Float
 ) {
     internal companion object {
         // Used when the cached server config is missing. Today that's the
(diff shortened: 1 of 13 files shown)


geofence/src/main/kotlin/io/customer/geofence/GeofenceConfig.kt
```diff
@@ -9,16 +9,24 @@ import kotlinx.serialization.Serializable
...
+    // Window (ms) for suppressing duplicate transition events for the same geofence.
     @SerialName("duplicateEventsExpiry")
     val duplicateEventsExpiry: Long,
+    // Max business geofences registered at once (the movement trigger is extra).
     @SerialName("maxBusinessGeofences")
-    val maxBusinessGeofences: Int
+    val maxBusinessGeofences: Int,
+    // Cap (m) on how far a business geofence can be from the device to be registered.
+    @SerialName("maxMonitoringDistance")
+    val maxMonitoringDistance: Float
```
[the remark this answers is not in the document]
mrehan27: Not applicable — `:geofence` is unreleased, so no shipped SDK has a persisted `cached_config`. When it releases, `maxMonitoringDistance` ships with every other config field. Keeping it required for consistency.

geofence/src/main/kotlin/io/customer/geofence/GeofenceDistanceFilter.kt
```diff
@@ -1,21 +1,25 @@
...
         latitude: Double,
         longitude: Double,
-        max: Int
+        max: Int,
+        maxDistanceMeters: Float
     ): List<GeofenceRegion> {
         if (max <= 0 || regions.isEmpty()) return emptyList()
         return regions
+            .filter { it.distanceTo(latitude, longitude) <= maxDistanceMeters }
             .sortedBy { it.distanceTo(latitude, longitude) }
```
mahmoud-elmorabea: Maybe we can map to `distanceTo` so we do it once instead of twice here

mahmoud-elmorabea (approved):

geofence/src/main/kotlin/io/customer/geofence/GeofenceDistanceFilter.kt
mrehan27: Good call, updated

PR #757, opened 2026-06-25
chore(geofence): emit numeric geofenceId on transition events

mrehan27:
Backend prefers a numeric `geofenceId`. `toEventProperties()` now emits it as a number when the id parses, falling back to the raw string otherwise:

```kotlin
put("geofenceId", geofenceId.toLongOrNull() ?: geofenceId)
```

`Long` (not `Int`) so large backend ids don't silently fall back to a string; same JSON wire format either way. Both delivery paths handle it — worker `JSONObject` serializes the number unquoted, pipeline `sanitizeValue` passes it through to `JsonAnySerializer`.

Tests: numeric → `Long`, non-numeric → string fallback, plus a worker wire-format check asserting the body carries `"geofenceId":12345` unquoted.

No public API change.

[MBL-1760](https://linear.app/customerio/issue/MBL-1760/android-execute-geofencing-validation-and-lifecycle-testing)

---

> [!NOTE]
> **Low Risk**
> Small serialization tweak in geofence analytics properties with backward-compatible string fallback; no auth or public API changes.
> 
> **Overview**
> **Geofence Transition** event properties now send **`geofenceId` as a JSON number** when the stored id parses as a **`Long`**, instead of always sending a string. Non-numeric ids still use the raw string via `geofenceId.toLongOrNull() ?: geofenceId`, so both the worker direct-HTTP path and the foreground flush (which share `toEventProperties()`) stay aligned with backend expectations.
> 
> Tests cover numeric → `Long`, non-numeric fallback, and worker wire format (`"geofenceId":12345` unquoted). No public API change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit daddcf241d0cee6b30826c94f8342b60848fbe2c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/geofence/src/main/kotlin/io/customer/geofence/store/PendingGeofenceDelivery.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/store/PendingGeofenceDelivery.kt
@@ -36,7 +36,9 @@ internal data class PendingGeofenceDelivery(
      */
     fun toEventProperties(): Map<String, Any> = buildMap {
         put("transition", transition.name.lowercase())
-        put("geofenceId", geofenceId)
+        // Backend prefers a numeric geofenceId; emit a number when it parses, else the raw
+        // string. Long (not Int) so large backend ids don't silently fall back to a string.
+        put("geofenceId", geofenceId.toLongOrNull() ?: geofenceId)
         geofenceName?.let { put("geofenceName", it) }
     }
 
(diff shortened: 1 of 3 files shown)


mrehan27:
Closing — backend is adding string support for geofenceId, so the numeric coercion isn't needed. The feature branch already emits geofenceId as a string (from #755).

PR #761, opened 2026-06-30
feat(in-app): add dark mode support for in-app messages

BernardGatt:
https://github.com/user-attachments/assets/be6b31f6-4703-4005-97c2-9e911eefa582

## Summary
- Adds `ColorScheme` enum (`LIGHT`, `DARK`, `AUTO`) as a public API on `MessagingInAppModuleConfig`
- `AUTO` (the default) resolves the current system/app dark mode state and sends `"light"` or `"dark"` to the Gist renderer via the existing `postMessage` protocol
- Live theme changes are detected via `View.onConfigurationChanged()` and pushed to active WebViews through the `updateColorScheme` action — matching the same message format used by gist-web
- `GistModalActivity` now declares `android:configChanges="uiMode"` so modal messages survive dark mode toggles instead of being dismissed
- Color scheme flows through the Redux state store (`Initialize` action → `InAppMessagingState`) for consistency and testability
- Exposes `ModuleMessagingInApp.setColorScheme()` for runtime changes — the new value persists in the state store for the session and is pushed to any active WebViews immediately

## Usage
```kotlin
// Default: follows system dark mode automatically
MessagingInAppModuleConfig.Builder(siteId, region).build()

// Force light or dark at init
MessagingInAppModuleConfig.Builder(siteId, region)
    .setColorScheme(ColorScheme.LIGHT)
    .build()

// Change at runtime (persists for the session)
ModuleMessagingInApp.instance().setColorScheme(ColorScheme.DARK)
```

## Test plan
- [ ] Trigger an in-app message with system dark mode OFF → message renders in light mode
- [ ] Toggle system dark mode ON → newly triggered messages render in dark mode
- [ ] With `ColorScheme.LIGHT` configured → messages always render light regardless of system setting
- [ ] With `ColorScheme.DARK` configured → messages always render dark regardless of system setting
- [ ] Toggle dark mode while an in-app message is displayed → message updates live without being dismissed
- [ ] Call `setColorScheme()` at runtime → active and future messages reflect the new setting
- [ ] Existing unit tests pass

---

> [!NOTE]
> **Medium Risk**
> Touches the in-app WebView bridge and modal activity lifecycle; regressions could affect message display or theme sync, but changes are additive with defaults preserved.
> 
> **Overview**
> Adds public **`ColorScheme`** (`LIGHT`, `DARK`, `AUTO`) on **`MessagingInAppModuleConfig`** (default `AUTO`) and **`ModuleMessagingInApp.setColorScheme()`** for runtime updates. The choice is stored in Redux (`Initialize` / `SetColorScheme`) and passed into **`EngineWebConfiguration`** as resolved `"light"` / `"dark"` for the Gist WebView.
> 
> **`EngineWebView`** pushes live updates via `updateColorScheme` (`MessageEvent` with `action: updateColorScheme`), subscribes to state changes, and reacts to **`onConfigurationChanged`** when mode is `AUTO`. **`GistModalActivity`** declares **`android:configChanges="uiMode"`** so modals are not recreated on theme toggles. Unit tests assert **`colorScheme`** on web configuration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c0f7be20cb3a076b8c7e3c8cf511142cd8462672. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
@@ -141,7 +141,10 @@ internal class EngineWebView @JvmOverloads constructor(
     override fun updateColorScheme(scheme: String) {
         lastResolvedColorScheme = scheme
         val json = Gson().toJson(mapOf("action" to "updateColorScheme", "colorScheme" to scheme))
-        webView?.evaluateJavascript("window.postMessage($json, '*');", null)
+        webView?.evaluateJavascript(
+            "window.dispatchEvent(new MessageEvent('message', { data: $json }));",
+            null
+        )
     }
 
     override fun onConfigurationChanged(newConfig: Configuration) {
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
@@ -170,8 +170,10 @@ internal class EngineWebView @JvmOverloads constructor(
             selector = { it.colorScheme }
         ) { colorScheme ->
             val resolved = colorScheme.resolve(context.resources.configuration.uiMode)
-            if (resolved != lastResolvedColorScheme) {
-                post { updateColorScheme(resolved) }
+            post {
+                if (resolved != lastResolvedColorScheme) {
+                    updateColorScheme(resolved)
+                }
             }
         }
     }
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
@@ -135,6 +135,8 @@ internal class EngineWebView @JvmOverloads constructor(
 
     override fun stopLoading() {
         webView?.stopLoading()
+        colorSchemeJob?.cancel()
+        colorSchemeJob = null
         // remove lifecycle observer to stop receiving further lifecycle events
         onLifecyclePaused()
         viewLifecycleOwner?.removeObserver(this)
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
@@ -157,7 +157,8 @@ internal class EngineWebView @JvmOverloads constructor(
 
     @SuppressLint("SetJavaScriptEnabled")
     override fun setup(configuration: EngineWebConfiguration) {
-        lastResolvedColorScheme = configuration.colorScheme
+        val initialColorScheme = configuration.colorScheme
+        lastResolvedColorScheme = initialColorScheme
         setupTimeout()
         elapsedTimer.start("Engine render for message: ${configuration.messageId}")
         val messageData = mapOf("options" to configuration)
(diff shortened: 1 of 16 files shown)


messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
```diff
@@ -128,15 +135,52 @@ internal class EngineWebView @JvmOverloads constructor(
...
+    }
+
+    private fun subscribeToColorSchemeChanges() {
+        colorSchemeJob?.cancel()
+        colorSchemeJob = inAppMessagingManager.subscribeToAttribute(
+            selector = { it.colorScheme }
+        ) { colorScheme ->
+            val resolved = colorScheme.resolve(context.resources.configuration.uiMode)
+            if (resolved != lastResolvedColorScheme) {
+                post { updateColorScheme(resolved) }
```
mahmoud-elmorabea: We should post the whole code block, that subscription could return on a different thread, posting the whole block (check + update) avoid a race condition between the config change / manual setter and a state update

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
```diff
@@ -128,15 +135,52 @@ internal class EngineWebView @JvmOverloads constructor(
...
+    override fun updateColorScheme(scheme: String) {
+        lastResolvedColorScheme = scheme
+        val json = Gson().toJson(mapOf("action" to "updateColorScheme", "colorScheme" to scheme))
+        webView?.evaluateJavascript(
+            "window.dispatchEvent(new MessageEvent('message', { data: $json }));",
+            null
+        )
+    }
+
+    override fun onConfigurationChanged(newConfig: Configuration) {
```
mahmoud-elmorabea, line 156: I see that we updated the config changes for `gist.presentation.GistModalActivity`, did we test or confirm inline views rendered in different activities that don't have that config setup will work?

messaginginapp/src/main/java/io/customer/messaginginapp/ui/controller/InAppMessageViewController.kt
```diff
@@ -100,14 +100,18 @@ internal abstract class InAppMessageViewController<ViewCallback : InAppMessageVi
     @UiThread
     internal fun loadMessage(message: Message) {
         val store = inAppMessagingManager.getCurrentState()
+        val viewContext = engineWebViewDelegate?.getView()?.context
+        val uiMode = (viewContext ?: SDKComponent.android().applicationContext)
+            .resources.configuration.uiMode
```
mahmoud-elmorabea: I think we should defer this down the line where we are sure to have view context available, instead of falling back to resolving ui config from a non-activity context

mahmoud-elmorabea (approved):
Looks good overall 💯 

Left few suggestions and the big question is does this work for inline?

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
BernardGatt: Fixed: moved the lastResolvedColorScheme check inside post {} so the read and update are both on the main thread.

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
BernardGatt: If the host activity doesn't declare `configChanges="uiMode"`, the activity is recreated on dark mode toggle.

The old EngineWebView is destroyed and a new one is created via loadMessage(), which reads the current colorScheme from the state and resolves it with the new configuration's uiMode. So inline views get the correct theme after recreation.

messaginginapp/src/main/java/io/customer/messaginginapp/ui/controller/InAppMessageViewController.kt
BernardGatt: Fixed: removed the fallback entirely. loadMessage() is @UiThread and always called after attachEngineWebView(), so the view context is always available. Falls back to 0 (resolves to light) only if the view is somehow null.

PR #750, opened 2026-06-23
chore(geofence): transition event shape updates + reboot re-registration

mrehan27:
[MBL-1760](https://linear.app/customerio/issue/MBL-1760/android-execute-geofencing-validation-and-lifecycle-testing)

### Why
Findings from geofencing validation + lifecycle testing: align the transition event shape with the backend/iOS, and close a reboot gap where business geofences could silently stop being monitored.

### What
**Transition event shape**
- Rename tracked events to `geofence_entered` / `geofence_exited` (applied in both delivery paths: analytics pipeline + WorkManager direct-HTTP).
- Add `geofence_name` to event properties — resolved once at the receiver and carried via the entry + WorkManager `inputData`, so both delivery paths agree on the name (and it's captured at transition time, like the `userId`/`timestamp` snapshots). Omitted when the region isn't cached.
- Migrate the geofence module to the shared `io.customer.android.publish-module` plugin (`artifactId "geofence"`).

**Reboot re-registration**
- GMS drops all geofences on reboot; only `GeofenceBootReceiver` re-registers them, but `BOOT_COMPLETED` isn't always delivered (stopped state, OEM battery managers, emulator). Stamp device uptime (`Clock.elapsedRealtime`) on each successful registration; when a later refresh sees a smaller uptime, the device rebooted since — treat OS state as empty and re-register all business geofences instead of trusting the surviving `registeredIds`.

### Testing
- `:geofence` + `:datapipelines` unit suites green; apiCheck + ktlint via pre-commit.
- On-device journey verified (enter/exit firing + exactly-once delivery). Reboot path under on-device validation.

---

> [!NOTE]
> **Medium Risk**
> Renaming geofence track events is a breaking change for existing Journeys/automation; reboot and re-registration logic touches core geofence sync but is covered by new unit tests.
> 
> **Overview**
> Aligns geofence transition analytics with backend/iOS and fixes monitoring gaps after reboot when `BOOT_COMPLETED` is missed.
> 
> **Transition events** rename tracked names from `CIO Geofence Entered` / `CIO Geofence Exited` to **`geofence_entered`** / **`geofence_exited`** (datapipelines + WorkManager direct HTTP). **`geofence_name`** is snapshotted at the receiver from the cached region, carried on `PendingGeofenceDelivery` and WorkManager `inputData`, and included in properties when known (omitted if uncached). The geofence module **`build.gradle`** switches to the shared **`io.customer.android.publish-module`** plugin.
> 
[...]

PR #732, opened 2026-06-12
chore: move on-device geofence implementation into :geofence module

mrehan27:
**Linear:** [MBL-1781](https://linear.app/customerio/issue/MBL-1781/android-extract-geofencing-into-a-dedicated-module)

**Stack** (merge bottom-up into `feature/geofence-on-device`):
- PR 1 — scaffold `:geofence` module (#730)
- PR 2 — publish location updates via EventBus (#731)
- **PR 3 — move geofence implementation into `:geofence` 👈 this PR**

Lifts geofence code out of `:location` into `:geofence`. Registering `ModuleLocation` alone no longer pulls in geofence machinery, the Play Services geofencing dependency, or the receiver manifest entries.

**Mostly `git mv`** (history-preserving renames of `io.customer.location.geofence.*` → `io.customer.geofence.*`). Reviewer focus is the small non-move delta:
- Geofence subscriptions (LocationAcquired / UserChanged / Reset / defensive-sync / foreground-flush) move from `ModuleLocation.initialize()` to `ModuleGeofence.initialize()`, verbatim.
- `LocationServices.getLastKnownLocation(): LocationCoordinates?` added (`@InternalCustomerIOApi`, hidden from customer `.api`); `LocationCoordinates` promoted internal→public.
- Foreground-flush extracted from `LocationLifecycleObserver` into new `GeofenceLifecycleObserver`.
- `ModuleGeofence` errors + bails if `ModuleLocation` isn't registered.
- Both sample apps depend on `:geofence` and register `ModuleGeofence`.

Behavior unchanged for customers registering both modules; location-only customers now correctly get tracking without geofencing.

---

> [!NOTE]
> **Medium Risk**
> Large modular split across location/geofence init order, event delivery, and OS receivers; intended behavior is unchanged when both modules are registered, but misconfiguration (geofence without location) or location-only integrators now get a different dependency surface.
> 
> **Overview**
> **Moves on-device geofencing out of `:location` into the dedicated `:geofence` artifact** (`io.customer.location.geofence.*` → `io.customer.geofence.*`), including manifest receivers and Play Services / WorkManager dependencies on the geofence side only.
> 
[...]

PR #730, opened 2026-06-12
chore(geofence): scaffold :geofence Gradle module

mrehan27:
**Linear:** [MBL-1781](https://linear.app/customerio/issue/MBL-1781/android-extract-geofencing-into-a-dedicated-module)

**Stack** (merge bottom-up into `feature/geofence-on-device`):
- **PR 1 — scaffold `:geofence` module 👈 this PR**
- PR 2 — publish location updates via EventBus (#731)
- PR 3 — move geofence implementation into `:geofence` (#732)

Adds an empty `:geofence` module (artifact `io.customer.android:geofence`) so the gradle wiring, manifest, proguard config, and publish-artifact name land independently of the code move. The module compiles in isolation; `ModuleGeofence.initialize()` is a no-op. Customers should not register it yet — the follow-up PRs move the implementation in and make it the registration point.

No behavior change to any existing module.

---

> [!NOTE]
> **Low Risk**
> Build-only scaffolding with empty module initialization; no changes to existing modules or customer-facing behavior.
> 
> **Overview**
> Introduces a new **`:geofence`** Android library (`io.customer.android:geofence`) as the first step in splitting geofencing out of `:location`. **`settings.gradle`** includes the module; Gradle applies the same publish, Android, codecov, and testing scripts as other SDK libraries, with dependencies on `:base`, `:core`, and `:location` plus Play location, lifecycle, coroutines, and WorkManager.
> 
> The public surface is a **`CustomerIOModule`** scaffold: **`GeofenceModuleConfig`** (empty builder for future options) and **`ModuleGeofence`** with a **no-op `initialize()`**—documented as not for customer registration until a follow-up moves implementation from `:location`. The manifest stays empty of location permissions (documented for host apps); standard ProGuard/API stubs are included.
> 
> **No runtime or behavioral changes** to existing modules.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 08db2b8c4e48d076ab77cf7ceeed9734f31beeb6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #728, opened 2026-06-11
chore(sample): port iOS APN sample's grant-background-location UX

mrehan27:
## Summary

Port the iOS APN UIKit sample's location-test screen UX 1:1 so the Android sample behaves the same way for geofence permission setup. Sample-app only — no SDK changes.

### What changed
- **BACKGROUND LOCATION section** moved to the top of the screen (above OPTION 1: Quick Presets). Matches iOS section order. `OPTION X:` prefixes preserved on the other sections.
- **Adaptive button + status label** with four states (driven by a `BgPermissionState` machine in `LocationTestActivity`):
  | State | Button | Status |
  |---|---|---|
  | Not determined | `Grant location access` | `Not determined. Tap to grant foreground access first.` |
  | Foreground-only | `Allow all the time` | `Foreground granted. Background geofence delivery requires "Allow all the time".` |
  | Background granted | `Background — granted` (disabled, dimmed) | `Background granted. Background geofence delivery is enabled.` |
  | Denied | `Open Settings` | `Denied. Enable location access in Settings.` |
- **`onResume` refresh** — so returning from Settings re-checks state without needing a re-tap. Mirrors iOS's `locationManagerDidChangeAuthorization`.
- **Rationale dialog** now offers `Cancel` / `Continue` / `Open Settings`, with the dialog title matching iOS (`Allow background location?`). The Settings action covers API 30+ where the OS silently denies the background prompt instead of showing a dialog.
- **Strings reworked** to mirror iOS wording with Android terminology (`"Allow all the time"` instead of `"Always"`, `"Android 11+"` instead of `"iOS 14+"`, etc.). Removed three now-redundant snackbar strings.
- **Triggers `requestLocationUpdate()` after a successful runtime grant** so the SDK fetches a fresh fix and registers geofences immediately. Without this, the SDK's `ON_APP_START` lifecycle hook has already run (at process start, before permission was granted) and geofencing wouldn't start until the next app launch — surprising for customers following this flow. Demonstrates the integration pattern customers should follow in their own apps.

### iOS reference
- [`LocationTestViewController+Location.swift`](https://github.com/customerio/customerio-ios/blob/main/Apps/APN-UIKit/APN%20UIKit/View/location/LocationTestViewController+Location.swift) — state machine + tap handlers
[...]

PR #763, opened 2026-06-30
chore(geofence): add support for geofencing feature

mrehan27:
### Summary

This PR adds support for the upcoming geofence feature, merging `feature/geofence-on-device` into `main`.

### Highlights

- **On-device geofence monitoring** — registers nearby geofences and tracks enter/exit transitions through the OS.
- **Movement-based tiered refresh** — re-ranks and re-fetches nearby geofences as the user moves (EXIT-triggered), within the OS geofence budget.
- **Reliable delivery** — transitions delivered via WorkManager (survives process death) with a foreground-flush fallback, arbitrated through a shared pending-delivery store; duplicate transitions suppressed by a cooldown filter.
- **Lifecycle resilience** — restores geofences after reboot, re-fetches on login, and snapshots `userId` per transition to prevent cross-user attribution.
- **New `:geofence` Gradle module** with the on-device implementation split out of `:location`.

### Changes

Includes previously approved/merged PRs (oldest first):

- #674
- #673
- #683
- #684
- #686
- #687
- #691
- #692
- #697
- #698
- #699
- #702
- #703
- #704
- #705
- #706
- #707
- #708
- #709
- #710
- #711
- #714
- #716
- #717
- #720
- #727
- #728
- #730
- #731
- #732
- #739
- #740
- #748
- #750
- #754
- #755
- #762
- #764
- #768
- #773
- #774
- #775
- #777
- #778
- #785
- #786
- #787
- #791
- #795
- #793
- #796
- #798

---

> [!NOTE]
> **High Risk**
> Large surface area touching background location, encrypted identity, disk-backed delivery races, and cross-module event attribution; regressions could drop transitions, duplicate events, or mis-attribute users after sign-out.
> 
> **Overview**
> Introduces **on-device geofencing** as a new `:geofence` Gradle module (with `:location` and `:core` dependencies), wired into CI lint/unit-test matrices alongside `:location`.
> 
> **Core platform changes** support the feature across modules: new event-bus types (`LocationAcquired`, `GeofenceTransitionEvent` with pinned `userId` and transition timestamp), `geofenceScope`, injectable `Clock`, shared **ISO-8601** timestamp formatting (replacing datapipelines’ `SegmentInstantFormatter`), **GET + query** support on `CustomerIOHttpClient`, and **`SecureUserStore`** (encrypted user id for workers/receivers). Pending delivery is generalized with atomic disk writes, `appendAll`, **`PendingDeliveryFlusher`** (at-most-once vs at-least-once), and **`claimSendRestore` / `sendRemoveOnSuccess`** for push/geofence arbitration. **CustomerIO** (datapipelines) persists/clears secure user id on identify/reset, subscribes to geofence transitions, and tracks **`Geofence Transition`** with attribution and timestamp enrichment.
> 
> The **geofence module** registers nearest workspace fences via GMS (movement-trigger tiered local re-rank vs remote fetch), handles transitions in broadcast/boot receivers (identified users only, cooldown dedupe, per-geoset fan-out), schedules **WorkManager** HTTP delivery, and flushes pending rows on foreground via the analytics pipeline. Sign-out resets OS registrations and user-scoped state while preserving workspace cache; reboot/app-update paths re-register when OS state is wiped.
> 
> A **manual test plan** documents transition delivery invariants (`transitionId` dedupe, offline → foreground flush).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b88b47f4d46ab481d247f75852e116389e2af71f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
```diff
@@ -0,0 +1,190 @@
...
+            val transition = when (gmsTransitionType) {
+                Geofence.GEOFENCE_TRANSITION_ENTER -> Event.GeofenceTransition.ENTER
+                Geofence.GEOFENCE_TRANSITION_EXIT -> Event.GeofenceTransition.EXIT
+                else -> {
+                    logger.logUnknownTransition(gmsTransitionType)
+                    return@forEach
+                }
+            }
+
+            if (!cooldownFilter.tryAcquire(geofenceId, transition)) {
```
Shahroz16: The cooldown is acquired before `appendAll`, but `PendingDeliveryStore.writeAll` logs and absorbs failures. If that write fails, there is no pending row to retry and this cooldown suppresses the same transition for its full window. Could we persist successfully before spending the cooldown, or release it when persistence fails?

core/src/main/kotlin/io/customer/sdk/data/store/PendingDeliveryFlusher.kt
```diff
@@ -0,0 +1,104 @@
...
+                            workManager.cancelUniqueWork(entry.key).await()
+                            callbacks.onWorkCancelled(entry)
+                        }
+                        // Atomically claim before publishing so this flush delivers
+                        // each entry at most once (and a worker that also claims —
+                        // push — can't deliver the same entry). Claiming per-entry
+                        // here, rather than batching a remove at the end, also means a
+                        // mid-loop CancellationException can't strand an already-
+                        // published entry for re-publish next flush.
+                        if (!store.claim(entry.key)) return@forEach
```
Shahroz16: Could we keep the pending row until the foreground hand-off is durable? `claim` removes it before `publish`, and `publish` is an in-memory `Channel.trySend`. If the process dies after this claim, or after WorkManager cancellation but before Analytics persists the event, neither delivery path owns the row anymore. The stable transition ID makes an at-least-once hand-off safe here.

geofence/src/main/kotlin/io/customer/geofence/GeofenceServices.kt
```diff
@@ -0,0 +1,175 @@
...
+        // clears it but runs on `scope`, so an in-process re-login (ResetEvent then
+        // UserChangedEvent) would read the previous user's center and rank the new
+        // user's geofences around it. Clearing it now makes the next identify fall
+        // back to a no-location skip and acquire a fresh fix instead.
+        regionStore.clearLastMovementTriggerLocation()
+        // Snapshot the userId synchronously — the datapipelines `ResetEvent`
+        // subscriber races to clear `secureUserStore`, so a deferred read
+        // inside `reset()` could see null and misclassify a normal sign-out
+        // as a re-login.
+        val signedOutUserId = secureUserStore.getUserId()
```
Shahroz16: Quick ordering question: `onUserSignedOut` runs from an asynchronous `ResetEvent` subscriber. A reset followed immediately by identify for user B could let this read B rather than signed-out A, and then reset B’s state. Do we have an ordering guarantee that rules this out? A regression test for reset followed by identify would make the intended behavior clear.

Shahroz16 (commented):
I have two delivery-reliability details and one question about the reset/re-identify ordering that i would love to know before 🟢

core/src/main/kotlin/io/customer/sdk/communication/Event.kt
```diff
@@ -50,5 +50,40 @@ sealed class Event {
...
+    }
+
+    /**
+     * Event published by the Location module when a geofence transition is received from the OS.
+     * Subscribers (e.g. data pipelines) translate this into a tracked event.
+     *
+     * [userId] is snapshotted at queue time, not the SDK's current identity — non-null means the
+     * subscriber must attribute the resulting track event to this userId (not whoever is identified
+     * at flush time), so a sign-out + sign-in between queue and delivery cannot reattribute. Null
+     * means anonymous at queue time; the subscriber falls back to the pipeline's anonymousId path.
```
Shahroz16, line 76: While tracing null-user delivery through the current backend, I found a contract mismatch here. This path publishes a Geofence Transition using the pipeline anonymous_id. The Customer.io destination maps that to Track v2 action: geofence, but Edge deliberately rejects anonymous geofence requests with HTTP 400. Can we align which side owns this? If geofencing is identified-only, this fallback should not publish and remove the row as a valid anonymous event. If anonymous geofencing is intended, Edge and its integration tests need to support it.

geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
mrehan27: Fixed on the feature branch: the fan-out now lives in `GeofenceTransitionEmitter`, which checks `appendAll`'s result and releases the cooldown when the persist fails (test: `emit_givenPersistFails_expectCooldownReleasedAndFalse`). The store write itself also became atomic temp+rename with the result propagated instead of absorbed (#778).

core/src/main/kotlin/io/customer/sdk/data/store/PendingDeliveryFlusher.kt
mrehan27: Fixed in #777: the geofence flush now runs `PendingDeliveryFlusher` in `AT_LEAST_ONCE` mode — publish first, remove the row only after publish returns, WorkManager cancel best-effort after that. A crash before publish keeps the row for the worker or the next flush; the duplicate window that ordering can create is deduped backend-side via the stable transitionId, exactly the at-least-once hand-off you describe.

geofence/src/main/kotlin/io/customer/geofence/GeofenceServices.kt
mrehan27: The ordering guarantee lives in `GeofenceRepository.reset()` rather than the subscriber: `clearIdentify()` clears the user store synchronously before `ResetEvent` is published, so when the async reset finally runs, a non-null user can only mean a newer sign-in — it skips the wipe instead of resetting B's state. Regression tests: `reset_givenUserSignedInAtResetTime_expectSkipWipe` and `refresh_givenIdentifyDuringInFlightResetWipe_expectDecisionWaitsAndRefreshes`. Residual: the subscriber's synchronous flag/anchor clears can still hit B, but that only downgrades B's next refresh to a fresh-fix acquisition (self-heals, no state loss).

core/src/main/kotlin/io/customer/sdk/communication/Event.kt
mrehan27: Alignment: geofencing is identified-only on the SDK side. The receiver drops anonymous transitions before persisting, so every stored row carries a snapshotted userId and this fallback is unreachable through live publishers today; the worker additionally defers null rows defensively. A small follow-up on the feature branch makes the contract explicit end-to-end: the subscriber drops a null-snapshot event (no anonymousId fallback, no attribution to the currently identified user) and the `GeofenceTransitionEvent` kdoc states identified-only — so nothing on our side can ever produce the anonymous track Edge rejects.

location/src/main/kotlin/io/customer/location/LocationServicesImpl.kt
```diff
@@ -43,12 +44,21 @@ internal class LocationServicesImpl(
...
+    }
+
+    @OptIn(InternalCustomerIOApi::class)
+    override fun requestLocationUpdateSilently() {
+        launchLocationRequest(orchestrator::requestLocationUpdateSilently)
+    }
+
+    private fun launchLocationRequest(request: suspend () -> Unit) {
         // If a request is already in flight, ignore the new call
         if (currentLocationJob?.isActive == true) return
```
Shahroz16: Could we preserve the tracked intent when the silent geofence request overlaps the normal location request? `UserChangedEvent` is collected independently by Location and Geofence, so on a first identified session the `ON_APP_START` path and automatic geofence bootstrap can both reach this slot. I reproduced the silent-first order in a focused test: this guard drops `requestLocationUpdate()`, the result goes through `onLocationReceivedWithoutTracking`, and no later retry emits the `CIO Location Update` for that process. Could a tracked request upgrade an in-flight silent request, or be queued to consume the same fix?

Shahroz16 (approved):

location/src/main/kotlin/io/customer/location/LocationServicesImpl.kt
mrehan27: Confirmed — reproduced exactly as you described, and Bugbot independently flagged the same slot. Fixed in #793: a tracked request now upgrades the in-flight intent so the same fix serves both consumers (routing decided at delivery, OFF gate re-applied there so an upgrade can't make a disabled mode emit analytics). Same semantics as the iOS fix (customerio-ios#1173).

location/src/main/kotlin/io/customer/location/LocationServicesImpl.kt
```diff
@@ -43,12 +44,21 @@ internal class LocationServicesImpl(
...
     override fun requestLocationUpdate() {
+        launchLocationRequest(orchestrator::requestLocationUpdate)
+    }
+
+    @OptIn(InternalCustomerIOApi::class)
+    override fun requestLocationUpdateSilently() {
+        launchLocationRequest(orchestrator::requestLocationUpdateSilently)
+    }
+
+    private fun launchLocationRequest(request: suspend () -> Unit) {
```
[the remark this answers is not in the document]
mrehan27: Valid — this matches Shahroz's earlier report on this file. Fixed in #793 via an intent upgrade: the tracked request promotes the in-flight silent fetch instead of being dropped, so one fix serves both callers and the CIO Location Update is emitted.

location/src/main/kotlin/io/customer/location/LocationServicesImpl.kt
mrehan27: Fixed in #793 (merged): a tracked request arriving while a silent fetch is in flight now upgrades the shared intent atomically and consumes the same fix — routing is decided at delivery, with the OFF gate re-checked there. A follow-up commit also closed the tail window: the intent is claimed atomically with the routing decision and closed on every exit path, so an upgrade landing after the fix was handed off is rejected and the caller starts its own fetch instead of being dropped. Both orders of your repro are pinned by tests.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
```diff
@@ -0,0 +1,483 @@
...
+     */
+    private fun osStateWipedByReboot(): Boolean =
+        store.getLastRegistrationUptime()?.let { clock.elapsedRealtime() < it } ?: false
+
+    @RequiresPermission(Manifest.permission.ACCESS_FINE_LOCATION)
+    override suspend fun handleMovement(latitude: Double, longitude: Double): Result<Unit> {
+        if (!refreshInProgress.compareAndSet(false, true)) {
+            logger.logSyncSkipped("refresh already in progress")
+            return Result.success(Unit)
+        }
```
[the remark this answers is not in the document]
mrehan27: Valid and triaged post-release: the loss is bounded — the movement EXIT that loses the gate is the same condition `refresh()` re-derives on the next identify/app-launch (ranking-staleness check), so it self-corrects on the next app open, and the planned shared pending-trigger latch will close it properly. Tracked for the next iteration rather than this release.

geofence/src/main/kotlin/io/customer/geofence/GeofenceServices.kt
```diff
@@ -0,0 +1,178 @@
...
+        // Act on a host-initiated refresh or the rising edge of a no-location skip;
+        // otherwise this becomes a per-update refresh storm on hosts that stream
+        // locations. Clear both flags so a single fix is consumed once.
+        val requested = explicitRefreshRequested.compareAndSet(true, false)
+        val rearmed = lastSkippedForNoLocation.compareAndSet(true, false)
+        if (!requested && !rearmed) return
+        val userId = secureUserStore.getUserId()
+        // No user yet: skip; a later identify re-triggers.
+        if (userId.isNullOrEmpty()) return
+        onUserIdentified(latitude, longitude)
```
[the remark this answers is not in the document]
mrehan27: By design: `refreshFromCurrentLocation` feeds the SDK a location (its primary role in MANUAL mode), it doesn't force a server fetch. The arriving fix goes through the normal tier decision — >5 km from the fetch anchor → remote fetch, >1 km from the last registration → local re-rank, otherwise SKIP because the registered set is already correct for that position. So the host refresh no-ops only when acting on it would change nothing.

geofence/src/main/kotlin/io/customer/geofence/GeofenceLifecycleObserver.kt
```diff
@@ -0,0 +1,57 @@
...
+                    logger.logForegroundFlushPublished(entry.geofenceId, entry.transition.name)
+                override fun onEntryFailed(entry: PendingGeofenceDelivery, cause: Throwable) =
+                    logger.logForegroundFlushEntryFailed(entry.geofenceId, entry.transition.name, cause.message)
+                override fun onComplete(count: Int) = logger.logForegroundFlushComplete(count)
+            },
+            guarantee = PendingDeliveryFlusher.DeliveryGuarantee.AT_LEAST_ONCE
+        ) { entry ->
+            eventBus.publish(
+                entry.withFreshestEventData(regionStore.getCachedRegion(entry.geofenceId)).toGeofenceTransitionEvent()
+            )
```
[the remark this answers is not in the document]
mrehan27: Unreachable via current write paths: the receiver drops anonymous transitions before persisting or spending a cooldown slot (geofencing is identified-only), so a row with an empty snapshotted userId can't be created — the worker's skip is defense-in-depth for exactly that reason, and no earlier SDK version shipped that could have left legacy rows. The asymmetry is fair to call out though: the foreground flush will get the same defensive guard as the worker in a follow-up.

core/src/main/kotlin/io/customer/sdk/data/store/PendingDeliveryClaim.kt
```diff
@@ -0,0 +1,108 @@
...
+    val result = send()
+    return when {
+        result.isSuccess -> PendingDeliveryResult.Delivered
+        isRetryable(result.exceptionOrNull()) -> {
+            append(entry)
+            PendingDeliveryResult.Retryable(result.exceptionOrNull())
+        }
+        else -> {
+            append(entry)
+            PendingDeliveryResult.Failed(result.exceptionOrNull())
```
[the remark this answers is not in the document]
mrehan27: Valid — losing the row needs a failed send and a failed atomic write in the same moment, so it's bounded to one transition under a double failure. Follow-up will check the append result and log loudly (optionally retry once), in line with the claim-durability hardening this path already got.

PR #764, opened 2026-07-01
chore(geofence): send device location for nearby geofence fetch requests

mrehan27:
## What

Re-introduce a location-based geofence sync (`GeofenceSyncMode.NEARBY`) alongside `FETCH_ALL`, and make it the default. NEARBY sends the device location to `/geofences/nearby` so the backend returns the set ranked around the device, and re-fetches once the device moves beyond `remoteFetchRefreshTriggerRadius` from the last fetch anchor. FETCH_ALL is unchanged (sends no location; backend returns the full capped set).

The device location is sent **as-is — there is no coarsening**. The `/geofences/nearby` request carries no user identity, so the transmitted coordinate can't be attributed to a user; snapping it to a grid added no privacy benefit.

## Why

- Nearby sync scales to workspaces with many geofences (backend ranks around the device) instead of shipping the whole set.
- The mode is a **compile-time** choice, never a runtime/server toggle — whether location is transmitted only changes in a deliberate SDK release.

## Notes

- No public API change (all `internal`).
- Both modes fully functional; existing FETCH_ALL behavior and tests unchanged.
- iOS parity: same NEARBY / FETCH_ALL sync logic.

## Linear

https://linear.app/customerio/issue/MBL-1945/use-coarse-location-for-nearby-geofence-fetch-requests

---

> [!NOTE]
> **Medium Risk**
> Default SDK behavior now transmits device coordinates on geofence API calls and changes when remote sync runs; changes are internal but affect privacy-sensitive location handling and sync semantics for all apps on this release.
> 
> **Overview**
> Re-introduces **`GeofenceSyncMode.NEARBY`** next to **`FETCH_ALL`** and switches the shipped default to **NEARBY** via `GeofenceSyncMode.active`.
> 
> In **NEARBY**, remote geofence sync sends the device’s latitude/longitude as query params on `GET /geofences/nearby` so the backend returns a location-ranked set; **`refresh`** and **`handleMovement`** trigger another remote fetch when the device moves beyond `remoteFetchRefreshTriggerRadius` from the last API anchor. **FETCH_ALL** is unchanged: no coordinates on the wire and movement never forces a re-fetch.
> 
> `GeofenceApiService.fetchGeofences` now takes an optional `GeofenceLocation?`; the repository passes coordinates only when sync mode is NEARBY. Tests cover wire format, mode-specific refresh/movement behavior, and the new default.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit db2874c122681345035f605c843650f06180d841. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


geofence/src/test/java/io/customer/geofence/GeofenceLocationTest.kt
```diff
@@ -0,0 +1,31 @@
...
+    @Test
+    fun coarsened_givenHighLatitude_expectLongitudeGridWidensToStayUniform() {
+        // The point of the grid over decimal rounding: at 60° (cos ≈ 0.5) a degree of longitude is
+        // ~half the ground distance, so the longitude grid roughly doubles in degrees to keep the
+        // ~500 m floor. Decimal rounding would instead over-refine longitude here.
+        val original = GeofenceLocation(latitude = 60.0, longitude = 10.123456)
+
+        val coarse = original.coarsened(500.0)
+
+        Math.abs(coarse.longitude - original.longitude) shouldBeLessThan 0.0090 // ~half of the ~0.009° cell
```
mahmoud-elmorabea: I think we should have an exact value here or an upper AND lower bound, I think if the `coarsened` logic isn't there or changed this test would remain green

mahmoud-elmorabea (approved):

geofence/src/test/java/io/customer/geofence/GeofenceLocationTest.kt
mrehan27: Good call. We decided not to use coarse location since the API doesn't send any user specific information. So we reverted the coarse location for now as this is no longer needed.

PR #768, opened 2026-07-02
chore(geofence): acquire location without emitting analytics events

mrehan27:
## What

Let hosts use geofencing without emitting "CIO Location Update" analytics events or enriching identify context — geofencing acquires the location it needs through an internal, analytics-free path.

- New internal `LocationServices.requestLocationUpdateSilently()` (`@InternalCustomerIOApi`): one-shot fix that publishes `LocationAcquired` and updates `lastKnownLocation` only — no track event, no persistence, not added to identify context. Requires location permission, but fetches regardless of tracking mode (geofencing needs a fix even when tracking is OFF).
- `LocationTracker` location split into `trackedLocation` (analytics / identify / persisted) and `lastKnownLocation` (any source incl. silent; backs `getLastKnownLocation`).
- New `GeofenceLocationMode { AUTOMATIC (default), MANUAL }` config. AUTOMATIC self-acquires a silent fix on identify and app launch when none is available; MANUAL leaves it to the host.
- New public `ModuleGeofence.instance()` + `ModuleGeofence.refreshFromCurrentLocation()` — the host drives a refresh after granting location permission.
- Identify/launch refresh anchors at the current registration center (falling back to the location cache), so a refresh after background movement doesn't re-rank from a stale cached fix.

## Why

Customers who only want geofencing shouldn't be forced to emit location analytics. Routing geofence's location needs through an internal path decouples the two: geofence fixes never reach the analytics pipeline or identify context.

## Notes

- `GeofenceServicesImpl` stays module-free so the broadcast/boot cold-start path never touches modules (they aren't registered then, e.g. in wrappers).
- No public *location* API change (silent primitive + `getLastKnownLocation` are `@InternalCustomerIOApi`). Public *geofence* API adds the enum, the config setter, `instance()`, and `refreshFromCurrentLocation()`.
- Refresh runs on identify and app launch (not on every foreground — intentionally, to avoid churn when the app is switched to often); the registration-center anchoring matches iOS.
- iOS parity: mirrors iOS field split, GeofenceLocationMode, refreshFromCurrentLocation, module-free broadcast path, silent bypasses tracking-mode OFF.
- Sample app drives geofencing via `refreshFromCurrentLocation()` after a background permission grant.

## Linear

https://linear.app/customerio/issue/MBL-1978/route-geofence-location-updates-through-an-internal-api

---

> [!NOTE]
> **Medium Risk**
> Changes cross-module location handling, user sign-out/session boundaries, and geofence refresh anchoring—areas where subtle bugs could affect privacy or wrong-user geofence ranking, though behavior is covered by new tests.
> 
> **Overview**
> Enables **geofencing without location analytics** by routing geofence-driven fixes through an internal silent location path and expanding geofence configuration and host APIs.
> 
> The location module adds **`requestLocationUpdateSilently()`** (internal): one-shot GPS that updates **`lastKnownLocation`** and publishes **`LocationAcquired`** only—no CIO Location Update track, persistence, or identify enrichment—and can run when tracking mode is **OFF**. **`LocationTracker`** splits **tracked** vs **last-known** state so silent fixes never touch analytics.
> 
> Geofencing adds **`GeofenceLocationMode`** (**AUTOMATIC** default / **MANUAL**), **`ModuleGeofence.instance()`**, and **`refreshFromCurrentLocation()`** for host-driven refreshes after permission grants. **AUTOMATIC** silently acquires a fix on identify/launch when no anchor exists; **MANUAL** relies on the host. Identify/launch refreshes anchor on the **registration center** (fallback: location cache). **`GeofenceServices`** arms explicit refreshes, consumes one fix per request, and on sign-out clears pending refresh flags and the registration anchor synchronously to avoid cross-user leakage.
> 
> The sample app triggers **`refreshFromCurrentLocation()`** after background location is granted instead of **`requestLocationUpdate()`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8e2ad669783d1ac4a612984cd6df33609683da57. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceLocationMode.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceLocationMode.kt
@@ -1,0 +1,13 @@
+package io.customer.geofence
+
+/**
+ * How the geofence module acquires the location it needs. Location acquired for
+ * geofencing is never sent to analytics.
+ */
+enum class GeofenceLocationMode {
+    /** The SDK acquires a fix itself when it needs one and none is available. Default. */
+    AUTOMATIC,
+
+    /** The host drives it via [ModuleGeofence.refreshFromCurrentLocation]. */
+    MANUAL
+}
--- a/geofence/api/geofence.api
+++ b/geofence/api/geofence.api
@@ -8,16 +8,28 @@ public final class io/customer/geofence/GeofenceBroadcastReceiver : android/cont
 	public fun onReceive (Landroid/content/Context;Landroid/content/Intent;)V
 }
 
+public final class io/customer/geofence/GeofenceLocationMode : java/lang/Enum {
+	public static final field AUTOMATIC Lio/customer/geofence/GeofenceLocationMode;
+	public static final field MANUAL Lio/customer/geofence/GeofenceLocationMode;
+	public static fun getEntries ()Lkotlin/enums/EnumEntries;
+	public static fun valueOf (Ljava/lang/String;)Lio/customer/geofence/GeofenceLocationMode;
+	public static fun values ()[Lio/customer/geofence/GeofenceLocationMode;
+}
+
 public final class io/customer/geofence/GeofenceModuleConfig : io/customer/sdk/core/module/CustomerIOModuleConfig {
+	public synthetic fun <init> (Lio/customer/geofence/GeofenceLocationMode;Lkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public final fun getLocationMode ()Lio/customer/geofence/GeofenceLocationMode;
 }
 
 public final class io/customer/geofence/GeofenceModuleConfig$Builder : io/customer/sdk/core/module/CustomerIOModuleConfig$Builder {
 	public fun <init> ()V
 	public fun build ()Lio/customer/geofence/GeofenceModuleConfig;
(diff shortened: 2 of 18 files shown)


geofence/src/main/kotlin/io/customer/geofence/GeofenceLocationMode.kt
```diff
@@ -0,0 +1,13 @@
...
+ * How the geofence module acquires the location it needs. Location acquired for
+ * geofencing is never sent to analytics.
+ */
+enum class GeofenceLocationMode {
+    /** The SDK acquires a fix itself when it needs one and none is available. Default. */
+    AUTOMATIC,
+
+    /** The host drives it via [ModuleGeofence.refreshFromCurrentLocation]. */
+    MANUAL
+}
```
mahmoud-elmorabea, line 13: I am concerned that the combination of location mode + geofencing mode would be confusing .. can we try to fold 'em into a single mode that defines both behaviors?

mahmoud-elmorabea (approved):

geofence/src/main/kotlin/io/customer/geofence/GeofenceLocationMode.kt
mrehan27: Discussing on slack, will address in follow up PR if needed

PR #769, opened 2026-07-03
feat: Visual Notification Inbox overlay (integration / build distribution)

mahmoud-elmorabea:
Draft integration PR for the Visual Notification Inbox overlay. Opened primarily to publish sample-app builds to Firebase App Distribution so another team can test the feature end-to-end.

Aggregates the already-reviewed sub-PRs: module skeleton, data layer, overlay UI, interaction/web-parity, message actions + branding colors, dismiss-resurrection guard, and sample-app listener demo.

Firebase App Distribution builds (java_layout + kotlin_compose) are posted as a comment on this PR once CI completes.

---

> [!NOTE]
> **Medium Risk**
> Touches in-app messaging state, queue/SSE sync, and new public Compose APIs; behavior is heavily unit-tested but race between HTTP backfill and SSE inbox updates is documented as a known gap.
> 
> **Overview**
> Introduces the **Visual Notification Inbox** end-to-end: a new **`messaginginbox`** library with Compose entry points (`NotificationInboxOverlay`, `NotificationInboxBell`, `NotificationInboxView`) that render messages via **Jist** and workspace **branding** (floating bell SVG, position, chrome).
> 
> **`messaginginapp`** gains the data layer: **`VisualInbox`** / **`InboxRepository`** fetch and cache `/api/v1/templates` and `/api/v1/branding`, gate visibility on **`X-CIO-Inbox-Enabled`**, and expose reactive flows for the overlay. Host apps can register **`InboxEventListener`** (build-time or runtime via `setInboxEventListener`) to intercept inbox actions. Inbox click metrics now include **`actionValue`** (new overload for binary compatibility). **Dismiss tombstones** stop deleted messages from reappearing on stale queue polls.
> 
> **SSE** no longer calls `fetchUserMessages()` on foreground alone; backfill runs once the server sends **`connected`**, matching gist-web and fixing empty inbox on fresh login. CI matrices add **`messaginginbox`**; build deps add **jist**, **Material3**, and **lifecycle-runtime-compose**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8576d0f558ac0819ee72a0db1d795d6ef516b594. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


messaginginbox/src/main/java/io/customer/messaginginbox/NotificationInboxOverlay.kt
```diff
@@ -0,0 +1,679 @@
...
+ */
+@Composable
+private fun InboxListContent(
+    controller: VisualInboxController,
+    colors: InboxColors,
+    modifier: Modifier = Modifier
+) {
+    val context = LocalContext.current
+    val state by remember(controller) { controller.uiStateFlow() }
+        .collectAsStateWithLifecycle(initialValue = VisualInboxUiState(loading = true))
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Partially valid — the mechanism is real but the stated impact is guarded.

`uiStateFlow()` is a cold flow, and both the parent (`NotificationInboxView`/`NotificationInboxOverlay`) and the child `InboxListContent` collect it, so mounting does start two collections that each hit `load()` via `.onStart`. However, `InboxRepository.loadTemplatesAndBranding()` prevents the *parallel load pipelines* described here: an atomic in-flight guard (`fetchInFlight.compareAndSet(false, true)` → early return with current state) plus a once-per-session revalidation gate mean a second concurrent load short-circuits with no extra network fetch.

So the residual cost is redundant flow collection (two subscriptions + snapshot reads), not parallel fetches or divergent UI. Worth a low-priority cleanup (hoist/share a single collection), not a correctness bug.

messaginginapp/src/main/java/io/customer/messaginginapp/inbox/VisualInbox.kt
```diff
@@ -0,0 +1,99 @@
...
+     * (including opened-state changes). Backed by the store's StateFlow, so it emits the
+     * current value immediately on collection (a late-mounting overlay still gets latest state).
+     * The overlay maps each emission to a fresh read of CACHED state via [getVisibility] /
+     * [getSelectedMessages]; this never triggers a network fetch.
+     */
+    fun observeInboxChanges(): Flow<Unit> =
+        inAppMessagingManager.state
+            .map { state -> InboxStateKey(state.isInboxEnabled, state.inboxMessages) }
+            .distinctUntilChanged()
+            .map { }
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Not fixing this — the read-time check is intentional.

The mechanism is exactly as described: `InboxSelection` applies `expiry <= now` at read time, `observeInboxChanges` keys only on `InboxStateKey(isInboxEnabled, inboxMessages)`, and nothing is time-driven — so a rendered message can outlive its expiry until the next poll or fetch (600s default) or any other store change.

We're accepting that deliberately, for a product reason rather than a technical one: expiring actively on a timer would make a message disappear while the user is reading it, or mid-scroll. That is a worse experience than briefly showing one that has just passed its expiry. The staleness is bounded and self-healing — it clears on the next poll, on any inbox store change, and on foreground/relaunch.

The same call applies to iOS, which has the identical read-time selection (`VisualInboxSelector.select(messages:now:)`), so the two platforms stay at parity rather than diverging on expiry behaviour.

mrehan27 (approved):

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/SseLifecycleManager.kt
```diff
@@ -92,6 +90,36 @@ internal class SseLifecycleManager(
...
+     * connection never establishes. Mirrors gist-web, which fetches from its `connected` listener.
+     *
+     * This narrows the window but does not close it: `fetchUserMessages()` is fire-and-forget, so an
+     * `inbox_messages` event arriving while the request is in flight can still be overwritten by the
+     * older HTTP snapshot, since both publish a full-state write. Ordering the two properly needs
+     * them versioned or merged; tracked separately.
+     */
+    private fun backfillOnConnect() {
+        sseLogger.logSseConnectionConfirmedBackfill()
+        gistQueue.fetchUserMessages()
```
Shahroz16, line 120: Could we track a follow-up to prevent the HTTP backfill from overwriting newer inbox state received through SSE? If an SSE update arrives while the backfill request is still running, the older HTTP response can replace it, and the inbox may stay stale until another SSE event or reconnect. I don’t think this needs to block this PR, but linking the follow-up here would make the remaining limitation clear.

messaginginapp/src/main/java/io/customer/messaginginapp/inbox/data/InboxRepository.kt
```diff
@@ -0,0 +1,431 @@
...
+        //  - there is actually something to serve. A failure with an empty cache has no fallback, so
+        //    closing would strand the inbox hidden for the rest of the process; leaving it open lets
+        //    the next poll retry. A cycle that resolved (fresh OR stale) still closes it, so a
+        //    failing server cannot cause a per-poll retry loop.
+        val hasServableAssets = (freshTemplates ?: staleTemplates) != null &&
+            (freshBranding ?: staleBranding) != null
+        if (inAppMessagingManager.getCurrentState().sessionId == sessionAtFetchStart && hasServableAssets) {
+            hasRevalidatedAssetsThisSession.set(true)
+        } else if (!hasServableAssets) {
+            logger.debug("$LOG_TAG revalidation produced nothing servable: gate stays open for the next poll")
```
Shahroz16, line 266: This looks non-blocking and could be followed up separately. If the initial asset fetch exhausts its retries while SSE remains connected, there is no independent retry, so the inbox can stay hidden until a later lifecycle or connection event. Could we track that follow-up?

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/SseLifecycleManager.kt
```diff
@@ -135,7 +163,7 @@ internal class SseLifecycleManager(
             if (state.shouldUseSse) {
                 // User became identified and SSE is enabled - start SSE connection
                 sseLogger.logEnablingSseForIdentifiedUser()
-                sseConnectionManager.startConnection()
+                startSseConnection()
```
Shahroz16, line 166: Could we handle a direct switch from one identified user to another by observing the actual user ID, clearing the previous inbox state, and reconnecting SSE? For A → B, `isUserIdentified` stays `true`, so this callback never runs. The existing stream remains authenticated as A while the SDK treats B as current, and later events from A can replace B’s inbox state. Since this can expose one user’s inbox to another, I think it should block the release. A focused A → B test would help cover the fix.

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/SseLifecycleManager.kt
mahmoud-elmorabea: The window is too narrow and it exceptionally unlikely that these two responses have different messages. I suggest we take as a follow up to sequence the two calls or have some kind of logic to merge them

messaginginapp/src/main/java/io/customer/messaginginapp/inbox/data/InboxRepository.kt
mahmoud-elmorabea: Sure, what do you suggest? I don't think we should keep retrying for the entire session or maybe only if its network related issue

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/SseLifecycleManager.kt
mahmoud-elmorabea: I think this is the whole of in-app, not just inbox. Maybe if we want to formally support that login while a user already logged in path we could. I suggest handling that as a follow up as well

PR #775, opened 2026-07-10
chore(geofence): enrich transition events with geofence metadata

mrehan27:
Enriches the "Geofence Transition" event with workspace-defined `metadata` from `POST /geofences/nearest`.

## What
- `metadata` (scalar string/number/bool map) from the nearby response is cached on the region, snapshotted onto the pending delivery at crossing time, and preferred fresh from cache at send — falling back to the snapshot when the geofence has left the cache (same hybrid as `geofenceName`).
- Emitted as a nested `metadata` object on the event, **always present** (empty `{}` when the fence has none). Values keep their type (string/long/double/boolean).
- Non-scalar values (null/array/object) are dropped at parse; defensive count (100) + payload (100 KB) caps as a runaway-payload backstop — the server does the real validation.
- `geosetId` now emitted as a plain string (dropped the numeric coercion).
- Dropped `radius` from the `/geofences/nearest` request body (now optional server-side).

## Notes
- Both delivery paths (WorkManager direct-HTTP + foreground flush) build an identical property set.
- The worker carries only the pending-store key through WorkManager `inputData` and loads the full row from the store, so metadata (which can be large) never risks the ~10 KB `Data` limit and the worker always sends the authoritative snapshot rather than a copy.
- No public API change (all internal); adds one internal helper `PendingDeliveryStore.get(key)` in `:core` (not in the API dump).
- iOS parity confirmed against the equivalent iOS change (metadata feature fully aligned; iOS is dropping `radius` too).

## Stack
```
main
 └─ feature/geofence-on-device
     └─ mbl-2045-geofence-transition-metadata  ← this PR
```

## Test plan
- `./gradlew :core:testDebugUnitTest :core:apiCheck :geofence:testDebugUnitTest :geofence:apiCheck :messagingpush:testDebugUnitTest` — green
- `make lint-no-format` — green

[MBL-2045](https://linear.app/customerio/issue/MBL-2045/mobile-enrich-geofence-transition-events-with-geofence-metadata)

---

> [!NOTE]
> **Medium Risk**
> Touches geofence event delivery, pending-store durability, and nearest-fetch request shape; behavior changes (metadata payload, persist-failure handling, no client radius) could affect transition delivery or backend fetch semantics if misaligned with server expectations.
> 
> **Overview**
> **Geofence Transition** events now carry workspace **`metadata`** from `POST /geofences/nearest`: parsed with tolerant sanitization (scalar-only, count/size caps), cached on `GeofenceRegion`, snapshotted on pending delivery, and merged at send time via `withFreshestEventData` (same pattern as `geofenceName`) on both WorkManager HTTP and foreground flush. Event properties always include a nested **`metadata`** object; **`geosetId`** is emitted as a string (no numeric coercion).
> 
> **Delivery reliability:** `PendingDeliveryStore.append`/`appendAll` return whether the disk write succeeded; `get(key)` supports worker lookup. On persist failure the broadcast receiver rolls back cooldown via `GeofenceCooldownFilter.release` and skips scheduling. **`GeofenceEventWorker`** passes only `entry_key` in WorkManager `Data` and loads the full row from the store (avoids ~10 KB limits and drift with large metadata).
> 
> **Fetch/sync:** `/geofences/nearest` no longer sends **`radius`** (removed `remoteSearchRadiusMeters`). OS re-register skips when only non-geometry fields change (`equalsForRegistration` replaces geoset-only equality), covering metadata, name, and `lastUpdated`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3986b2e009c01debb9bb4608a6413109b756cccc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


geofence/src/main/kotlin/io/customer/geofence/worker/GeofenceEventWorker.kt
```diff
@@ -38,16 +30,12 @@ internal class GeofenceEventScheduler(
...
-            .putString(KEY_GEOFENCE_ID, entry.geofenceId)
-            .putString(KEY_TRANSITION, entry.transition.name)
-            .putString(KEY_TRANSITION_ID, entry.transitionId)
-            .putLong(KEY_TIMESTAMP, entry.timestamp)
-            .apply {
-                entry.userId?.let { putString(KEY_USER_ID, it) }
-                entry.geofenceName?.let { putString(KEY_GEOFENCE_NAME, it) }
-                entry.geosetId?.let { putString(KEY_GEOSET_ID, it) }
-            }
+            .putString(KEY_ENTRY_KEY, entry.key)
```
Shahroz16, line 38: Now that WorkManager receives only this key, the pending file is the sole copy of the event. appendAll currently returns Unit and writeAll logs and swallows I/O failures, so the receiver schedules this key even when no row was saved. The worker then sees get(entryKey) return null, assumes another path delivered it, and returns success. Could append success be made observable and scheduling happen only after confirmed persistence, with a fallback or cooldown rollback when the write fails? Otherwise a transient disk error silently loses the transition.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRegion.kt
```diff
@@ -70,8 +73,9 @@ internal fun GeofenceRegion.toGmsTransitionTypes(): Int {
...
 /**
- * Equal for OS-registration purposes: everything but [geosetIds], which drive event fan-out, not
- * GMS registration. A geoset-only change updates the cache without a needless re-register.
+ * True when two regions are identical for OS-registration purposes — matching on everything except
+ * the event-only fields ([geosetIds], [metadata]), which drive event payload rather than GMS geometry.
+ * A change to only those updates the cache without a needless re-register.
  */
-internal fun GeofenceRegion.equalsIgnoringGeosetIds(other: GeofenceRegion): Boolean =
-    copy(geosetIds = other.geosetIds) == other
+internal fun GeofenceRegion.equalsForRegistration(other: GeofenceRegion): Boolean =
```
Shahroz16, line 80: Could this comparison use only the fields passed to Geofence.Builder, which are ID, coordinates, radius, and transition types? name is event payload too, but data-class equality still compares it. A backend rename therefore falls out of existingBusinessIds, is re-added with INITIAL_TRIGGER_ENTER, and can emit a synthetic enter for a device already inside. A name-only regression test alongside the metadata and geoset tests would pin the intended behavior.

geofence/src/main/kotlin/io/customer/geofence/worker/GeofenceEventWorker.kt
mrehan27: Updated to remove cooldown if write fails

geofence/src/main/kotlin/io/customer/geofence/GeofenceRegion.kt
mrehan27: Updated

Shahroz16 (approved):

PR #776, opened 2026-07-13
chore(inbox): overlay bottom sheet + branding bell, position & web-parity actions

mahmoud-elmorabea:
Reworks the Android Visual Inbox drop-in overlay to a native Material3 bottom sheet with a branding-driven bell (dynamic SVG + fallback), branding bell position, and web-parity message actions (openUrl/openDeeplink/performAction/dismiss), mirroring the iOS overlay.

---

> [!NOTE]
> **Medium Risk**
> Touches public Compose API, navigation/intents, and action-resolution semantics (stricter behavior parsing); overlay UX changes from side panel to bottom sheet may affect integrators.
> 
> **Overview**
> Reworks the Android Visual Inbox **drop-in overlay** from a custom scrim + slide-out side panel to a **Material3 `ModalBottomSheet`**, aligned with the iOS sheet experience. The floating bell now honors branding **corner position**, optional **custom bell SVG** (with crash-safe parse + bundled fallback), and **`unreadIndicator.showAlert`** for the badge.
> 
> **Public API:** `NotificationInboxView` adds an optional **`onDismissRequest`** so hosts can close their own sheet when a message navigates away; the binary API dump reflects this.
> 
> **Message actions (web/iOS parity):** Resolution is a strict **`data.behavior`** switch (`openUrl`, `openDeeplink`, `performAction`, `dismiss`) using **`data.action`** (legacy `data.url` fallback) and **`data.name`** for the host listener—no URL-scheme guessing. **`openDeeplink`** is now handled by the SDK (host app first, then external), with manifest **`queries`** for http(s) `VIEW` so `resolveActivity` works on API 30+. Successful **openUrl/openDeeplink** (and host-handled nav actions) **dismiss the sheet**; failed opens leave it open.
> 
> Adds **Material3** dependency, updates the default bell drawable, expands unit/Robolectric tests, and switches the **kotlin_compose** sample to **`NotificationInboxOverlay()`** instead of a manual bell + sheet.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8bdc890fe3177583ab7b09cd6679c516d08cde4b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/messaginginbox/src/main/java/io/customer/messaginginbox/InboxBellIcon.kt
+++ b/messaginginbox/src/main/java/io/customer/messaginginbox/InboxBellIcon.kt
@@ -1,0 +1,112 @@
+package io.customer.messaginginbox
+
+import androidx.compose.foundation.Canvas
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.geometry.Offset
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.Path
+import androidx.compose.ui.graphics.PathFillType
+import androidx.compose.ui.graphics.drawscope.scale
+import androidx.compose.ui.graphics.drawscope.translate
+import androidx.compose.ui.graphics.vector.PathParser
+import kotlin.math.min
+
+/**
+ * Renders a pre-built branding bell glyph ([InboxBellSvg.BellArt]) — the workspace's configured
+ * `patterns.inbox.floatingIcon.svg`, parsed once into a combined even-odd path — scaled-to-fit the
+ * target size (aspect preserved) and filled with [tint].
+ *
+ * The art is built (and validated) up front by [InboxBellSvg.buildArt]; callers pass the result only
+ * when non-null and fall back to the bundled drawable otherwise, so a malformed SVG never reaches
+ * here. Only `<path d>` geometry + `viewBox` are honored (presentation attributes are ignored — the
+ * bell is a single tinted silhouette).
+ */
+@Composable
+internal fun InboxBellIcon(
+    art: InboxBellSvg.BellArt,
+    tint: Color,
+    modifier: Modifier = Modifier
+) {
+    Canvas(modifier = modifier) {
+        val scale = min(size.width / art.width, size.height / art.height)
+        // Center the scaled artwork in the target rect.
+        val left = (size.width - art.width * scale) / 2f
+        val top = (size.height - art.height * scale) / 2f
+        translate(left = left, top = top) {
+            scale(scaleX = scale, scaleY = scale, pivot = Offset.Zero) {
+                // Shift the viewBox origin to (0,0) before scaling so a non-zero minX/minY viewBox fits.
+                translate(left = -art.minX, top = -art.minY) {
+                    drawPath(path = art.path, color = tint)
+                }
+            }
+        }
+    }
+}
+
+/**
+ * Branding-SVG bell parsing for [InboxBellIcon]. [parse] is a pure (non-Compose) regex extraction of
+ * the `viewBox` + each `<path>`'s `d`, unit-testable on the JVM. [buildArt] turns that into a
+ * ready-to-draw [BellArt] using Compose's built-in [PathParser] (no third-party dep — unlike iOS,
+ * which vendors a parser); it is crash-safe (see below) and its result is `remember`-cached by the
+ * caller so the SVG is parsed once, not per frame.
+ */
+internal object InboxBellSvg {
+    /** A built, ready-to-draw bell glyph: the combined even-odd [Path] plus its source viewBox. */
+    data class BellArt(
+        val path: Path,
+        val minX: Float,
+        val minY: Float,
+        val width: Float,
+        val height: Float
+    )
+
+    /** Parsed geometry: viewBox origin/size + the `d` string of each `<path>`. */
+    data class Parsed(
+        val minX: Float,
+        val minY: Float,
+        val width: Float,
+        val height: Float,
+        val pathData: List<String>
+    )
+
+    private val VIEW_BOX = Regex("""viewBox\s*=\s*["']\s*([^"']+?)\s*["']""", RegexOption.IGNORE_CASE)
+    private val PATH_D = Regex("""<path\b[^>]*?\bd\s*=\s*["']([^"']+)["']""", setOf(RegexOption.IGNORE_CASE, RegexOption.DOT_MATCHES_ALL))
+
+    /**
+     * Parse [svg], or return null when there is nothing renderable (no `<path d>`). A missing
+     * `viewBox` defaults to a 24×24 box so a bare `<svg><path/></svg>` still fits.
+     */
+    fun parse(svg: String): Parsed? {
+        val pathData = PATH_D.findAll(svg).map { it.groupValues[1].trim() }.filter { it.isNotEmpty() }.toList()
+        if (pathData.isEmpty()) return null
+
+        val box = VIEW_BOX.find(svg)?.groupValues?.get(1)
+            ?.split(Regex("[\\s,]+"))
+            ?.mapNotNull { it.toFloatOrNull() }
+        return if (box != null && box.size == 4 && box[2] > 0f && box[3] > 0f) {
+            Parsed(minX = box[0], minY = box[1], width = box[2], height = box[3], pathData = pathData)
+        } else {
+            Parsed(minX = 0f, minY = 0f, width = 24f, height = 24f, pathData = pathData)
+        }
+    }
+
+    /**
+     * Build a ready-to-draw [BellArt] from raw [svg], or null when it cannot be rendered (no
+     * `<path d>`, or the path data is malformed). [PathParser.parsePathString] THROWS
+     * `IllegalArgumentException` on invalid path commands (e.g. `"M0 0 X1 1"`), so the build is
+     * guarded — a malformed branding SVG falls back to the bundled bell instead of crashing.
+     */
+    fun buildArt(svg: String): BellArt? {
+        val parsed = parse(svg) ?: return null
+        return try {
+            val path = Path().apply {
+                fillType = PathFillType.EvenOdd
+                parsed.pathData.forEach { d -> addPath(PathParser().parsePathString(d).toPath()) }
+            }
+            if (path.isEmpty) null else BellArt(path, parsed.minX, parsed.minY, parsed.width, parsed.height)
+        } catch (ex: Exception) {
+            null
+        }
+    }
+}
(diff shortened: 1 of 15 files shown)


messaginginbox/src/main/java/io/customer/messaginginbox/InboxBellIcon.kt
```diff
@@ -0,0 +1,99 @@
...
+    modifier: Modifier = Modifier
+) {
+    val parsed = remember(svg) { InboxBellSvg.parse(svg) } ?: return
+    Canvas(modifier = modifier) {
+        val path = Path().apply {
+            fillType = PathFillType.EvenOdd
+            parsed.pathData.forEach { d ->
+                // PathParser is robust to malformed data (it stops at the bad token); a fully
+                // unparseable `d` simply contributes nothing rather than throwing.
+                addPath(PathParser().parsePathString(d).toPath())
```
Shahroz16: Could we validate the path data before choosing the branded icon? `canRender()` currently only checks that extraction found a nonempty `<path d>`, while the Compose parser runs later during drawing. I was able to make `<path d="M0 0 X1 1">` pass `canRender()` and then throw `IllegalArgumentException`, so malformed branding could crash instead of using the bundled bell. Would it make sense to parse safely up front, fall back for errors or empty paths, and add an invalid command test? I also could not find the device or Compose test mentioned in the test KDoc, so that coverage note may need updating.

messaginginbox/src/main/java/io/customer/messaginginbox/VisualInboxOverlayState.kt
```diff
@@ -232,34 +233,41 @@ internal class VisualInboxController(
...
         // SDK default navigation (item 12), unless the host handled it.
         val navigation = if (handledByHost) {
             InboxNavigation.None
         } else {
             when (resolution) {
                 is InboxAction.OpenUrl -> InboxNavigation.OpenUrl(resolution.url)
-                is InboxAction.Deeplink -> {
+                // openDeeplink: route through the app's deep-link handling (the overlay resolves the
+                // host app first, then external), mirroring push/in-app deep links.
+                is InboxAction.Deeplink -> InboxNavigation.OpenDeeplink(resolution.url)
```
Shahroz16, line 263: Could we preserve the original `openDeeplink` behavior even when the host listener handles it? Right now that outcome becomes `InboxNavigation.None`, so a host that performs single activity Compose or fragment navigation can leave the SDK sheet over the destination. iOS closes the sheet for the host handled outcome as well. Perhaps the result could carry both the behavior and whether default navigation was suppressed, which would let the overlay dismiss without opening the link twice.

messaginginbox/src/main/java/io/customer/messaginginbox/NotificationInboxOverlay.kt
```diff
@@ -163,120 +167,109 @@ fun NotificationInboxOverlay(
...
+    // so it is composed outside the anchor Box. Gate on canShowChrome too: if the inbox empties while
+    // the sheet is open (dismissing the last message -> Hidden), the sheet closes with the inbox.
+    if (sheetExpanded && canShowChrome) {
+        val sheetState = rememberModalBottomSheetState()
+        ModalBottomSheet(
+            onDismissRequest = { sheetExpanded = false },
+            sheetState = sheetState,
+            shape = RoundedCornerShape(topStart = colors.cornerRadius, topEnd = colors.cornerRadius),
+            containerColor = colors.panelColor,
+            dragHandle = { BottomSheetDefaults.DragHandle() }
```
Shahroz16: Do we need to pass the drag handle explicitly here? Material3 already uses this same handle by default. The captureless lambda also produces the new `ComposableSingletons$NotificationInboxOverlayKt` entry with a hash named getter in `messaginginbox.api`. If we omit the argument, could we remove that generated implementation detail from the public API dump?

messaginginbox/src/main/java/io/customer/messaginginbox/VisualInboxOverlayState.kt
```diff
@@ -353,79 +361,93 @@ internal class VisualInboxController(
...
- * string. Safe casts throughout: a non-object `data` or non-primitive `url` yields null.
+ * Extracts the action value from a Jist action event's data object — the web-schema `data.action`,
+ * falling back to the legacy `data.url` — or null if absent / not a string. Safe casts throughout:
+ * a non-object `data` or non-primitive value yields null.
  */
-private fun actionUrl(event: JistActionEvent): String? =
-    ((event.data as? JsonObject)?.get("url") as? JsonPrimitive)?.contentOrNull
+private fun actionValue(event: JistActionEvent): String? {
+    val data = event.data as? JsonObject ?: return null
+    return (data["action"] as? JsonPrimitive)?.contentOrNull
```
Shahroz16: Could we make these reads explicitly string only? `contentOrNull` also returns text for numeric and Boolean primitives, so a value such as `action: true` can reach navigation as `"true"`, and numeric names or behaviors can reach tracking or resolution. The web parser rejects those shapes. Checking `JsonPrimitive.isString` for action, url, name, and behavior, then adding a couple of malformed field tests, would keep the native contract aligned. We could keep the Boolean and string compatibility for dismiss if that is intentional.

messaginginbox/src/main/java/io/customer/messaginginbox/VisualInboxOverlayState.kt
```diff
@@ -353,79 +361,93 @@ internal class VisualInboxController(
...
+    /** Route [url] through the app's deep-link handling (host app first, then external). Closes the overlay. */
+    data class OpenDeeplink(val url: String) : InboxNavigation
 }
 
 /**
- * Maps a Jist action event to the SDK's [InboxAction]. Determined from the action `name`,
- * `data.behavior` and `data.url`:
+ * Maps a Jist action event to the SDK's [InboxAction] (web parity). Uses a STRICT `data.behavior`
+ * switch, mirroring the iOS overlay's `resolve` — there is deliberately no http/scheme guessing and
+ * no legacy `deeplink`/`newTab` alias, so behavior matches iOS exactly. Value is `data.action`
```
Shahroz16: Could we clarify which legacy forms we intend to support here? This comment says there is no `newTab` alias, but the resolver below accepts one, while web models `newTab` as a Boolean field rather than a behavior. The `OpenUrl` and `Deeplink` comments above also mention URL shape inference even though an absent behavior now becomes `Unknown`. Aligning the comments, tests, and resolver would make the compatibility contract much easier to follow.

Shahroz16 (commented):
Thanks for the native sheet and parity updates. Could we take another look at the inline cases before merging? The focused messaginginbox tests and assembleDebug pass. The only failing tracking migration check appears unrelated because the Robolectric dependency download failed, so a rerun may be enough there?. 

My inline questions are mainly around malformed branding, sheet dismissal after host handled navigation, and a few API and schema cleanup points.

I also verified Cursor’s existing https://github.com/customerio/customerio-android/pull/776#discussion_r3583714491 remains valid. It covers a host owned presentation when the SDK handles the deep link, which is separate from my inline question about host handled navigation in the SDK owned overlay.

messaginginbox/src/test/java/io/customer/messaginginbox/InboxBellSvgTest.kt
```diff
@@ -0,0 +1,78 @@
+package io.customer.messaginginbox
+
+import org.amshove.kluent.shouldBeEqualTo
+import org.amshove.kluent.shouldBeNull
+import org.amshove.kluent.shouldNotBeNull
+import org.junit.Test
+
+/**
+ * Unit tests for [InboxBellSvg] — the pure SVG extraction (viewBox + `<path d>`) used by the
+ * branding bell glyph. Building the actual Path is Compose/`android.graphics`-bound and covered on
```
Shahroz16, line 10: For future improvement, now that `buildArt` catches the parser failure and returns `null`, maybe we should add a regression test that feeds it an invalid command such as `M0 0 X1 1` and verifies the fallback result? This suite currently only exercises `parse`, and I could not find the device coverage mentioned here. A small Robolectric or instrumented test asserting `buildArt(...) == null` would keep the fix from regressing.

Shahroz16 (approved):
Thus cursor bot issue: https://github.com/customerio/customerio-android/pull/776#discussion_r3587588850

and kotlin lint action failure are the only two points left that needs to be tackled before we merge it.

messaginginbox/src/main/java/io/customer/messaginginbox/InboxBellIcon.kt
mahmoud-elmorabea: Fixed — `buildArt` parses safely up front (`runCatching`) and falls back to the bundled bell on malformed path data.

messaginginbox/src/main/java/io/customer/messaginginbox/VisualInboxOverlayState.kt
mahmoud-elmorabea: Fixed — host-handled navigating actions now resolve to `InboxNavigation.Dismiss`, so the sheet closes even when the host performs the navigation (matches iOS).

messaginginbox/src/main/java/io/customer/messaginginbox/NotificationInboxOverlay.kt
mahmoud-elmorabea: Fixed — removed the explicit drag handle; the `ComposableSingletons` entry is gone from the API dump.

messaginginbox/src/main/java/io/customer/messaginginbox/VisualInboxOverlayState.kt
mahmoud-elmorabea: Fixed — reads now guard on `JsonPrimitive.isString` (action/url/name/behavior); added malformed-field tests.

messaginginbox/src/main/java/io/customer/messaginginbox/VisualInboxOverlayState.kt
mahmoud-elmorabea: Fixed — aligned comments/resolver/tests; `newTab` is a boolean (not a behavior) and an absent behavior is `Unknown`.

messaginginbox/src/test/java/io/customer/messaginginbox/InboxBellSvgTest.kt
mahmoud-elmorabea: Done — added `InboxBellArtTest` (Robolectric): malformed path (`M0 0 X1 1`) → `buildArt` returns null.

PR #777, opened 2026-07-14
chore(geofence): harden transition delivery and sign-out reset

mrehan27:
### Summary

Addresses review feedback on the geofence integration PR (#763) — three delivery/attribution durability fixes plus two small cleanups. Base is `feature/geofence-on-device`, not `main`.

**Foreground flush is now at-least-once (was at-most-once).** `PendingDeliveryFlusher` gained a `DeliveryGuarantee`; the geofence foreground flush now publishes **then** removes, so a crash between the two re-delivers instead of dropping the only durable copy. Duplicates are deduped downstream by `transitionId`. Push delivery is unchanged (`AT_MOST_ONCE`).

**Sign-out reset no longer clobbers a fast switch-account.** `clearIdentify` now clears `secureUserStore` **synchronously before** publishing `ResetEvent`, so a reset subscriber reading identity sees the signed-out state (not a stale value or a fast re-identify). The geofence repo tracks the user that owns the live registration (in-memory, under the existing lock) and `reset()` skips the wipe only when the current user already owns it — wiping on a plain sign-out and when a different user hasn't re-synced yet. No public API change.

**Anonymous transitions are dropped at the receiver.** Geofencing is identified-only (Edge rejects anonymous geofence tracks), so a null/empty-user transition has no deliverable path — we drop it before spending a cooldown slot or persisting a row no channel could send, instead of queuing an undeliverable row.

**Cleanups:** `GeofenceRegion.name` is now optional (`String? = null`, iOS parity); `withFreshestEventData` takes the freshest cached name at send.

### Reviewer mapping (#763)
- Foreground flush claims before durable hand-off (HIGH, Shahroz + Bugbot) → at-least-once flush.
- Identify-before-reset snapshot (HIGH, Shahroz + Bugbot) → clear-before-publish + registered-owner check.
- Anonymous flush hits rejected track (MEDIUM, Shahroz + Bugbot) → drop at receiver.

### Stack
```
main
└─ feature/geofence-on-device
   └─ geofence-delivery-durability-fixes  ← this PR
```

### Testing
- Unit tests added/updated: flusher at-least-once (happy path + throw-retains-entry), repo `reset` owner classification (plain wipe / skip-when-owned / wipe-when-different-owner / manager-fail-preserve), `clearIdentify` synchronous store clear, anonymous-drop at receiver, `name` optional parse + send.
- Green: `core`, `geofence`, `datapipelines`, `messaginginapp`, `messagingpush` unit tests, lint, `apiCheck` (no surface change), ktlint.

Linear: https://linear.app/customerio/issue/MBL-2087/mobile-address-geosets-prs-review-feedback

---

> [!NOTE]
> **Medium Risk**
> Touches identity timing on reset, pending-delivery ordering, and geofence wipe vs. fast account switch—behavioral changes in event delivery and sign-out paths, mitigated by focused unit tests and unchanged public API.
> 
> **Overview**
> Hardens geofence **transition delivery** and **sign-out / identity** behavior without public API changes.
> 
> **Foreground flush (at-least-once).** `PendingDeliveryFlusher` adds a `DeliveryGuarantee` parameter: geofence foreground drain uses **publish then remove** (`AT_LEAST_ONCE`) instead of claim-before-publish, so a crash after hand-off can retry; push stays **claim then publish** (`AT_MOST_ONCE`). End-to-end dedupe still relies on `transitionId` plus the send-then-remove worker.
> 
> **Sign-out reset.** `clearIdentify` now clears `secureUserStore` **synchronously before** `ResetEvent` (the async subscriber that cleared it was removed). `GeofenceRepository.reset()` no longer takes a signed-out user snapshot: it **skips the wipe only when someone is still signed in** (fast re-identify / account switch), and wipes when the store reads unsigned-out after that synchronous clear.
> 
> **Anonymous transitions.** The broadcast receiver drops null/empty `userId` transitions **before** cooldown or persistence (identified-only; no undeliverable rows or foreground-only worker skips).
> 
> **Small model cleanup.** `GeofenceRegion.name` is optional (`String?`); `withFreshestEventData` uses the cached name when present (including clearing a stale snapshot when the cache has no name).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 38f8f0e752750a3fef3d218531a90f8d2bd96ac8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/core/src/main/kotlin/io/customer/sdk/data/store/PendingDeliveryFlusher.kt
+++ b/core/src/main/kotlin/io/customer/sdk/data/store/PendingDeliveryFlusher.kt
@@ -13,17 +13,11 @@ import kotlinx.coroutines.launch
  * (typically an app-foreground handoff), coordinating with the durable
  * WorkManager channel that also consumes the same store.
  *
- * For each pending entry it: cancels that entry's WorkManager unique work — so
- * the worker can't also deliver — then atomically
- * [claims][PendingDeliveryStore.claim] it and hands it to the caller's
- * [publish] block. Cancel happens before the claim on purpose: an `ENQUEUED`
- * or running worker flips to `CANCELLED` immediately, narrowing the window in
- * which both channels could race. Cancellation is best-effort and can lose
- * across process death, so the claim is what makes this flush deliver each
- * entry at most once. Whether the *system* is exactly- or at-least-once depends
- * on the paired worker: one that also claims before sending (push) is
- * exactly-once; one that sends-then-removes (geofence) is at-least-once and
- * relies on a stable id in the payload for downstream dedupe.
+ * For each pending entry it cancels that entry's WorkManager unique work — so the worker can't
+ * also deliver — then spends it against the store per the caller's [DeliveryGuarantee] and hands
+ * it to [publish]. Cancel is best-effort (can lose across process death), so the store op is what
+ * actually bounds duplicates; match the guarantee to the paired worker (push claims →
+ * [DeliveryGuarantee.AT_MOST_ONCE]; geofence sends-then-removes → [DeliveryGuarantee.AT_LEAST_ONCE]).
  *
  * Entries are processed in isolation: one failed cancel/publish does not abort
  * the batch, and an unclaimed or failed entry survives for the next flush. The
@@ -39,6 +33,22 @@ class PendingDeliveryFlusher<T : PendingDeliveryStore.PendingDeliveryEntry>(
     private val workManagerProvider: CustomerIOWorkManagerProvider,
     private val dispatchersProvider: DispatchersProvider
 ) {
+
+    /** How the foreground flush spends a pending entry against a possible mid-publish crash. */
+    enum class DeliveryGuarantee {
+        /**
+         * Claim (remove) before publishing. If the process dies after the claim the entry is gone,
+         * so it delivers at most once. Pair with a worker that also claims — together exactly-once.
+         */
+        AT_MOST_ONCE,
+
+        /**
+         * Publish, then remove only after it returns. A crash between the two leaves the entry for
+         * the next flush to re-publish, so it delivers at least once; requires a stable payload id
+         * for downstream dedupe. Pair with a send-then-remove worker.
+         */
+        AT_LEAST_ONCE
+    }
 
     /**
      * Per-feature observation hooks. All default to no-ops so callers only
@@ -66,7 +76,11 @@ class PendingDeliveryFlusher<T : PendingDeliveryStore.PendingDeliveryEntry>(
      * on [DispatchersProvider.background]. Safe to call on every foreground
      * transition — an empty store is a cheap no-op.
      */
-    fun flush(callbacks: Callbacks<T> = Callbacks(), publish: (T) -> Unit) {
+    fun flush(
+        callbacks: Callbacks<T> = Callbacks(),
+        guarantee: DeliveryGuarantee = DeliveryGuarantee.AT_MOST_ONCE,
+        publish: (T) -> Unit
+    ) {
         CoroutineScope(dispatchersProvider.background).launch {
             runCatching {
                 val pending = store.loadAll()
@@ -81,14 +95,17 @@ class PendingDeliveryFlusher<T : PendingDeliveryStore.PendingDeliveryEntry>(
                             workManager.cancelUniqueWork(entry.key).await()
                             callbacks.onWorkCancelled(entry)
                         }
-                        // Atomically claim before publishing so this flush delivers
-                        // each entry at most once (and a worker that also claims —
-                        // push — can't deliver the same entry). Claiming per-entry
-                        // here, rather than batching a remove at the end, also means a
-                        // mid-loop CancellationException can't strand an already-
-                        // published entry for re-publish next flush.
-                        if (!store.claim(entry.key)) return@forEach
-                        publish(entry)
+                        // Order claim vs. publish per the caller's crash-safety guarantee (see [DeliveryGuarantee]).
+                        when (guarantee) {
+                            DeliveryGuarantee.AT_MOST_ONCE -> {
+                                if (!store.claim(entry.key)) return@forEach
+                                publish(entry)
+                            }
+                            DeliveryGuarantee.AT_LEAST_ONCE -> {
+                                publish(entry)
+                                store.remove(entry.key)
+                            }
+                        }
                         callbacks.onPublished(entry)
                         publishedCount++
                     } catch (ce: CancellationException) {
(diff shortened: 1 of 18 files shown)


geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
```diff
@@ -371,10 +380,14 @@ internal class GeofenceRepositoryImpl(
...
-        val currentUserId = secureUserStore.getUserId()
-        if (!currentUserId.isNullOrBlank() && currentUserId != signedOutUserId) {
+    override suspend fun reset(): Result<Unit> = stateMutex.withLock {
+        // Skip only when the user signed in *now* already owns the live registration — a
+        // switch-account where the new user re-synced first (registeredOwnerUserId is set under
+        // this same lock). Current user is read after clearIdentify cleared it, so null here means
+        // a plain sign-out → wipe. A different current user with a stale owner → wipe (the new
+        // user's sync re-registers).
+        val currentUserId = secureUserStore.getUserId()?.takeIf { it.isNotEmpty() }
+        if (currentUserId != null && currentUserId == registeredOwnerUserId) {
```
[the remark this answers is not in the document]
mrehan27: By design — a same-user re-identify still owns the live registration, so we skip the teardown. Wiping + their re-sync would re-register the same fences and fire a spurious `INITIAL_TRIGGER_ENTER`; their re-identify refresh keeps it current. Documented in `reset_givenSameUserReidentifiedBeforeReset_expectSkipWipe`.

core/src/main/kotlin/io/customer/sdk/data/store/PendingDeliveryFlusher.kt
```diff
@@ -81,14 +95,17 @@ class PendingDeliveryFlusher<T : PendingDeliveryStore.PendingDeliveryEntry>(
...
-                        publish(entry)
+                        // Order claim vs. publish per the caller's crash-safety guarantee (see [DeliveryGuarantee]).
+                        when (guarantee) {
+                            DeliveryGuarantee.AT_MOST_ONCE -> {
+                                if (!store.claim(entry.key)) return@forEach
+                                publish(entry)
+                            }
+                            DeliveryGuarantee.AT_LEAST_ONCE -> {
+                                publish(entry)
+                                store.remove(entry.key)
```
Shahroz16, line 114: Could we keep the row until this handoff is acknowledged by a durable queue? In this call site, `publish` is `EventBus.publish`, which only calls `eventChannel.trySend` on an in-memory channel and returns `Unit`; the DataPipeline subscriber and its queue write happen later. `PendingDeliveryFlusher` removes the disk row as soon as this callback returns, so a process death after `trySend` but before the subscriber persists the event still loses it. This ordering narrows the old window, but it does not provide the `AT_LEAST_ONCE` guarantee described above.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
Shahroz16: I think there is still a quick account-switch ordering where the new user refresh can finish without taking ownership. `registeredOwnerUserId` is assigned only after an actual registration; `RefreshAction.SKIP` leaves it as the previous owner, or nil after process restart. Because reset and refresh run concurrently, user B can identify and take the fresh-cache SKIP path before reset acquires the mutex. Reset then sees B != owner, clears the OS registrations and user-scoped state, and the refresh has already completed, so nothing re-registers until a later trigger. Could the SKIP path transfer ownership of the still-valid registration, or could reset schedule or await a refresh when it clears while another user is identified? A regression test with refresh(SKIP) completing before reset would cover this ordering.

core/src/main/kotlin/io/customer/sdk/data/store/PendingDeliveryFlusher.kt
```diff
@@ -40,6 +34,22 @@ class PendingDeliveryFlusher<T : PendingDeliveryStore.PendingDeliveryEntry>(
...
+    enum class DeliveryGuarantee {
+        /**
+         * Claim (remove) before publishing. If the process dies after the claim the entry is gone,
+         * so it delivers at most once. Pair with a worker that also claims — together exactly-once.
+         */
+        AT_MOST_ONCE,
+
+        /**
+         * Publish, then remove only after it returns. A crash between the two leaves the entry for
+         * the next flush to re-publish, so it delivers at least once; requires a stable payload id
```
Shahroz16: Could we either make the pending-store rewrite crash-atomic before calling this `AT_LEAST_ONCE`, or narrow the crash-safety guarantee in these docs? `store.remove` eventually uses `File.writeText`, which truncates and rewrites the JSON file. A process death during this post-publish removal can leave invalid JSON, and `readAll` then treats the queue as empty. That leaves a narrow but real loss path after publish; serializing the batch under a mutex does not make the file replacement crash-atomic.

Shahroz16 (approved):
I left three comments about foreground delivery durability and one concurrent account-switch ordering. The same-user case Cursor raised is protected when the registration owner is present; the remaining case is when the fresh-cache SKIP path completes without transferring ownership.

core/src/main/kotlin/io/customer/sdk/data/store/PendingDeliveryFlusher.kt
```diff
@@ -81,14 +103,17 @@ class PendingDeliveryFlusher<T : PendingDeliveryStore.PendingDeliveryEntry>(
...
+                        // Order claim vs. publish per the caller's crash-safety guarantee (see [DeliveryGuarantee]).
+                        when (guarantee) {
+                            DeliveryGuarantee.AT_MOST_ONCE -> {
+                                if (!store.claim(entry.key)) return@forEach
+                                publish(entry)
+                            }
+                            DeliveryGuarantee.AT_LEAST_ONCE -> {
+                                publish(entry)
+                                store.remove(entry.key)
+                            }
```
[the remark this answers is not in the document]
mrehan27: Accurate, but this is within the `AT_LEAST_ONCE` contract rather than a correctness bug. Duplicates are deduped downstream by `transitionId` (backend dedupes on `(transitionId, geoset_id)`), so no duplicate event reaches the customer — it's the same duplicate-tolerance the design already relies on between the worker (direct HTTP) and the foreground flush. The `AT_MOST_ONCE` path (push) avoids it via the atomic `claim`; `AT_LEAST_ONCE` trades that for no-loss, accepting dedupable duplicates. The only cost is a bit of wasted work when two foreground drains overlap with a non-empty store, so we're keeping it as-is.

PR #778, opened 2026-07-15
chore(core): fail pending-delivery claim when its removing write fails

mrehan27:
## What

Addresses a Bugbot comment on #763 ("Claim ignores failed store write"). No Linear ticket — this is PR-review-comment handling for the geofence feature branch.

`PendingDeliveryStore.claim()` returned `true` even when the underlying `writeAll()` failed. On a write failure the entry is still on disk, so an `AT_MOST_ONCE` foreground flush (push) would treat the row as claimed and publish it while the WorkManager worker — which also sees the still-present row — could deliver it too. That breaks at-most-once coordination.

Fix: `claim()` returns the write result. A failed claim now makes the caller back off and leave the still-present entry for the other channel, so exactly one channel delivers.

## Scope

- One-line source change in `:core` (`PendingDeliveryStore.claim`) + return-contract KDoc.
- Affects the push at-most-once path too, not just geofence.
- No public API change (`claim` already returned `Boolean`; `apiCheck` clean).

## Tests

- `claim_givenRemovingWriteFails_expectFalseAndEntryPreserved` — store returns `false` and keeps the row on write failure.
- `flush_givenAtMostOnceAndClaimWriteFails_expectNotPublishedAndEntryRetained` — the flusher backs off (no publish, rows retained) end-to-end. This back-off branch was previously untested at any layer.
- Both verified to fail without the fix. `:core`, `:datapipelines`, `:messagingpush` suites green.

## Deferred (not in this PR)

Bugbot's companion comment ("Null userId rows never drain") is unreachable: the receiver drops null/empty-userId transitions before persisting, and the feature is unreleased so there are no legacy rows — the worker/flush null-userId branches are dead defensive code.

---

> [!NOTE]
> **Medium Risk**
> Changes delivery deduplication and disk persistence for push/geofence pending metrics; behavior shifts on IO failures but reduces duplicate-delivery risk rather than expanding attack surface.
> 
> **Overview**
> Fixes **at-most-once** coordination when a foreground flush and WorkManager both use the pending-delivery store: `PendingDeliveryStore.claim()` now returns **`false` unless the removing persist succeeds**, so a failed claim no longer implies ownership while the row is still on disk.
> 
> **`writeAll`** persists via a **temp file + atomic rename**, so interrupted writes keep the previous JSON intact and align with “failed claim ⇒ entry still present.”
> 
> **`claimSendRestore`** treats a failed claim as **`AlreadyClaimed` only if the entry is gone**; if it is still present (e.g. write failed), it returns **`Retryable`** and does not call `send`.
> 
> **`PendingDeliveryFlusher`** adjusts ordering: **AT_MOST_ONCE** cancels the worker then **backs off without publishing** when claim fails; **AT_LEAST_ONCE** **publishes before remove**, cancels the worker only after a successful publish, and **ignores cancel failures** so a delivered row is not left behind or marked failed.
> 
> New **core** and **messagingpush** tests cover failed claim writes and the updated flush/worker paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 15d0d34d34cd6073b1a9400e367b79983aa80476. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/core/src/main/kotlin/io/customer/sdk/data/store/PendingDeliveryStore.kt
+++ b/core/src/main/kotlin/io/customer/sdk/data/store/PendingDeliveryStore.kt
@@ -108,14 +108,16 @@ class PendingDeliveryStore<T : PendingDeliveryStore.PendingDeliveryEntry>(
      * channel whose [claim] returns true owns the send, the other backs off.
      * A read-only check is not enough — claim-then-send must be atomic, or a
      * slow send lets both channels act on the same still-present entry.
-     * Returns true if this call removed the entry, false if it was already gone.
+     * Returns true only if this call actually removed the entry from storage:
+     * false when it was already gone, and also false when the removing write
+     * failed, so the caller backs off and leaves the still-present entry for
+     * the other channel rather than sending a duplicate.
      */
     fun claim(key: String): Boolean = lock.withLock {
         val entries = readAll()
         val filtered = entries.filterNot { it.key == key }
         if (filtered.size == entries.size) return@withLock false
         writeAll(filtered)
-        true
     }
 
     /**
--- a/core/src/main/kotlin/io/customer/sdk/data/store/PendingDeliveryStore.kt
+++ b/core/src/main/kotlin/io/customer/sdk/data/store/PendingDeliveryStore.kt
@@ -108,10 +108,11 @@ class PendingDeliveryStore<T : PendingDeliveryStore.PendingDeliveryEntry>(
      * channel whose [claim] returns true owns the send, the other backs off.
      * A read-only check is not enough — claim-then-send must be atomic, or a
      * slow send lets both channels act on the same still-present entry.
-     * Returns true only if this call actually removed the entry from storage:
-     * false when it was already gone, and also false when the removing write
-     * failed, so the caller backs off and leaves the still-present entry for
-     * the other channel rather than sending a duplicate.
+     * Returns true only when the removing write completed and this call owns
+     * the send; false when the entry was already gone or the write failed. A
+     * false return means "you did not claim it — back off", not that the entry
+     * definitely survived: [writeAll] isn't crash-atomic, so a mid-write failure
+     * can still leave the file blank or partial.
      */
     fun claim(key: String): Boolean = lock.withLock {
         val entries = readAll()
--- a/core/src/main/kotlin/io/customer/sdk/data/store/PendingDeliveryStore.kt
+++ b/core/src/main/kotlin/io/customer/sdk/data/store/PendingDeliveryStore.kt
@@ -4,6 +4,7 @@ import android.content.Context
 import io.customer.base.internal.InternalCustomerIOApi
 import io.customer.sdk.core.util.Logger
 import java.io.File
+import java.io.IOException
 import java.util.concurrent.locks.ReentrantLock
 import kotlin.concurrent.withLock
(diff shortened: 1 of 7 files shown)


core/src/main/kotlin/io/customer/sdk/data/store/PendingDeliveryStore.kt
```diff
@@ -108,14 +108,16 @@ class PendingDeliveryStore<T : PendingDeliveryStore.PendingDeliveryEntry>(
      * channel whose [claim] returns true owns the send, the other backs off.
      * A read-only check is not enough — claim-then-send must be atomic, or a
      * slow send lets both channels act on the same still-present entry.
-     * Returns true if this call removed the entry, false if it was already gone.
+     * Returns true only if this call actually removed the entry from storage:
+     * false when it was already gone, and also false when the removing write
+     * failed, so the caller backs off and leaves the still-present entry for
```
Shahroz16: Could we narrow this wording or make writeAll atomic? The read-only test fails before the file is opened, so the old JSON survives. File.writeText can also fail after truncation and leave a blank or partial file; in that case false does not guarantee a still-present entry for any channel. A temp-file plus atomic replace would make this new contract true; otherwise the KDoc should describe the narrower guarantee.

core/src/test/java/io/customer/sdk/data/store/PendingDeliveryFlusherTest.kt
```diff
@@ -138,6 +143,28 @@ class PendingDeliveryFlusherTest : RobolectricTest() {
...
 
+    @Test
+    fun flush_givenAtMostOnceAndClaimWriteFails_expectNotPublishedAndEntryRetained() {
+        val store = newStore()
+        listOf("a", "b").forEach { store.append(TestEntry(it)) }
+        val callbacks = RecordingCallbacks()
+        val publishedKeys = mutableListOf<String>()
+
+        // Entries stay readable but the claiming write can't land, so claim reports failure and the
+        // flush must back off — leaving the rows for the worker rather than publishing a duplicate.
```
Shahroz16: Could we keep WorkManager available in this regression test, or add a companion case for that production path? The flusher currently awaits cancellation before claim, so when this new false branch is a write failure the row remains but its worker has already been cancelled. The next foreground flush can still retry, so I do not see this as blocking the fix, but the comment's 'leaving the rows for the worker' is not what the current ordering does. It would be useful to either assert and document that delayed retry, or avoid cancelling the AT_MOST_ONCE worker until after the foreground claim wins.

Shahroz16 (approved):
The focused core tests pass. I left two follow-ups about what the new failure contract guarantees and what the regression test currently exercises. There is a failing `kotlin_compose` instrumentation job, which should be unrelated here i believe

core/src/main/kotlin/io/customer/sdk/data/store/PendingDeliveryStore.kt
mrehan27: Updated

core/src/test/java/io/customer/sdk/data/store/PendingDeliveryFlusherTest.kt
mrehan27: Updated

PR #782, opened 2026-07-16
fix(inbox): overlay QA fixes, custom fonts + dedicated CDP metric events

mahmoud-elmorabea:
Visual Inbox overlay QA fixes surfaced during device testing, plus web-parity inbox metric reporting and host-supplied custom fonts.

- **MBL-2123** — render the branding bell by filling each `<path>` independently (even-odd per path) so overlapping subpaths don't flip winding parity.
- **MBL-2124** — compact drag handle / top margin on the sheet.
- **MBL-2126** — badge background + text color driven by branding.
- **MBL-2127** — badge proportions/offset.
- **MBL-2125** — report inbox opened/clicked via the generic `Report Delivery Event` (`metric: opened|clicked`), matching the web SDK (`cdp-analytics-js`). The CDP backend renders these as "Opened/Clicked Inbox Message" for an inbox delivery. Click carries `actionName` + `actionValue` (web parity). **No** separate named CDP events and **no** client-side "delivered" — the backend synthesizes delivered.
- Expose a public `fonts` map on `NotificationInboxOverlay` so hosts can supply custom font families for Jist theme resolution (regenerated public API dump).

Pre-flight: `:messaginginbox:assembleDebug`, `:messaginginapp:assembleDebug`, `:messaginginbox:testDebugUnitTest`, `:messaginginapp:testDebugUnitTest`, `apiDump`, `ktlint --android` all green locally.

> Custom fonts render end-to-end only once the Jist quote-stripping fix (#43) lands and republishes; sample-app font scaffolding is held for a follow-up PR.

---

> [!NOTE]
> **Medium Risk**
> Touches public SDK/Compose APIs and changes inbox click-metric payload and dedupe semantics, which can affect analytics parity and integrators; UI/branding changes are localized to the inbox overlay.
> 
> **Overview**
> Visual inbox overlay fixes from device QA, plus **web-aligned inbox analytics** and an optional **custom-font** hook for Jist.
> 
> **Analytics (MBL-2125):** Inbox clicks now flow through the existing generic delivery metric (`Metric.Clicked`) with both `actionName` and `actionValue` when present; empty strings are omitted. A new **three-argument** `trackMessageClicked` overload carries the URL/deeplink without changing the two-arg `$default` synthetic (binary compatibility). The overlay controller forwards the resolved action value and **dedupes per distinct action** (queue + name + value) so multiple CTAs on one message each report, while identical taps stay deduped.
> 
> **Overlay UI:** Branding bell SVGs render **one Compose path per `<path>`** with per-path fill rules (MBL-2123). Sheet uses a **compact drag handle** and tighter list/message padding (MBL-2124). Unread badge **color, text color, and `fontSize`** come from `unreadIndicator` branding, with dark-mode overrides and sensible contrast fallbacks (MBL-2126/2127). Default and branded bells share a fixed glyph size.
> 
> **Public API:** `NotificationInboxOverlay` / `NotificationInboxView` accept an optional `fonts: Map<String, FontFamily>` wired through `JistTheme` for theme `fontFamily` resolution (API dump updated).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f023a767e0d04770ea045fe2a7022164d7345701. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/inbox/NotificationInbox.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/inbox/NotificationInbox.kt
@@ -239,13 +239,16 @@ class NotificationInbox internal constructor(
      *
      * @param message The inbox message that was clicked
      * @param actionName Optional name of the action clicked (e.g., "view_details", "dismiss")
+     * @param actionValue Optional value of the action clicked (e.g., the deep link / URL). Included
+     *   in the `Report Delivery Event` (metric: clicked) alongside the name, matching web.
      */
     @JvmOverloads
-    fun trackMessageClicked(message: InboxMessage, actionName: String? = null) {
+    fun trackMessageClicked(message: InboxMessage, actionName: String? = null, actionValue: String? = null) {
         inAppMessagingManager.dispatch(
             InAppMessagingAction.InboxAction.TrackClicked(
                 message = message,
-                actionName = actionName
+                actionName = actionName,
+                actionValue = actionValue
             )
         )
     }
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/inbox/NotificationInbox.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/inbox/NotificationInbox.kt
@@ -239,11 +239,27 @@ class NotificationInbox internal constructor(
      *
      * @param message The inbox message that was clicked
      * @param actionName Optional name of the action clicked (e.g., "view_details", "dismiss")
-     * @param actionValue Optional value of the action clicked (e.g., the deep link / URL). Included
-     *   in the `Report Delivery Event` (metric: clicked) alongside the name, matching web.
      */
     @JvmOverloads
-    fun trackMessageClicked(message: InboxMessage, actionName: String? = null, actionValue: String? = null) {
+    fun trackMessageClicked(message: InboxMessage, actionName: String? = null) {
+        trackMessageClicked(message, actionName, actionValue = null)
+    }
+
+    /**
+     * Tracks a click event for an inbox message, including the clicked action's value.
+     * Sends metric event to data pipelines to track message interaction.
+     *
+     * This is a separate overload rather than a third defaulted parameter on the two-arg method
+     * above: adding a default to that method would change its generated `$default` synthetic
+     * signature and break precompiled Kotlin integrations that call `trackMessageClicked(message)`
+     * / `trackMessageClicked(message, actionName)` against a released SDK (`NoSuchMethodError`).
+     *
+     * @param message The inbox message that was clicked
+     * @param actionName Name of the action clicked (e.g., "view_details", "dismiss")
+     * @param actionValue Value of the action clicked (e.g., the deep link / URL). Included in the
+     *   `Report Delivery Event` (metric: clicked) alongside the name, matching web.
+     */
+    fun trackMessageClicked(message: InboxMessage, actionName: String?, actionValue: String?) {
         inAppMessagingManager.dispatch(
             InAppMessagingAction.InboxAction.TrackClicked(
                 message = message,
--- a/messaginginbox/src/main/java/io/customer/messaginginbox/NotificationInboxOverlay.kt
+++ b/messaginginbox/src/main/java/io/customer/messaginginbox/NotificationInboxOverlay.kt
@@ -14,6 +14,7 @@ import androidx.compose.foundation.clickable
 import androidx.compose.foundation.isSystemInDarkTheme
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.fillMaxHeight
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
@@ -22,6 +23,7 @@ import androidx.compose.foundation.layout.heightIn
 import androidx.compose.foundation.layout.offset
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.layout.widthIn
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.items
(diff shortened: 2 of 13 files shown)


messaginginbox/src/main/java/io/customer/messaginginbox/VisualInboxOverlayState.kt
```diff
@@ -169,6 +169,8 @@ internal class VisualInboxController(
                 .forEach { message ->
                     markedOpenedQueueIds.add(message.queueId)
                     visualInbox.markMessageOpened(message)
+                    // Web-parity CDP event (MBL-2125): `Opened Inbox Message` { delivery_id }.
+                    visualInbox.trackInboxMessageOpenedEvent(message)
```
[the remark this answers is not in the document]
mahmoud-elmorabea: The `Delivered Inbox Message` event was removed in the metric rework (commit 4ce6929a) — the backend synthesizes it, matching web — so the Opened-vs-Delivered ordering no longer applies. The residual point (`markOpenMessagesOpened` marking all visible messages opened when the sheet opens) is pre-existing behavior unrelated to this PR's metric changes; we'll track that separately.

messaginginbox/src/main/java/io/customer/messaginginbox/NotificationInboxOverlay.kt
```diff
@@ -710,9 +823,17 @@ private fun rememberInboxColors(branding: Branding? = null): InboxColors {
...
+        // off-brand literal red, so the badge stays on-brand. (If the branded value is a token
+        // reference string rather than a hex color, it can't be dereferenced here — see fix report.)
         val badgeColor = dark.childStr("unreadIndicator", "background").toColorOrNull()
             ?: light?.unreadIndicator?.background.toColorOrNull()
-            ?: Color(0xFFE53935)
+            ?: bellColor
+        // MBL-2126: badge label color from `unreadIndicator.text.color`, else white for contrast.
+        val badgeTextColor = dark.grandChildStr("unreadIndicator", "text", "color").toColorOrNull()
+            ?: (light?.unreadIndicator?.text?.get("color") as? String).toColorOrNull()
+            ?: Color.White
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Fixed in 4ce6929a. When `unreadIndicator.background` is unset, the badge text color now falls back to a luminance-contrast color (black/white) against the resolved badge background — mirroring `bellIconColor` — instead of hard white, so the count stays legible on a light bell.

messaginginapp/src/main/java/io/customer/messaginginapp/inbox/NotificationInbox.kt
```diff
@@ -239,13 +239,16 @@ class NotificationInbox internal constructor(
      *
      * @param message The inbox message that was clicked
      * @param actionName Optional name of the action clicked (e.g., "view_details", "dismiss")
+     * @param actionValue Optional value of the action clicked (e.g., the deep link / URL). Included
+     *   in the `Report Delivery Event` (metric: clicked) alongside the name, matching web.
      */
     @JvmOverloads
-    fun trackMessageClicked(message: InboxMessage, actionName: String? = null) {
+    fun trackMessageClicked(message: InboxMessage, actionName: String? = null, actionValue: String? = null) {
```
Shahroz16: Could we keep the existing two-parameter method as a wrapper and add a separate three-parameter overload for `actionValue`? We should still send `actionValue`; the concern is that adding another default parameter changes the generated JVM method. A precompiled Kotlin integration that called `trackMessageClicked(message)` against an older SDK expects the previous `$default` signature. With the current change, that signature disappears and the integration could fail at runtime with `NoSuchMethodError`. Keeping the wrapper would preserve compatibility while the new overload carries `actionValue`.

messaginginbox/src/main/java/io/customer/messaginginbox/VisualInboxOverlayState.kt
```diff
@@ -302,17 +305,29 @@ internal class VisualInboxController(
...
+        visibility: InboxVisibility,
+        message: JistInboxMessage,
+        actionName: String?,
+        actionValue: String?
+    ) {
         val visible = visibility as? InboxVisibility.Visible ?: return
         val tracked = visible.messages.firstOrNull { it.queueId == message.queueId } ?: return
         // Reserve only AFTER confirming the message exists, so a failed lookup never permanently
         // dedupes (blocks) the click metric for later taps on the same message.
         if (!clickedQueueIds.add(message.queueId)) return
```
Shahroz16: Could we avoid keeping `clickedQueueIds` for the lifetime of the controller? A message can contain more than one CTA, so this sends only the first action’s name and value to CDP. Web emits an inbox click for every action, and the backend records each click response while deduplicating only the delivery’s first-click metric. If this guard is only intended to prevent rapid double taps, could it be in-flight or short-lived instead, with a test covering two different CTAs on the same message?

messaginginapp/src/main/java/io/customer/messaginginapp/inbox/NotificationInbox.kt
mahmoud-elmorabea: Good catch — confirmed this is a real break for released Kotlin integrations. `trackMessageClicked(message, actionName? = null)` is `@JvmOverloads` and has shipped since 4.16.0 (current main is 4.19.0), so the `$default` synthetic `(…, InboxMessage, String, int, Object)` is public API. Adding a third defaulted param changed that synthetic to `(…, String, String, int, Object)` — Java callers bind to the generated overloads and are fine, but precompiled *Kotlin* callers that omit `actionName` route through `$default` and would hit `NoSuchMethodError`.

Fixed in f023a767: the two-arg `@JvmOverloads` method is restored exactly as on main (delegating), and `actionValue` now lives on a separate three-arg overload `trackMessageClicked(message, actionName, actionValue)` with no defaults. `apiCheck` diff is now purely additive — only `(InboxMessage, String, String)` is added; the `$default` signature reverts to main. Added a test for the 3-arg overload.

messaginginbox/src/main/java/io/customer/messaginginbox/VisualInboxOverlayState.kt
mahmoud-elmorabea: Fixed in f023a767. The dedup is now keyed by the DISTINCT action (`Triple(queueId, actionName, actionValue)`) rather than by `queueId` alone, so a message with multiple CTAs reports a click for each — matching web (the backend dedupes the delivery’s first-click metric). A repeat of the *same* action (rapid double-tap of one CTA) is still suppressed. Added `handleAction_givenTwoDifferentCtasSameMessage_expectEachTracked` (two CTAs → two tracks) alongside the existing same-action-twice → tracked-once test.

I kept it as a set rather than a time-window/in-flight guard so the behavior stays deterministic and unit-testable; happy to switch to a short-lived window instead if you’d prefer strict "emit every tap, let the backend dedupe" parity.

Shahroz16 (approved):

PR #785, opened 2026-07-17
chore(sample): align sample default location mode with SDK default

mrehan27:
## What

Two sample-only alignments for the geofence/location screens:

1. **Default location mode** — the java sample defaulted to `ON_APP_START`, which differs from the SDK's own default of `MANUAL`. Align it with out-of-the-box SDK behavior. `ON_APP_START` and `OFF` remain selectable in settings — only the initial default changes.
2. **Background-location prompt** — drop the redundant "Open Settings" button from the background-location rationale dialog. On Android 11+ "Continue" already deep-links to the app's exact Location permission screen (and shows the inline dialog on API 29); "Open Settings" only reached the generic App info page — extra taps for no benefit. The separate foreground-denied and `DENIED` recovery dialogs keep their "Open Settings" (there, no in-app request can open settings).

## Changes

- `CustomerIOSDKConfig.getDefaultConfigurations()`: `LocationTrackingMode.ON_APP_START` → `LocationTrackingMode.MANUAL`.
- `LocationTestActivity.showBackgroundLocationRationale()`: remove the neutral "Open Settings" button; reword the rationale string to point at Continue.

Sample-only; no SDK code touched.

Base: `feature/geofence-on-device`.

[MBL-2148](https://linear.app/customerio/issue/MBL-2148/mobile-align-sample-app-location-mode-with-sdk-defaults)

---

> [!NOTE]
> **Low Risk**
> Changes are confined to the Java sample app (config default, strings, and one dialog); no SDK or production library code is modified.
> 
> **Overview**
> **Sample-only** tweaks so the Java layout app matches SDK defaults and clarifies background-location onboarding.
> 
> The default in `CustomerIOSDKConfig.getDefaultConfigurations()` changes from `LocationTrackingMode.ON_APP_START` to **`MANUAL`**, matching the location module builder default; other modes stay selectable in settings.
> 
> On **Location Test**, the background-location rationale dialog no longer offers a neutral **Open Settings** button—**Continue** alone launches the background permission flow (app location screen on Android 11+). Copy and comments are updated to describe that flow; foreground-denied recovery still uses Open Settings on the main permission button when appropriate.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7688d97a66eba3be18ffa731ebba61b94c6e27ab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #786, opened 2026-07-20
chore(geofence): harden movement refresh, cooldown, and init threading

mrehan27:
## What & why

Batch of review follow-ups on the geofence module (no public API change, all internal).

**Delivery — movement-trigger EXIT**
`onMovementTriggerExit` now returns the refresh `Job` so the broadcast receiver can hold its `goAsync` window open until the refresh lands. The movement trigger usually fires with the process backgrounded, where the OS can kill it mid-fetch / mid-re-registration the moment the receiver returns. The wait spends only what's left of an 8s dispatch budget measured from dispatch entry — the persistence/GMS awaits count against it, so the end-to-end bound holds under the ~10s goAsync window (headroom included). Only the *wait* is bounded — the refresh itself is never cancelled and self-completes if the process survives. Sequenced after business-transition persistence so waiting can't starve it.

**Refresh vs. sign-out reset race**
`refresh()` now makes its REMOTE/LOCAL/SKIP decision under `stateMutex` (pref reads only; the network call stays outside the lock). Previously a sign-out reset that had passed its no-user check but was still awaiting the GMS clear could let a concurrent identify read the pre-wipe freshness/registrations, SKIP, and then lose its registrations to the wipe — leaving the newly identified user unmonitored for the session. Either ordering is now safe: a decision after the wipe sees stale state and refreshes; a decision before the reset is superseded by the reset's signed-in recheck.

**Cooldown lifecycle**
- Suppression windows are now **user-scoped** (matching iOS): keys are `userId:geofenceId:transition`, so after an account switch the previous user's window can never mask the new user's transition on the same fence. The receiver already snapshots `userId` before the cooldown check (anonymous transitions drop earlier), so the value is always present.
- Cooldown history is cleared inside the repository's guarded reset instead of the `ResetEvent` subscriber, so it shares the superseded-reset guard; on a genuine sign-out it's wiped even if the GMS clear fails (with user-scoped keys this is data hygiene, not cross-user correctness).
- Added `GeofenceCooldownStore.pruneOlderThan`, swept once per allowed emit, to bound the store as fence definitions churn. Cutoff is `MAX_DUPLICATE_EVENTS_EXPIRY_MS` (24h) — the coerced upper bound of `duplicateEventsExpiry`, so nothing prunable could ever still suppress. It also sweeps orphaned pre-user-scoping keys.

**App-launch init threading**
The persisted-user read + defensive sync run off the main thread now (`getUserId()` hits SharedPreferences + a Keystore decrypt, which can block for hundreds of ms on some OEMs). The one-shot scope cancels itself when done. `ProcessLifecycleOwner` observer registration stays on the main thread.

**API decode robustness**
`GeofenceApiRegion.radius` decodes as `Double` so a fractional backend value (e.g. `150.5`) can't fail the whole `/geofences/nearest` response; the domain type stays `Float`.

**Docs / comments**
- Worker + tracker KDoc now state the receiver drops anonymous transitions before persisting, so a null `userId` there is defensive-only (not a real path).
- Rewrote the manual-test doc from the old exactly-once model to the current at-least-once + backend-dedupe-by-`transitionId` model.
- Added a `SecureUserStore` round-trip test.

## Testing
- `:geofence` + `:core` unit tests green (added coverage: goAsync wait bounded/not-cancelled/skipped-when-budget-spent, reset-vs-identify race regression with a gated GMS clear, cross-user cooldown independence at filter + store level, prune sweep on allowed vs suppressed emit, cutoff exclusivity + underscore-safe keys, fractional-radius decode, cooldown-survives-superseded-reset, SecureUserStore round-trip).
- ktlint + `apiCheck` clean.

## Ticket
https://linear.app/customerio/issue/MBL-2152/android-address-geofencing-improvements

## Stack
```
main
 └─ feature/geofence-on-device
     └─ mbl-2152-geofence-improvements  ← this PR
```
Base: `feature/geofence-on-device`. Rebased onto feature after #785 merges so the diff stays clean.

---

> [!NOTE]
> **Medium Risk**
> Touches background delivery (broadcast receiver timing), refresh/sign-out concurrency, and duplicate-event suppression—important for monitoring correctness but no public API surface and heavily tested.
> 
> **Overview**
> Internal geofence hardening with **no public API changes**.
> 
> **Movement-trigger EXIT + `goAsync`:** `onMovementTriggerExit` now returns the refresh `Job`; the broadcast receiver waits (up to remaining time in an **8s** dispatch budget) so background processes are less likely to die mid re-registration. The refresh itself is never cancelled—only the wait is bounded.
> 
> **Refresh vs sign-out:** `refresh()` chooses REMOTE/LOCAL/SKIP under `stateMutex` so a concurrent sign-out wipe cannot produce a SKIP on stale state and leave a newly identified user without monitoring.
> 
> **Cooldown:** Duplicate suppression is **user-scoped** (`userId:geofenceId:transition`, aligned with iOS). Cooldown is cleared inside `GeofenceRepository.reset()` (with the superseded-reset guard), including when GMS clear fails. Allowed emits trigger `pruneOlderThan` to cap store growth.
> 
> **Init:** Persisted-user read + defensive app-launch sync run off the main thread (Keystore/SharedPreferences); lifecycle observer registration stays on main.
> 
> **Decode:** API `radius` decodes as `Double` so fractional values do not fail the whole nearest-geofences response.
> 
> **Docs/tests:** Manual plan updated to at-least-once delivery + `Geofence Transition` events; added `SecureUserStore` round-trip tests and expanded unit coverage for the above behaviors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5f2c94a47e8fe432f53c750050334f638b86cdc2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
```diff
@@ -372,25 +370,23 @@ internal class GeofenceRepositoryImpl(
...
+        // Clear OS-side first. On failure, preserve store state so the next refresh's stale-cleanup
+        // diff can retry removal (unremoved OS regs would otherwise orphan). Cached regions/config
+        // are kept; the freshness timestamp is dropped so the next login re-fetches.
         manager.clearAll().also { result ->
             if (result.isSuccess) {
                 store.clearUserScopedState()
             }
+            // Cooldown is user-scoped suppression: wipe it on any genuine sign-out even if the OS
+            // clear failed, so the next user can't inherit the prior user's stale windows.
+            cooldownFilter.clearAll()
```
Shahroz16, line 393: Could we make this guard cover the `clearAll()` suspension too? I can reproduce this ordering: reset reads no user and starts `manager.clearAll()`, user B identifies, and B's refresh sees the previous fresh timestamp and registered IDs and returns `SKIP` because that decision is outside `stateMutex`. When `clearAll()` then succeeds, reset removes the registrations and user-scoped state, leaving B with no movement trigger or business fences until another launch or explicit refresh. This needs a quick `clearIdentify()` followed by `identify()` while the asynchronous GMS clear is in flight, so it is not every reset, but it is a practical silent coverage gap. Could we serialize refresh's identity and freshness decision with `stateMutex` and add this interleaving as a regression test?

geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
```diff
@@ -189,6 +192,11 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
                 }
             }
         }
+
+        // Hold the goAsync window open until the refresh lands so the OS doesn't kill a
+        // backgrounded process mid-re-registration. Bounded; a timeout ends the wait only,
+        // not the refresh (it runs on the longer-lived services scope).
+        movementRefreshJob?.let { job -> withTimeoutOrNull(MOVEMENT_REFRESH_WAIT_TIMEOUT_MS) { job.join() } }
```
Shahroz16: Could we make this timeout use the remaining receiver budget? This eight-second timer starts only after unknown-ID cleanup and all business entries have been persisted and scheduled. The movement job starts concurrently, so normal callbacks should finish well within the budget, but a slow refresh combined with a cold multi-geoset WorkManager batch can still spend the earlier work plus the full eight-second join and exceed the roughly ten-second `goAsync` window mentioned below. Capturing elapsed time at receiver entry and joining only for the remaining budget would make this bound hold end to end. This is a tail case and is not blocking for me.

Shahroz16 (approved):
I found one account-switch race in the reset path that can leave the newly identified user without registered fences for the rest of the session. I reproduced it with a focused coroutine test and left the ordering inline. I think that one needs addressing before approval. I also left one smaller receiver-budget suggestion; that second comment would not block approval on its own.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
mrehan27: Good catch — fixed in 44b01460. The freshness/identity decision now runs under `stateMutex` (pref reads only; the network call stays outside the lock), so it either waits for the reset and sees the wiped state, or — if it decided first — the reset's signed-in recheck supersedes the wipe. Added your interleaving as a regression test (`refresh_givenIdentifyDuringInFlightResetWipe_expectDecisionWaitsAndRefreshes`) with a gated `manager.clearAll()` and identify mid-reset.

geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
mrehan27: Done in 44b01460 — dispatch captures uptime at entry and the join now waits only for the remaining budget (skipped entirely when already spent), so the persistence/GMS awaits and the join share one 8s budget inside the goAsync window. Added a test simulating a spent budget (`dispatchTransition_givenDispatchBudgetAlreadySpent_expectNoWaitForRefreshJob`).

geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
```diff
@@ -189,6 +193,17 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
...
+
+        // Hold the goAsync window open until the refresh lands so the OS doesn't kill a
+        // backgrounded process mid-re-registration. Waits only for what's left of the
+        // dispatch budget — the persistence/GMS awaits above count against it. A timeout
+        // ends the wait only, not the refresh (it runs on the longer-lived services scope).
+        movementRefreshJob?.let { job ->
+            val remainingBudgetMs = DISPATCH_WAIT_BUDGET_MS - (SDKComponent.clock.elapsedRealtime() - dispatchStartUptimeMs)
+            if (remainingBudgetMs > 0) {
+                withTimeoutOrNull(remainingBudgetMs) { job.join() }
+            }
```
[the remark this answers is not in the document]
mrehan27: The pre-dispatch work outside the budget (DI setup, intent parsing, coroutine launch) is all in-memory — no disk, GMS, or IPC awaits — so it costs milliseconds even on a cold start. Every await that can meaningfully consume time (store persistence, GMS removal, WorkManager enqueue, the refresh join) happens after the timestamp and is counted. The ~2s headroom between the 8s budget and the ~10s goAsync window is there to absorb exactly this fixed overhead, so keeping the timestamp at dispatch entry — moving it to onReceive would add state/parameter plumbing to reclaim a few ms of accounting.

PR #787, opened 2026-07-20
chore(geofence): emit ENTER when already inside a geofence at registration

mrehan27:
## What & why

Android relied on GMS `INITIAL_TRIGGER_ENTER` to report a geofence the user is already inside when it's registered. GMS drops this unreliably — confirmed on-device: a fresh install standing inside a business fence registered 6 fences but fired **no** ENTER (only the movement trigger's did). So a user who logs in / opens the app already inside a fence never got the ENTER.

This synthesizes the ENTER ourselves, matching iOS's `emitInitialEnters`.

## How

After a successful registration, for each business fence that is:
- **newly monitored** (not already registered with unchanged params — a re-register of an unchanged set stays silent; a re-added param change counts as new),
- **monitors ENTER**, and
- **contains the device** (point-in-circle against the sync anchor),

emit an ENTER through the normal delivery path.

- **Cooldown-deduped**: if GMS *does* fire its own ENTER, the two collapse to a single event (no doubles).
- **Login parity**: sign-out clears the registered-id set, so the next sign-in re-fires.
- **Boot restore & reboot excluded**: the anchor may predate the reboot / moving-while-off, so containment can't be trusted (post-reboot registration still re-adds everything).
- **Identity rechecked** after the awaited GMS call — a sign-out/switch mid-registration never queues a synthetic ENTER.
- **Refactor**: extracted the transition fan-out (cooldown → per-geoset persist → schedule) into a shared `GeofenceTransitionEmitter` used by both the broadcast receiver and this new path, so both emit identical event rows.

Derived from an iOS↔Android parity sweep — this was the one substantive iOS-only construct; everything else (event shape, cooldown, delivery, reset, config) was already aligned.

## Scope / safety

- No public API change (`apiCheck` green); no DB/schema change; `:geofence` only.
- Behavior change: **more ENTER events** when identifying/opening while already inside a fence — the intended fix.
- Extraction is behavior-preserving: existing receiver/repo tests pass unedited.
- Threading: `emitInitialEnters` runs under the existing `stateMutex` (coroutine mutex), with identity rechecked after the GMS await. No new locks, deadlocks, or scopes; the receiver's hot path takes no `stateMutex`.

## Testing

Unit: 10 repository initial-enter cases (new+inside→emit, params-changed→emit, reboot-wiped→re-registers but none, reboot+stale-cache remote→none, already-registered→none, outside→none, EXIT-only→none, register-fail→none, user-changed-mid-register→none, boot-restore→none) + 5 `GeofenceTransitionEmitter` cases (suppress / single / multi-geoset shared transitionId / scheduler-throw isolation / persist-fail rollback). `:geofence` + `:core` tests, `apiCheck`, ktlint all green.

**On-device (pre-merge):** fresh install inside a fence → expect exactly one `Geofence Transition` (`enter`) per geoset + log `device already inside a newly-registered fence — synthesizing ENTER`; verify no double when GMS also fires; verify sign-out→sign-in re-fires; verify outside→none and boot-restore→no spam.

## Ticket
https://linear.app/customerio/issue/MBL-2152/android-address-geofencing-improvements

## Stack
```
main
 └─ feature/geofence-on-device
     └─ mbl-2152-geofence-initial-enter  ← this PR
```
Base: `feature/geofence-on-device`.

---

> [!NOTE]
> **Medium Risk**
> Intentionally increases ENTER analytics when identifying inside fences; delivery logic is refactored but shared via one emitter with extensive tests—wrong containment or guard bugs could duplicate or miss events.
> 
> **Overview**
> Fixes missing **ENTER** events when the user is already inside a business geofence at registration time—GMS `INITIAL_TRIGGER_ENTER` is unreliable, so the SDK now emits ENTER itself after a successful sync registration (iOS parity).
> 
> **Shared delivery path:** Transition fan-out (cooldown → atomic per-geoset persist → WorkManager schedule) moves into **`GeofenceTransitionEmitter`**. The broadcast receiver delegates OS transitions to it; the repository uses the same emitter so synthetic and real crossings produce identical pending rows and share cooldown dedupe.
> 
> **When synthesis runs:** After registration succeeds, for each nearest business fence that is newly monitored (not unchanged re-register), monitors ENTER, and contains the sync anchor. Skipped on reboot-detected refresh (untrusted anchor), boot restore, failed registration, EXIT-only fences, device outside radius, or user identity change after the GMS await.
> 
> Adds **`logInitialEnterInside`** and broad unit coverage for emitter behavior and repository guardrails.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 673a481dabbf1b85771f060b917151dee794bfed. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
@@ -8,6 +8,7 @@ import io.customer.geofence.api.toDomainRegions
 import io.customer.geofence.store.GeofenceRegionStore
 import io.customer.geofence.store.getCachedConfigOrFallback
 import io.customer.location.LocationCoordinates
+import io.customer.sdk.communication.Event
 import io.customer.sdk.core.util.Clock
 import io.customer.sdk.data.store.SecureUserStore
 import java.util.concurrent.atomic.AtomicBoolean
@@ -56,6 +57,7 @@ internal class GeofenceRepositoryImpl(
     private val manager: GeofenceManager,
     private val secureUserStore: SecureUserStore,
     private val cooldownFilter: GeofenceCooldownFilter,
+    private val transitionEmitter: GeofenceTransitionEmitter,
     private val clock: Clock,
     private val logger: GeofenceLogger
 ) : GeofenceRepository {
@@ -211,7 +213,10 @@ internal class GeofenceRepositoryImpl(
             latitude = effectiveLocation.latitude,
             longitude = effectiveLocation.longitude,
             cachedConfig = cachedConfig,
-            register = manager::replaceGeofencesForBootRestore
+            register = manager::replaceGeofencesForBootRestore,
+            // No initial-enter on boot restore: the cached anchor may be stale if the device moved
+            // while off, so containment can't be trusted.
+            emitInitialEnter = false
         )
     }
 
@@ -261,14 +266,16 @@ internal class GeofenceRepositoryImpl(
         latitude: Double,
         longitude: Double,
         cachedConfig: GeofenceConfig,
-        register: suspend (List<GeofenceRegion>) -> Result<Unit> = ::registerWithBusinessDiff
+        register: suspend (List<GeofenceRegion>) -> Result<Unit> = ::registerWithBusinessDiff,
+        emitInitialEnter: Boolean = true
     ): Result<Unit> = registerNearestAndPersist(
         userId = userId,
         latitude = latitude,
         longitude = longitude,
         regions = store.getCachedRegions(),
         config = cachedConfig,
-        register = register
+        register = register,
+        emitInitialEnter = emitInitialEnter
     )
 
     /**
@@ -308,7 +315,8 @@ internal class GeofenceRepositoryImpl(
         regions: List<GeofenceRegion>,
         config: GeofenceConfig,
         register: suspend (List<GeofenceRegion>) -> Result<Unit> = ::registerWithBusinessDiff,
-        onRegistered: () -> Unit = {}
+        onRegistered: () -> Unit = {},
+        emitInitialEnter: Boolean = true
     ): Result<Unit> {
         // Pure mapping + filter — no shared state, kept outside the lock.
         val nearest = distanceFilter.nearest(
@@ -336,7 +344,14 @@ internal class GeofenceRepositoryImpl(
                 logger.logSyncSkipped("user changed during refresh")
                 return@withLock Result.success(Unit)
             }
-            register(regionsToRegister).also { result ->
+            // Snapshot already-monitored fences before register overwrites the set — initial-enter
+            // fires only for newly-monitored ones. A reboot wipes OS state, so treat all as new.
+            val previouslyRegisteredBusinessIds = if (osStateWipedByReboot()) {
+                emptySet()
+            } else {
+                store.getRegisteredIds() - GeofenceConstants.MOVEMENT_TRIGGER_ID
+            }
+            val registrationResult = register(regionsToRegister).also { result ->
                 if (result.isSuccess) {
                     // Stale cleanup — Manager added new−existing, we remove
                     // existing−new. Runs only on add success; on failure leave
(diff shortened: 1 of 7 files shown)


geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
```diff
@@ -336,7 +344,14 @@ internal class GeofenceRepositoryImpl(
                 logger.logSyncSkipped("user changed during refresh")
                 return@withLock Result.success(Unit)
             }
-            register(regionsToRegister).also { result ->
+            // Snapshot already-monitored fences before register overwrites the set — initial-enter
+            // fires only for newly-monitored ones. A reboot wipes OS state, so treat all as new.
+            val previouslyRegisteredBusinessIds = if (osStateWipedByReboot()) {
```
Shahroz16: Could we avoid synthesizing from this reboot fallback unless the location is known to be current? `ModuleGeofence.refreshAnchor()` prefers the persisted movement-trigger center and does not request a fresh fix while that value exists. After a reboot this branch treats every fence as new, then containment is checked against that persisted point. If the device moved while it was off and `BOOT_COMPLETED` was missed or races app launch, we can emit ENTER for the old location. `restoreFromCache()` already disables synthesis for exactly this stale-anchor reason. Could we either skip synthesis on this reboot path or acquire a fresh fix first?

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
```diff
@@ -370,6 +385,44 @@ internal class GeofenceRepositoryImpl(
                     logger.logSyncSucceeded(nearest.size)
                 }
             }
+            if (registrationResult.isSuccess && emitInitialEnter) {
```
Shahroz16, line 397: Can we recheck `secureUserStore` after `register(...)` returns and before calling `emitInitialEnters`? The current check happens before the suspending GMS call. `stateMutex` makes reset wait, but it does not stop `clearIdentify()` or `identify(B)` from changing the store during that await. When registration completes, this can queue synthetic ENTERs with captured user A even though A has signed out or B is current, and reset does not clear the pending delivery row. iOS stops its synthetic loop as soon as the expected identity changes, and a deferred GMS completion should make this deterministic to test.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
```diff
@@ -370,6 +385,44 @@ internal class GeofenceRepositoryImpl(
...
+    private suspend fun emitInitialEnters(
+        candidates: List<GeofenceRegion>,
+        previouslyRegisteredBusinessIds: Set<String>,
+        userId: String,
+        latitude: Double,
+        longitude: Double
+    ) {
+        val timestamp = clock.currentTimeSeconds()
+        candidates.forEach { region ->
+            val newlyRegistered = region.id !in previouslyRegisteredBusinessIds
```
Shahroz16: Not blocking on this by itself, but could we base `newlyRegistered` on the registration diff rather than raw ID presence? `registerWithBusinessDiff()` re-adds a same-ID fence when its center, radius, or transitions change, while `previouslyRegisteredBusinessIds` still contains that ID. A radius expansion or moved fence that now contains the device can therefore miss the initial ENTER this PR is compensating for when GMS drops its callback. If same-ID geometry changes are intentionally silent, could we document that contract; otherwise passing the actual re-added IDs here seems safer.

Shahroz16 (approved):

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
mrehan27: Went with skipping synthesis on the reboot path — same reasoning as `restoreFromCache`: the launch anchor can predate the reboot, so containment can't be trusted. The re-register itself still happens (GMS gets its `INITIAL_TRIGGER_ENTER` chance), and the case where this drops an event needs the user to have stayed inside through the reboot — where the pre-reboot ENTER usually already fired and cooldown would suppress a re-emit. Acquiring a fresh fix first would need the refresh pipeline to wait on async location acquisition (it currently only consumes a location), which felt too heavy for this edge. Fixed in f01f0d53 (`refresh_givenRebootWipedOsState_expectReRegistrationButNoInitialEnter`).

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
mrehan27: Good catch — identity is now rechecked under the mutex after `register(...)` returns; synthesis skips with a log when it changed. Regression test with a deferred GMS completion: `refresh_givenUserChangesDuringRegistration_expectNoInitialEnter`. Fixed in f01f0d53.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
mrehan27: Done — the synthesis baseline now uses the same skip-safe diff as registration (extracted `unchangedRegisteredIds`), so a re-added param change synthesizes like a new fence. The reboot override deliberately stays registration-only (see the stale-anchor thread). Test: `refresh_givenRegisteredFenceParamsChangedDeviceInside_expectInitialEnter`. Fixed in f01f0d53.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
```diff
@@ -370,6 +387,50 @@ internal class GeofenceRepositoryImpl(
...
+            logger.logInitialEnterInside(region.id)
+            transitionEmitter.emit(
+                geofenceId = region.id,
+                transition = Event.GeofenceTransition.ENTER,
+                userId = userId,
+                timestampSeconds = timestamp,
+                geofenceName = region.name,
+                metadata = region.metadata,
+                geosetIds = region.geosetIds
+            )
```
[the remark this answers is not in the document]
mrehan27: Fixed in 673a481d — `refresh()` now disables synthesis outright while the reboot flag is up, on both the LOCAL and REMOTE branches (a reboot + expired cache re-fetches around the same untrusted anchor). Tests: `refresh_givenRebootWipedOsState_expectReRegistrationButNoInitialEnter` (includes a never-registered fence) and `refresh_givenRebootAndStaleCache_expectRemoteFetchButNoInitialEnter`.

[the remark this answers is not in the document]
mrehan27: Dismissing this one: a per-candidate recheck can't close the window, only shrink it — there is always a gap between the last read and the persist, even inside `emit()`. The recheck after the GMS await already collapsed the exposure from a multi-second suspension to the microseconds of an in-memory loop; closing it fully would require identity writes to take `stateMutex` (cross-module). The residual is an event that would have been legitimate had sign-out landed a moment later.

PR #789, opened 2026-07-21
fix(live-notifications): identify→start race + logout-during-render writeback

mahmoud-elmorabea:
Fixes two live-notification races on the local path.

- **identify()→start() drop:** `DataPipeline.isUserIdentified` is now updated synchronously on the caller thread during identify()/clearIdentify(), so a start/update/end issued immediately after identify() is reported instead of dropped by an async-lagging identity flag. The lifecycle client gates on `pipeline.isUserIdentified` instead of an EventBus-fed cached flag.
- **logout during a slow render:** the handler re-checks the render generation immediately before posting the notification and writing the activity type back, so a logout landing during the branding-logo download / app renderer can't resurface the previous user's activity.

Unit tests updated/added; `:messagingpush` and `:datapipelines` unit suites pass locally.

---

> [!NOTE]
> **Medium Risk**
> Touches identity semantics used across the data pipeline and live-notification lifecycle; behavior change is intentional but affects login/logout timing-sensitive paths.
> 
> **Overview**
> Fixes two **local live-notification races**.
> 
> **identify() → start() drop:** `CustomerIO` now keeps a caller-thread `syncUserIdentified` mirror set in `identify()` / `clearIdentify()`, so `DataPipeline.isUserIdentified` is correct as soon as those calls return (instead of lagging behind async `analytics.userId()`). `LiveNotificationLifecycleClient` gates start/update/end on `pipeline.isUserIdentified` and drops the separate `setIdentified` / EventBus-fed cache.
> 
> **Logout during slow render:** `LiveNotificationHandler` accepts an `isSuperseded` callback checked right before `notify` and `setActivityType`, so a reset during branding download or custom rendering cannot repost or write back into a cleared store. `LiveNotificationManager` wires this to its `renderGeneration` bump on logout.
> 
> Adds `IsUserIdentifiedSyncTests` and updates handler/registrar/lifecycle tests for the new contracts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c41d2d171f8496b013970d5b88f2edec03b7e83d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
+++ b/datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
@@ -117,6 +117,16 @@ class CustomerIO private constructor(
     // Reset on clearIdentify so a subsequent identify of the same userId is not deduped.
     private var lastIdentifiedUserIdThisSession: String? = null
 
+    // Caller-thread mirror of whether a user is currently identified, backing the synchronous
+    // [isUserIdentified]. analytics.userId() is updated asynchronously and lags a synchronous
+    // identify(), so a consumer that gates on it right after login (e.g. a live-notification
+    // start() on the next line) would otherwise read stale state and drop the event. Set on the
+    // caller's thread during identify()/clearIdentify() so it is accurate the instant they return.
+    // null = no identify/reset has happened this process yet, so fall back to the (possibly
+    // restored-from-persistence) analytics userId; true = identified; false = anonymous.
+    @Volatile
+    private var syncUserIdentified: Boolean? = null
+
     init {
         // Set analytics logger and debug logs based on SDK logger configuration
         Analytics.debugLogsEnabled = logger.logLevel == CioLogLevel.DEBUG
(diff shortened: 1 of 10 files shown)


datapipelines/src/main/kotlin/io/customer/sdk/CustomerIO.kt
```diff
@@ -117,6 +117,16 @@ class CustomerIO private constructor(
...
 
+    // Caller-thread mirror of whether a user is currently identified, backing the synchronous
+    // [isUserIdentified]. analytics.userId() is updated asynchronously and lags a synchronous
+    // identify(), so a consumer that gates on it right after login (e.g. a live-notification
+    // start() on the next line) would otherwise read stale state and drop the event. Set on the
+    // caller's thread during identify()/clearIdentify() so it is accurate the instant they return.
+    // null = no identify/reset has happened this process yet, so fall back to the (possibly
+    // restored-from-persistence) analytics userId; true = identified; false = anonymous.
+    @Volatile
+    private var syncUserIdentified: Boolean? = null
```
Shahroz16, line 128: Non-blocking: could we add one CustomerIO-level test for this mirror? The messagingpush tests mock `isUserIdentified`, so they don't prove that the new contract is true immediately after `identify()`, false immediately after `clearIdentify()`, and still restores the persisted identity on cold launch. The production logic looks sound and I wouldn't hold the PR for this, but it would keep the regression fix covered at the layer that owns it.

Shahroz16 (approved):

PR #791, opened 2026-07-22
chore(geofence): drop regions the OS would reject instead of crashing the sync

mrehan27:
## Summary

A single malformed region in the `/geofences/nearest` response could crash the host app: `Geofence.Builder.build()` throws `IllegalArgumentException` for `radius <= 0`, `|lat| > 90`, or `|lng| > 180`, the whole batch was built outside any try/catch, and the geofence coroutine scope has no exception handler. Since the region cache only saves on success, every retry re-fetched the same bad region — a crash loop until the backend data changes.

Linear: [MBL-2157](https://linear.app/customerio/issue/MBL-2157/mobile-validate-geofence-data-before-registering-with-the-os)

## Changes

- **Parse-time validation**: `toDomain()` returns null for regions violating `Geofence.Builder` preconditions — dropped with a log, so one bad region costs itself, not the whole sync. Parse time is the only safe layer for per-region drops: dropping after the registration diff would mark regions as registered that never reached the OS.
- **Unusable-response guard**: an unexpected mapper throw drops only the affected region (logged distinctly from invalid values); a non-empty response whose regions *all* drop — by throw or by validation — fails the refresh instead of succeeding empty, preserving live registrations and cache. A legitimately empty response still clears normally, keeping the kill-switch semantics.
- **Batch-build guard**: `GeofencingRequest` construction is wrapped so a GMS rejection fails the sync as a normal `Result.failure` instead of throwing.
- **Validator reuse**: `isValidCoordinate` moved from `LocationServicesImpl`'s companion to `LocationCoordinates.isValid()` (already a public `@InternalCustomerIOApi` type), shared by `:location` manual input validation and geofence parse validation. No API-dump change — the marker is in `nonPublicMarkers` (the companion needs its own annotation; the validator checks nested classes individually).

Malformed JSON was already safe (decoded inside `mapCatching` → `Result.failure`).

## Stack

```
main ← feature/geofence-on-device ← this PR
```

## Tests

- Invalid regions (zero/negative radius, out-of-range lat/lng) dropped with a log while valid ones are kept; boundary values (poles, antimeridian, sub-meter radius) kept.
- GMS batch rejection returns failure without a client call (region built directly, bypassing parse validation, so the guard is pinned genuinely).
- `LocationCoordinates.isValid()` unit tests moved from `LocationServicesImplTest` to `LocationCoordinatesTest`.

---

> [!NOTE]
> **Medium Risk**
> Changes core geofence refresh and OS registration paths; behavior is intentionally defensive but incorrect guards could leave stale registrations or block sync until backend data fixes.
> 
> **Overview**
> **Hardens geofence sync** so malformed `/geofences/nearest` data cannot crash the host app or wipe live OS registrations.
> 
> **API → domain mapping** now drops individual regions that violate GMS preconditions (non-positive radius, out-of-range lat/lng) via shared `LocationCoordinates.isValid()`, with new debug/error logs. A non-empty response where *every* region is dropped or mapping throws is treated as unusable: mapping throws, `GeofenceRepository` catches it and returns `Result.failure` without registering or persisting—so existing fences stay in place. Legitimately empty `geofences: []` still clears as before.
> 
> **GMS batch build** in `GeofenceManager.registerBatch` is wrapped in `try/catch` so `IllegalArgumentException` from `GeofencingRequest` construction becomes `Result.failure` instead of an uncaught throw in the geofence coroutine scope.
> 
> **Location module** moves coordinate validation from `LocationServicesImpl` to `LocationCoordinates.isValid()` for reuse by geofence parsing; tests move to `LocationCoordinatesTest`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 415bf49d2f20397d6fa9c6a481fe9b0e65311192. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/geofence/src/main/kotlin/io/customer/geofence/api/GeofenceApiResponse.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/api/GeofenceApiResponse.kt
@@ -84,14 +84,26 @@ internal data class GeofenceApiRegion(
 internal fun GeofenceApiResponse.toDomainConfig(): GeofenceConfig? =
     config?.toDomain()
 
-internal fun GeofenceApiResponse.toDomainRegions(): List<GeofenceRegion> =
-    geofences.mapNotNull { region ->
-        // Unknown-unknowns net: an escaping throw would crash the host (scope has no exception handler).
-        runCatching { region.toDomain() }.getOrNull() ?: run {
-            SDKComponent.geofenceLogger.logInvalidRegionDropped(region.id)
+// One bad region costs itself (dropped + logged). A non-empty response whose regions ALL drop
+// is unusable — throw so the caller fails the refresh instead of clearing live state.
+internal fun GeofenceApiResponse.toDomainRegions(): List<GeofenceRegion> {
+    if (geofences.isEmpty()) return emptyList()
+
+    val mapped = geofences.mapNotNull { region ->
+        try {
+            region.toDomain() ?: run {
+                SDKComponent.geofenceLogger.logInvalidRegionDropped(region.id)
+                null
+            }
+        } catch (e: Exception) {
+            SDKComponent.geofenceLogger.logRegionMappingFailed(region.id, e.message)
             null
         }
     }
+
+    if (mapped.isEmpty()) error("all ${geofences.size} regions dropped")
+    return mapped
+}
 
 // Coerces raw server values into sane bounds so a misconfigured backend can't push the SDK into a
 // pathological state: non-positive values fall back; positive out-of-range values clamp.
@@ -136,7 +148,7 @@ private fun GeofenceApiConfig.toDomain(): GeofenceConfig {
 }
 
 // Null when the region violates Geofence.Builder preconditions; one bad region must not cost the whole sync.
-private fun GeofenceApiRegion.toDomain(): GeofenceRegion? {
+internal fun GeofenceApiRegion.toDomain(): GeofenceRegion? {
     if (radius <= 0 || !LocationCoordinates.isValid(latitude, longitude)) return null
     return GeofenceRegion(
         id = id,
--- a/geofence/src/main/kotlin/io/customer/geofence/api/GeofenceApiResponse.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/api/GeofenceApiResponse.kt
@@ -5,6 +5,7 @@ import io.customer.geofence.GeofenceConstants
 import io.customer.geofence.GeofenceRegion
 import io.customer.geofence.GeofenceTransitionType
 import io.customer.geofence.di.geofenceLogger
+import io.customer.location.LocationCoordinates
 import io.customer.sdk.core.di.SDKComponent
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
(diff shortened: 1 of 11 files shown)


geofence/src/main/kotlin/io/customer/geofence/api/GeofenceApiResponse.kt
```diff
@@ -84,7 +85,13 @@ internal fun GeofenceApiResponse.toDomainConfig(): GeofenceConfig? =
     config?.toDomain()
 
 internal fun GeofenceApiResponse.toDomainRegions(): List<GeofenceRegion> =
-    geofences.map { it.toDomain() }
+    geofences.mapNotNull { region ->
+        // Unknown-unknowns net: an escaping throw would crash the host (scope has no exception handler).
+        runCatching { region.toDomain() }.getOrNull() ?: run {
```
Shahroz16: Do we still need the `runCatching` here now that the expected invalid values are handled explicitly by `toDomain()`? It catches every thrown exception, so a future mapper defect affecting all regions could turn the response into a successful empty result. That would remove the existing registrations and persist an empty cache, which seems harder to recover from than failing the refresh and preserving the current state. I think keeping nullable validation while allowing unexpected mapping errors to fail the refresh would be safer.

Shahroz16 (approved):

geofence/src/main/kotlin/io/customer/geofence/api/GeofenceApiResponse.kt
mrehan27: Good catch — addressed in 5641bcfc, with one refinement after weighing the single-region case. An unexpected mapper throw now drops only the affected region (logged distinctly via `logRegionMappingFailed`, so a code defect doesn't read as bad server data), keeping one exotic region from taking down the whole sync. The empty-success hazard you flagged is handled by an all-dropped guard: a non-empty response whose regions *all* drop — by throw or by validation — is treated as unusable; `toDomainRegions` throws and the repository fails the refresh, preserving current registrations and cache (and the freshness timestamp never stamps, so the next trigger retries remotely). A legitimately empty response still clears normally, so the kill switch is unaffected. Pinned by `parseAndMap_givenOneRegionMappingThrows_expectOthersKept`, `parseAndMap_givenAllRegionsMappingThrow_expectThrowsNotEmptyList`, `parseAndMap_givenAllRegionsInvalid_expectThrowsNotEmptyList`, and `refresh_givenResponseMappingThrows_expectFailureNotEmptySuccess`.

geofence/src/main/kotlin/io/customer/geofence/api/GeofenceApiResponse.kt
```diff
@@ -128,18 +147,22 @@ private fun GeofenceApiConfig.toDomain(): GeofenceConfig {
...
-    longitude = longitude,
-    radius = radius.toFloat(),
-    transitionTypes = resolveTransitionTypes(transitionTypes),
-    lastUpdated = lastUpdated ?: 0L,
-    geosetIds = geosetIds,
-    metadata = sanitizeMetadata(metadata)
-)
+// Null when the region violates Geofence.Builder preconditions; one bad region must not cost the whole sync.
+internal fun GeofenceApiRegion.toDomain(): GeofenceRegion? {
+    if (radius <= 0 || !LocationCoordinates.isValid(latitude, longitude)) return null
```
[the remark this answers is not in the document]
mrehan27: False positive — NaN/Infinity can't reach `toDomain()`: the serializer has no `allowSpecialFloatingPointValues`, so kotlinx.serialization rejects special values at decode (checked on the parsed result, so lenient-mode quoted strings like `"radius": "NaN"` are covered too), and `GeofenceApiRegion` has no producer besides that decode. A failed decode surfaces as a failed sync via `mapCatching`, never a wipe or crash. Pinned in 415bf49d by `parse_givenNaNOrInfinityValues_expectDecodeFails` so the assumption is enforced rather than remembered.

PR #799, opened 2026-07-27
fix(inbox): correctness and cross-platform parity fixes

mahmoud-elmorabea:
Addresses the correctness findings from the cross-platform review of the inbox overlay work. Paired with the equivalent iOS PR so both SDKs behave identically.

- An absent or unparseable `X-CIO-Inbox-Enabled` header no longer hides the inbox. 304s and error responses routinely omit it, so the previous coercion to `false` made the bell disappear until the next clean 200.
- The once-per-session revalidation gate reopens on a new store session id, so a logout can't serve the previous user's templates/branding.
- Expiry is now inclusive (`expiry <= now`), matching iOS.
- `fromCache` is removed from `InboxVisibility`/`InboxFetchOutcome`; nothing consumed it and it was miscomputed when one asset came from cache and the other was refetched.
- The host `messageOpened` callback reports the post-action opened state.
- Removed the unnecessary `@OptIn(InternalCustomerIOApi)` from both samples (verified they compile without it).
- Deleted the stale `MessagingInboxModuleTest` placeholder.

Verified locally: `messaginginapp` + `messaginginbox` unit tests, ktlint, `apiCheck`, and both sample apps compile.

---

> [!NOTE]
> **Medium Risk**
> Changes affect inbox visibility, session/logout asset isolation, and poll-time enablement parsing—user-visible inbox chrome and cross-user cache leakage are the main risks, mitigated by expanded unit tests.
> 
> **Overview**
> Aligns Android visual inbox behavior with iOS and hardens edge cases around enablement, visibility, and session-scoped assets.
> 
> **Enablement & visibility:** `X-CIO-Inbox-Enabled` no longer defaults to off when the header is missing or unparseable (e.g. 304/error responses)—the last known value is kept. An enabled inbox with templates/branding but **no messages** is **Visible** with an empty list (“caught up”); message count no longer gates visibility (overlay bell still requires renderable messages). Message expiry uses inclusive `expiry <= now`.
> 
> **Templates/branding session gate:** Revalidation reopens on store `sessionId` change (logout/reset) so the next user does not inherit cached assets. The gate only closes when the fetch cycle’s session is still current and there are servable assets; failed fetches with empty cache leave the gate open for retry. Branding JSON parsing is memoized. Unused `fromCache` is removed from `InboxFetchOutcome` / `InboxVisibility`.
> 
> **Overlay layer:** `VisualInbox.getSelectedMessages()` is removed; Jist mapping happens once from `InboxVisibility.Visible.messages` in the controller. `uiStateFlow` is shared via `shareIn` so bell, sheet, and embedded views share one upstream. `VisualInboxController` is a SDK singleton (not per-composable). `messageOpened` callbacks pass `opened = true`. Samples drop unnecessary `@OptIn`; placeholder `MessagingInboxModuleTest` is deleted.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7f0a3f77f94c245a42ba918e3f1a3307ace0cf53. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


messaginginapp/src/main/java/io/customer/messaginginapp/inbox/data/InboxSelection.kt
```diff
@@ -44,6 +44,6 @@ internal object InboxSelection {
 
     private fun isExpired(message: InboxMessage, now: Date): Boolean {
         val expiry = message.expiry ?: return false
-        return expiry.before(now)
+        return expiry <= now
```
Shahroz16, line 47: The `<=` check looks right. Could we keep a follow-up for time-driven expiry? Selection only reruns after another update, so an expired message can remain visible during a quiet SSE session.

messaginginbox/src/main/java/io/customer/messaginginbox/VisualInboxOverlayState.kt
```diff
@@ -136,11 +162,13 @@ internal class VisualInboxController(
...
-        val messages = if (visibility is InboxVisibility.Visible) {
-            visualInbox.getSelectedMessages()
-        } else {
-            emptyList()
-        }
+        //
+        // Reuse the list the visibility decision was made from rather than re-selecting: a second
+        // read would repeat the selection and the Jist deep-copy, and could observe a store change
+        // in between and disagree with the state published alongside it.
+        val messages = (visibility as? InboxVisibility.Visible)
```
Shahroz16, line 170: Could we remove `getSelectedMessages()` and its stale test setup? `snapshot()` now uses `visibility.messages`, so this method no longer has a production caller.

messaginginapp/src/main/java/io/customer/messaginginapp/inbox/data/InboxRepository.kt
```diff
@@ -74,12 +78,19 @@ internal class InboxRepository(
...
+     * (and both assets persisted) loads serve the persisted value with no network call. The repo
+     * is a DI singleton, so this reopens on process restart or on a new store session id (Reset /
+     * logout), which is what stops one user's render assets from being served to the next.
      */
     private val hasRevalidatedAssetsThisSession = AtomicBoolean(false)
 
+    init {
+        inAppMessagingManager.subscribeToAttribute({ it.sessionId }) {
+            hasRevalidatedAssetsThisSession.set(false)
+        }
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Confirmed and fixed in 70568c4.

The race is real: `hasRevalidatedAssetsThisSession` was set after the awaited fetch, so a session reset landing mid-fetch reopened the gate only for the completing cycle to close it again. The gate now closes only when the session that started the cycle is still current.

Two notes on scope, for the record:

- **Impact is narrower than "the prior user's templates/branding."** Both assets are keyed off `gistQueue.baseUrl`, so they are workspace/environment scoped, not per-user — a logout/login inside one workspace resolves to the same URLs. The real consequence was a skipped conditional GET, leaving assets up to one session stale, not a cross-user leak.
- **iOS had the identical race**, which this review could not see. `VisualInboxRepositoryImpl.performRevalidation` set its gate after the await too, and actor reentrancy let the logout handler interleave the same way. Fixed in the paired PR (customerio-ios#1181) with a session-generation counter.

Covered by `loadTemplatesAndBranding_givenSessionChangesMidFetch_expectGateStaysOpen`, verified to fail without the guard.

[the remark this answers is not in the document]
mahmoud-elmorabea: Confirmed and fixed in 6733865.

`subscribeToAttribute` is `scope.launch { storeStateFlow.map(selector).distinctUntilChanged().collect { listener(it) } }` — a StateFlow always replays its current value on collect, and the `scope.launch` makes that delivery asynchronous, so it can indeed land after a load has already closed the gate. The listener now resets only on a change away from a session id it has already seen.

Supporting evidence for the diagnosis: every other `subscribeToAttribute` call site in the module (`SseLifecycleManager` ×2, `PollingLifecycleManager`) already carries an explicit `return@subscribeToAttribute` first-emission guard. This subscription was the odd one out.

On severity — the observable cost is one redundant conditional GET (a 304) inside the race window, not incorrect state, so this reads to me as a perf nit rather than Medium. Fixing it regardless, since it is cheap and it makes the reset rule match iOS's ("only a transition away from an already-known value counts").

Covered by `loadTemplatesAndBranding_givenInitialSessionReplayAfterLoad_expectGateStaysClosed`, verified to fail without the guard. The two existing session tests now simulate the replay before the real change so they exercise the realistic delivery sequence.

messaginginbox/src/main/java/io/customer/messaginginbox/VisualInboxOverlayState.kt
mahmoud-elmorabea: Done in 1b0df36 — confirmed, and it was worse than dead code.

`getSelectedMessages()` is gone from `VisualInbox` along with all eight references in `VisualInboxControllerTest`: five `every { ... } returns` stubs, one `returnsMany`, and two `verify(exactly = 0)` assertions that were only meaningful while the method existed. `snapshot_givenHiddenButSelectableMessagesExist_...` is renamed to `snapshot_givenHidden_...` since the real invariant it guards is that only `InboxVisibility.Visible` carries messages.

Checked before removing: `@InternalCustomerIOApi`, absent from `messaginginapp.api`, and unreferenced by the React Native and Flutter wrappers — so no baseline change and no wrapper break. `apiDump` validation is clean.

Worth passing on, because it nearly cost me: removing the method broke `:messaginginbox` test COMPILATION, and this repo's test task has `ignoreFailures`, so `./gradlew test` printed `BUILD SUCCESSFUL` while that module silently produced no JUnit XML at all — 75 tests just vanished from the run rather than failing. I only caught it because I count tests out of the XML. Both modules now report: 592 tests, 0 failures.

messaginginapp/src/main/java/io/customer/messaginginapp/inbox/data/InboxSelection.kt
mahmoud-elmorabea: Agreed, and I have deliberately left this one out of these PRs — it is a real gap but a behavioral change rather than a fix to what this PR touched, and it needs a design decision.

Confirming the diagnosis: expiry is applied on READ, in `InboxSelection`/`VisualInboxSelector`, and a read only happens when something re-triggers it — a queue poll, an SSE store update, or a templates/branding fetch completing. Nothing is time-driven, so during a quiet stretch an expired message stays on screen until the next unrelated update. iOS is identical (`VisualInboxSelector.select(messages:now:)` off the same read path), so any fix wants to land on both.

The open question is what drives re-evaluation. A wall-clock timer in the SDK is the obvious answer but means a periodic wake purely to hide a row, and it interacts with the polling interval the server already dictates. Scheduling a single one-shot at the earliest upcoming expiry — recomputed whenever the selection changes — gets the same result with one pending timer and no idle work, which is what I would propose.

Happy to open a follow-up with that shape unless you would rather it ride along here. One caveat either way: the expiry check is now inclusive (`expiry <= now`), so a timer firing exactly at the boundary does drop the message.

Shahroz16, line 47: The one-shot timer at the earliest expiry sounds like the right follow-up. Could you open and link that issue here so it is not lost? This does not need to block this PR.

messaginginbox/src/main/java/io/customer/messaginginbox/VisualInboxOverlayState.kt
```diff
@@ -136,11 +162,13 @@ internal class VisualInboxController(
         // renderable (Visible == enabled + templates + branding + >=1 message). Otherwise the
         // panel could render the list with null templates/theme (Jist renders empty) and
         // markOpenMessagesOpened would no-op, leaving viewed messages unopened.
-        val messages = if (visibility is InboxVisibility.Visible) {
-            visualInbox.getSelectedMessages()
-        } else {
-            emptyList()
-        }
+        //
+        // Reuse the list the visibility decision was made from rather than re-selecting: a second
```
Shahroz16, line 167: Could we update the remaining comments to match the new empty-inbox behavior? This file still says visibility requires a message and that dismissing the last row makes the inbox hidden. The old rule also remains in `InboxRepository.computeVisibility`, `InboxFetchOutcome`, and the repository test header.

Shahroz16 (approved):

messaginginapp/src/main/java/io/customer/messaginginapp/inbox/data/InboxSelection.kt
mahmoud-elmorabea: Checked gist-web and iOS before filing this, and the answer argues against the follow-up I proposed — so I'd rather not open it.

**gist-web has no timer.** Expiry is filtered on read in `getInboxMessagesFromLocalStore()` (`inbox-message-manager.ts:46-48`, `expiryDate > now`), called from `Gist.getInboxMessages()` and the inbox component's render, which re-renders off the `messageInboxUpdated` event.

**iOS has no timer either** — read-path only through `VisualInboxSelector.isExpired`, same as Android.

So all three platforms already behave identically here, and adding a timer to mobile would be a deliberate divergence from web rather than parity work.

We're keeping the current behavior. A new message or a foreground/background transition re-runs selection and trims expired rows, which covers the realistic cases, and we specifically don't want rows vanishing under the user mid-read — a timer would introduce exactly that.

messaginginbox/src/main/java/io/customer/messaginginbox/VisualInboxOverlayState.kt
mahmoud-elmorabea: Fixed in 7f0a3f7.

All four sites you named are corrected, and I swept for others rather than just patching the anchors:

- `InboxRepository.computeVisibility` — dropped "at least one selected message exists" from the VISIBLE contract and noted that the selected messages ride along on `InboxVisibility.Visible` without gating it.
- `InboxFetchOutcome` — both blocks: the file header ("visibility ALSO requires `isInboxEnabled` + a selected message") and the `Visible` case ("subject to the enabled + messages checks").
- `VisualInboxOverlayState.snapshot()` — "Visible == enabled + templates + branding + >=1 message", plus the dedupe-guard comment below it that described being "stuck after the last row is dismissed → Hidden", which is no longer a state that occurs.
- The `InboxRepositoryTest` header — "VISIBLE iff enabled + >=1 selected message + templates + branding".

Deliberately left alone: the two comments in `NotificationInboxOverlay` about the bell needing at least one renderable message. Those are still accurate — chrome gating is separate from visibility, and it is what keeps a bell from appearing over an empty panel.

592 tests, 0 failures across both modules; ktlint and apiCheck clean.

messaginginapp/src/main/java/io/customer/messaginginapp/inbox/data/InboxSelection.kt
mahmoud-elmorabea: https://github.com/customerio/customerio-android/pull/799#discussion_r3662862791

PR #803, opened 2026-07-29
fix(live-notifications): recover the opt-in set in a cold process

mahmoud-elmorabea:
Live notifications delivered after ordinary process death were dropped, because the enabled-types gate could not distinguish an app that never opted in from a process that had not loaded the opt-in yet.

## Why

Android starts a process solely to deliver an FCM push. No app code runs there — `CustomerIOFirebaseMessagingService` only calls `SDKComponent.setupAndroidComponent(context)` — so `ModuleMessagingPushFCM` is never registered and `SDKComponent.pushModuleConfig` falls back to `MessagingPushModuleConfig.default()`, whose type set is empty. The gate in `LiveNotificationHandler` read that as "not enabled" and returned.

A dropped `end` was worse than a missed render: the matching `start` was posted by an earlier live process as an ongoing, non-dismissible notification, so it stayed on screen with no way for the user to clear it.

## What

- `LiveNotificationStore` persists the app-wide opt-in set, replacing on each write and clearing when empty, so disabling a type does not leave a stale opt-in behind. The key is deliberately unprefixed to stay clear of the per-activity reclamation sweeps.
- `ModuleMessagingPushFCM.initialize()` records the set on every initialization, empty included.
- The gate prefers the live config and falls back to the persisted set.

A populated config always wins, so the opt-in guarantee still holds for apps that never enabled the feature, and a newly disabled type takes effect immediately rather than waiting for the persisted copy to be rewritten.

Custom types additionally need a render callback, which cannot be persisted; that half is tracked separately. Their `end` is already improved here, since the handler cancels a stranded notification when there is nothing to render.

## Verification

- `messagingpush` unit suite: **279 tests, 0 failures**, including 9 new ones covering the cold-process gate both ways, `end` cancelling a stranded notification, config-wins-over-stale-persisted, replace/clear semantics, and survival of the reclamation sweeps.
- `ktlint` clean; `apiCheck` passes with no `.api` changes (all touched code is `internal`).

Not device-tested.

---

> [!NOTE]
> **Medium Risk**
> Changes push delivery gating and SharedPreferences opt-in state; behavior is well-covered by new tests but affects notification lifecycle after process death.
> 
> **Overview**
> Fixes live notifications being dropped when FCM wakes a **cold process** after the app was killed. In that case `pushModuleConfig.liveNotificationTypes` is empty, so the handler treated opted-in apps the same as apps that never enabled the feature—including **`end` events**, which could leave a non-dismissible ongoing notification stuck on screen.
> 
> **Persistence:** `LiveNotificationStore` now saves the app-wide enabled activity-type set (replace on write, clear when empty). `ModuleMessagingPushFCM.initialize()` writes that set on every init, including when empty, so disabling types does not leave a stale opt-in.
> 
> **Gating:** `LiveNotificationHandler` uses `enabledActivityTypes()`—live config when non-empty, otherwise the persisted set—so cold FCM delivery can still honor the opt-in while apps that never enabled live notifications stay ignored.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c73d8cc9dee0d778e34522ccf0ab3615ac89a1b5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Shahroz16 (approved):
Looks good to me. I traced this through the killed-process FCM path, and the saved type set is replaced/cleared correctly, with the live config taking over once it exists. That fixes the practical drop for the bundled templates, including cold end delivery.

One known follow-up, not something I think should hold this PR: a wrapper cold process still does not have the runtime callback or branding config, so custom start/update cannot render and bundled templates use fallback styling. A custom end is improved here because it can now clear the stranded notification. This was already outside this minimal fix, so I think we can handle it separately after release.

PR #805, opened 2026-07-31
chore(geofence): gate geofence transitions on tracked containment state

mrehan27:
## What

Adds per-region containment bookkeeping so a geofence transition is delivered only when it reflects an
actual boundary crossing.

Two Play Services re-evaluation behaviours produce transitions that don't correspond to the user moving:

**1. EXIT with no prior ENTER.** Evaluating a newly registered region against a coarse fix, Play
Services can treat it as already entered without reporting an ENTER, then report EXIT once an accurate
fix arrives. The device was never inside, but an EXIT-triggered campaign fires.

**2. Repeat ENTER after reboot or app update.** Both wipe OS-side registrations, so the SDK
re-registers with `INITIAL_TRIGGER_ENTER` and Play Services reports ENTER for every region the device
is sitting in — a second "arrived" for a visit the user never left.

Both need an OS-side re-registration *plus* real movement afterwards, so they surface in multi-day
physical-device runs rather than emulator or single-session testing. Reproduced on a Pixel 6 on 30 July
with the full region geometry captured, and both guards then observed firing on a drive on 2 August.

## Reviewing this

371 lines of production code across 6 files (the rest of the diff is tests). It reads best from the
store outward — each file below depends only on the ones above it:

| # | File | What to check |
|---|---|---|
| 1 | `store/GeofenceRegionStore` | The two sets and their invariants. Everything else is a caller. |
| 2 | `GeofenceBroadcastReceiver` | Where the phantom-EXIT guard decides, and its three exemptions. |
| 3 | `GeofenceTransitionEmitter` | Where the repeat-ENTER guard decides, and why it sits before the cooldown. |
| 4 | `GeofenceRepository` | Who seeds containment, and the epoch that orders it against reported exits. |
| 5 | `GeofenceServices` | Crash safety only — independent of the guards. |
| 6 | `GeofenceLogger`, `di/DIGraphGeofence` | New log lines and one constructor argument. |

Commits 1-2 are the guards themselves; 3-8 are responses to review and Bugbot findings, so
"changes since last review" is the more useful diff if you have already read the first two.

## How

**Two persisted sets, deliberately not merged into one:**

- `enteredIds` — where the device **is**. Seeded from our own geometry at every sync, so a record
  exists even when initial-enter synthesis is suppressed.
- `emittedEnterIds` — what was already **reported to the backend**. Written only after the pending row
  is durably persisted, so a rolled-back write can't suppress its own retry.

A sync seeds containment 20-46 ms before the OS reports the matching ENTER (measured in a drive log),
so a single set keyed on containment would drop genuine arrivals.

**`emittedEnterIds` is user-scoped.** A direct A-to-B identify publishes no `ResetEvent`, so without
that, B would inherit A's marks for every region that stayed registered and lose their first arrival. A
read from a different owner returns false; a mark from a different owner replaces the set. `claimExit`
keeps its signature, since it deliberately runs ahead of the identity checks — containment is a
physical fact.

**Both sets are pruned to the registered set on every sync.** A region dropped from the monitored set
while the device is inside never reports the EXIT that would clear its mark.

**The seed is ordered against reported exits by an epoch.** Registration awaits GMS for seconds, so a
departure can land in that window; writing the seed afterwards would resurrect containment the OS has
since told us to drop, leaving the device recorded inside a fence it left and disarming the guard until
its next departure. The store counts claimed exits, a sync reads that count *with* its coordinates, and
reconcile drops any fence claimed since. Only a claim that consumed a record counts — one that found
nothing is a suspected OS artifact, and letting it block the seed would leave the fence with no
containment at all. The per-fence records are trimmed to the registration snapshot on every reconcile.

**The synthesized initial ENTER requires the seeded set and this fix's geometry to agree.** Geometry
alone could fire for a fence reconcile deliberately left out, and that ENTER would never balance: its
genuine EXIT finds no containment record and the guard drops it, while the mark stays set and suppresses
later arrivals. Containment alone can outlive the visit, since reconcile carries a record forward for any
fence that stays registered — a later param-drift re-add would then fire for somewhere the device has
since left. Both read the same fix, so on the happy path the second condition is a no-op; it bites only
on a carried-forward record.

**Bookkeeping is serialized across broadcasts** by a process-wide `Mutex`. Each broadcast runs on its own
scope, so an ENTER and an EXIT for one region could otherwise interleave read-decide-persist: the ENTER
reads the mark the EXIT is about to clear, is dropped, and the EXIT then drops containment — leaving the
device inside with no record. This removes interleaving, not reordering: an ENTER that acquires the lock
ahead of an earlier EXIT is still dropped. `Mutex` is FIFO by `lock()` order and the coroutines launch in
delivery order, so that needs the dispatcher to start them out of order; the residual cost is one dropped
ENTER, with containment self-healing at the next sync.

## Cases that stay unchanged

- **Exit-only regions** — never monitored for ENTER, so no record can exist; exempt from the
  phantom-EXIT guard.
- **Enter-only regions** — never report an EXIT, so nothing could clear the mark; exempt from the
  repeat-ENTER guard, which would otherwise suppress every arrival after the first for the life of the
  registration. These fall back to the `duplicateEventsExpiry` cooldown alone, so a reboot more than one
  window after arrival can deliver one duplicate. Not reachable today: CDP omits `transition_types`, so
  every region decodes as ENTER+EXIT.
- **Regions with no cache row** — unknown transition types can't justify dropping a crossing.
- **Installs upgrading from a version without these sets** — every probe fails open (a missing owner
  reads as "not mine"), so the first transition per region is delivered exactly as before. The sets seed
  at the first registration. **No migration step.**

## Crash safety

The sync runs on a scope with a `SupervisorJob` and no exception handler, so an exception escaping it
reached the thread's default handler and took the host app down. `Location.distanceBetween` throws on
NaN, infinite or out-of-range coordinates, and the ranking and containment maths call it part-way
through a sync — while nothing validated the device's own fix, since the coordinate check was only ever
applied to regions arriving from the API. Not reachable from a fused-provider fix, but a mock or
misbehaving OEM provider could produce one.

Such a fix is now rejected at the sync boundary, rearming as for a missing one, and the launched work is
wrapped so no exception escapes the scope. Fatal `Error`s still propagate.

## Testing

1687 unit tests across all modules, 0 failures. `apiCheck` clean — everything added is `internal`.
ktlint clean. `lintDebug` reports 0 issues. Sample app builds.

**Mutation-verified.** Inverting each guard fails the test written for it and only that test. The same
holds for every review-round change: dropping the enter-only exemption, the lock, the owner check, the
synthesized ENTER's monitoring shape, the seed's epoch filter, reading the epoch after the GMS await
instead of before, dropping either half of the synthesized ENTER's containment-and-geometry check, the unusable-fix check, or the scope
backstop.

**Two upgrade paths validated on an emulator** with an injected location track: in place over a build
that had already written `emitted_enter_ids`, and over a 4.20.0-shaped store where
`registeredIds`/`cachedRegions` exist but none of the new keys do. Covered re-registration after the app
update, the synthesized ENTER for the region the device sits in, the OS's own repeat ENTER dropped by the
guard, the EXIT clearing both the containment record and the mark, and the owner key written with the
identified user. No crashes or receiver errors in any phase.

**Field-validated on 2 August** (Pixel 6, Bahria Town → airport → back): the repeat-ENTER guard fired
after a reboot inside a fence, and the phantom-EXIT guard dropped two unmatched EXITs on arrival home. A
released 4.20.0 build on the same phone through the same drive delivered three unbalanced EXITs; this
build delivered zero, with every genuine crossing still delivered.

The concurrency and identity-switch paths are covered by unit tests rather than on-device — neither is
reachable through the sample app's UI.

## Notes

Independent of #806; both branch from `main` with no ordering requirement. #807 stacks on this one.

---

> [!NOTE]
> **Medium Risk**
> Changes core geofence event delivery and persisted state; wrong guard logic could drop real ENTER/EXIT or send spurious campaigns, though exemptions and upgrade fail-open behavior limit blast radius.
> 
> **Overview**
> Geofence delivery is gated on **tracked containment** so transitions reflect real crossings instead of Play Services reconciliation (EXIT with no prior ENTER) or re-registration after reboot/app update (duplicate ENTER while still inside).
> 
> **`GeofenceRegionStore`** gains two persisted, user-scoped sets: `enteredIds` (device inside) and `emittedEnterIds` plus owner (ENTER already sent to the backend), with a **containment epoch** so sync geometry does not overwrite an EXIT claimed while GMS registration was in flight. Sign-out clears both; registration calls **`reconcileEnteredIds`** / **`pruneEmittedEnterIds`** when fences move or drop off the monitored set.
> 
> **`GeofenceBroadcastReceiver`** records ENTER, **`claimExit`** on EXIT (drops phantom EXITs when containment and cache say ENTER was monitored), and serializes per-fence bookkeeping with a process-wide **`Mutex`**. **`GeofenceTransitionEmitter`** drops redundant ENTER when `monitorsExit` and a durable persist already marked the fence, then marks after successful persist.
> 
> **`GeofenceRepository`** snapshots epoch at sync start, seeds containment after successful registration, tightens **initial ENTER synthesis** to require seeded containment and current geometry, and passes **`monitorsExit`** into emit. **`GeofenceServices`** rejects invalid coordinates and wraps sync/reset work so exceptions do not crash the host app.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0950ed3a1b85075dde5f30983046b2c4cabb4d7e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
@@ -130,8 +130,8 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
                 }
             }
 
-            // Serialized across broadcasts: each runs on its own scope, so two transitions for the
-            // same fence would otherwise interleave read-decide-persist and lose a containment record.
+            // Each broadcast runs on its own scope, so two transitions for one fence would
+            // otherwise interleave read-decide-persist and lose a containment record.
             transitionMutex.withLock {
                 handleBusinessTransition(
                     geofenceId = geofenceId,
@@ -162,28 +162,21 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
         androidComponent: AndroidSDKComponent,
         logger: GeofenceLogger
     ) {
-        // Containment bookkeeping runs before the identity checks below: being inside a fence is
-        // a physical fact, independent of whether the transition is deliverable.
+        // Ahead of the identity checks below: being inside a fence is a physical fact, independent of
+        // whether the transition is deliverable.
         val store = androidComponent.geofenceRegionStore
         val cachedRegion = store.getCachedRegion(geofenceId)
         when (transition) {
             Event.GeofenceTransition.ENTER -> store.recordEntered(geofenceId)
-            // claimExit runs first either way so the record is consumed atomically; the checks
-            // after it only decide whether an unmatched EXIT is trusted.
-            //
-            // Two cases where an absent record proves nothing, so the EXIT is delivered:
-            // a region registered without ENTER monitoring can never produce one to record, and
-            // an install upgraded from an SDK without the set has no data until the first
-            // registration seeds it. A missing cache row is treated the same way — unknown
-            // transition types can't justify dropping a crossing.
+            // An unmatched EXIT is a GMS reconciliation artifact, not a crossing — delivering it
+            // fires EXIT campaigns at people who were never there. The two guards after claimExit
+            // cover the cases where an absent record proves nothing: an upgraded install with no set
+            // yet, and a region that never monitored ENTER (a missing cache row counts as unknown).
             Event.GeofenceTransition.EXIT -> if (
                 !store.claimExit(geofenceId) &&
                 store.hasContainmentRecord() &&
                 cachedRegion?.transitionTypes?.contains(GeofenceTransitionType.ENTER) == true
             ) {
-                // GMS reports EXIT for a fence it never reported as entered — an artifact of its
-                // own state reconciliation, not a crossing. Delivering it fires EXIT-triggered
-                // campaigns at people who were never there.
                 logger.logExitDroppedNeverEntered(geofenceId)
                 return
             }
@@ -192,9 +185,8 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
         // Snapshot userId so a sign-out + sign-in before delivery can't reattribute this
         // transition. Empty userId is treated as "not identified" per `isUserIdentified`.
         val userId = androidComponent.secureUserStore.getUserId()?.takeIf { it.isNotEmpty() }
-        // Geofencing is identified-only: the backend rejects anonymous geofence tracks, so an
-        // anonymous transition has no deliverable path. Drop it before spending a cooldown slot
-        // or persisting a row neither channel could ever send.
+        // Identified-only: the backend rejects anonymous geofence tracks, so drop before spending a
+        // cooldown slot or persisting a row neither channel could send.
         if (userId == null) {
             logger.logTransitionDroppedAnonymous(geofenceId, transition.name)
             return
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
@@ -128,6 +128,33 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
                 }
             }
 
+            // Containment bookkeeping runs before the identity checks below: being inside a fence is
+            // a physical fact, independent of whether the transition is deliverable.
+            val store = androidComponent.geofenceRegionStore
+            val cachedRegion = store.getCachedRegion(geofenceId)
+            when (transition) {
+                Event.GeofenceTransition.ENTER -> store.recordEntered(geofenceId)
+                // claimExit runs first either way so the record is consumed atomically; the checks
+                // after it only decide whether an unmatched EXIT is trusted.
+                //
+                // Two cases where an absent record proves nothing, so the EXIT is delivered:
+                // a region registered without ENTER monitoring can never produce one to record, and
+                // an install upgraded from an SDK without the set has no data until the first
+                // registration seeds it. A missing cache row is treated the same way — unknown
+                // transition types can't justify dropping a crossing.
+                Event.GeofenceTransition.EXIT -> if (
+                    !store.claimExit(geofenceId) &&
+                    store.hasContainmentRecord() &&
+                    cachedRegion?.transitionTypes?.contains(GeofenceTransitionType.ENTER) == true
+                ) {
+                    // GMS reports EXIT for a fence it never reported as entered — an artifact of its
+                    // own state reconciliation, not a crossing. Delivering it fires EXIT-triggered
+                    // campaigns at people who were never there.
+                    logger.logExitDroppedNeverEntered(geofenceId)
+                    return@forEach
+                }
+            }
+
             // Snapshot userId so a sign-out + sign-in before delivery can't reattribute this
             // transition. Empty userId is treated as "not identified" per `isUserIdentified`.
             val userId = androidComponent.secureUserStore.getUserId()?.takeIf { it.isNotEmpty() }
@@ -139,7 +166,6 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
                 return@forEach
             }
 
-            val cachedRegion = androidComponent.geofenceRegionStore.getCachedRegion(geofenceId)
             transitionEmitter.emit(
                 geofenceId = geofenceId,
                 transition = transition,
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
@@ -12,6 +12,7 @@ import io.customer.geofence.di.geofenceRegionStore
 import io.customer.geofence.di.geofenceServices
 import io.customer.geofence.di.geofenceTransitionEmitter
 import io.customer.sdk.communication.Event
+import io.customer.sdk.core.di.AndroidSDKComponent
 import io.customer.sdk.core.di.SDKComponent
 import io.customer.sdk.core.di.clock
 import io.customer.sdk.core.di.setupAndroidComponent
@@ -20,6 +21,8 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
 import kotlinx.coroutines.withTimeoutOrNull
 
 /** Receives OS geofence transition callbacks and dispatches them to the SDK. */
@@ -94,7 +97,6 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
         val timestamp = SDKComponent.clock.currentTimeSeconds()
         val dispatchStartUptimeMs = SDKComponent.clock.elapsedRealtime()
         val androidComponent = SDKComponent.android()
-        val transitionEmitter = androidComponent.geofenceTransitionEmitter
         // Defense-in-depth against orphans (failed clearAll, app-data wipe, SDK
         // ID-format changes): events for unregistered IDs are dropped and the OS-side
         // registration is removed so it stops firing.
@@ -128,53 +130,17 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
                 }
             }
 
-            // Containment bookkeeping runs before the identity checks below: being inside a fence is
-            // a physical fact, independent of whether the transition is deliverable.
-            val store = androidComponent.geofenceRegionStore
-            val cachedRegion = store.getCachedRegion(geofenceId)
-            when (transition) {
-                Event.GeofenceTransition.ENTER -> store.recordEntered(geofenceId)
-                // claimExit runs first either way so the record is consumed atomically; the checks
-                // after it only decide whether an unmatched EXIT is trusted.
-                //
-                // Two cases where an absent record proves nothing, so the EXIT is delivered:
-                // a region registered without ENTER monitoring can never produce one to record, and
-                // an install upgraded from an SDK without the set has no data until the first
-                // registration seeds it. A missing cache row is treated the same way — unknown
-                // transition types can't justify dropping a crossing.
-                Event.GeofenceTransition.EXIT -> if (
-                    !store.claimExit(geofenceId) &&
-                    store.hasContainmentRecord() &&
-                    cachedRegion?.transitionTypes?.contains(GeofenceTransitionType.ENTER) == true
-                ) {
-                    // GMS reports EXIT for a fence it never reported as entered — an artifact of its
-                    // own state reconciliation, not a crossing. Delivering it fires EXIT-triggered
-                    // campaigns at people who were never there.
-                    logger.logExitDroppedNeverEntered(geofenceId)
-                    return@forEach
-                }
+            // Serialized across broadcasts: each runs on its own scope, so two transitions for the
+            // same fence would otherwise interleave read-decide-persist and lose a containment record.
+            transitionMutex.withLock {
+                handleBusinessTransition(
+                    geofenceId = geofenceId,
+                    transition = transition,
+                    timestamp = timestamp,
+                    androidComponent = androidComponent,
+                    logger = logger
+                )
             }
-
-            // Snapshot userId so a sign-out + sign-in before delivery can't reattribute this
-            // transition. Empty userId is treated as "not identified" per `isUserIdentified`.
-            val userId = androidComponent.secureUserStore.getUserId()?.takeIf { it.isNotEmpty() }
-            // Geofencing is identified-only: the backend rejects anonymous geofence tracks, so an
-            // anonymous transition has no deliverable path. Drop it before spending a cooldown slot
-            // or persisting a row neither channel could ever send.
-            if (userId == null) {
-                logger.logTransitionDroppedAnonymous(geofenceId, transition.name)
-                return@forEach
-            }
-
-            transitionEmitter.emit(
-                geofenceId = geofenceId,
-                transition = transition,
-                userId = userId,
-                timestampSeconds = timestamp,
-                geofenceName = cachedRegion?.name,
-                metadata = cachedRegion?.metadata ?: emptyMap(),
-                geosetIds = cachedRegion?.geosetIds ?: emptyList()
-            )
         }
 
         // Hold the goAsync window open until the refresh lands so the OS doesn't kill a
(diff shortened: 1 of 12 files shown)


geofence/src/main/kotlin/io/customer/geofence/GeofenceTransitionEmitter.kt
```diff
@@ -29,6 +38,14 @@ internal class GeofenceTransitionEmitter(
         metadata: Map<String, JsonElement>,
         geosetIds: List<String>
     ): Boolean {
+        // Every reboot and app update re-registers with INITIAL_TRIGGER_ENTER, so the OS re-reports
+        // ENTER for a fence the device never left. Ahead of the cooldown so the drop doesn't spend
+        // the fence's slot. Read-then-mark isn't atomic — the mark waits for a durable write below —
+        // but tryAcquire is, so two concurrent duplicates still collapse to one.
+        if (transition == Event.GeofenceTransition.ENTER && regionStore.hasEmittedEnter(geofenceId)) {
```
Shahroz16: Could we either limit this persistent gate to fences that also monitor EXIT, or monitor EXIT internally and filter its delivery? Current CDP omits `transition_types`, so production defaults to ENTER+EXIT today. But the SDK already accepts ENTER-only payloads; once that backend field is enabled, GMS will never call `claimExit()` for those fences, and every genuine re-entry after the first will stay suppressed until eviction or user-state reset.

geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
```diff
@@ -128,6 +128,33 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
                 }
             }
 
+            // Containment bookkeeping runs before the identity checks below: being inside a fence is
+            // a physical fact, independent of whether the transition is deliverable.
+            val store = androidComponent.geofenceRegionStore
+            val cachedRegion = store.getCachedRegion(geofenceId)
+            when (transition) {
```
Shahroz16: Could transition bookkeeping be serialized, at least per fence? Each broadcast is handled in a new Default-dispatcher scope, so a quick EXIT then ENTER can complete in reverse order. The ENTER sees the old marker and is dropped, then the EXIT clears containment, leaving the device inside with no record and causing its next EXIT to be dropped too. Boundary jitter makes this reachable without an unusually fast revisit.

geofence/src/main/kotlin/io/customer/geofence/store/GeofenceRegionStore.kt
```diff
@@ -115,6 +181,61 @@ internal class GeofenceRegionStoreImpl(
...
+        // anonymous drop, and a stranded mark silences the next genuine arrival.
+        clearEnterEmitted(geofenceId)
+        true
+    }
+
+    override fun reconcileEnteredIds(registeredIds: Set<String>, inside: Set<String>) = synchronized(enteredLock) {
+        writeJson(KEY_ENTERED_IDS, ID_SET_SERIALIZER, (getEnteredIds() intersect registeredIds) + inside)
+    }
+
+    override fun hasEmittedEnter(geofenceId: String): Boolean =
```
Shahroz16: Could this marker be scoped by `userId`, or rotated synchronously when the identified profile changes? It currently keys only on the fence. A direct A to B identify does not publish `ResetEvent`, and a quick re-identify can also make the async reset skip its wipe once B is present. If the fence stays registered, B can inherit A’s standing ENTER marker and have a synthesized or OS ENTER suppressed even though B never received one.

geofence/src/main/kotlin/io/customer/geofence/GeofenceTransitionEmitter.kt
mrehan27: Fixed in fd2c02eb. The gate now applies only when the region also monitors EXIT, and the mark isn't written for enter-only fences either — one that can never be cleared is worse than none. It mirrors the exemption the EXIT-side guard already has for regions that never report an ENTER, so the two guards are now symmetric about what the region actually monitors.

Took the narrower of your two options. Registering EXIT internally and filtering delivery is the better end state — it keeps containment tracking complete for every fence — but it changes what we hand GMS for every region, which I didn't want to do in a patch release. Worth doing before the backend enables the field.

Trade-off after this change, stated plainly: an enter-only fence falls back to the `duplicateEventsExpiry` cooldown alone, so a reboot more than one window after arrival can deliver one duplicate ENTER. That's against losing every re-entry permanently today, so it's the better failure, but it is a weaker guarantee than both-transition fences get.

geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
mrehan27: Fixed in fd2c02eb. The per-fence bookkeeping and emit now run under a process-wide `Mutex`, so two broadcasts can't interleave their read-decide-persist. The test launches an ENTER and an EXIT for the same fence concurrently and asserts the critical section is never re-entered; removing the lock fails that test and nothing else.

Two limits worth recording rather than leaving implied:

The lock removes interleaving, not reordering. If the ENTER coroutine acquires the lock ahead of an earlier EXIT it still reads the stale mark and is dropped. `Mutex` is FIFO by `lock()` call order and the coroutines are launched in broadcast-delivery order, so that needs the dispatcher to start them out of order — possible, not likely. The residual cost is one dropped ENTER, and containment self-heals at the next sync via `reconcileEnteredIds`, which recomputes `inside` from the current location. Making it genuinely ordered needs per-fence transition timestamps, which I left out of a patch.

It also serializes receiver-against-receiver only. The sync path still touches the same store concurrently, exactly as before — each store operation is individually atomic, but the sequences aren't serialized against each other.

geofence/src/main/kotlin/io/customer/geofence/store/GeofenceRegionStore.kt
mrehan27: Fixed in fd2c02eb. The set now carries an owner userId: a read from a different owner returns false, and a mark from a different owner replaces the set rather than adding to it.

Chose that over keying the set by userId so `claimExit` keeps its signature. It deliberately runs ahead of the identity checks in the receiver — being inside a fence is a physical fact, independent of who is signed in — and threading a userId into it would have entangled the two.

On upgrade the owner key is absent, which reads as "not mine", so the first ENTER per fence is delivered rather than suppressed. Verified on device by upgrading in place over a build that had already written `emitted_enter_ids`: the pre-existing marks are ignored once and the cooldown absorbed the repeat, and the owner key is written on the next genuine ENTER.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
```diff
@@ -411,6 +411,19 @@ internal class GeofenceRepositoryImpl(
...
+                    // Seed containment from our own geometry rather than waiting on a GMS ENTER:
+                    // synthesis is suppressed after a reboot or app update, so no ENTER is emitted
+                    // for a fence the device is already sitting inside, and its later genuine EXIT
+                    // would otherwise look unentered and be dropped.
+                    store.reconcileEnteredIds(
+                        registeredIds = idsToSave,
+                        inside = nearest.filter { it.distanceTo(latitude, longitude) <= it.radius }
+                            .map { it.id }
+                            .toSet()
+                    )
```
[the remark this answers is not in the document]
mrehan27: Fixed in 0e858156, though not for the reason described — worth recording what the measurement showed.

A shared lock would not have helped. `claimExit` and `reconcileEnteredIds` are each already atomic under the store's own lock; nothing interleaves inside either. The defect is staleness, not mutual exclusion: `inside` is geometry derived from the fix the sync started with, and registration then awaits GMS for seconds. If a departure is reported in that window, writing the seed afterwards resurrects containment the OS has since told us to drop. A lock only orders the two operations — with reconcile second, the stale seed still lands.

It is also narrower than it looks. The two crossings have to be reported in *different* GMS evaluation passes, with the trigger's pass at a fix still inside the fence. When both land in the same pass — the common case, and what our field logs show — the reported fix is by definition outside the fence, so `inside` never contains it and there is nothing to resurrect. A designed reproduction on an emulator missed: the sync completed in ~200 ms while GMS took 5 s to report the second crossing.

The fix matches the actual failure. The store counts claimed exits; a sync reads the count together with its coordinates and hands it back at reconcile, and any fence claimed since is dropped from the seed. Only a claim that consumed a record counts — one that found nothing is a suspected reconciliation artifact, and letting it block the seed would leave the fence with no containment at all, which is the failure the seed exists to prevent. Both directions are covered by tests, and reverting either half fails only its own test.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
```diff
@@ -469,7 +495,8 @@ internal class GeofenceRepositoryImpl(
                 timestampSeconds = timestamp,
                 geofenceName = region.name,
                 metadata = region.metadata,
-                geosetIds = region.geosetIds
+                geosetIds = region.geosetIds,
+                monitorsExit = GeofenceTransitionType.EXIT in region.transitionTypes
```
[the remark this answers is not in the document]
mrehan27: Confirmed and fixed in 5e356426 — this one was a genuine regression from the epoch change, not a pre-existing gap, so thanks for catching it on the same pass.

To be precise about the damage, since it is worse than "an ENTER without an EXIT": before the epoch change the stale re-seed and the synthesized ENTER agreed with each other, so a later genuine EXIT still found containment and balanced the pair. Suppressing the re-seed without suppressing the synthesis broke that symmetry — the ENTER goes out, its genuine EXIT is dropped by the phantom guard, and `emittedEnterIds` stays set so later arrivals at that fence are suppressed too until it leaves the monitored set.

The fix removes the second geometry pass rather than teaching it about the epoch: synthesis now reads the seeded set back from the store, so containment and the synthesized ENTER are derived from one computation and cannot disagree. A departure arriving between the reconcile and the read only makes the answer fresher. Reverting to a recomputed `distanceTo <= radius` fails the new test and only that test.

Reachability for the record: it needs the fence to be *newly* registered in the same sync that loses its containment, which excludes the reboot and app-update paths since synthesis is disabled outright while the OS-state-wiped flag is up.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
```diff
@@ -411,6 +420,19 @@ internal class GeofenceRepositoryImpl(
...
+                    store.reconcileEnteredIds(
+                        registeredIds = idsToSave,
+                        inside = nearest.filter { it.distanceTo(latitude, longitude) <= it.radius }
+                            .map { it.id }
+                            .toSet(),
+                        // Registration awaited GMS, so an EXIT since this fix is newer evidence.
+                        sinceEpoch = containmentEpoch
+                    )
+                    // Same snapshot: a dropped fence never reports the EXIT that would re-arm it.
+                    store.pruneEmittedEnterIds(idsToSave)
```
Shahroz16: Could we reset the transition bookkeeping for IDs whose registration parameters changed? A same-ID fence remains in `idsToSave`, so `reconcileEnteredIds` carries its old containment and this prune keeps its old emitted-ENTER mark. If the fence is moved while the device was recorded inside, GMS does not guarantee an EXIT for the old circle. The first real ENTER at the new circle is then dropped as already reported, and its EXIT can be delivered unmatched. Clearing the old containment and emitted marker for changed IDs, then reseeding from the new geometry, would keep edits from losing the next visit. A test using the real emitter across a same-ID geometry change would cover it.

Shahroz16 (commented):
I found one state carryover case when a live fence keeps its ID but changes geometry.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
mrehan27: Confirmed and fixed in 13f76ef3 — and worth being precise that this is a regression this PR introduced, not a pre-existing corner: `hasEmittedEnter` doesn't exist in 4.20.0, so today that first arrival at the edited circle is delivered. This PR is what loses it.

The mechanism is exactly as you describe. `equalsForRegistration` compares lat/lng/radius/transitionTypes, so an edited fence is excluded from `unchangedRegisteredIds` and re-added to GMS, but its ID stays in `idsToSave` — so `reconcileEnteredIds` carried its old containment and the prune kept its mark. GMS guarantees no EXIT for a circle it no longer holds, so the arrival at the new circle was dropped as already reported and its departure then delivered unmatched.

The fix computes the drifted set in the same pre-mutation snapshot as `unchangedRegistered` (registered before this pass, not cache-equal) and resets containment and the mark for it.

One qualification on "then reseeding from the new geometry": the reset is conditional on the new geometry putting the device **outside**. Resetting unconditionally would make a radius *widening* while the device sits inside reseed containment, fire synthesis, and deliver a duplicate arrival where today there is none. Mutation-tested — dropping that condition fails `refresh_givenFenceReRegisteredWhileDeviceInside_expectRecordAndMarkKept` and nothing else.

I did not add the real-emitter composition test you suggested. The chain is covered link-by-link instead — the emitter suite's relaxed default is `hasEmittedEnter = false`, so every delivery test already covers "no mark → delivered", and the mark-suppresses-ENTER direction has its own test. Flagging that as a deliberate choice rather than an oversight, in case you'd rather have the end-to-end one.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
```diff
@@ -411,6 +425,25 @@ internal class GeofenceRepositoryImpl(
...
+                    val movedAwayIds = reRegisteredIds - insideIds
+                    store.reconcileEnteredIds(
+                        registeredIds = idsToSave,
+                        inside = insideIds,
+                        // Registration awaited GMS, so an EXIT since this fix is newer evidence.
+                        sinceEpoch = containmentEpoch,
+                        resetIds = movedAwayIds
+                    )
+                    // Same snapshot: a dropped fence never reports the EXIT that would re-arm it.
+                    store.pruneEmittedEnterIds(idsToSave - movedAwayIds)
```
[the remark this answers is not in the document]
mrehan27: Confirmed and fixed in 0950ed3a. This one is a regression from the geometry-drift fix two commits earlier, so thanks for catching it on the same pass.

It's also more reachable than "the device physically crosses during the await". The reset is decided from the fix the sync started with, but registration awaits GMS for seconds and GMS evaluates the new circle against its *own*, fresher position — so it only takes our fix and GMS's to disagree about the new boundary, not a real crossing. GMS then reports the arrival, the receiver records containment and delivers the ENTER, and the reset dropped both the record and its mark.

The damage is worse than an unbalanced pair: the genuine EXIT finds no containment, the never-entered guard drops it, and the backend is left with an arrival the device never leaves until the next sync re-seeds from a fresh fix.

Two halves to the fix:

The store now stamps reported entries on the same counter as claimed exits, and declines a reset for any fence that reported one after the caller's `sinceEpoch`. That's the mirror of what already protects the geometry *seed* from a mid-sync departure — the OS placing the device inside outranks a sync's older geometry, in both directions.

`reconcileEnteredIds` also returns the ids it actually dropped, so `pruneEmittedEnterIds` acts on the store's decision rather than the caller's guess. Without that the two disagree — containment kept, mark cleared — and the mark belongs to the arrival that was just reported, so clearing it would let the next report through as a fresh one. The two disagreeing is the shape of both of the last regressions here, so it seemed worth making it structurally impossible rather than matching the sets by hand at the call site.

Both halves are mutation-tested: removing the epoch filter fails only the new store test, pruning from the unfiltered set fails only the new repository test.

Shahroz16 (approved):

PR #806, opened 2026-07-31
chore(geofence): rank nearest regions by boundary distance, not centre

mrehan27:
## What

Nearest-set selection ranked and capped on distance to each region's **centre**, ignoring radius.

A region the device is inside therefore ranks at roughly its own radius, so a large occupied region is evicted from the monitored set as soon as `maxBusinessGeofences` regions have nearer centres — and an unmonitored region can never report its exit. The `maxMonitoringDistance` cap had the same flaw: a region whose centre fell outside the cap was filtered out even when its area contained the device.

## How

New `GeofenceRegion.edgeDistanceTo()` = `(distanceTo - radius).coerceAtLeast(0f)`, used for both the distance filter and the ranking, so within the fetched set an occupied region sorts first and survives both the cap and the count limit.

Distances round to whole metres and ties break on ascending id, so equidistant regions at the eviction boundary order deterministically instead of following the server's response order.

## Deliberate behaviour changes

- Equal-radius ranking is **unchanged** — subtracting a constant preserves order.
- The distance cap **widens by one radius**: regions just outside the old `maxMonitoringDistance` now register. The default is 1 000 km, so this is inert unless a workspace sets a tight cap.
- Regions the device is outside are now ordered by boundary proximity, so a large region up the road outranks a small one further away.

## Known limitation

This ranks the set the server returned; it cannot recover a region the server never sent. `/geofences/nearest` omits `limit`, so CDP applies its default of 100 and orders by **centre** distance before truncating. In a workspace dense enough to exceed that, a large region containing the device can be cut server-side and never reach this sort. Aligning the server-side limit and ordering with boundary distance is a backend follow-up; until then "occupied regions are retained" holds within the fetched set, not absolutely.

## Not changed

The containment checks in `GeofenceRepository` stay centre-based (`distanceTo <= radius`) — edge distance would make them always true.

## Testing

1642 unit tests across all modules, 0 failures. `apiCheck` clean — everything added is `internal`. ktlint clean. Sample app builds.

6 new tests, mutation-verified by reverting to `distanceTo`: that fails all three new filter tests with the exact symptom (occupied region evicted / filtered out entirely / ranking inverted) while leaving the pre-existing filter tests passing.

## Notes

Independent of the other geofence patch PR in flight; both branch from `main` with no ordering requirement.

---

> [!NOTE]
> **Medium Risk**
> Changes which geofences are registered under count and distance limits—a correctness fix for missed exit events, but workspaces with tight distance caps may monitor a slightly wider set of regions.
> 
> **Overview**
> **Nearest geofence selection** now ranks and applies the monitoring distance cap using **boundary** distance (`edgeDistanceTo`) instead of distance to the region center, so a region the device is inside always ranks first and is not dropped when the OS geofence budget or `maxMonitoringDistance` would have evicted it under center-based math.
> 
> `GeofenceDistanceFilter.nearest` rounds distances to whole meters and breaks ties on ascending region `id` (aligned with iOS) so equidistant regions at the cap are deterministic rather than following server list order.
> 
> New unit tests cover occupied-region retention, distance-cap behavior when the center is outside the cap, boundary-vs-center ranking for large regions, and sub-meter rounding tiebreaks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4b175e15f936ace96dc00e805b9e1d4799d3a29c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceDistanceFilter.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceDistanceFilter.kt
@@ -1,13 +1,23 @@
 package io.customer.geofence
 
+import kotlin.math.round
+
 /**
  * Selects the geofence regions closest to a reference location, capped at a maximum count and
  * (optionally) a maximum distance.
  *
  * The OS limits the number of geofences an app can register simultaneously; this filter
  * picks the most relevant subset based on straight-line distance from the device's
- * current location. Regions beyond [maxDistanceMeters] are excluded entirely; local re-ranking
- * re-includes them as the device approaches.
+ * current location. Regions whose boundary is beyond [maxDistanceMeters] are excluded entirely;
+ * local re-ranking re-includes them as the device approaches.
+ *
+ * Ranking and the cap both measure to the region's *boundary* ([edgeDistanceTo]), so a region the
+ * device is inside always sorts first and survives both the count limit and the distance cap.
+ *
+ * Ties break on ascending [GeofenceRegion.id], and distances round to whole meters first:
+ * `Location.distanceBetween` can return sub-meter-varying results for the same inputs, which would
+ * otherwise defeat the tiebreak and leave equidistant regions ordered by the server's response
+ * order. Matches iOS so both platforms pick the same set at the cap.
  */
 internal class GeofenceDistanceFilter {
     fun nearest(
@@ -19,9 +29,9 @@ internal class GeofenceDistanceFilter {
     ): List<GeofenceRegion> {
         if (max <= 0 || regions.isEmpty()) return emptyList()
         return regions
-            .map { it to it.distanceTo(latitude, longitude) }
+            .map { it to round(it.edgeDistanceTo(latitude, longitude)) }
             .filter { (_, distance) -> distance <= maxDistanceMeters }
-            .sortedBy { (_, distance) -> distance }
+            .sortedWith(compareBy({ (_, distance) -> distance }, { (region, _) -> region.id }))
             .take(max)
             .map { (region, _) -> region }
     }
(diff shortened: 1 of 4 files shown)


geofence/src/main/kotlin/io/customer/geofence/GeofenceDistanceFilter.kt
```diff
@@ -19,9 +29,9 @@ internal class GeofenceDistanceFilter {
     ): List<GeofenceRegion> {
         if (max <= 0 || regions.isEmpty()) return emptyList()
         return regions
```
Shahroz16, line 31: I checked the current CDP and Services path and there is one earlier cap to account for. The SDK omits `limit`, CDP defaults to 100, and Services orders by center distance before applying that limit. In a dense workspace, a large fence containing the device can fall outside those 100 and never reach this boundary-distance sort. Could we align the server-side limit with boundary distance, or track that as a required backend follow-up before describing occupied fences as always retained?

Shahroz16 (approved):

geofence/src/main/kotlin/io/customer/geofence/GeofenceDistanceFilter.kt
mrehan27: You're right, and there's no SDK-side fix that closes it. Sending an explicit `limit` narrows the window but the server ordering is still by centre distance, so a large region containing the device can still be cut before it ever reaches this sort.

I've dropped the absolute claim from the description — it now says occupied regions are retained within the fetched set, not absolutely — and recorded aligning the server-side limit and ordering with boundary distance as a required backend follow-up. Left the code as is.

One number worth adding for whoever picks that up: `maxMonitoringDistance` defaults to 1 000 km, so in a dense workspace CDP's 100-region default is the binding constraint long before our own cap is.

mahmoud-elmorabea (approved):

PR #807, opened 2026-08-04
chore(geofence): refresh discovery from a live fix on foreground entry

mrehan27:
## What

Refreshes the nearby geofence set from a live location fix on foreground entry, so the SDK notices it
has moved without waiting on the OS.

Discovery today has a single input: an OS geofence callback. When those stop arriving, nothing else
re-evaluates where the device is — every other refresh path anchors at the last registration center,
which is by definition where the device used to be, so the tier table sees no movement and skips.

Observed on a 20 km drive on 2 August: Play Services logged nothing device-wide for 90 minutes (zero
`Geofencer` lines, across all apps), and two installs resumed within 43 ms of each other when it came
back. The device sat 23 km from the nearest fence it knew about, and no code path could ask where it
was.

## How

On foreground entry, in `GeofenceLocationMode.AUTOMATIC`, arm a refresh and request a silent
(no-analytics) fix — the pair `refreshFromCurrentLocation()` already uses. The returning fix drives
the refresh through `onLocationAcquired`, anchored at the device's real position where
`movedBeyondFetchRadius` can finally see the move.

The fresh fix does double duty: Play Services evaluates geofences off location updates rather than
polling, so a fix flowing through the fused provider also gives it a reason to re-evaluate regions it
is holding stale state for.

Two cases keep their existing behaviour, both decided inside `refreshOnForeground` so tests cover
them rather than an untestable lifecycle lambda:

- **MANUAL** hands acquisition to the host, so it never arms — a flag no fix consumes would sit set
  until some unrelated fix landed.
- **A sync already stuck without a location** keeps its narrower recovery, which requests a live fix
  itself where one is needed and falls back to the anchor where that is the best available.

No throttle. Repeat entries are gated by the existing tier table, which returns SKIP when the device
hasn't moved — measured at ~5 ms with no network and no OS re-registration.

## Behaviour

Only `AUTOMATIC` mode changes, and only on foreground entry. No persisted state is added, so there is
nothing to migrate on upgrade and no key an older install could be missing.

## Testing

Three unit tests. Four mutations verified, each caught only by its own test: dropping the AUTOMATIC
guard, dropping the stuck-sync guard, arming after the request instead of before, and never arming.

Emulator with a mock-location provider, 23 km jump with the app backgrounded, then foregrounded —
patched vs an unpatched control build on the same app data:

|  | patched | control |
|---|---|---|
| silent fix on foreground entry | yes, +32 ms after the flush | none |
| sync triggered | yes, +39 ms | none |
| re-registered at the new position | yes | not until the OS trigger fired |

The patched build also beat the OS's own movement trigger by 3.3 s. Five foreground entries while
stationary each acquired a fix and each hit tier SKIP in ~5 ms.

1678 unit tests across all modules, ktlint, `apiCheck` and `lintDebug` clean.

## Known limitation

This does not make Play Services deliver callbacks; a device that never opens the app still sees the
gap. The emulator cannot reproduce hours-long dormancy, so what is measured above is the mechanism
and the latency, not a fix for the dormancy itself. Keeping a monitored boundary near the device at
all times — concentric discovery rings — is the only lever with a mechanism against that, and it is
feature-sized: it spends slots from the 19+1 budget and changes registration, eviction and the tier
table.

## Stack

```
main
 └── geofence/nearest-edge-distance      #806
 └── geofence/containment-guards         #805
      └── geofence/foreground-discovery  this PR
```

Shares `ModuleGeofence`'s foreground hook with #805, so merge #806 → #805 → this.

---

> [!NOTE]
> **Medium Risk**
> Touches foreground lifecycle, location acquisition, and identity reads; behavior is gated but repeat foregrounds can trigger silent fixes (tier table should SKIP when stationary).
> 
> **Overview**
> **AUTOMATIC** geofence mode now tries to **rediscover nearby fences from a real GPS fix** whenever the app returns to the foreground, instead of only running the stuck-sync retry path.
> 
> On foreground, work moves to **`geofenceScope`** (identity read / Keystore decrypt off the main thread). New **`refreshOnForeground`** arms `onRefreshRequested()` then requests a silent location update—same pattern as `refreshFromCurrentLocation()`—so movement can be detected when OS geofence callbacks go quiet. It is skipped for **MANUAL** mode, when **no user is identified**, or when a sync is **already awaiting location** (existing `retrySyncAwaitingLocation` still handles that).
> 
> Unit tests cover those guards and **arm-before-request** ordering.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d8c56aec0b9cb99ceb7757350504545c504d7d8b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/geofence/src/main/kotlin/io/customer/geofence/ModuleGeofence.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/ModuleGeofence.kt
@@ -17,6 +17,7 @@ import io.customer.sdk.core.di.SDKComponent
 import io.customer.sdk.core.module.CustomerIOModule
 import io.customer.sdk.core.util.HandlerMainThreadPoster
 import io.customer.sdk.core.util.MainThreadPoster
+import io.customer.sdk.data.store.SecureUserStore
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.launch
 
@@ -170,8 +171,17 @@ class ModuleGeofence @JvmOverloads constructor(
                     regionStore = sdkAndroid.geofenceRegionStore,
                     logger = logger,
                     onForeground = {
-                        if (!refreshOnForeground(sdkAndroid.geofenceServices, locationModule)) {
-                            retrySyncAwaitingLocation(sdkAndroid, locationModule)
+                        // Off the main thread: deciding whether to take a fix needs the identity read,
+                        // a Keystore decrypt that can block for hundreds of ms on some OEMs.
+                        val foregroundScope = SDKComponent.scopeProvider.geofenceScope
+                        foregroundScope.launch {
+                            try {
+                                if (!refreshOnForeground(sdkAndroid.geofenceServices, sdkAndroid.secureUserStore, locationModule)) {
+                                    retrySyncAwaitingLocation(sdkAndroid, locationModule)
+                                }
+                            } finally {
+                                foregroundScope.cancel()
+                            }
                         }
                     }
                 )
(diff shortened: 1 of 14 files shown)


geofence/src/main/kotlin/io/customer/geofence/ModuleGeofence.kt
```diff
@@ -200,6 +204,25 @@ class ModuleGeofence @JvmOverloads constructor(
...
+     * hours, and every other path anchors at the last registration center — where the device used to
+     * be — so nothing else notices it has moved.
+     *
+     * Returns false when no fix was taken and [retrySyncAwaitingLocation] handles the entry instead:
+     * MANUAL leaves fixes to the host, and a sync already stuck without one has its own recovery.
+     */
+    @VisibleForTesting
+    @OptIn(InternalCustomerIOApi::class)
+    internal fun refreshOnForeground(services: GeofenceServices, locationModule: ModuleLocation): Boolean {
+        if (moduleConfig.locationMode != GeofenceLocationMode.AUTOMATIC) return false
```
Shahroz16, line 232: Could we avoid requesting this foreground fix until a user is identified? In automatic mode this runs on every resume before identify and after sign-out. `onLocationAcquired` then consumes the flag and returns at the null-user check, so the fix cannot drive a sync. The launch path already checks the persisted user. Adding the same guard here would avoid unnecessary location access and battery use.

Shahroz16 (approved):

geofence/src/main/kotlin/io/customer/geofence/ModuleGeofence.kt
mrehan27: Agreed and fixed in d8c56aec. The after-sign-out case is the one that convinced me: it isn't a one-off before identify, it's a silent fix taken on *every* resume for the rest of the process, with no user to attribute it to.

The gate went into `refreshOnForeground` rather than the call site, so all three "don't take a fix" reasons sit together.

The reason it isn't a one-line addition: this runs from `onStart`, and `SecureUserStore.getUserId()` is a SharedPreferences read plus a Keystore decrypt — the same cost the launch path a few lines below explicitly refuses to pay on the main thread. So the foreground decision moved onto the geofence scope with the check.

One thing worth confirming, since it's load-bearing: returning `false` for the no-user case lets the entry fall through to `retrySyncAwaitingLocation`, which looks like it would then take a fix anyway. It can't — its first line returns unless `isAwaitingLocation()`, and we only reach the identity check when that's already false. So the fallback is a provable no-op here, and `false` keeps the return value meaning what its KDoc says ("no fix was taken") rather than "handled".

PR #808, opened 2026-08-04
chore(geofence): wait for the refresh slot instead of dropping a pass that carries a live fix

mrehan27:
A sync pass carrying the device's real position could be silently dropped in favour of one working from a stored anchor — a movement-trigger EXIT, or a location fix the SDK had just asked for. In the worst case that stops geofence discovery outright until the app is next opened, rebooted, or updated.

## Stack

```
release-july-launch-fixes
└── #805  gate transitions on tracked containment state
    └── #807  refresh discovery from a live fix on foreground entry
        └── #808  wait for the refresh slot  ← this PR
```

(#806 is already merged into the release branch.)

## The failure

`refresh()` and `handleMovement()` share one `refreshInProgress` gate, and until now the loser dropped its pass on the floor. That dedupe is right for `refresh()` — identify and app-launch fire together on every cold start and only one needs to run — but wrong for `handleMovement()`, because the two carry different information. A `refresh()` pass measures distance from the **last registration centre**, i.e. where the device used to be. Only the movement pass carries the OS's live triggering fix.

On a cold start driven by a trigger EXIT, the ordering is fixed and unlucky: the broadcast starts the process, module init publishes the restored identity first, that `refresh()` takes the gate, measures `distFromReg=0.0m`, and SKIPs — having consumed the slot to do nothing. The movement pass then finds the gate taken and returns.

**Why that is permanent rather than a missed beat:** once a GMS geofence has fired EXIT, the OS holds it in the outside state. It cannot fire again until it is re-registered. So a movement pass that returns without re-centring the trigger ends discovery outright — no further EXIT, no re-rank, no fetch. Recovery only comes from an app open (#807), a reboot, or an app update.

Field evidence, Pixel 6, Bahria Town drive:

```
18:47:10.406  BroadcastReceiver invoked by OS
18:47:10.414  sync triggered: user-identified
18:47:10.436  sync triggered: app-launch
18:47:10.448  skipped: refresh already in progress
18:47:10.487  sync triggered: movement-trigger-exit
18:47:10.492  skipped: refresh already in progress      ← the one that mattered
18:47:10.501  Refresh tier=SKIP (distFromFetch=0.0m, distFromReg=0.0m)
```

No geofence activity followed. The captured prefs showed `last_sync_timestamp` and `registered_ids` still pinned to the previous sync, with the device 2.3 km outside the trigger it had already exited.

## The fix

`awaitRefreshSlot()` — a movement pass now waits for the gate instead of dropping. `refresh()` keeps its existing drop-fast behaviour, so the identify/app-launch collapse is unchanged. Concurrency is still exactly one pass at a time.

The wait is bounded by `withTimeoutOrNull`, and on expiry the behaviour is precisely what it is today, so the change can only add work in a case that previously did none.

`MOVEMENT_SLOT_WAIT` was first set to 3 seconds, bracketed above a local pass (~90 ms measured) and below the receiver's `DISPATCH_WAIT_BUDGET_MS`. **Review corrected that** (thanks @Shahroz16): the dispatch budget is the wrong ceiling, because it bounds only how long the receiver *joins* — the sync itself runs on the longer-lived services scope and survives the receiver returning. Three seconds also cannot cover a remote holder, which may spend the HTTP client's connect plus read timeout (10 s each) before it releases; and a holder that then *fails* registers nothing, so it never re-centres the fired trigger on its way out. That is the case that strands the trigger, and giving up on it hands the job to nobody. The wait is now **30 seconds**, deliberately past the dispatch budget: a pass that lands late still re-centres, where a dropped one leaves nothing to re-centre.

It cannot block or ANR: the receiver caps its own join at the remaining budget and calls `pendingResult.finish()` regardless, both gate holders clear the flag in a `finally`, nothing holds `stateMutex` while waiting on the gate, and the wait is a `delay()` on `Dispatchers.default` rather than a held thread.

## Test correction

`handleMovement_givenSecondCallWhileFirstInFlight_expectSecondDropped` encoded the wrong guarantee and is renamed to `...expectSerializedNotDropped`. Two movement passes should serialize but both must run — each carries its own fix, and each has to re-centre the trigger.

## Verification

- Full suite green; `apiCheck`, `ktlint`, `lintDebug` clean.
- Two mutations — restoring the drop-on-contention, and a zero wait budget — each fail both movement tests and nothing else.
- On-device A/B, identical scenario (mock provider, `am kill`, cross the trigger boundary). Buggy build: movement dropped, `tier=SKIP`, no trigger re-registration. Fixed build: movement waits, `Movement refresh tier=LOCAL`, trigger removed and re-added, and a following `ENTER [cio_movement_trigger]` confirms it is armed again.

The natural race window is ~30–90 ms and the emulator usually wins it, so reproduction required deliberately widening it.

## Second commit: the stale-center case, now closed too

Waiting for the slot fixes the dropped pass, but not the pass that *wins* it and re-registers the trigger on a stale anchor — the trigger lands where the device isn't, and the OS holds it there. That is now handled at registration: the movement trigger uses `INITIAL_TRIGGER_EXIT`, so the OS evaluates the device's position at register time and a misplaced center fires EXIT immediately, which re-centers on a real fix. Boot restore already worked this way; the normal path did not.

ENTER is dropped rather than combined with EXIT. It only delivered a transition the receiver ignores, and an on-device A/B showed the combination stops GMS delivering the trigger's later EXIT altogether — the trigger went silent for 420 s+ where ENTER-only fired in 5 s, with a business fence's EXIT still delivering in the same run to prove the rig was live. Every assertion below was re-run with a positive control for that reason.

Still open: acting on business-fence transitions as a second discovery input. The 4 Aug drive showed it would have re-armed 4m43s earlier, held back because more frequent re-registration risks more phantom events.

## Device verification of the registration change

| scenario | result |
|---|---|
| EXIT-only trigger, device inside at register, then moves out | EXIT fired in 5 s |
| EXIT-only trigger registered on a stale center, device elsewhere | EXIT fired immediately, re-centered on the live fix |
| normal path EXIT-only, device moves out (business EXIT as control) | both fired in 6 s |
| `ENTER or EXIT` | trigger never fired — rejected |

## Third commit: the same defect on the other entry point

`refresh()` keeping its drop-fast behaviour is right for identify and app-launch — but not for a returning location fix, which reaches `refresh()` too. That pass carries the device's real position while the holder works from a stored anchor, and `onLocationAcquired` consumes the arming flags the moment the fix lands, so a dropped pass spends the fix without using it and nothing requests another.

Field evidence, Pixel 6, 5 Aug — the first app open after moving:

```
16:33:34.369  sync triggered: user-identified        ← takes the slot, runs REMOTE at the stored anchor
16:33:34.483  Foreground entry: requested a live fix
16:33:34.768  Location update received: 31.3686501, 74.1715
16:33:34.787  sync triggered: user-identified
16:33:34.788  skipped: refresh already in progress   ← the fix, discarded
16:33:36.156  Nearest filter: 8 candidates → 4 selected ids=[6, 3, 2, 1]
```

The set was registered around an anchor **2016 m** from the device and stayed that way for **2 m 24 s**, recovering only because this PR's `INITIAL_TRIGGER_EXIT` change fired. On 4.20.0 the trigger would have been registered at that stale centre with `INITIAL_TRIGGER_ENTER`, with the device already outside it, so no crossing would ever occur.

`refresh()` and a new `refreshFromLiveFix()` now differ only in how they take the slot — try versus await — and share one body. `onLocationAcquired` dispatches the new one under its own reason string, so device logs finally distinguish a returning fix from an identify.

**Cost:** two remote fetches on a cold start where there was one. It is self-limiting — the waiting pass re-reads freshness at the live coordinates, so it only fetches when the positions actually disagree. Measured: a fix injected at the anchor gives `tier=SKIP`, `distFromFetch=0.0m`, and zero extra network calls.

### Verification of the third commit

- 1696 tests across all 9 CI modules, 0 failures; `apiCheck`, `lintDebug`, ktlint clean. The existing test pinning `refresh()`'s drop behaviour still passes untouched.
- Mutation: switching `awaitRefreshSlot()` back to `tryTakeRefreshSlot()` fails the new regression test and only that test.
- Emulator A/B, identical seeded state, a fix injected 2 km from the anchor ~700 ms into a holder's pass:

| build | result |
|---|---|
| without the fix | `skipped: refresh already in progress`; ends monitoring three fences 2 km away |
| with the fix | `took the slot after 2034ms`, re-fetch at the live position, ends correctly with the trigger re-centred |

- The same run shows both policies coexisting: `app-launch` still dropped while `location-acquired` waited.
- The give-up branch was forced with a throwaway 10 ms build: bounded, logged, returns success, and a following pass still acquires the slot — expiry does not latch the gate.

Play services on the emulator ignores mock providers and `emu geo fix` (silent requests just time out), so fixes were injected through `setLastKnownLocation`, the path a MANUAL-mode host uses. The upstream request-to-arrival leg is covered by the field log above.

---

> [!NOTE]
> **Medium Risk**
> Touches core geofence sync concurrency and GMS movement-trigger behavior; wrong slot or trigger logic could strand discovery or mis-register fences, but identify/app-launch dedupe is unchanged and coverage is strong.
> 
> **Overview**
> Fixes cases where geofence discovery could stall until reboot or app reopen: movement-trigger EXIT and returning GPS fixes were dropped when they lost a race to identify/app-launch `refresh()`, and the movement trigger could stay registered on a stale center.
> 
> **Refresh slot:** `handleMovement()` and a new `refreshFromLiveFix()` wait up to **30s** (`awaitRefreshSlot`) for an in-flight sync instead of returning immediately; identify/app-launch `refresh()` still drops on contention. Shared logic lives in `runRefresh()`.
> 
> **Services:** `onLocationAcquired` calls `refreshFromLiveFix` (reason `location-acquired`) so a one-shot fix is not wasted if a concurrent refresh holds the slot.
> 
> **GMS registration:** The movement trigger always uses `INITIAL_TRIGGER_EXIT` (not ENTER); boot restore delegates to the same path as normal `replaceGeofences`.
> 
> Tests cover serialization, waiting through slow/failed holders, and live-fix registration at the real coordinates.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 97e33c3a6590b1c094c9ba991e490ae837f9cacc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
@@ -86,8 +86,7 @@ internal class GeofenceRepositoryImpl(
             }
 
             val config = store.getCachedConfigOrFallback()
-            // Captured with the coordinates, before any network or GMS await: an exit claimed while
-            // this sync is in flight must beat the geometry derived from this fix.
+            // Captured with the coordinates: an exit claimed mid-sync must beat this fix's geometry.
             val containmentEpoch = store.containmentEpoch()
             // Decided under stateMutex so a concurrent sign-out reset can't wipe state right
             // after this reads pre-wipe freshness/registrations and SKIPs — that would leave
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
@@ -24,12 +24,23 @@ import kotlinx.coroutines.withTimeoutOrNull
  *
  * - [refresh] — for identify / app-launch. Reuses the cached set within the freshness window
  *   (re-registering locally or skipping); otherwise fetches fresh from the API.
+ * - [refreshFromLiveFix] — same pass, for a fix the SDK asked for and received.
  * - [handleMovement] — for movement-trigger EXIT. Re-ranks the cached regions for the new
  *   location.
  */
 internal interface GeofenceRepository {
     @RequiresPermission(Manifest.permission.ACCESS_FINE_LOCATION)
     suspend fun refresh(latitude: Double, longitude: Double): Result<Unit>
+
+    /**
+     * [refresh] for a just-acquired fix, which waits for an in-flight sync instead of dropping.
+     * The two run the same pass; they differ only in what a collision costs. Identify and
+     * app-launch collide constantly and share one anchor, so dropping loses nothing — but this
+     * caller holds the device's real position while the sync in flight is working from a stored
+     * anchor that can be kilometres stale, and the fix is consumed whether or not it is used.
+     */
+    @RequiresPermission(Manifest.permission.ACCESS_FINE_LOCATION)
+    suspend fun refreshFromLiveFix(latitude: Double, longitude: Double): Result<Unit>
 
     @RequiresPermission(Manifest.permission.ACCESS_FINE_LOCATION)
     suspend fun handleMovement(latitude: Double, longitude: Double): Result<Unit>
@@ -67,9 +78,10 @@ internal class GeofenceRepositoryImpl(
     private val logger: GeofenceLogger
 ) : GeofenceRepository {
 
-    // One sync pass at a time. refresh() drops when the slot is taken — identify and app-launch
-    // fire together and only one needs to run — while a movement pass waits for it instead, see
-    // [awaitRefreshSlot]. Released in `finally` so a failure or cancellation can't latch the gate.
+    // One sync pass at a time. A pass carrying its own live fix waits for the slot rather than
+    // dropping — see [awaitRefreshSlot]; [refresh] drops, because identify and app-launch fire
+    // together on the same anchor and only one needs to run. Released in `finally` so a failure
+    // or cancellation can't latch the gate.
     private val refreshInProgress = AtomicBoolean(false)
 
     // Serializes state-mutation against reset() (sign-out). Held only around the
@@ -88,6 +100,20 @@ internal class GeofenceRepositoryImpl(
             logger.logSyncSkipped("refresh already in progress")
             return Result.success(Unit)
         }
+        return runRefresh(latitude, longitude)
+    }
+
+    @RequiresPermission(Manifest.permission.ACCESS_FINE_LOCATION)
+    override suspend fun refreshFromLiveFix(latitude: Double, longitude: Double): Result<Unit> {
+        if (!awaitRefreshSlot()) {
+            logger.logSyncSkipped("refresh already in progress after waiting")
+            return Result.success(Unit)
+        }
+        return runRefresh(latitude, longitude)
+    }
+
+    @RequiresPermission(Manifest.permission.ACCESS_FINE_LOCATION)
+    private suspend fun runRefresh(latitude: Double, longitude: Double): Result<Unit> {
         try {
             val userId = secureUserStore.getUserId()
             if (userId.isNullOrBlank()) {
@@ -187,9 +213,10 @@ internal class GeofenceRepositoryImpl(
     private fun osStateWiped(): Boolean = osStateWipedByReboot() || osStateWipedByAppUpdate()
 
     /**
-     * Takes the in-flight slot for a movement pass, waiting rather than dropping: the OS holds a
-     * fired trigger in the outside state, so a pass that returns without re-centring it stops
-     * discovery until the next app open, reboot or update.
+     * Takes the in-flight slot for a pass that can't be dropped, waiting for the holder instead.
+     * For a movement pass, the OS holds a fired trigger in the outside state, so returning without
+     * re-centring it stops discovery until the next app open, reboot or update; for a live fix, the
+     * fix is spent on arrival and nothing re-requests one.
      */
     private suspend fun awaitRefreshSlot(): Boolean {
         if (tryTakeRefreshSlot()) return true
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
@@ -401,6 +401,11 @@ internal class GeofenceRepositoryImpl(
             // override here (unlike the registration diff): callers disable synthesis outright
             // while the reboot flag is up — the anchor can predate the reboot.
             val unchangedRegistered = unchangedRegisteredIds(nearest)
+            // Registered before this pass but not cache-equal: the ID survived a geometry edit.
+            val previouslyRegistered = store.getRegisteredIds()
+            val reRegisteredIds = nearest.map { it.id }
+                .filter { it in previouslyRegistered && it !in unchangedRegistered }
+                .toSet()
             val registrationResult = register(regionsToRegister).also { result ->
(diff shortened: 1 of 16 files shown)


geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
```diff
@@ -176,10 +186,27 @@ internal class GeofenceRepositoryImpl(
...
     private fun osStateWiped(): Boolean = osStateWipedByReboot() || osStateWipedByAppUpdate()
 
+    /**
+     * Takes the in-flight slot for a movement pass, waiting rather than dropping: the OS holds a
+     * fired trigger in the outside state, so a pass that returns without re-centring it stops
+     * discovery until the next app open, reboot or update.
+     */
+    private suspend fun awaitRefreshSlot(): Boolean {
+        if (tryTakeRefreshSlot()) return true
+        return withTimeoutOrNull(MOVEMENT_SLOT_WAIT) {
```
Shahroz16, line 223: Could we keep the movement pass pending instead of dropping it after three seconds? A cold-start refresh can hold the slot in a remote fetch beyond this timeout. If that fetch fails, `refresh()` registers nothing, so the new `INITIAL_TRIGGER_EXIT` recovery never runs; this movement fix has already returned and the fired trigger stays stranded until another foreground, reboot, or update path. The receiver already bounds only how long it joins the services-scope job. Replaying the latest movement fix when the slot opens, or falling back to a local re-rank when the holder fails, plus a test where the holder exceeds three seconds and fails, would close the remaining loss case.

Shahroz16 (approved):

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
mrehan27: Agreed and fixed in 1a6b10f0. Three seconds was sized against a local pass (~90ms measured) with the receiver's dispatch budget as the ceiling. That ceiling was the wrong constraint: a remote holder can spend the HTTP client's connect plus read timeout (10s each) before it releases, so no value inside an 8s budget could ever cover it.

The wait now outlasts that holder, deliberately past `DISPATCH_WAIT_BUDGET_MS`. As you note, the receiver bounds only how long it joins — the sync runs on the services scope and survives the receiver returning, so a pass that lands late still re-centres, where a dropped one leaves nothing to re-centre. Your framing is what made me stop treating the two bounds as one.

I took the wait over the replay you suggested. Once the wait covers the holder, the movement pass gets the slot and its existing re-arm-on-failure path runs — which is the local re-rank fallback, so both of your options end in the same place, and replay would be a second mechanism for a guarantee the first now provides. If you'd rather have the pass survive independently of the holder's duration I'm happy to revisit, but I didn't want two overlapping recovery paths in a patch.

Test added as you asked: the holder goes remote, delays past the old three seconds, and fails. It asserts the trigger is still re-centred on the movement fix and that nothing logged the skip. Reverting the constant to `3.seconds` fails only that test.

Residual, stated plainly: if the process is killed after the receiver returns, the pending pass is still lost. That was true at three seconds too, and there's nothing to hold the process past `goAsync` short of moving the pass to WorkManager, which I'd rather not do in a patch.

PR #809, opened 2026-08-05
fix(geofence): correct transition accuracy and region discovery reliability

mrehan27:
> [!IMPORTANT]
> **Squash-merge this PR.** All four commits below are `chore`, so a merge commit would put no releasable message on `main` and cut no release at all — silently, with no error. Squashing makes this PR's title the single commit message semantic-release reads, which is why it is a `fix(...)`, and why the title deserves a deliberate read before merging: it becomes the version bump *and* the changelog entry for the whole patch.

### What this is

A release branch collecting post-4.20.0 geofence fixes so one release is cut for all of them rather than one per merge. Nothing here is new work — the diff is exactly the union of the four PRs below, each reviewed and merged on its own.

| PR | What | Reviewed |
|---|---|---|
| #806 | Nearest-set selection ranks by boundary distance instead of centre, so a large region the device is inside is not evicted by nearer centres | Approved |
| #805 | Transitions gated on tracked containment: an EXIT for a region never entered is dropped, and a re-reported ENTER no longer becomes a second arrival | Approved, 4 review rounds + Bugbot |
| #807 | Discovery refreshes from a live fix on foreground entry, and is skipped entirely when no user is identified | Approved |
| #808 | A sync pass carrying the device's real position waits for the in-flight slot instead of being dropped; a stale movement-trigger centre now self-heals | Approved, 2 review rounds + Bugbot |

**Read those four for the reasoning, trade-offs and review history.** This description deliberately doesn't restate them.

### Reviewing this PR

The diff is the concatenation of four already-reviewed changes: 4 commits, 32 files, and `git log main..release-july-launch-fixes` contains nothing but the four squash commits. Everything in it has been through review on the PR it came from, and the discussion lives there.

What is genuinely new at this level is the *combination*. #805 and #808 both edit `GeofenceRepository.kt`, and three pairs interact in ways no sub-PR could show on its own:

- **#807 depends on #808 to be useful on a cold start.** #807 requests a live fix on foreground entry; before #808's third commit that fix's pass was dropped whenever module init still held the sync slot. Both halves of this are in one field log from 5 Aug: at `16:33` the fix was discarded and the set was registered around an anchor 2016 m away for 2 m 24 s; at `16:35` the same device, with the slot free, registered correctly. #808's live-fix commit is what closes the first case.
- **#806 reduces how often #805's mark-prune fires.** #805 drops a region's reported-ENTER mark when it leaves the monitored set, which is what would otherwise let a duplicate arrival through on re-registration. #806 stops evicting regions the device is standing inside, so the occupied-region case that made that prune dangerous now largely doesn't arise. They compound in the same direction.
- **#808's EXIT-only trigger increases re-registration, which #805's guards absorb.** A trigger registered on a stale centre now fires EXIT immediately and re-registers, so initial-ENTER synthesis runs more often. Without #805's redundant-ENTER guard that would mean more duplicate arrivals; with it, the repeats are suppressed. Observed 4 times on the 5 Aug drive, each correctly dropped.
- **Sequential passes over the containment state.** #808 makes a second pass run immediately after a first, so `reconcileEnteredIds` can now run twice in quick succession with different coordinates. #805's epoch mechanism is what keeps that safe: a transition reported between the two passes outranks the older pass's geometry.

### Validation of the combined branch

- Full suite green on the merged branch — **1703 tests, 0 failures** across all 9 CI modules. `apiCheck`, `lintDebug` and ktlint clean; no public API change; no compiler warnings in the touched modules.
- **On-device drive, Pixel 6, 5 Aug (2 h 15 m):** 10/10 movement triggers delivered, 6 complete ENTER/EXIT pairs, **zero false EXITs**, and 4 duplicate arrivals correctly suppressed. The cold-start stranding fix was observed recovering in 1.3 s. That build carried #805, #806, #807 and #808's first two commits.
- **#808's third commit (the live-fix wait) landed after that drive** and was validated by an emulator A/B instead: same seeded state, a fix injected 2 km from the anchor mid-pass. Without it, the fix is discarded and the device ends up monitoring regions 2 km away; with it, the pass waits 2034 ms and re-registers at the live position. The same run shows `app-launch` still dropping while `location-acquired` waits, which is the distinction the change rests on.
- **#805's geometry-drift path** was exercised on the emulator in both directions: a region whose server geometry moves away from the device resets containment and its mark; one that still contains the device keeps both, and the synthesized ENTER is suppressed rather than delivered as a duplicate.
- The 30-second slot wait's give-up branch was forced with a throwaway short-wait build: bounded, logged, returns success, and a following pass still acquires the slot — expiry does not latch the gate.

### Adding to this release

Target this branch, squash-merge, and title it `chore(...)` like the four above — individual titles never reach `main`, so they cannot affect the version. Before this PR merges:

- If anything landing here is a **feature**, retitle this PR to `feat(...)` or a minor ships as a patch.
- Widen this PR's title if the contents stop being geofence-only.
- If `main` moves during the window, merge it into this branch so CI tests what will actually ship.

---

> [!NOTE]
> **Medium Risk**
> Touches core geofence event delivery, OS registration, and location-driven sync with new persisted state and concurrency; behavior is heavily tested but regressions would affect campaign triggers in the field.
> 
> **Overview**
> **Geofence transition handling** gains persisted *containment* and *reported ENTER* state so GMS reconciliation artifacts are filtered: unmatched EXITs drop (with carve-outs for exit-only fences, uncached regions, and pre-upgrade installs), redundant ENTERs drop when an EXIT-monitored fence was already delivered, and a process-wide mutex serializes per-fence bookkeeping across concurrent broadcasts. Delivery runs through `handleBusinessTransition`, which records ENTER before identity checks and passes `monitorsExit` into the emitter.
> 
> **Discovery and sync** change how regions are chosen and when passes run. Nearest-set selection uses **boundary** distance (`edgeDistanceTo`), meter rounding, and id tie-breaks (aligned with iOS). Movement and **live-fix** sync paths **wait up to ~30s** for the refresh slot via `refreshFromLiveFix` instead of dropping; identify/app-launch still drop on collision. Foreground (automatic mode, identified user) arms a silent fix refresh; acquired locations route to `refreshFromLiveFix`. Registration seeds/reconciles containment with a **containment epoch**, resets state when geometry moves away, and tightens initial-ENTER synthesis to require store + geometry agreement.
> 
> **GMS registration** uses `INITIAL_TRIGGER_EXIT` for the movement trigger (boot restore unified with normal replace). **GeofenceServices** rejects invalid coordinates, and coroutine work is wrapped so sync/reset failures log instead of crashing the host.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 37c55a63ae63f6da1685c21f2821f04c3c35b033. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
@@ -12,6 +12,7 @@ import io.customer.geofence.di.geofenceRegionStore
 import io.customer.geofence.di.geofenceServices
 import io.customer.geofence.di.geofenceTransitionEmitter
 import io.customer.sdk.communication.Event
+import io.customer.sdk.core.di.AndroidSDKComponent
 import io.customer.sdk.core.di.SDKComponent
 import io.customer.sdk.core.di.clock
 import io.customer.sdk.core.di.setupAndroidComponent
@@ -20,6 +21,8 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
 import kotlinx.coroutines.withTimeoutOrNull
 
 /** Receives OS geofence transition callbacks and dispatches them to the SDK. */
@@ -94,7 +97,6 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
         val timestamp = SDKComponent.clock.currentTimeSeconds()
         val dispatchStartUptimeMs = SDKComponent.clock.elapsedRealtime()
         val androidComponent = SDKComponent.android()
-        val transitionEmitter = androidComponent.geofenceTransitionEmitter
         // Defense-in-depth against orphans (failed clearAll, app-data wipe, SDK
         // ID-format changes): events for unregistered IDs are dropped and the OS-side
         // registration is removed so it stops firing.
@@ -128,27 +130,17 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
                 }
             }
 
-            // Snapshot userId so a sign-out + sign-in before delivery can't reattribute this
-            // transition. Empty userId is treated as "not identified" per `isUserIdentified`.
-            val userId = androidComponent.secureUserStore.getUserId()?.takeIf { it.isNotEmpty() }
-            // Geofencing is identified-only: the backend rejects anonymous geofence tracks, so an
-            // anonymous transition has no deliverable path. Drop it before spending a cooldown slot
-            // or persisting a row neither channel could ever send.
-            if (userId == null) {
-                logger.logTransitionDroppedAnonymous(geofenceId, transition.name)
-                return@forEach
+            // Each broadcast runs on its own scope, so two transitions for one fence would
+            // otherwise interleave read-decide-persist and lose a containment record.
+            transitionMutex.withLock {
+                handleBusinessTransition(
+                    geofenceId = geofenceId,
+                    transition = transition,
+                    timestamp = timestamp,
+                    androidComponent = androidComponent,
+                    logger = logger
+                )
             }
-
-            val cachedRegion = androidComponent.geofenceRegionStore.getCachedRegion(geofenceId)
-            transitionEmitter.emit(
-                geofenceId = geofenceId,
-                transition = transition,
-                userId = userId,
-                timestampSeconds = timestamp,
-                geofenceName = cachedRegion?.name,
-                metadata = cachedRegion?.metadata ?: emptyMap(),
-                geosetIds = cachedRegion?.geosetIds ?: emptyList()
-            )
         }
 
         // Hold the goAsync window open until the refresh lands so the OS doesn't kill a
@@ -163,6 +155,55 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
         }
     }
 
+    private suspend fun handleBusinessTransition(
+        geofenceId: String,
+        transition: Event.GeofenceTransition,
+        timestamp: Long,
+        androidComponent: AndroidSDKComponent,
+        logger: GeofenceLogger
+    ) {
+        // Ahead of the identity checks below: being inside a fence is a physical fact, independent of
+        // whether the transition is deliverable.
+        val store = androidComponent.geofenceRegionStore
+        val cachedRegion = store.getCachedRegion(geofenceId)
+        when (transition) {
+            Event.GeofenceTransition.ENTER -> store.recordEntered(geofenceId)
+            // An unmatched EXIT is a GMS reconciliation artifact, not a crossing — delivering it
+            // fires EXIT campaigns at people who were never there. The two guards after claimExit
+            // cover the cases where an absent record proves nothing: an upgraded install with no set
+            // yet, and a region that never monitored ENTER (a missing cache row counts as unknown).
+            Event.GeofenceTransition.EXIT -> if (
+                !store.claimExit(geofenceId) &&
+                store.hasContainmentRecord() &&
+                cachedRegion?.transitionTypes?.contains(GeofenceTransitionType.ENTER) == true
+            ) {
+                logger.logExitDroppedNeverEntered(geofenceId)
+                return
+            }
+        }
+
+        // Snapshot userId so a sign-out + sign-in before delivery can't reattribute this
+        // transition. Empty userId is treated as "not identified" per `isUserIdentified`.
+        val userId = androidComponent.secureUserStore.getUserId()?.takeIf { it.isNotEmpty() }
+        // Identified-only: the backend rejects anonymous geofence tracks, so drop before spending a
+        // cooldown slot or persisting a row neither channel could send.
+        if (userId == null) {
+            logger.logTransitionDroppedAnonymous(geofenceId, transition.name)
+            return
+        }
+
+        androidComponent.geofenceTransitionEmitter.emit(
+            geofenceId = geofenceId,
+            transition = transition,
+            userId = userId,
+            timestampSeconds = timestamp,
+            geofenceName = cachedRegion?.name,
+            metadata = cachedRegion?.metadata ?: emptyMap(),
+            geosetIds = cachedRegion?.geosetIds ?: emptyList(),
+            monitorsExit = cachedRegion?.transitionTypes?.contains(GeofenceTransitionType.EXIT) == true
+        )
+    }
+
     private fun transitionName(gmsTransitionType: Int): String = when (gmsTransitionType) {
         Geofence.GEOFENCE_TRANSITION_ENTER -> "ENTER"
         Geofence.GEOFENCE_TRANSITION_EXIT -> "EXIT"
@@ -174,5 +215,8 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
         // goAsync grants ~10s before the OS considers the receiver blocked; total budget for
         // one dispatch (persistence + GMS awaits + movement-refresh wait), with headroom.
         private const val DISPATCH_WAIT_BUDGET_MS = 8_000L
+
+        // Process-wide: a receiver instance lives for one broadcast, so the lock has to outlive it.
+        private val transitionMutex = Mutex()
     }
 }
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
@@ -12,20 +12,35 @@ import io.customer.sdk.communication.Event
 import io.customer.sdk.core.util.Clock
 import io.customer.sdk.data.store.SecureUserStore
 import java.util.concurrent.atomic.AtomicBoolean
+import kotlin.time.Duration.Companion.milliseconds
+import kotlin.time.Duration.Companion.seconds
+import kotlinx.coroutines.delay
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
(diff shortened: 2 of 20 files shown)


geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
```diff
@@ -163,6 +155,55 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
...
+        androidComponent.geofenceTransitionEmitter.emit(
+            geofenceId = geofenceId,
+            transition = transition,
+            userId = userId,
+            timestampSeconds = timestamp,
+            geofenceName = cachedRegion?.name,
+            metadata = cachedRegion?.metadata ?: emptyMap(),
+            geosetIds = cachedRegion?.geosetIds ?: emptyList(),
+            monitorsExit = cachedRegion?.transitionTypes?.contains(GeofenceTransitionType.EXIT) == true
+        )
```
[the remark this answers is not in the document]
mrehan27: Half of this is by design, half is real — and the real half is sharper than described. (For context, this is a re-raise of the first finding on #805, resolved there; a release PR re-reviews the union, so resolved sub-PR threads come back.)

**The cooldown case is not a loss.** Suppression means the first EXIT in the window was already delivered and this one is a repeat. Clearing containment is correct regardless of deliverability: the device left, which is a physical fact, and the record tracks where the device *is* rather than what the backend has been told. That is why `claimExit` sits ahead of the identity and cooldown checks rather than after them.

**The persist-failure case is real, and the asymmetry is the interesting part.** On a failed `appendAll` the emitter rolls back the cooldown so "the crossing retries". That holds for ENTER — GMS re-reports it on the next re-registration. It does not hold for EXIT: once a fence has fired EXIT the OS holds it in the outside state until it is re-registered, so no retry is coming. And since `claimExit` has already consumed the record, a later reconciliation EXIT is dropped by the never-entered guard as well. So that EXIT is gone, and the backend keeps an unbalanced ENTER until the user's next arrival.

Not fixing it in this patch, on impact rather than provenance:

- The trigger is `appendAll` failing, which is an atomic temp-and-rename write hardened in #778 — realistically disk-full or IO error, and if that is happening ENTERs are failing in the same breath.
- No occurrences in the 2 h 15 m field drive behind this release; every transition logged `delivered via WorkManager`.
- The fix is not the rollback implied here. Restoring containment on *cooldown* suppression would be actively wrong — it would resurrect the device into a fence it has left, and the next sync would carry that record forward. So `emit` would first have to distinguish "suppressed" from "failed"; today it returns a bare `false` for redundant-ENTER, cooldown and persist failure alike, and the receiver ignores the value entirely. That is a change to the emitter's contract plus receiver wiring, on a path with no field evidence, in a patch whose four PRs are already reviewed and merged.

Documenting the asymmetry rather than filing it. The backend already cannot rely on a guaranteed EXIT from a mobile SDK — uninstall, revoked permission, a wiped device and an OS-dropped registration all end the same way — so an at-most-once EXIT under a failed disk write is a new cause of a failure mode the server side must already tolerate, not a new class of it. If persistence failures ever do show up in the field, the fix is a typed emit outcome with containment restored only on genuine failure.

geofence/src/main/kotlin/io/customer/geofence/store/GeofenceRegionStore.kt
```diff
@@ -166,6 +326,9 @@ internal class GeofenceRegionStoreImpl(
             remove(KEY_LAST_API_FETCH_LOCATION)
             remove(KEY_LAST_MOVEMENT_TRIGGER_LOCATION)
             remove(KEY_REGISTERED_IDS)
+            remove(KEY_ENTERED_IDS)
+            remove(KEY_EMITTED_ENTER_IDS)
+            remove(KEY_EMITTED_ENTER_OWNER)
```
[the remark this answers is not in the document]
mrehan27: The interleaving is real but every outcome it can produce is already safe, so a lock here would add ordering theater rather than correctness. Resurrected `entered_ids` is physically true (containment is deliberately identity-free — `claimExit` runs ahead of the identity checks for this reason) and the next reconcile prunes/re-seeds it; resurrected marks carry the old user's owner snapshot, which `hasEmittedEnter`'s owner scoping rejects for the next user; a claim racing the wipe fails open at `hasContainmentRecord`. The epoch maps are monotonic, so pre-sign-out entries always read as older than any later sync's capture and are trimmed on the next reconcile. The deeper reason not to lock: a transition landing just *after* an atomic wipe writes the same state back legitimately — sign-out vs. independent OS broadcasts cannot be ordered at this boundary, so the design makes all post-wipe writes safe instead.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
```diff
@@ -174,10 +212,28 @@ internal class GeofenceRepositoryImpl(
...
 
+    /**
+     * Takes the in-flight slot for a pass that can't be dropped, waiting for the holder instead.
+     * For a movement pass, the OS holds a fired trigger in the outside state, so returning without
+     * re-centring it stops discovery until the next app open, reboot or update; for a live fix, the
+     * fix is spent on arrival and nothing re-requests one.
+     */
+    private suspend fun awaitRefreshSlot(): Boolean {
+        if (tryTakeRefreshSlot()) return true
+        return withTimeoutOrNull(MOVEMENT_SLOT_WAIT) {
```
Shahroz16: Could we make this handoff cancellation-safe? `withTimeoutOrNull` can time out asynchronously even just before the block returns. If the current pass releases near the 30-second deadline, this waiter can win the CAS and still return `false`, leaving `refreshInProgress` set forever. Every later launch, movement, and live-fix refresh then drops or times out until process restart. A permit whose acquisition and release share one `try/finally`, plus a boundary test, would avoid latching the pipeline.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
```diff
@@ -506,4 +601,14 @@ internal class GeofenceRepositoryImpl(
         radius = radiusMeters,
         transitionTypes = listOf(GeofenceTransitionType.EXIT)
     )
+
+    private companion object {
+        // Long enough to outlast the slowest holder: a remote pass can spend the HTTP client's
+        // connect plus read timeout (10s each) before it releases, and giving up on one that then
+        // fails to register is the case that strands the trigger. Deliberately past the receiver's
```
Shahroz16, line 609: Would this late work need durable scheduling instead of relying on the services scope? Once the receiver's 8-second `goAsync()` budget ends and `finish()` runs, Android may kill a receiver-only process. A slow holder or registration can therefore terminate this movement pass before it re-centres the trigger, leaving discovery stranded until another app lifecycle event. Queueing the movement refresh durably before the receiver finishes, or keeping all required work within its window, would close that gap.

Shahroz16 (approved):
I found two reliability gaps that can still strand refreshes

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
mrehan27: Valid, and the inline comment defending it was wrong. Fixed in 37c55a63.

That comment claimed nothing suspends between the winning CAS and the return, so the deadline couldn't land in between. The reasoning doesn't hold: `withTimeoutOrNull` enforces the deadline by cancelling the `TimeoutCoroutine` from a scheduled task, and a block that completes normally into a job already in `Cancelling` loses — its value is discarded. Same mechanism as the familiar `withTimeout { mutex.lock() }` leak. Absence of a suspension point doesn't protect it.

Rather than compensate for a lost win by releasing after the fact, I removed the handoff. The wait is now bounded polling, so the CAS result *is* the return value and the slot can only be held by a caller that got `true`:

```kotlin
repeat(MOVEMENT_SLOT_ATTEMPTS) {
    if (tryTakeRefreshSlot()) return true
    delay(MOVEMENT_SLOT_POLL)
}
return tryTakeRefreshSlot()
```

The bound derives from the same 30s/50ms constants. One behavioural trade: scheduling delay now stretches the wait instead of cutting it short. That's the right direction here — the point is to outlast the holder, not to give up punctually.

On the boundary test you asked for, I want to be straight that I couldn't write one. The window needs real thread interleaving, and a single-threaded test scheduler can't produce it, so no black-box test distinguishes a correct implementation from a broken one. I wrote the compensating version first and a mutation confirmed its test had no teeth, which is what pushed me to the structural fix. The test that landed covers the give-up bound and that the slot stays usable afterwards — verified by mutation — but not the race itself. The safety there is by construction, not by coverage.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
mrehan27: Real gap. I'm not going to lean on it being pre-existing — that's provenance, not impact. Answering on impact, and leaving it out of this release.

The exposed window is narrower than the whole pass. The receiver joins for up to `DISPATCH_WAIT_BUDGET_MS`, so only the tail beyond ~8s runs unprotected, and delivery is already durable — `goAsync` is held until WorkManager has committed the work spec. What's at risk is the re-centring alone, and losing it degrades to "stranded until the next app open, fix, reboot or update" rather than to something permanent. Reaching it needs a holder slow enough to push past 8s *and* the OS to kill a cached process inside that tail.

The fix also doesn't cut cleanly in our favour. Scheduling the movement refresh through WorkManager subjects a geofence re-centring to doze deferral, and a re-centre delayed by an hour is worse for discovery than a best-effort one that usually lands in seconds. Tightening the wait to fit inside the 8s budget instead would reintroduce exactly the stranding this PR fixes. So the real options are "accept the tail" or "design durable re-centring properly" — and the second should cover the whole movement pass rather than the wait alone, and be validated on a device rather than bolted onto a release branch.

Taking the first for this release, deliberately.

PR #814, opened 2026-08-17
feat: add polygon wire contract and geometry foundation

Shahroz16:
## Summary

Adds the Android wire and geometry foundation for polygon geofences while preserving the existing circular path.

- decodes an explicit `shape`, GeoJSON `geometry`, and backend-provided `enclosing_circle`
- keeps `latitude`, `longitude`, and `radius` as the unchanged circle contract
- accepts only the V1 geometry profile: one convex outer ring, with no holes or multipolygons
- validates that the canonical backend circle contains every polygon vertex, then applies the Android wake margin
- never derives or replaces the canonical enclosing circle on-device
- provides exact point-in-polygon, boundary-distance, accuracy, and transition-confirmation primitives
- fails closed for missing, mixed, unknown, invalid, or unsupported polygon payloads
- persists polygon geometry through the existing cache

This PR does not register polygon wake circles, request location, or emit polygon transitions. Those runtime behaviors are isolated in the next stacked PR.

## V1 flow

```mermaid
flowchart LR
  A["Nearest response"] --> B{"shape"}
  B -->|"circle"| C["Existing OS circular path"]
  B -->|"polygon"| D["Validate convex ring and backend enclosing circle"]
  D --> E["Persist exact geometry and Android wake circle"]
  E --> F["#815 runtime"]
```

## Stack

1. **#814:** wire contract and geometry foundation
2. [#815](https://github.com/customerio/customerio-android/pull/815): wake-scoped responsive runtime

The earlier continuous-tracking prototype is outside V1. The all-in-one prototype branch remains untouched for historical testing and is not part of this review stack.

## Validation

- `:geofence:testDebugUnitTest`
- `:geofence:compileDebugAndroidTestKotlin`
- public API and repository binary validation
- Kotlin formatting and pre-push lint

Linear: [MBL-2287](https://linear.app/customerio/issue/MBL-2287/android-add-polygon-wire-contract-and-geometry-foundation)


geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonWakeCircle.kt
```diff
@@ -0,0 +1,77 @@
...
+internal class PolygonWakeCircleValidator {
+    fun prepare(
+        geometry: PolygonGeometry,
+        wakeCircle: PolygonWakeCircle
+    ): PolygonTriggerCircle {
+        require(wakeCircle.baseRadiusMeters.isFinite() && wakeCircle.baseRadiusMeters > 0.0) {
+            "polygon wake-circle radius must be finite and positive"
+        }
+        geometry.vertices.forEach { vertex ->
+            require(wakeCircle.center.distanceMetersTo(vertex) <= wakeCircle.baseRadiusMeters + CONTAINMENT_TOLERANCE_METERS) {
```
Shahroz16, line 34: The backend calculates base_radius_m on WGS84, while this check uses a 6,400 km sphere with only 1 m of tolerance. On a multi-kilometre fence, the client result can be tens of metres larger, so we can reject a polygon the backend has already admitted. Can we use the Android WGS84 distance calculation here, or apply the agreed relative tolerance?

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonGeometry.kt
```diff
@@ -0,0 +1,251 @@
...
+import kotlinx.serialization.Serializable
+
+@Serializable
+internal data class PolygonCoordinate(
+    @SerialName("latitude")
+    val latitude: Double,
+    @SerialName("longitude")
+    val longitude: Double
+) {
+    init {
```
Shahroz16, line 18: Could we keep PolygonCoordinate passive and validate ranges inside PolygonGeometry.from? init runs while the cached List<GeofenceRegion> is being decoded. One corrupt vertex therefore throws before fromOrNull can isolate that ring, and readJson removes the entire cached catalog, including valid circles.

mrehan27:
Closing in favour of #833, which landed on [`feature/geofence-polygons`](https://github.com/customerio/customerio-android/tree/feature/geofence-polygons) as `634cd850`. The foundation here is what shipped: wire contract, `PolygonGeometry`, `PolygonWakeCircle`, the `shape` discriminator and the `PolygonSupport` gate, with authorship preserved.

Two changes: the evaluator, route processor and transition state machine moved to the monitoring PR since nothing here calls them, and the four review findings on #833 are fixed.

#815 stays open until its replacement is up.

PR #819, opened 2026-08-24
feat: support FID based device registration for FCM

### LINKED ISSUE #756: FCM "token" deprecations
**Is your feature request related to a problem? Please describe.**
Firebase Android SDK recently deprecated FCM token APIs. Apps are now required to use the Firebase Installation ID (FID) as a replacement. The Customer.io SDK only supports the old/deprecated tokens.

**Describe the solution you'd like**
Add support for the new FID.

jcpsimmons:
Closes #756

## What

Firebase Cloud Messaging deprecated its registration token APIs in firebase-messaging 25.1.0 in favor of registering app instances by Firebase Installation ID (FID). This PR adds FID support to the messaging push module while keeping the existing token behavior for apps that have not opted in.

FID registration is opt-in through a manifest flag in the host app:

```xml
<meta-data android:name="firebase_messaging_installation_id_enabled" android:value="true" />
```

Enabling the flag makes the legacy token APIs throw, so an app that opts in is broken with the SDK as it is today: `FirebaseMessaging.getToken()` returns a failed task and no device token is ever registered.

## API replacements

| Deprecated | Replacement |
| --- | --- |
| `FirebaseMessaging.getToken()` | `FirebaseMessaging.register()` + FID from Firebase Installations |
| `FirebaseMessagingService.onNewToken(String)` | `FirebaseMessagingService.onRegistered(String)` |
| `FirebaseMessaging.deleteToken()` | `FirebaseMessaging.unregister()` (not used by the SDK) |

## How

- `FCMTokenProviderImpl` reads the same manifest flag FCM uses to pick the registration mode. With the flag enabled it calls `register()` to make sure the FCM backend knows about the app instance, then reads the FID from Firebase Installations and uses it as the device token. Without the flag it keeps the existing `getToken()` flow, which stays the only working API for apps that have not opted in.
- `CustomerIOFirebaseMessagingService` overrides `onRegistered()` and treats the delivered FID like a refreshed token. A static `onRegistered(context, installationId)` helper mirrors the existing `onNewToken(context, token)` one for apps that use their own messaging service.
- `firebase-messaging` updated from 24.1.2 to 25.1.2, which is where the new registration APIs live. `FirebaseInstallations` comes in transitively from firebase-messaging, no new dependency.
- Unit tests cover both registration modes: flag missing, flag disabled, flag enabled with success and failure of `register()` and of the FID lookup, and manifest lookup failures falling back to the token flow.

## Breaking change: minSdk 21 to 23

firebase-messaging 25.1+ requires minSdk 23 (Firebase raised the minimum across all its Android libraries in June 2025), so this bumps the SDK minimum from API 21 to API 23. The commit carries a `BREAKING CHANGE` footer so semantic-release treats it as a major bump. If we want to keep minSdk 21 for now, the alternative is staying on firebase-messaging 24.x, which caps us to a partial FID implementation without `register()`/`onRegistered()`.

## Notes for review

- Delivering pushes to a device registered by FID requires FID sends on the delivery side (the `fid` field in the FCM HTTP v1 API / Admin SDKs). Until that is in place end to end, customers should keep the flag disabled; nothing changes for them with this PR.
- FCM invokes `onRegistered()` with the identifier its registration store holds, so the service callback stays authoritative if a device ever falls back to legacy registration internally (very old Google Play services). The provider path and the callback path both publish `RegisterDeviceTokenEvent`, same as before.
- `messagingpush.api` regenerated with `apiDump`; `apiCheck`, ktlint, Android lint, and the unit test suites for all nine modules pass locally. The java_layout sample needed the same minSdk bump.

---

> [!NOTE]
> **Medium Risk**
> Breaking minSdk bump and push device identity changes when FID mode is enabled; delivery must support FID sends end-to-end. Legacy apps without the manifest flag should see unchanged token behavior.
> 
> **Overview**
> Adds **Firebase Installation ID (FID)** registration for FCM alongside the legacy token flow, and bumps **firebase-messaging** to **25.1.2** with **minSdk 23** (from 21).
> 
> When the host app sets manifest metadata `firebase_messaging_installation_id_enabled`, `FCMTokenProviderImpl` uses `FirebaseMessaging.register()` and the FID from Firebase Installations instead of deprecated `getToken()`. Without the flag, behavior stays on the legacy token path. `CustomerIOFirebaseMessagingService` adds `onRegistered()` (and a static helper) so FID updates publish the same `RegisterDeviceTokenEvent` as `onNewToken`.
> 
> Public API surface grows via regenerated `messagingpush.api`. Unit tests cover both registration modes, failures, and manifest lookup fallback. The java_layout sample aligns with minSdk 23.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c95367775e064a2243b6a8c0478543c471536096. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/buildSrc/src/main/kotlin/io.customer/android/Versions.kt
+++ b/buildSrc/src/main/kotlin/io.customer/android/Versions.kt
@@ -4,7 +4,10 @@ object Versions {
     // Android SDK versions
     const val COMPILE_SDK = 36
     const val TARGET_SDK = 36
-    const val MIN_SDK = 21
+
+    // firebase-messaging 25.1+ (required for FID based registration) raised its minimum to API 23,
+    // matching the Firebase wide minSdk bump from June 2025.
+    const val MIN_SDK = 23
 
     // When updating AGP version, make sure to also update workflow: gradle-compatibility-builds
     // and script: update-gradle-compatibility as needed.
@@ -24,7 +27,7 @@ object Versions {
     internal const val DOKKA = "1.9.20"
     internal const val JUNIT_BOM = "5.9.3"
     internal const val ESPRESSO = "3.7.0"
-    internal const val FIREBASE_MESSAGING = "24.1.2"
+    internal const val FIREBASE_MESSAGING = "25.1.2"
     internal const val GRADLE_NEXUS_PUBLISH_PLUGIN = "2.0.0"
     internal const val ANDROID_PUBLISH_PLUGIN = "0.1.0"
     internal const val GOOGLE_PLAY_SERVICES_BASE = "18.7.2"
--- a/messagingpush/api/messagingpush.api
+++ b/messagingpush/api/messagingpush.api
@@ -11,6 +11,8 @@ public class io/customer/messagingpush/CustomerIOFirebaseMessagingService : com/
 	public fun onMessageReceived (Lcom/google/firebase/messaging/RemoteMessage;)V
 	public static final fun onNewToken (Landroid/content/Context;Ljava/lang/String;)V
 	public fun onNewToken (Ljava/lang/String;)V
+	public static final fun onRegistered (Landroid/content/Context;Ljava/lang/String;)V
+	public fun onRegistered (Ljava/lang/String;)V
 }
 
 public final class io/customer/messagingpush/CustomerIOFirebaseMessagingService$Companion {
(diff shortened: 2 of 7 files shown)


mahmoud-elmorabea (changes requested):
Suggesting we hold off on merging this and put it back as draft.

FID isn't a drop-in replacement for a push token, and adopting it is bigger than the SDK, there is a major lifecycle change that we need to adapt for system wide.

jcpsimmons:
Agreed, converting this back to draft. FID adoption needs the system-wide lifecycle plan first, so this holds until that's scoped. Will pick it back up then.

PR #821, opened 2026-08-25
chore(core): log uncaught SDK coroutine failures instead of crashing the host

mrehan27:
## What

An exception escaping a coroutine launched on an SDK scope currently crashes the host app. This adds a last-resort `CoroutineExceptionHandler` to all five `SdkScopeProvider` scopes that logs the failure and drops it, plus attributable catches at the launch sites that need their own handling.

## Changes

- **`SdkScopeProvider` (core)**: every scope now carries a shared `CoroutineExceptionHandler` that logs via `SDKComponent.logger`. Cancellation is unaffected (coroutines never route `CancellationException` to handlers).
- **`ModuleGeofence`**: the foreground-hook, app-launch, and retry launches catch and log with the failing operation's name, so a failure is attributable instead of falling to the scope's generic log. `CancellationException` is rethrown.
- **`AsyncGeofenceEventTracker`**: the fire-and-forget delivery launch runs on an ad-hoc `CoroutineScope` that has **no** handler at all — an escaping throw crashed the host outright. Now caught and logged; the entry stays in the pending store, so the existing foreground flush retries delivery.

## Behavior

No logic change on success paths. Failure paths that previously crashed the host now log; async delivery failures additionally keep the entry for the existing retry path. Worst case of the backstop itself: an exception that would have crashed the app is logged and dropped — call sites still own their real error handling.

## Testing

- New `SdkScopeProviderTest`: all five scopes log-not-crash, and a failed coroutine doesn't affect siblings on the same scope.
- New `AsyncGeofenceEventTrackerTest` case: store throw is logged, entry kept, no crash.
- Full suite: 3406 tests / 0 failures (debug + release, all modules), `apiCheck` and ktlint green.

### DIFF
--- a/core/src/main/kotlin/io/customer/sdk/core/util/ScopeProvider.kt
+++ b/core/src/main/kotlin/io/customer/sdk/core/util/ScopeProvider.kt
@@ -18,7 +18,9 @@ class SdkScopeProvider(private val dispatchers: DispatchersProvider) : ScopeProv
     // Last-resort net: an exception escaping a coroutine on an SDK scope would crash the
     // host app. This only logs and drops — call sites still own their error handling.
     private val exceptionHandler = CoroutineExceptionHandler { _, throwable ->
-        SDKComponent.logger.error("Uncaught exception in SDK coroutine", throwable = throwable)
+        // Exception spelled into the message: custom log dispatchers (wrapper SDKs) receive only
+        // the message string, not the throwable.
+        SDKComponent.logger.error("Uncaught exception in SDK coroutine: $throwable", throwable = throwable)
     }
 
     override val eventBusScope: CoroutineScope
--- a/core/src/main/kotlin/io/customer/sdk/core/util/ScopeProvider.kt
+++ b/core/src/main/kotlin/io/customer/sdk/core/util/ScopeProvider.kt
@@ -1,5 +1,7 @@
 package io.customer.sdk.core.util
 
+import io.customer.sdk.core.di.SDKComponent
+import kotlinx.coroutines.CoroutineExceptionHandler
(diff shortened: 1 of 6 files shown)


core/src/main/kotlin/io/customer/sdk/core/util/ScopeProvider.kt
```diff
@@ -12,14 +14,21 @@ interface ScopeProvider {
 }
 
 class SdkScopeProvider(private val dispatchers: DispatchersProvider) : ScopeProvider {
+
+    // Last-resort net: an exception escaping a coroutine on an SDK scope would crash the
+    // host app. This only logs and drops — call sites still own their error handling.
+    private val exceptionHandler = CoroutineExceptionHandler { _, throwable ->
+        SDKComponent.logger.error("Uncaught exception in SDK coroutine", throwable = throwable)
```
Shahroz16: When a custom log dispatcher is installed, LoggerImpl only forwards the message and drops the throwable. Could we include the exception type or message in the primary log message so wrapper consumers retain enough context to diagnose the failure?

Shahroz16 (approved):
Looks good overall.

core/src/main/kotlin/io/customer/sdk/core/util/ScopeProvider.kt
mrehan27: Good catch — confirmed: `LoggerImpl` only hands the dispatcher `(level, message)`, so the throwable never reached wrapper consumers. Fixed in 34604e73: the exception is now spelled into the message (`: $throwable` → type + message), and the throwable is still passed along so Logcat keeps the full stack trace.

PR #822, opened 2026-08-28
chore(geofence): make the emitted-mark write atomic and bound GMS awaits

mrehan27:
https://linear.app/customerio/issue/MBL-2319/improve-geofence-transition-reliability

Two ways an interrupted operation could strand geofence state. Neither changes
the events the SDK emits; both remove a way the existing behaviour can be lost.

### The emitted-ENTER mark could be persisted under the wrong owner

`markEnterEmitted` wrote the mark's owner and its id set as two separate prefs
commits. A process death between them persists the new owner alongside the
previous user's id set. Owner scoping then reads those marks as the new user's,
silencing their first arrival at every fence in the set — and the next mark
write carries the stale ids forward, so the contamination outlives the crash
that caused it.

Both keys now go in a single editor block, which is one atomic file write, so
the owner can never be persisted apart from the set it owns.
Replace-on-owner-change and idempotency semantics are unchanged.

`claimExit` had the same window over the same two keys — it consumed the
containment record and cleared the mark in two commits, so a torn write strands
the mark and the next genuine arrival is dropped by the redundant-ENTER guard.
Its single caller is inlined so both keys land in one commit. (Review follow-up.)

### A GMS call that never calls back suspended the sync pipeline for the process lifetime

`addGeofences`, `removeGeofences` and `clearAll` awaited their GMS Task with an
unbounded `suspendCancellableCoroutine`. GMS normally calls back in
milliseconds, but a Task never resolves at all when Play Services is mid-update
or has crashed. Every caller holds the refresh slot — and `reset()` waits on
`stateMutex` behind it — so one unresolved Task wedged every later sync for the
life of the process.

Each await now carries a 5s deadline and fails the call instead, landing in the
retry paths a real GMS failure already uses. 5s is generous against the
millisecond norm and fits inside the broadcast receiver's own 8s `goAsync`
budget, so on the movement-refresh path the deadline can still fire before the
receiver gives up waiting.

`withTimeoutOrNull` is safe in this position, unlike around the refresh-slot
CAS: nothing holds a resource on the block's return value, so a deadline racing
a late GMS success only converts it into a spurious failure, and the cancelled
continuation makes the late callback a no-op through the existing `isActive`
guards.

### Test plan

- New `replaceGeofences_givenGmsTaskNeverCallsBack_expectTimeoutFailureInsteadOfIndefiniteSuspend`,
  using a Task that registers listeners and never invokes them; asserts the call
  fails with `TimeoutException` rather than suspending. `runTest` virtual time
  makes the deadline instant.
- `removeGeofencesByIds` and `clearAll` share the deadline helper and now have
  their own timeout tests.
- No new test for the atomic writes: the semantics it preserves are already
  covered by `markEnterEmitted_givenNewOwner_expectPreviousUsersMarksDiscarded`
  and `markEnterEmitted_givenCalledTwice_expectStillReportedAndIdempotent`, and
  the atomicity itself is by construction — `GeofenceRegionStoreImpl` builds its
  prefs from `context`, so reaching the crash window from a test would mean
  adding an injection seam used by nothing else. The semantics `claimExit`
  preserves are covered by `claimExit_givenEnterReported_expectMarkClearedWithContainment`
  and `claimExit_givenNeverEntered_expectMarkRetained`.
- 409 tests / 0 failures, ktlint and apiCheck green.

### Stack

```
feature/geofence-polygons
└── mbl-2319-geofence-atomic-mark-and-bounded-awaits   ← this PR
```

### DIFF
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceManager.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceManager.kt
@@ -225,7 +225,9 @@ internal class GeofenceManager(
     }
 
     private companion object {
-        // Generous against the millisecond norm, and inside the receiver's 8s goAsync budget.
+        // Per call, generous against the millisecond norm. A pass chains up to four (remove and
+        // re-add movement, add business, roll back), so a wedged GMS can still outlast the
+        // receiver's 8s goAsync budget — the join there is best-effort and the refresh continues.
         val GMS_CALL_TIMEOUT = 5.seconds
     }
 }
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceManager.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceManager.kt
@@ -9,8 +9,12 @@ import androidx.annotation.RequiresPermission
 import com.google.android.gms.location.Geofence
 import com.google.android.gms.location.GeofencingClient
 import com.google.android.gms.location.GeofencingRequest
+import java.util.concurrent.TimeoutException
 import kotlin.coroutines.resume
+import kotlin.time.Duration.Companion.seconds
+import kotlinx.coroutines.CancellableContinuation
 import kotlinx.coroutines.suspendCancellableCoroutine
+import kotlinx.coroutines.withTimeoutOrNull
 
 /** Wraps GeofencingClient to register/remove geofences with the OS. */
 internal class GeofenceManager(
@@ -137,7 +141,7 @@ internal class GeofenceManager(
             return Result.failure(e)
         }
 
-        return suspendCancellableCoroutine { cont ->
+        return awaitGmsCall("addGeofences") { cont ->
             try {
                 client.addGeofences(request, pendingIntent)
                     .addOnSuccessListener {
@@ -159,7 +163,7 @@ internal class GeofenceManager(
     suspend fun removeGeofencesByIds(ids: List<String>): Result<Unit> {
         if (ids.isEmpty()) return Result.success(Unit)
 
-        return suspendCancellableCoroutine { cont ->
+        return awaitGmsCall("removeGeofences") { cont ->
             try {
                 client.removeGeofences(ids)
                     .addOnSuccessListener {
@@ -180,7 +184,7 @@ internal class GeofenceManager(
     }
 
     suspend fun clearAll(): Result<Unit> {
-        return suspendCancellableCoroutine { cont ->
+        return awaitGmsCall("removeGeofences (all)") { cont ->
             client.removeGeofences(pendingIntent)
                 .addOnSuccessListener {
                     if (!cont.isActive) return@addOnSuccessListener
@@ -196,6 +200,21 @@ internal class GeofenceManager(
         }
     }
 
+    /**
+     * A Task never calls back when Play Services is mid-update, and the caller holds the refresh
+     * slot — so an unbounded await wedges every later sync. Safe here unlike around the slot CAS:
+     * nothing holds a resource on the result, so a late success just becomes a spurious failure.
+     */
+    private suspend fun awaitGmsCall(
+        description: String,
+        block: (CancellableContinuation<Result<Unit>>) -> Unit
+    ): Result<Unit> =
+        withTimeoutOrNull(GMS_CALL_TIMEOUT) { suspendCancellableCoroutine(block) }
+            ?: run {
+                logger.logGmsCallTimedOut(description)
+                Result.failure(TimeoutException("GMS $description gave no callback within $GMS_CALL_TIMEOUT"))
+            }
+
     private fun GeofenceRegion.toGmsGeofence(): Geofence {
         return Geofence.Builder()
             .setRequestId(id)
@@ -203,5 +222,10 @@ internal class GeofenceManager(
             .setTransitionTypes(toGmsTransitionTypes())
             .setExpirationDuration(GeofenceConstants.GEOFENCE_EXPIRATION_NEVER)
             .build()
+    }
+
+    private companion object {
+        // Generous against the millisecond norm, and inside the receiver's 8s goAsync budget.
+        val GMS_CALL_TIMEOUT = 5.seconds
     }
 }
--- a/geofence/src/main/kotlin/io/customer/geofence/store/GeofenceRegionStore.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/store/GeofenceRegionStore.kt
@@ -247,11 +247,13 @@ internal class GeofenceRegionStoreImpl(
         // One identity owns the set at a time, so a different owner makes it stale — replace, not add.
         val sameOwner = readEmittedEnterOwner() == userId
         val current = if (sameOwner) readEmittedEnterIds() else emptySet()
-        if (!sameOwner) {
-            prefs.edit { putString(KEY_EMITTED_ENTER_OWNER, userId) }
-        }
-        if (geofenceId !in current) {
-            writeJson(KEY_EMITTED_ENTER_IDS, ID_SET_SERIALIZER, current + geofenceId)
+        if (sameOwner && geofenceId in current) return@synchronized
+        // One commit: a death between separate writes leaves the new owner holding the old set.
+        prefs.edit {
+            if (!sameOwner) {
+                putString(KEY_EMITTED_ENTER_OWNER, userId)
+            }
+            putString(KEY_EMITTED_ENTER_IDS, jsonSerializer.encode(ID_SET_SERIALIZER, current + geofenceId))
         }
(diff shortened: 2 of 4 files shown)


geofence/src/main/kotlin/io/customer/geofence/store/GeofenceRegionStore.kt
```diff
@@ -247,11 +247,13 @@ internal class GeofenceRegionStoreImpl(
...
-        if (geofenceId !in current) {
-            writeJson(KEY_EMITTED_ENTER_IDS, ID_SET_SERIALIZER, current + geofenceId)
+        if (sameOwner && geofenceId in current) return@synchronized
+        // One commit: a death between separate writes leaves the new owner holding the old set.
+        prefs.edit {
+            if (!sameOwner) {
+                putString(KEY_EMITTED_ENTER_OWNER, userId)
+            }
+            putString(KEY_EMITTED_ENTER_IDS, jsonSerializer.encode(ID_SET_SERIALIZER, current + geofenceId))
         }
```
mahmoud-elmorabea, line 262: 🔴 **`claimExit` still has this exact window, and it's the more damaging half of the pair.**

Lines 208-213, over the same two keys this block just made atomic:

```kotlin
writeJson(KEY_ENTERED_IDS, ID_SET_SERIALIZER, current - geofenceId)  // editor commit #1
clearEnterEmitted(geofenceId)                                        // editor commit #2
```

geofence/src/main/kotlin/io/customer/geofence/GeofenceManager.kt
```diff
@@ -204,4 +223,9 @@ internal class GeofenceManager(
             .setExpirationDuration(GeofenceConstants.GEOFENCE_EXPIRATION_NEVER)
             .build()
     }
+
+    private companion object {
+        // Generous against the millisecond norm, and inside the receiver's 8s goAsync budget.
+        val GMS_CALL_TIMEOUT = 5.seconds
```
mahmoud-elmorabea, line 231: 🟠 **The deadline is per-call; this comment reads as per-pass.**

The movement path issues at least two sequential calls — `removeGeofencesByIds(MOVEMENT_TRIGGER_ID)` at line 102, then `registerBatch(movement)` — so a wedged Play Services burns ≥10s against the receiver's `DISPATCH_WAIT_BUDGET_MS = 8_000L`. Worst case through `replaceGeofencesInternal` (remove movement / add movement / add business / rollback) is 20s.

Not a regression — the receiver's `job.join()` is best-effort and the wait was unbounded before — but this comment is the whole justification for picking 5s, so it's worth making the per-call scope explicit rather than implying one deadline fits inside the receiver's budget.

geofence/src/main/kotlin/io/customer/geofence/GeofenceManager.kt
```diff
@@ -196,6 +200,21 @@ internal class GeofenceManager(
...
+     */
+    private suspend fun awaitGmsCall(
+        description: String,
+        block: (CancellableContinuation<Result<Unit>>) -> Unit
+    ): Result<Unit> =
+        withTimeoutOrNull(GMS_CALL_TIMEOUT) { suspendCancellableCoroutine(block) }
+            ?: run {
+                logger.logGmsCallTimedOut(description)
+                Result.failure(TimeoutException("GMS $description gave no callback within $GMS_CALL_TIMEOUT"))
+            }
```
mahmoud-elmorabea, line 216: 🟠 **New way to reach the ignored result at line 102.**

`replaceGeofencesInternal` discards the outcome of `removeGeofencesByIds(listOf(MOVEMENT_TRIGGER_ID))` and then re-adds the same ID. Against a slow-but-alive GMS the removal can now return a timeout failure while the request is still being processed, and the re-add lands on an ID that was never torn down — which is the "stuck OUTSIDE, blocks future EXITs" state the removal exists to prevent, and it stops discovery until the next app open or reboot.

GMS serializes requests per client in submission order, so the ordering itself should hold; the exposure is that a *spurious* deadline is a new way in. Low likelihood at 5s against a millisecond norm — flagging because the result is dropped on the floor, so nothing downstream can notice.

geofence/src/test/java/io/customer/geofence/GeofenceManagerTest.kt
```diff
@@ -360,6 +361,18 @@ class GeofenceManagerTest : RobolectricTest() {
...
+    fun replaceGeofences_givenGmsTaskNeverCallsBack_expectTimeoutFailureInsteadOfIndefiniteSuspend() = runTest {
+        // A Task that never resolves previously suspended the await for the life of the process.
+        grantAllPermissions()
+        every { client.addGeofences(any<GeofencingRequest>(), any()) } returns silentTask()
+
+        val result = manager.replaceGeofences(listOf(buildRegion(id = "business-1")))
+
+        result.isFailure.shouldBeTrue()
+        (result.exceptionOrNull() is TimeoutException).shouldBeTrue()
+    }
```
mahmoud-elmorabea: 🟠 Two small things.

**Coverage:** `awaitGmsCall` has three call sites; only `addGeofences` is exercised. A regression that dropped the deadline from `removeGeofencesByIds` or `clearAll` would go unnoticed, and `silentTask()` already exists — two more tests.

**Placement:** this landed inside the `// -- Helpers: permission --` section, between `grantAllPermissions()` and the `// -- Helpers: GMS Task stubs --` marker. Belongs with the other `@Test`s, above the `// -- Helpers: permission --` marker at line 349.

geofence/src/main/kotlin/io/customer/geofence/store/GeofenceRegionStore.kt
mrehan27: Good catch, and it was the worse one. Fixed in e29b692d.

`clearEnterEmitted` had exactly one caller, so it is inlined and both keys now land in a single `prefs.edit`. The thing that makes this clearly a defect rather than a judgement call is the comment that was already sitting on it — `// Same step:` — asserting an atomicity the code did not have. A torn write strands the mark and the next genuine arrival is then dropped by the redundant-ENTER guard, so the customer arrives and no ENTER fires.

No new test, same reasoning as the `markEnterEmitted` half: the store builds its prefs from `context`, so reaching the crash window would mean adding an injection seam used by nothing else. The semantics are covered by `claimExit_givenEnterReported_expectMarkClearedWithContainment` and `claimExit_givenNeverEntered_expectMarkRetained`.

geofence/src/main/kotlin/io/customer/geofence/GeofenceManager.kt
mrehan27: Agreed — the comment overstated it. Fixed in e29b692d.

It now reads per-call and states the chaining explicitly: a pass issues up to four (remove movement, re-add movement, add business, roll back), so a wedged Play Services can still outlast the receiver's 8s budget. As you say, not a regression — the join is best-effort and the wait was unbounded before.

Worth noting the constant does not change: 5s still bounds the worst case at 20s where 10s would have made it 40s. It was the justification that was wrong, not the number.

geofence/src/main/kotlin/io/customer/geofence/GeofenceManager.kt
mrehan27: Thought about this one and I am not changing it — mostly on the strength of your own premise.

If GMS serializes per client in submission order, the removal is still processed before the re-add. Our deadline changes when *we* stop waiting, not what GMS does, so the "re-added onto an ID that was never torn down" state does not actually materialise. And the result at line 102 is already discarded for genuine GMS failures, which is a path that has shipped for months; the deadline adds a rarer route to a path that already exists.

That serialization guarantee is the load-bearing assumption, so if you have reason to doubt it the conclusion flips and this becomes real. Flagging that rather than quietly relying on it.

The separate point — that the result is dropped on the floor so nothing downstream can notice — is fair and pre-existing. Happy to file it if you think it is worth doing properly, but it is a behaviour change beyond this PR.

geofence/src/test/java/io/customer/geofence/GeofenceManagerTest.kt
mrehan27: Both fixed in e29b692d.

Coverage: added `removeGeofencesByIds_givenGmsTaskNeverCallsBack_expectTimeoutFailure` and `clearAll_givenGmsTaskNeverCallsBack_expectTimeoutFailure`, reusing `silentTask()`. Codecov had independently flagged the same gap as 2 partials in that file.

Placement: you are right, it landed inside the permission-helpers section. All three tests are now above the `// -- Helpers: permission --` marker.

mahmoud-elmorabea (approved):
Looks good 💯 

Thanks for addressing the feedback 🥇

PR #824, opened 2026-08-28
chore(geofence): trust geometry only from live fixes and stamp epochs with the fix

mrehan27:
Part of: [MBL-2319](https://linear.app/customerio/issue/MBL-2319/improve-geofence-transition-reliability)

## Stack

- 👉 [#824 trust geometry only from live fixes](https://github.com/customerio/customerio-android/pull/824)
- [#832 qualify a fix before letting it judge containment](https://github.com/customerio/customerio-android/pull/832)
- [#833 polygon wire contract and geometry foundation](https://github.com/customerio/customerio-android/pull/833)

## Problem

A sync pass does not always run from a fix the OS just gave us. `refresh()` —
identify, app-launch, foreground retry — and boot restore run from the
**persisted anchor**: the last registration centre, or the OS location cache.
That is where the device *was*, and it can be arbitrarily stale. The SDK treated
it as trustworthy geometry, which breaks in both directions.

**It fabricates arrivals.** Containment was seeded by testing the pass's
coordinates against each fence, and an initial ENTER was synthesized for anything
the device landed inside. With an anchor, "inside" only means the device was once
there. The trigger is a customer rolling out new fences: every user whose stored
anchor happens to sit inside one gets an arrival on their next launch.

**It also swallows real ones.** Synthesis was gated on
`emitInitialEnter = !osStateWiped()`, suppressed after a reboot or app update
because the anchor could not be trusted then. But the seeding ran anyway from that
same untrusted anchor, so later passes saw the fence as already occupied and
stayed quiet. The genuine arrival is never reported until the user leaves and
comes back.

Gating on OS-state-wiped was the wrong axis. The question is not whether GMS was
reset, it is whether the pass is holding a real fix.

## Change

`FixSource { LIVE, ANCHOR, MOVEMENT }` is threaded through the pass.

| | source | containment seed | moved-circle reset | initial-ENTER synthesis |
|---|---|---|---|---|
| **LIVE** | requested silent fix | yes | yes | yes, unless OS state was wiped |
| **MOVEMENT** | movement-trigger EXIT | yes | yes | yes |
| **ANCHOR** | `refresh()`, boot restore | no | yes | no |

A movement fix is the OS's own, produced because the device moved, so no wipe can have staled
it — and that is the path the backstop exists for, since the `INITIAL_TRIGGER_ENTER` it would
otherwise defer to is the callback that gets missed.

An anchor pass ranks and prunes, and may retire a record it can no longer vouch for —
but it never seeds one. Synthesis keeps the `!osStateWiped()` condition it had before
this branch: a reboot or app update re-registers every fence with
`INITIAL_TRIGGER_ENTER`, so GMS re-reports those arrivals itself, and the backstop is
redundant exactly when the stored record is least likely to be true.

Two supporting changes.

**The containment epoch moves to before the slot wait.** It belongs to the fix, so
a crossing reported while a pass queues behind a holder outranks that pass's
geometry instead of being folded into its baseline.

**`reconcileEnteredIds(inside:)` becomes nullable**, to keep two cases apart that
`emptySet()` used to conflate:

- `null` — this pass has no live fix and cannot judge containment.
- `emptySet()` — a live fix looked and found the device inside nothing.

That distinction matters because key-absence of `entered_ids` is the upgrade grace
`hasContainmentRecord()` reads, and the unmatched-EXIT guard defers while the grace
holds. So a prune-only pass must not create the key — arming the guard off an
anchor would drop the genuine EXIT of anyone upgrading from 4.20.0, which tracked
no containment at all. A live fix must create it even when nothing is inside, since
that is a real reading; otherwise a user who has never been inside a fence never
arms the guard and the phantom EXITs it exists to filter come back.

## Behaviour notes

**An anchor pass no longer seeds containment; the next live fix does.** An anchor pass that
registers a new fence around a genuinely stationary device does not seed containment or
synthesize its ENTER — it cannot tell "inside" from "was once inside". GMS registers business
fences with `INITIAL_TRIGGER_ENTER` and reports that arrival itself, so synthesis is only the
backstop for when GMS drops it.

The part that is easy to miss is that `entered_ids` has **two** readers. Besides
`emitInitialEnters`, the unmatched-EXIT guard reads it as
`!claimExit(id) && hasContainmentRecord()`. So deferring the seed defers both: until a live fix
judges, a dropped `INITIAL_TRIGGER_ENTER` has no backstop, and the later genuine EXIT reads as a
phantom.

**The EXIT guard is armed by the next live fix — including one that changes no registration.**
A successful anchor pass stamps the sync timestamp, the movement location, the registration uptime
and the package update time, so the live fix that follows on a cold start reaches `refreshAction`
and takes `SKIP`. `SKIP` now runs a containment-only pass for any fix that trusts geometry: no
fetch, no registration, just the judgement the anchor could not make. Raised by @Shahroz16; an
earlier revision of this section argued the loss was permanent, which was wrong.

**That pass seeds; it does not synthesize.** It runs on every foreground fix in AUTOMATIC, so
synthesizing there would move the backstop from "at registration" to "at every resume", decided by
an exact point check on a fix this path measures at 100 m indoors. Near a boundary that fabricates
an ENTER *and* arms the emitted mark, and since reconcile carries a record for a still-registered
fence, the record is sticky: the user's real next arrival is then dropped as
`isRedundantEnter` and only its EXIT gets through. A wrong seed has no equivalent cost — it can
only let an EXIT through, which is the direction this path already fails open in. So the
synthesis backstop stays tied to a pass that registers, and a dropped `INITIAL_TRIGGER_ENTER` for
an anchor-registered fence stays lost; GMS reports that arrival itself, and synthesis only ever
covered the case where it does not.

**The grace is now wider than upgrades.** `clearUserScopedState` removes
`entered_ids`, so a sign-out re-enters the grace too, as does a `readJson`
parse-failure wipe. Until the next live pass, an unmatched EXIT is delivered rather
than dropped — fail-open, and correct when we genuinely have no record. The root
cause is that `hasContainmentRecord()` answers "does this install track
containment?" by testing a *user-scoped* key; a dedicated marker that survives
sign-out would decouple them. Left out of this PR deliberately — candidate
follow-up on the same ticket.

## Review follow-ups

Findings from @Shahroz16's review, all fixed here.

**Fixed here — a stale arrival mark survives a geometry edit.** Gating the
moved-circle reset on a live fix means an anchor pass never clears the mark, so a
fence re-registered with new geometry keeps saying "already reported". GMS fires
`INITIAL_TRIGGER_ENTER` for the new circle, the emitter drops it as `isRedundantEnter`,
and the first arrival at the edited geometry is swallowed — its later EXIT then arrives
unmatched. That is a regression this PR introduces, and it breaks GMS's own report
rather than the backstop, so it is fixed here.

The gate was too broad. An anchor cannot be trusted to *create* evidence: seeding
containment or synthesizing an ENTER invents an arrival with no OS event behind it. A
reset only removes a record, and the record it removes describes a circle GMS no longer
holds. **An anchor may retire evidence it can no longer vouch for; it just may not
seed.**

So `movedAwayIds` is computed from the pass's own geometry on every pass, while the
containment seed and the synthesis stay live-only. That restores exactly what 4.20.1 and
4.20.2 already do for this decision, and `reconcileEnteredIds` still declines any reset a
fence has outranked with a newer reported arrival.

One case survives, unchanged from 4.20.2: an anchor that wrongly reads "inside the new
circle" keeps the mark and still swallows that arrival. It needs the anchor to be wrong,
not merely stale, and no reset can do better without a fix to judge against.

**Fixed here — synthesis after a reboot or app update (58f70e2e, corrected by dbbf329c).**
Gating synthesis on `FixSource` dropped the `emitInitialEnter = !osStateWiped()` condition it
replaced, so a live fix arriving post-wipe synthesized an ENTER where no pass previously did.
That is the one path this branch opened rather than narrowed, so the condition is restored.

Worth reading as two commits: the first put the condition back at the call site, where it was
a **no-op** — a successful registration stamps uptime and package update time, and those two
writes are exactly what make the wipe detectors return false, so the guard always saw "not
wiped". Caught by Bugbot. The second snapshots the value before the registration block, next
to the synthesis baseline, which is snapshotted for the same reason. The test had passed
because a fixed mock stub returned "wiped" regardless of the write; it is stateful now and
fails against the first commit. Movement
passes reached synthesis through the old parameter default rather than the guard, so they
lose it here too; the window is the gap between a wipe and the first successful
registration, which boot restore performs.

**Fix quality — shipped as #832.** Raised by @Shahroz16. The 11–31 m figures first cited here
were GMS `triggeringLocation` callbacks recorded during a drive, not the
`getCurrentLocation(PRIORITY_LOW_POWER)` this path uses, which reports 100 m on every fix
indoors; and `getCurrentLocation` does return cached fixes, measured to 124 s. #832 carries the
fix's monotonic timestamp through and runs no pass for a fix older than 5 minutes. It does not
gate on accuracy: a margin on the inside test silences the indoor backstop, and a margin on
retirement keeps a stale mark that swallows the next arrival, so containment stays the exact
point check this PR uses.

Net effect of this branch on that surface is still a narrowing: anchor passes used to seed
containment too, and only live ones can now.

## Upgrade path

No migration needed. No key added, renamed or removed, and no serialized schema
change — the only persistence delta is whether one existing
`writeJson(KEY_ENTERED_IDS, …)` runs.

- **From 4.20.1 / 4.20.2** (`entered_ids` present): reconcile still carries
  `enteredIds ∩ registeredIds`, so containment survives and the guard stays armed.
  Synthesis was already suppressed on an app update. The upgrade pass is identical.
- **From 4.20.0** (key absent, and it had no EXIT guard at all): the grace holds and
  the guard fails open — 4.20.0's own behaviour. GMS's `INITIAL_TRIGGER_ENTER`
  re-report creates the key within seconds for any fence the device is standing in.
  Inside no fence, the guard arms at the first live pass that does work: the first
  movement-trigger EXIT, or 24 h staleness.
- **Better in one case**: 4.20.1 wrote `[]` from a possibly-wrong anchor, arming the
  guard on a guess and then dropping that upgrader's genuine EXIT. Now delivered.
- **Rollback safe**: 4.20.1 handles an absent key and recreates it unconditionally.

## Verification

Emulator differential on a Pixel 10 Pro AVD against a real workspace (8 fences
around Bahria Town Lahore). Anchor seeded inside fence 5, location provider removed
so only the anchor could drive a pass:

| | before | after |
|---|---|---|
| `entered_ids` after an anchor-only relaunch | `["5"]` | absent |
| ENTER delivered | none | none |

The old build silently records "inside fence 5" with no event — the swallowed
arrival. Positive control on the new build: a live fix inside fence 5 registers,
records containment and delivers the ENTER.

Field verification of the anchor path rides the polygon drive.

## Test plan

- Anchor pass: nothing seeded, no moved-away resets, no synthesis; registration and
  pruning still happen.
- Anchor pass with a carried record and a re-registered fence: still no synthesis, but the
  record and mark of a circle that moved away are reset.
- Live pass: all three still happen. Movement passes are live too.
- Live pass just after a reboot: registers and seeds, but synthesizes nothing.
- Epoch: a transition landing while the pass is queued for the slot, and one landing
  during the GMS await, both outrank the pass's geometry.
- Fresh-input live pass: `SKIP` judges containment and emits nothing — including after an EXIT
  cleared the record, which is the chain a synthesizing SKIP would open. An anchor `SKIP` judges
  nothing; a sign-out mid-pass writes nothing; a cached-but-unregistered fence is left out.
- Grace: prune-only leaves `entered_ids` absent; a live fix inside nothing creates
  it; a prune-only pass with an existing record still prunes.
- 429 tests / 0 failures; ktlint, apiCheck and `lintDebug` green.
- Each new guard was mutation-tested — inverted in isolation to confirm a test fails.
  Two tests were passing for the wrong reason and were rewritten; re-applying the gate
  above fails the review-follow-up test and nothing else.

### DIFF
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
@@ -214,12 +214,28 @@ internal class GeofenceRepositoryImpl(
     private fun osStateWiped(): Boolean = osStateWipedByReboot() || osStateWipedByAppUpdate()
 
     /**
-     * How far a pass's geometry can be trusted. [LIVE] is a fix the OS delivered for this pass — it
-     * may seed containment, reset moved-circle records and synthesize initial ENTERs. [ANCHOR] is
-     * stored coordinates describing where the device once was, so it may only rank and prune; the
-     * next LIVE pass owns the reseed.
+     * How far a pass's geometry can be trusted.
+     *
+     * [LIVE] and [MOVEMENT] both carry a real fix, so either may seed containment, reset
+     * moved-circle records and synthesize initial ENTERs. [ANCHOR] is stored coordinates describing
+     * where the device once was, so it may only rank and prune; the next real fix owns the reseed.
      */
-    private enum class FixSource { LIVE, ANCHOR }
+    private enum class FixSource {
+        /** A fix requested for this pass. */
+        LIVE,
+
+        /** Stored coordinates from the last registration, or the OS location cache. */
+        ANCHOR,
+
+        /** The fix the OS delivered with a movement trigger, because the device moved. */
+        MOVEMENT;
+
+        /** An anchor proves nothing about where the device is now, so it may not judge containment. */
+        val trustsGeometry: Boolean get() = this != ANCHOR
+
+        /** True when the OS produced this fix by observing the device, not when we asked for it. */
+        val isOsObserved: Boolean get() = this == MOVEMENT
+    }
 
     /**
      * Takes the in-flight slot for a pass that can't be dropped, waiting for the holder instead.
@@ -265,16 +281,16 @@ internal class GeofenceRepositoryImpl(
                 movedBeyondFetchRadius(distanceFromAnchor, config)
             // The triggering fix is the OS's own, so a reboot/app-update wipe can't make it stale.
             return if (needsRemoteFetch) {
-                val remote = performRemoteRefresh(userId, latitude, longitude, containmentEpoch, FixSource.LIVE)
+                val remote = performRemoteRefresh(userId, latitude, longitude, containmentEpoch, FixSource.MOVEMENT)
                 if (remote.isFailure) {
                     // The trigger already fired, and a failed pass never re-centres it — leaving it
                     // where the device just exited, so nothing can fire again. Re-rank to re-arm it.
                     logger.logMovementRearmedAfterFailedRefresh()
-                    performLocalRefresh(userId, latitude, longitude, config, containmentEpoch, FixSource.LIVE)
+                    performLocalRefresh(userId, latitude, longitude, config, containmentEpoch, FixSource.MOVEMENT)
                 }
                 remote
             } else {
-                performLocalRefresh(userId, latitude, longitude, config, containmentEpoch, FixSource.LIVE)
+                performLocalRefresh(userId, latitude, longitude, config, containmentEpoch, FixSource.MOVEMENT)
             }
         } finally {
             releaseRefreshSlot()
@@ -497,7 +513,7 @@ internal class GeofenceRepositoryImpl(
                         .toSet()
                     // Without a record the later genuine EXIT looks unentered and gets dropped, but
                     // an anchor would seed a place the device left days ago — so it judges nothing.
-                    val insideIds = insideNow.takeIf { fixSource == FixSource.LIVE }
+                    val insideIds = insideNow.takeIf { fixSource.trustsGeometry }
                     // A moved circle keeps its ID, so its record and mark can describe a fence that
                     // no longer exists, and the mark then suppresses GMS's own arrival at the new
                     // one. An anchor may retire evidence it can't vouch for; it just may not seed.
@@ -530,10 +546,15 @@ internal class GeofenceRepositoryImpl(
                 }
             }
             // An anchor inside a fence proves nothing about where the device is now, so a record
-            // carried across a geometry edit must not become a fresh arrival. A wipe re-registers
-            // every fence with INITIAL_TRIGGER_ENTER, so GMS re-reports these itself and the
-            // backstop is redundant exactly when its inputs are least trustworthy.
-            if (registrationResult.isSuccess && fixSource == FixSource.LIVE && !osStateWasWiped) {
+            // carried across a geometry edit must not become a fresh arrival. After a wipe a
+            // requested fix may also describe a stretch we never observed, so synthesis defers to
+            // the re-registration's own INITIAL_TRIGGER_ENTER — but only for a fix we asked for. A
+            // movement fix is the OS's own, produced because the device just moved, so no wipe can
+            // have made it stale, and suppressing it here would drop the backstop on the one path
+            // that exists because that callback gets missed.
+            val maySynthesize = fixSource.trustsGeometry &&
+                (fixSource.isOsObserved || !osStateWasWiped)
+            if (registrationResult.isSuccess && maySynthesize) {
                 // Identity can change during the awaited GMS call, and reset doesn't clear pending
                 // delivery rows — never queue a synthetic ENTER for a signed-out/switched user.
                 if (secureUserStore.getUserId() == userId) {
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
@@ -465,6 +465,9 @@ internal class GeofenceRepositoryImpl(
             // override here (unlike the registration diff): a reboot re-registers with
             // INITIAL_TRIGGER_ENTER, so the OS re-reports these itself.
             val unchangedRegistered = unchangedRegisteredIds(nearest)
+            // Snapshotted for the same reason: a successful registration stamps uptime and package
+            // update time, which is what clears the wipe signals this reads.
+            val osStateWasWiped = osStateWiped()
             // Registered before this pass but not cache-equal: the ID survived a geometry edit.
             val previouslyRegistered = store.getRegisteredIds()
             val reRegisteredIds = nearest.map { it.id }
@@ -530,7 +533,7 @@ internal class GeofenceRepositoryImpl(
             // carried across a geometry edit must not become a fresh arrival. A wipe re-registers
             // every fence with INITIAL_TRIGGER_ENTER, so GMS re-reports these itself and the
             // backstop is redundant exactly when its inputs are least trustworthy.
-            if (registrationResult.isSuccess && fixSource == FixSource.LIVE && !osStateWiped()) {
+            if (registrationResult.isSuccess && fixSource == FixSource.LIVE && !osStateWasWiped) {
                 // Identity can change during the awaited GMS call, and reset doesn't clear pending
                 // delivery rows — never queue a synthetic ENTER for a signed-out/switched user.
                 if (secureUserStore.getUserId() == userId) {
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
@@ -558,13 +558,7 @@ internal class GeofenceRepositoryImpl(
                 // Identity can change during the awaited GMS call, and reset doesn't clear pending
                 // delivery rows — never queue a synthetic ENTER for a signed-out/switched user.
                 if (secureUserStore.getUserId() == userId) {
-                    emitInitialEnters(
-                        candidates = nearest,
-                        newIds = nearest.map { it.id }.toSet() - unchangedRegistered,
-                        userId = userId,
-                        latitude = latitude,
-                        longitude = longitude
-                    )
+                    emitInitialEnters(nearest, unchangedRegistered, userId, latitude, longitude)
                 } else {
                     logger.logSyncSkipped("user changed during refresh — initial-enter synthesis skipped")
                 }
@@ -577,10 +571,13 @@ internal class GeofenceRepositoryImpl(
      * Judges containment from a fix whose registration is already current.
      *
      * The anchor pass that made the inputs look fresh could not judge geometry, so without this the
-     * live fix queued behind it is discarded and the entered set stays unseeded: the receiver then
-     * reads a genuine EXIT as unmatched and drops it, and an `INITIAL_TRIGGER_ENTER` GMS dropped for
-     * a fence that pass registered stays lost. Registration and the cache are left alone — this only
-     * decides where the device is.
+     * live fix queued behind it is discarded and the entered set stays unseeded — the receiver then
+     * reads a genuine EXIT as unmatched and drops it. Registration and the cache are left alone.
+     *
+     * Seeding only. Synthesis stays tied to a pass that registers: this one runs on every foreground
+     * fix, and an exact point check on a fix that can be 100 m out would fabricate an arrival near a
+     * boundary, then leave a mark that swallows the real one. A wrong seed is the fail-open
+     * direction — it lets an EXIT through — so it does not carry that cost.
      */
     private suspend fun reconcileContainment(
         userId: String,
@@ -603,40 +600,29 @@ internal class GeofenceRepositoryImpl(
                 .filter { it.distanceTo(latitude, longitude) <= it.radius }
                 .map { it.id }
                 .toSet()
-            val containedBefore = store.getEnteredIds()
             store.reconcileEnteredIds(
                 registeredIds = registeredIds,
                 inside = insideNow,
                 sinceEpoch = containmentEpoch
             )
             logger.logContainmentJudged(insideNow.size)
-            // Only where the record is new: a fence we already knew we were inside has had its
-            // arrival reported, and an enter-only one carries no mark to stop this repeating. No
-            // wipe gate — refreshAction routes a wiped install to LOCAL, so this pass never sees one.
-            emitInitialEnters(
-                candidates = monitored,
-                newIds = insideNow - containedBefore,
-                userId = userId,
-                latitude = latitude,
-                longitude = longitude
-            )
             Result.success(Unit)
         }
     }
 
     /**
-     * Synthesizes an ENTER for each fence in [newIds] the device is already inside — GMS's
+     * Synthesizes an ENTER for each newly-registered fence the device is already inside — GMS's
      * `INITIAL_TRIGGER_ENTER` unreliably drops this for a region added around a stationary device.
-     * [newIds] is what this pass has a reason to check: newly registered where it registered,
-     * newly contained where the registration was already current. Cooldown-deduped, so a real GMS
-     * ENTER and this one collapse to one event.
+     * "New" = not in [unchangedRegisteredIds]: brand-new fences and re-added param changes fire,
+     * an unchanged re-register stays silent. Cooldown-deduped, so a real GMS ENTER and this one
+     * collapse to one event.
      *
      * Requires the stored containment record and this fix's geometry to agree — reconcile carries a
      * record forward for a still-registered fence, so it can outlive the visit.
      */
     private suspend fun emitInitialEnters(
         candidates: List<GeofenceRegion>,
-        newIds: Set<String>,
+        unchangedRegisteredIds: Set<String>,
         userId: String,
         latitude: Double,
         longitude: Double
@@ -644,12 +630,12 @@ internal class GeofenceRepositoryImpl(
         val timestamp = clock.currentTimeSeconds()
         val contained = store.getEnteredIds()
         candidates.forEach { region ->
-            val isNew = region.id in newIds
+            val newlyRegistered = region.id !in unchangedRegisteredIds
             val monitorsEnter = GeofenceTransitionType.ENTER in region.transitionTypes
             // Both: a carried-forward record can outlive the visit, and geometry alone ignores an
             // EXIT reported while GMS was awaited.
             val insideNow = region.distanceTo(latitude, longitude) <= region.radius
-            if (!isNew || !monitorsEnter || region.id !in contained || !insideNow) return@forEach
+            if (!newlyRegistered || !monitorsEnter || region.id !in contained || !insideNow) return@forEach
             logger.logInitialEnterInside(region.id)
             transitionEmitter.emit(
                 geofenceId = region.id,
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
@@ -138,7 +138,7 @@ internal class GeofenceRepositoryImpl(
                 RefreshAction.LOCAL -> performLocalRefresh(userId, latitude, longitude, config, containmentEpoch, fixSource)
                 RefreshAction.SKIP -> {
                     logger.logSyncSkippedFresh()
-                    Result.success(Unit)
+                    reconcileContainment(userId, latitude, longitude, containmentEpoch, fixSource)
                 }
             }
         } finally {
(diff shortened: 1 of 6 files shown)


geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
```diff
@@ -121,19 +129,13 @@ internal class GeofenceRepositoryImpl(
...
-            // A launch/identify refresh runs with the persisted anchor, which a reboot or app update
-            // can leave pointing at a stale position — containment can't be trusted, so no synthesis
-            // this pass (mirrors restoreFromCache). Drops once registration re-stamps.
-            val emitInitialEnter = !osStateWiped()
             return when (action) {
-                RefreshAction.REMOTE -> performRemoteRefresh(userId, latitude, longitude, containmentEpoch, emitInitialEnter)
-                RefreshAction.LOCAL -> performLocalRefresh(userId, latitude, longitude, config, containmentEpoch, emitInitialEnter = emitInitialEnter)
+                RefreshAction.REMOTE -> performRemoteRefresh(userId, latitude, longitude, containmentEpoch, fixSource)
+                RefreshAction.LOCAL -> performLocalRefresh(userId, latitude, longitude, config, containmentEpoch, fixSource)
                 RefreshAction.SKIP -> {
```
Shahroz16, line 139: Could we carry pending live reconciliation across an anchor registration? A successful anchor pass makes every refresh input fresh, so the live fix requested during the normal cold-start flow can take `SKIP`. Even when it does run, the anchor pass has already made new or edited fences look unchanged. That means a dropped GMS initial ENTER stays lost, while an edited same-ID fence can keep the previous emitted-ENTER mark and suppress the first arrival at its new geometry. Persisting the IDs that still need live seeding, reset, or synthesis, and consuming them on the next live fix even when registration is fresh, would close this.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
```diff
@@ -480,16 +489,19 @@ internal class GeofenceRepositoryImpl(
...
-                    val insideIds = nearest.filter { it.distanceTo(latitude, longitude) <= it.radius }
-                        .map { it.id }
-                        .toSet()
-                    // A moved circle keeps its ID, so containment and the reported-ENTER mark can
-                    // describe where the fence used to be. Reset both, but only once the new geometry
-                    // agrees the device is out, or trimming a radius manufactures a second arrival.
-                    val movedAwayIds = reRegisteredIds - insideIds
+                    // Without a record the later genuine EXIT looks unentered and gets dropped, but
+                    // an anchor would seed a place the device left days ago — so it judges nothing.
+                    val insideIds = if (fixSource == FixSource.LIVE) {
```
Shahroz16: Could we carry the location timestamp and horizontal accuracy into this decision before treating a fix as `LIVE`? The request uses low-power location, documented here as roughly 10 km accuracy, and `getCurrentLocation` may return a cached point, but both fields are discarded before this code seeds containment, resets marks, or synthesizes ENTERs against much smaller fences. Treating an unverified fix as anchor-only, or requiring a recent fix with an accuracy and edge margin for each fence, would avoid fabricating or losing transitions.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
mrehan27: Fixed in 397ede94 — and simpler than either of us proposed, with no new persisted state.

The gate was too broad rather than in the wrong place. An anchor can't be trusted to *create* evidence: seeding containment or synthesizing an ENTER invents an arrival with no OS event behind it. But a reset only *removes* a record, and the record it removes describes a circle GMS no longer holds. So `movedAwayIds` is computed from the pass's own geometry on every pass, while the seed and the synthesis stay live-only. That's what 4.20.1/4.20.2 already did for this decision.

**Correcting myself on the other half.** I said an anchor pass costs only the synthesis backstop. That understates it, and I only found out by probing the store rather than reasoning about it. `entered_ids` has two readers, not one:

- `emitInitialEnters` — the backstop, which I did account for.
- the unmatched-EXIT guard in `GeofenceBroadcastReceiver` — `!claimExit(id) && hasContainmentRecord()`.

Not seeding on an anchor pass silences both. So if GMS also drops the `INITIAL_TRIGGER_ENTER` for a fence registered on that pass, `claimExit` later returns `false`, the guard fires, and the **genuine EXIT is dropped too**. Measured on a real store: `claimExit` returns `true` pre-PR and `false` now, for the same sequence.

I still don't think it should be fixed, and I'd rather argue that than quietly widen the change:

- The two situations are indistinguishable by construction. "GMS missed the arrival, this EXIT is real" and "GMS never had us inside, this EXIT is a phantom" leave identical state. That is the same phantom-EXIT shape we chased in 4.20.1.
- The old behaviour wasn't the good outcome. That record was manufactured from stale coordinates — it let phantom EXITs through as readily as genuine ones, which is the bug this branch exists to remove.
- Fixing it needs per-fence "not yet judged" state, i.e. the queue you proposed. That is the one thing the evidence says we shouldn't add.

It's now written into the PR description's Behaviour notes rather than left implicit. Happy to be pushed on it.

One case also survives unchanged from 4.20.2: an anchor that wrongly reads "inside the new circle" keeps the mark and still swallows that arrival. It needs the anchor to be wrong, not merely stale, and no reset can do better without a fix to judge against.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
mrehan27: Two parts to this, and they've landed differently. Updating rather than leaving my earlier answer standing, since I've since changed my mind on one of them.

**The path this branch actually opened — fixed here, in 58f70e2e + dbbf329c.** Synthesis used to be gated on `emitInitialEnter = !osStateWiped()`, so nothing synthesized after a reboot or app update. Replacing that with `FixSource` dropped the condition, and a live fix arriving post-wipe began synthesizing where no pass previously did. Restored: a wipe re-registers every fence with `INITIAL_TRIGGER_ENTER`, so GMS re-reports those arrivals itself, and the backstop is redundant exactly when the stored record is least likely to be true — the device may have moved while the phone was off. Movement passes reached synthesis through the old parameter default rather than the guard, so they lose it here too; the window is the gap between a wipe and the first successful registration, which boot restore performs.

Two commits because I got it wrong the first time, and it's worth knowing which: putting the condition back at the call site was a **no-op**, since a successful registration stamps uptime and package update time and those writes are precisely what make the wipe detectors return false. Bugbot caught it. `dbbf329c` snapshots the value before the registration block instead. My test had passed on a fixed mock stub that returned "wiped" regardless of the write — it's stateful now, and fails against the first commit.

**Validating fix quality — I built it, measured it, and I'm not shipping it.** I said above I'd gate on accuracy per fence, `distanceTo(centre) + horizontalAccuracy <= radius`. That works and it needs no threshold constant, but the numbers don't justify it. Our field data is 11–31 m fixes against fences of 100 m and up, so the gate only changes the answer inside a ~30 m annulus near the edge — and there it trades a rare fabricated ENTER for a rare missing containment record, since not seeding also disarms the EXIT guard (see the other thread). The cost was `horizontalAccuracy` on a public `Event` plus plumbing through the location module and the broadcast receiver: ~205 lines across three modules for a wash.

So there is no follow-up ticket for it. I'd rather say that plainly than leave one open that nobody intends to pick up. The branch is kept locally if you want to see it.

Worth separating from the above: the recency half needs no gate either. `FusedLocationProvider` calls `getCurrentLocation(priority, token)`, not `getLastLocation()`, so it computes a fix rather than returning whatever was cached — our five background fixes measured ages of 0.2–2.7 s.

Net effect of the branch on this surface is a narrowing: anchor passes used to seed containment too, and now only live ones can. If you think the annulus case is worth the API change anyway, say so and I'll reopen it — the work is done, it's the trade I'm unconvinced by.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
```diff
@@ -211,6 +213,14 @@ internal class GeofenceRepositoryImpl(
 
     private fun osStateWiped(): Boolean = osStateWipedByReboot() || osStateWipedByAppUpdate()
 
+    /**
+     * How far a pass's geometry can be trusted. [LIVE] is a fix the OS delivered for this pass — it
```
Shahroz16: Could we keep the timestamp and horizontal accuracy with `LIVE`, rather than treating every callback as trusted geometry? `getCurrentLocation(priority, token)` can return a recent cached fix, and this path still requests low-power priority. The five field samples are useful, but they do not change that platform contract. Without those fields this can still seed containment, clear a mark, or synthesize an ENTER from a fix that cannot resolve a 100 m fence. I think we should qualify the fix against age and accuracy, or treat it as anchor-only when it cannot make a confident boundary decision.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
```diff
@@ -517,7 +529,11 @@ internal class GeofenceRepositoryImpl(
                     logger.logSyncSucceeded(nearest.size, movementTriggerRegistered = monitoringEnabled)
                 }
             }
-            if (registrationResult.isSuccess && emitInitialEnter) {
+            // An anchor inside a fence proves nothing about where the device is now, so a record
+            // carried across a geometry edit must not become a fresh arrival. A wipe re-registers
+            // every fence with INITIAL_TRIGGER_ENTER, so GMS re-reports these itself and the
+            // backstop is redundant exactly when its inputs are least trustworthy.
+            if (registrationResult.isSuccess && fixSource == FixSource.LIVE && !osStateWasWiped) {
```
Shahroz16: Could we keep the pre-PR behavior for movement-trigger passes while the wipe detector is set? `handleMovement` has the OS triggering fix and used to leave synthesis on intentionally, but this shared guard now suppresses it. I added a focused probe for movement plus wiped state plus a newly registered containing fence, and the emitter was called zero times. That leaves us relying only on `INITIAL_TRIGGER_ENTER` in the exact path where the backstop exists because that callback can be missed. Since the emitter already dedupes the OS report and synthetic report, I think movement needs its own policy here, or a third source, instead of inheriting the live-refresh wipe gate.

mrehan27: You're right, and it's a regression this PR introduced rather than something it inherited — fixed in `419a2144`.

`FixSource` gains `MOVEMENT`, exempt from the wipe gate on synthesis only; seeding and moved-circle resets still go through `trustsGeometry`, so an anchor still can't judge containment. The gate's reasoning was about anchor staleness, which the fix source already covers — and the movement branch's own comment had said as much: the triggering fix is the OS's, so a reboot or app update can't have staled it.

Your probe is now a test — movement pass, wiped state, newly registered containing fence, expect one synthesized ENTER. I inverted each of the three conditions and all three fail without the fix.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
mrehan27: Conceded — you're right, and my field numbers didn't answer your point.

Those 11–31 m samples were GMS `triggeringLocation` callbacks recorded during a drive, not `getCurrentLocation(PRIORITY_LOW_POWER)`, which is the source this path actually uses. I measured that path afterwards and it reports 100 m accuracy on every fix stationary indoors — so a fix that cannot resolve a 100 m fence really can seed containment, clear a mark, or synthesize an ENTER. That also withdraws the "no follow-up ticket for it" in my earlier reply on the other thread.

It lands as its own PR rather than here, since this one is scoped to fix trust. Worth flagging that the two paths differ a lot in cost: the movement trigger already hands the receiver the OS `Location`, so age and accuracy are available there with no new surface, while the requested-fix path drops both before the geofence module sees them. I'll size it against that rather than assume the widest option.

On the rule itself I'd take fix age plus an edge margin against reported accuracy, and treat anything that cannot make a confident boundary decision as anchor-only — which is what `FixSource` already expresses, so it needs no second concept. Say if you'd rather set the thresholds.

mrehan27: Opened as #832, stacked on this branch.

It went further than I said it would. Sizing it against the real regime showed my earlier numbers were the wrong ones twice over: the 11-31m samples were drive callbacks, and "getCurrentLocation computes a fix rather than returning a cached one" was just wrong — measured ages reach 124s. I have struck that section from this PR's description rather than leave it standing.

One thing worth your eye before that one lands: at the 100m accuracy this path actually returns indoors, the margin makes a 100m fence unjudgable at any distance, so the initial-ENTER backstop goes quiet there. That is the honest consequence of the data rather than a threshold I can tune away, and #832 states it in full with the narrower alternative if you would rather keep the backstop.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
mrehan27: Superseded — resolving to keep this PR's state legible.

The "not shipping, no follow-up ticket" in my reply above is withdrawn, and the reasoning under it was wrong twice: the 11-31m samples were drive callbacks rather than this path, and `getCurrentLocation` does return cached fixes (measured to 124s). The gate shipped as #832. The other accuracy thread carries the full correction.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
Shahroz16: Following up because the latest exact-head review still reproduces the unresolved half here. The problem is not asking the anchor to seed anything. When the requested live fix waits behind the anchor pass, that pass writes the sync timestamp and movement location; the live call then reaches `refreshAction` and returns `SKIP`, so the real fix never gets to `reconcileEnteredIds` or `emitInitialEnters`. That breaks the "next real fix owns the reseed" contract even though a fix did arrive. Could `LIVE + SKIP` run a containment-only or local pass? A regression test where the holder makes the second pass fresh would pin this down.

mrehan27: You're right, and it closes more than the reseed — `36ba49d5`.

`SKIP` now runs a containment-only pass for any fix that trusts geometry: no fetch, no registration, just `reconcileEnteredIds` from the fix's own geometry, then synthesis for the fences it *newly* places the device inside. "Newly" is what makes it safe on every foreground fix — an enter-only fence never sets `hasEmittedEnter`, so the record is the only thing that stops a repeat.

That retires the "left as-is deliberately" note in Behaviour notes above, which I've rewritten. With the seed restored, a fence an anchor pass registered gets both readers back: the dropped `INITIAL_TRIGGER_ENTER` is synthesized, and the later genuine EXIT matches `claimExit` instead of reading as a phantom. The loss is a deferral to the next live fix — in AUTOMATIC, the next foreground entry — rather than permanent.

Your test shape is there (`refreshFromLiveFix_givenAnchorPassMadeInputsFresh_…`: anchor registers and stamps, live fix takes SKIP, containment seeded and one ENTER emitted), plus four more — sign-out mid-pass, device outside, a cached-but-unregistered fence, and an anchor SKIP still judging nothing. Six mutations, each confirmed applied and each killed by a named test: the SKIP call, the fix-source guard, the newly-contained narrowing, the registered-ids filter, the `isNew` gate, the identity recheck. 425 geofence / 103 core / 61 location, 0 failures; ktlint, apiCheck, lintDebug green.

mrehan27: Correcting my previous reply on this thread — the synthesis half was wrong and is removed in `30480f4e`.

`SKIP` seeds containment and emits nothing. Synthesizing there was the mistake: this pass runs on every foreground fix in AUTOMATIC, so it moved the backstop from "at registration" to "at every resume", decided by an exact point check on a fix this path measures at 100 m indoors. Near a boundary the chain is: a real EXIT clears the record and the mark, the next foreground fix reads a hair inside and fabricates an ENTER *and* re-arms the mark; reconcile carries a record for a still-registered fence, so a later pass reading "outside" never clears it; the user's real next arrival is then dropped as `isRedundantEnter` and only its EXIT gets through.

Seeding has no equivalent cost. A wrong seed can only let an EXIT through — the direction this path already fails open in — and it sets no mark, so it cannot swallow a real arrival. So your line-139 concern is answered by the seed: the guard is armed by the live fix that follows the anchor, not left unarmed. The synthesis backstop stays tied to a pass that registers, and a dropped `INITIAL_TRIGGER_ENTER` for an anchor-registered fence stays lost — GMS reports that arrival itself, and synthesis only ever covered the case where it does not. Behaviour notes rewritten to say exactly that.

A test pins the chain (`refreshFromLiveFix_givenRecordClearedByAnExit_expectReseedButNoSynthesizedEnter`), and one of the five mutations re-adds the synthesis call and dies on it, so this cannot come back quietly. 425 geofence / 103 core / 61 location, 0 failures; ktlint, apiCheck, lintDebug green.

Shahroz16 (approved):
The live-fix SKIP path now seeds containment without synthesizing an ENTER, and the regression coverage matches the intended lifecycle.

PR #825, opened 2026-08-29
ci: add fast and full PR validation lanes

Shahroz16:
## Summary
- remove duplicate feature-branch push runs while retaining PR, main, and beta validation
- gate sample distribution, Maestro E2E, binary size, snapshot publishing, and AGP compatibility behind `ci:full`
- preserve release-published sample builds and scheduled/manual E2E
- isolate unrelated label events so they cannot cancel a requested full run

## Validation
- `./gradlew assembleDebug` succeeded (377 tasks)
- repository binary validation and Kotlin lint hooks passed
- changed workflow YAML passes differential Actionlint validation
- no release or deploy workflow changed

### DIFF
--- a/.github/scripts/validate-release-policy.sh
+++ b/.github/scripts/validate-release-policy.sh
@@ -55,9 +55,15 @@ while IFS= read -r -d '' commit_message; do
   fi
 done < "$commit_messages_file"
 
+full_validation=false
+requires_full_label=false
+
+if [[ "$pr_release" == 'true' || "$commit_release" == 'true' || "${CI_FULL:-false}" == 'true' ]]; then
+  full_validation=true
+fi
+
 if [[ "$commit_release" == 'true' && "$pr_release" != 'true' && "${CI_FULL:-false}" != 'true' ]]; then
-  echo "::error::This PR contains a release-bearing commit, but its PR title/body is non-release. Retitle it with fix:, feat:, perf:, revert:, or a breaking-change marker, or apply ci:full, so release validation runs before merge."
-  exit 1
+  requires_full_label=true
 fi
 
 if [[ ("$pr_release" == 'true' || "$commit_release" == 'true') && ("$HEAD_REPOSITORY" != "$TARGET_REPOSITORY" || "$PR_AUTHOR" == 'dependabot[bot]') ]]; then
@@ -65,4 +71,17 @@ if [[ ("$pr_release" == 'true' || "$commit_release" == 'true') && ("$HEAD_REPOSI
   exit 1
 fi
 
-echo "Release validation policy satisfied (pr_release=$pr_release, commit_release=$commit_release)."
+if [[ -n "${GITHUB_OUTPUT:-}" ]]; then
+  {
+    echo "pr_release=$pr_release"
+    echo "commit_release=$commit_release"
+    echo "full_validation=$full_validation"
+    echo "requires_full_label=$requires_full_label"
+  } >> "$GITHUB_OUTPUT"
+fi
+
+if [[ "$requires_full_label" == 'true' ]]; then
+  echo "::notice::A release-bearing commit was detected behind a non-release PR title. The semantic PR helper will apply ci:full automatically."
+fi
+
+echo "Release validation policy satisfied (pr_release=$pr_release, commit_release=$commit_release, full_validation=$full_validation)."
--- a/.github/scripts/validate-release-policy.sh
+++ b/.github/scripts/validate-release-policy.sh
@@ -66,7 +66,9 @@ if [[ "$commit_release" == 'true' && "$pr_release" != 'true' && "${CI_FULL:-fals
   requires_full_label=true
 fi
 
-if [[ ("$pr_release" == 'true' || "$commit_release" == 'true') && ("$HEAD_REPOSITORY" != "$TARGET_REPOSITORY" || "$PR_AUTHOR" == 'dependabot[bot]') ]]; then
+if [[ "${ENFORCE_TRUSTED_RELEASE:-true}" == 'true' &&
+  ("$pr_release" == 'true' || "$commit_release" == 'true') &&
+  ("$HEAD_REPOSITORY" != "$TARGET_REPOSITORY" || "$PR_AUTHOR" == 'dependabot[bot]') ]]; then
   echo "::error::Release-bearing PRs require secret-dependent validation from a trusted repository branch. Recreate or update this change on an internal branch before merge."
   exit 1
 fi
--- a/.github/scripts/validate-release-policy.sh
+++ b/.github/scripts/validate-release-policy.sh
@@ -1,0 +1,68 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+: "${PR_TITLE:?PR_TITLE is required}"
+: "${BASE_SHA:?BASE_SHA is required}"
+: "${HEAD_SHA:?HEAD_SHA is required}"
+: "${HEAD_REPOSITORY:?HEAD_REPOSITORY is required}"
+: "${TARGET_REPOSITORY:?TARGET_REPOSITORY is required}"
+: "${PR_AUTHOR:?PR_AUTHOR is required}"
+
+is_release_message() {
+  local message="$1"
+  local header
+  header=$(sed -n '/[^[:space:]]/{p;q;}' <<< "$message")
+
+  case "$header" in
+    fix:*|fix\(*|feat:*|feat\(*|perf:*|perf\(*|revert:*|revert\(*|Revert\ \"*)
+      return 0
+      ;;
+  esac
+
+  if grep -Eq '^[[:alpha:]]+(\([^)]*\))?!:' <<< "$header"; then
+    return 0
+  fi
+
+  if grep -Eq '(^|[[:space:]])(BREAKING CHANGES?:|BREAKING-CHANGE:)' <<< "$message"; then
+    return 0
+  fi
+
+  if grep -Eq '(^|[[:space:]])This reverts commit[[:space:]]' <<< "$message"; then
+    return 0
+  fi
+
+  return 1
+}
+
+pr_message="$PR_TITLE"$'\n\n'"${PR_BODY:-}"
+pr_release=false
+commit_release=false
+
+if is_release_message "$pr_message"; then
+  pr_release=true
+fi
+
+git cat-file -e "$BASE_SHA^{commit}"
+git cat-file -e "$HEAD_SHA^{commit}"
+commit_messages_file=$(mktemp)
+trap 'rm -f "$commit_messages_file"' EXIT
+git log --format='%B%x00' "$BASE_SHA..$HEAD_SHA" > "$commit_messages_file"
+
+while IFS= read -r -d '' commit_message; do
+  if is_release_message "$commit_message"; then
+    commit_release=true
+    break
+  fi
+done < "$commit_messages_file"
+
+if [[ "$commit_release" == 'true' && "$pr_release" != 'true' && "${CI_FULL:-false}" != 'true' ]]; then
+  echo "::error::This PR contains a release-bearing commit, but its PR title/body is non-release. Retitle it with fix:, feat:, perf:, revert:, or a breaking-change marker, or apply ci:full, so release validation runs before merge."
+  exit 1
+fi
+
+if [[ ("$pr_release" == 'true' || "$commit_release" == 'true') && ("$HEAD_REPOSITORY" != "$TARGET_REPOSITORY" || "$PR_AUTHOR" == 'dependabot[bot]') ]]; then
+  echo "::error::Release-bearing PRs require secret-dependent validation from a trusted repository branch. Recreate or update this change on an internal branch before merge."
+  exit 1
+fi
+
+echo "Release validation policy satisfied (pr_release=$pr_release, commit_release=$commit_release)."
--- a/.github/scripts/validate-release-policy.sh
+++ b/.github/scripts/validate-release-policy.sh
@@ -4,10 +4,6 @@ set -euo pipefail
 : "${PR_TITLE:?PR_TITLE is required}"
 : "${BASE_SHA:?BASE_SHA is required}"
 : "${HEAD_SHA:?HEAD_SHA is required}"
-: "${HEAD_REPOSITORY:?HEAD_REPOSITORY is required}"
-: "${TARGET_REPOSITORY:?TARGET_REPOSITORY is required}"
-: "${PR_AUTHOR:?PR_AUTHOR is required}"
-
 is_release_message() {
   local message="$1"
   local header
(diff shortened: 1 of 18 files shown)


Shahroz16:
## Follow-up: Android compatibility, teardown, and CI decision

I investigated the failures rather than suppressing or retrying them.

### Compatibility decision

This SDK currently uses compileSdk/targetSdk 36, AGP 8.9.3, Kotlin 2.1.21, AndroidX Core 1.17.0, and Gradle 8.11.1. Android's compatibility table puts API 36's minimum at AGP 8.9.1; AGP 8.9 requires Gradle 8.11.1; AndroidX Core 1.17 requires Kotlin Gradle Plugin 2.0 or newer.

The old compatibility script returned success immediately for every requested `8.*` AGP version. The four jobs named AGP 8.0 and 8.6.1 therefore rebuilt the repository default, AGP 8.9.3, without testing the claimed versions. That was false coverage.

Decision:

- test the actual supported floor, AGP 8.9.1 with Gradle 8.11.1
- keep both Groovy and Kotlin DSL consumer builds
- compile both consumers in one Gradle invocation so shared SDK modules are built once
- rely on the normal full sample build for the current AGP 8.9.3 configuration
- do not downgrade AndroidX or Kotlin to preserve an unsupported compatibility label

References: [Android API and AGP minimums](https://developer.android.com/build/releases/about-agp), [AGP 8.9 compatibility](https://developer.android.com/build/releases/agp-8-9-0-release-notes), [AndroidX Core 1.17](https://developer.android.com/jetpack/androidx/releases/core#1.17.0).

### Why the instrumentation test was breaking

The baseline Kotlin Compose instrumentation job failed twice, then passed on its third attempt. `MemoryLeakageInstrumentedTest` cleared the SDK dependency graph while queued in-app work was still running. An OkHttp dispatcher then tried to resolve `AndroidSDKComponent` from the already-reset graph and crashed with `AndroidSDKComponent is not initialized`.

The fix gives SDK coroutine scopes one parent lifecycle, while each returned scope remains independently cancellable. Teardown now cancels and joins SDK-owned work before resetting DI, and Data Pipelines is shut down first. Regression tests cover both global shutdown and one-scope isolation. This follows Kotlin's structured-concurrency ownership model. No sleeps or retries were added. Reference: [Kotlin coroutine scope and cancellation](https://kotlinlang.org/docs/coroutines-basics.html#scope-builder-and-concurrency).

### Cleanup and measured result

- removed unused Ruby/Bundler and `sd` setup
- removed duplicate manual Gradle caching because `setup-gradle` owns Gradle User Home caching
- reused the repository Android setup action
- removed compile-only workspace credentials
- reduced four fake compatibility jobs to one real compatibility job
- bounded unit-test jobs at 20 minutes after one hosted worker hung; its targeted diagnostic rerun passed in 4:31 and the final Messaging In-App job passed in 3:51

Measured snapshot plus compatibility lane from GitHub job timestamps:

- before: 23.4 runner-minutes, 12:09 wall time ([run](https://github.com/customerio/customerio-android/actions/runs/33278229418))
- after: 14.6 runner-minutes, 14:38 wall time ([run](https://github.com/customerio/customerio-android/actions/runs/33284004096))
- runner load reduction: 37.7%

The optional `ci:full` lane is 2:29 longer wall-clock because it now performs real AGP 8.9.1 release builds instead of instant-success compatibility checks. I am not counting that as a speed win. The default PR lane remains the fast path.

Final validation:

- exact local AGP 8.9.1 and Gradle 8.11.1 combined release build passed
- format, lint, ShellCheck, Actionlint, `assembleDebug`, and API check passed
- full Messaging In-App unit and JaCoCo reruns passed locally
- four consecutive hosted Kotlin Compose instrumentation passes after the lifecycle fix
- final [full test workflow](https://github.com/customerio/customerio-android/actions/runs/33284004088), [Maestro smoke flow](https://github.com/customerio/customerio-android/actions/runs/33284004094), and all PR checks passed

### Outdated-action follow-up

The hosted logs also show Node 20 deprecation warnings and HTTP 400 Gradle-cache restores. The repository already has Dependabot PR #629 for action majors, but it is stale, conflicted, and covers shared actions used by deployment. I deliberately did not mix an unvalidated release-runtime migration into this CI PR. The next isolated maintenance change should replace that PR with current pinned action versions and compare cache-hit rates before and after. GitHub has moved actions to Node 24: [GitHub runner notice](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/).

Snapshot publication behavior and stable release/deploy workflows are unchanged.

Shahroz16:
## Revised CI decision after correctness review

Exact head: `b49fd78`

The review feedback was correct: semantic-release intent is not a correctness boundary. A `chore:`, `ci:`, or dependency PR can break Gradle, public API, native consumers, SDK size, or sample apps. This head keeps those checks automatic.

- Unit, instrumentation, API, lint, binary-size, and supported AGP-floor sample builds remain ordinary PR correctness checks.
- The AGP 8.9.1 floor now performs one real combined Linux build of both Groovy and Kotlin DSL sample consumers. It no longer reports four false compatibility greens.
- The repository wrapper owns Gradle execution and caching. The unused standalone Gradle installation was removed.
- `ci:full` gates Maestro end-to-end coverage. Schedules and manual dispatch remain available.
- PR title and commit semantics control only snapshot publication and signed sample distribution side effects.
- Stable SDK release and deploy workflows are unchanged.

This PR has no `ci:full` label. Its commit range contains `fix:`, so release-side snapshot and sample distribution ran automatically, while Maestro skipped as designed. Final hosted result: **36 successful, 2 intentional skips, 1 neutral, 0 failures**. The real AGP compatibility build passed in **8m27s**, and all unit and API-31 instrumentation cells passed.

The controlled before-and-after validation reduced snapshot plus compatibility runner usage from **23.4 to 14.6 minutes, 37.7%**, while replacing false compatibility checks with real builds. Cache variability put the final head at 15m53s for those two jobs, so the percentage is not being presented as every-run wall time.

Claude Code Opus rejected the earlier semantic lane split, then returned **APPROVE** after correctness became automatic and path-based. All changed workflows pass actionlint and the final hosted head is green.

.github/scripts/validate-release-policy.sh
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env bash
```
mrehan27, line 1: **P2** — this script and `reusable-release-policy.yml` are byte-identical across all five SDK repos (sha256 match), with nothing keeping them in sync. `customerio/mobile-ci-tools` already hosts shared reusable workflows — this change calls one. Moving it there is mechanical with zero drift today; after merge, every policy fix becomes a five-repo change.

.github/workflows/maestro-e2e.yml
```diff
@@ -29,11 +30,37 @@ on:
...
+
+  # Metadata edits rerun full so a newer skipped check cannot replace a failure.
   prepare:
     name: Select Android suites
+    needs: release_policy
+    if: >-
+      always() &&
+      (vars.MOBILE_E2E_ENABLED == 'true' || github.event_name == 'workflow_dispatch') &&
+      (github.event_name != 'pull_request' ||
+      ((github.event.action == 'edited' ||
```
mrehan27: **P2** — `github.event.action == 'edited' || github.event.action == 'labeled'` forces the full lane regardless of classification, so a title typo fix or any unrelated label runs the expensive lane on a non-release PR. Gating `labeled` on `github.event.label.name == 'ci:full'` keeps the intent without that case.

.github/workflows/maestro-e2e.yml
```diff
@@ -29,11 +30,37 @@ on:
 
 concurrency:
   group: maestro-android-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```
mrehan27, line 40: **P2** — concurrency policy diverges across the five repos:

- android: `cancel-in-progress: ${{ github.event_name == 'pull_request' }}`
- ios, flutter: same, plus `|| startsWith(github.ref, 'refs/heads/feature/')`
- reactnative: ungated `cancel-in-progress: true`
- expo-plugin: both forms, in the same PR

Worth converging, or a comment if the differences are deliberate.

mrehan27 (commented):
Cross-repo review — the same change across five repos.

The lane gating looks right: omitting `always()` so a skipped `prepare` transitively blocks E2E, and moving the dependabot guard to `github.event.pull_request.user.login` (on a `labeled` event `github.actor` is the labeler, not the author).

Main ask: the release-policy script and reusable workflow are byte-identical in all five repos today (sha256 match), so extracting them to `mobile-ci-tools` is free now and a five-repo change after merge.

.github/workflows/build-sample-apps.yml
```diff
@@ -1,19 +1,42 @@
...
+  pull_request:
+    types: [opened, synchronize, reopened, edited, labeled]
   push:
     branches: [ main, feature/* ]
   release: # build sample apps for every git tag created. These are known as "stable" builds that are suitable for people outside the mobile team. 
     types: [ published ]
+  workflow_dispatch:
 
 concurrency: # cancel previous workflow run if one exists. 
   group: ${{ github.workflow }}-${{ github.ref }}
```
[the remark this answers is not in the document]
Shahroz16: Fixed in `8ab60277` across all three affected workflows. Unrelated labels and body-only edits now use a separate `metadata` concurrency group, so they cannot cancel an in-flight `ci:full` sample distribution, snapshot publication, or Maestro run. `ci:full`, title/base changes, and new commits remain in `validation`, preserving cancellation of genuinely stale work. `actionlint` passes, and the corrected cross-repository event model received an explicit Opus approval.

.github/workflows/maestro-e2e.yml
Shahroz16: Addressed on current head `8ab60277`. A bare metadata event no longer requests the full lane: release classification still controls eligibility, `ci:full` must be the label that fired the labeled event, and unrelated labels/body edits are isolated in the `metadata` concurrency group.

.github/workflows/reusable_build_sample_apps.yml
```diff
@@ -102,8 +102,7 @@ jobs:
           echo "BRANCH_NAME=${{ github.head_ref || github.ref_name }}" >> $GITHUB_ENV
           COMMIT_HASH="${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}"
           echo "COMMIT_HASH=${COMMIT_HASH:0:7}" >> $GITHUB_ENV
-          # Get the latest tag, making sure we get all tags with --tags flag (in case tag was just created)
-          git fetch --tags
```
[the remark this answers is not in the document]
Shahroz16: Addressed on the current head. `git fetch --tags` is explicitly restored in the Capture Git Context step before `git describe`, so the workflow does not rely on checkout's tag behavior.

.github/scripts/validate-release-policy.sh
Shahroz16: Agreed that five copies create drift risk. I am deliberately not extracting this classifier in these PRs because it gates signed/published side effects, and moving it immediately to the existing mutable shared-workflow dependency would enlarge the blast radius of a policy change. The safe follow-up is to version and pin the shared policy, canary one SDK, then migrate the other callers. For this rollout the local copies are byte-identical and fixture-tested.

.github/workflows/maestro-e2e.yml
Shahroz16: Reviewed across all five SDKs. The PR cancellation contract is consistent, including the new `metadata` versus `validation` isolation. The remaining `cancel-in-progress` syntax differs because non-PR trigger scope differs: iOS/Flutter retain cancellation for feature-branch pushes, Android preserves non-PR sample/release runs, Expo's full wrapper is PR-only, and React Native preserves scheduled/manual/reusable E2E. Converging the surface syntax here would change existing non-PR behavior.

.github/workflows/snapshot-release.yml
```diff
@@ -1,13 +1,50 @@
...
+  # ci:full run. They share a separate group from meaningful validation events.
+  group: >-
+    ${{ github.workflow }}-${{ github.ref }}-${{
+      github.event_name == 'pull_request' &&
+      ((github.event.action == 'labeled' && github.event.label.name != 'ci:full') ||
+      (github.event.action == 'edited' &&
+      github.event.changes.title.from == '' &&
+      github.event.changes.base.ref.from == '')) &&
+      'metadata' || 'validation'
+    }}
```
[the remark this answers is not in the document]
Shahroz16: Confirmed, and intentionally retained as the fail-closed tradeoff. A metadata run re-executes this Linux correctness job honestly; skipping the same-named job on the same SHA could replace a real failed check in the checks summary with a passing skip. Metadata events are rare and bounded to one concurrent group. Opus verified that no required-status-check rule exists today, so this protects the visible result and future gating rather than a current hard merge gate. A later split of correctness and publication into separate workflows can remove the duplicate without that masking risk.

.github/workflows/build-sample-apps.yml
```diff
@@ -1,19 +1,50 @@
...
-  cancel-in-progress: true
+  # Unrelated labels and body-only edits must not cancel an in-flight release or
+  # ci:full run. They share a separate group from meaningful validation events.
+  group: >-
+    ${{ github.workflow }}-${{ github.ref }}-${{
+      github.event_name == 'pull_request' &&
+      ((github.event.action == 'labeled' && github.event.label.name != 'ci:full') ||
+      (github.event.action == 'edited' &&
+      !github.event.changes.title &&
+      !github.event.changes.base)) &&
```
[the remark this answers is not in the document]
Shahroz16: Addressed on current head `6baf088c`. Although GitHub's loose null/empty comparison already made the prior body-only guard fail closed, all edited-event guards now use explicit `changes.title != null || changes.base != null` booleans, matching the concurrency predicate exactly. Body-only edits cannot start distribution/publish jobs; title edits and base retargets can replace full work only when release policy or `ci:full` qualifies the PR. `actionlint` passes and Opus approved the aligned truth table.

mahmoud-elmorabea (approved):

mahmoud-elmorabea:
- AGP pinned at 8.9.1 in the workflow while Versions.kt default is 8.9.3; nothing ties them together, and Versions.kt still says to keep the workflow in sync.
- Gating snapshot publish behind full removes per-PR branch-SNAPSHOT artifacts that wrapper testing has used
- release_policy / classify runs twice (two callers, each a fetch-depth: 0 clone).

Shahroz16:
Addressed the review feedback in b57dcdd6.\n\n- Restored branch SNAPSHOT publication for every internal, non-documentation PR. Wrapper SDK consumers no longer depend on release naming or ci:full to receive the PR artifact.\n- Removed the second release-policy classification and full-history checkout from this workflow.\n- Clarified that AGP 8.9.3 is the repository default while AGP 8.9.1 plus Gradle 8.11.1 is the explicit compileSdk 36 compatibility floor.\n- Split job concurrency so a multi-module Maven upload cannot be canceled halfway through and expose a mixed SNAPSHOT. Superseded compatibility builds remain cancelable.\n- Preserved base-retarget validation while skipping title-only and body-only edits.\n\nHosted proof: [snapshot publication and AGP-floor sample builds both passed](https://github.com/customerio/customerio-android/actions/runs/33562063003).

PR #826, opened 2026-08-30
chore: report WebView render process termination as an in-app failure

mahmoud-elmorabea:
## MBL-2310 · Android stack — step 2 of 4

| Step | PR | |
| --- | --- | --- |
| **2** | **#826 — `chore:` report WebView render process termination as an in-app failure** | **← you are here** |
| 3 | #827 — `chore:` classify and log in-app message failures internally |  |
| 4 | #828 — `fix:` surface the failure reason on `errorWithMessage` | cuts the Android release |

---
Handles the WebView renderer dying mid-message, which the SDK previously ignored entirely.

Stacked on #823. Groundwork for MBL-2310 — it closes the one platform gap in the in-app failure map.

### Why

`WebViewClient.onRenderProcessGone` had no override anywhere in the repo, so the platform default applied: **the host app's process was killed along with the renderer**, and the host was told nothing about the message. iOS has handled the equivalent since forever via `EngineWeb.webViewWebContentProcessDidTerminate`.

Returning `true` is the point of the override — it tells the platform we absorbed the loss. The failure is then reported through the listener like any other load error, so `MessageLoadingFailed` dispatches, the message is dismissed, and teardown runs normally.

`releaseResources()` is deliberately *not* called from here: it bails out early while the view is still attached to its parent, which it always is at this point, so it would be a silent no-op.

### Verification

New `EngineWebViewRenderProcessGoneTest` drives the real `WebViewClient` through Robolectric and asserts both halves — the call returns `true`, and the listener sees exactly one `error()`.

- `:messaginginapp:testDebugUnitTest` green (1 test, 0 failures).
- `:messaginginapp:lintDebug` green, zero `NewApi` findings. `onRenderProcessGone` is API 26+ against `minSdk` 21; the platform only invokes it on 26+, and `didCrash()` is version-guarded.

### DIFF
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
@@ -5,7 +5,9 @@ import android.content.Context
 import android.content.res.Configuration
 import android.graphics.Color
 import android.net.http.SslError
+import android.os.Build
 import android.util.AttributeSet
+import android.webkit.RenderProcessGoneDetail
 import android.webkit.SslErrorHandler
 import android.webkit.WebResourceError
 import android.webkit.WebResourceRequest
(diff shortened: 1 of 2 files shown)


messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
```diff
@@ -268,6 +270,34 @@ internal class EngineWebView @JvmOverloads constructor(
...
+                    } else {
+                        null
+                    }
+                    logger.error(
+                        "WebView render process gone (didCrash: $didCrash), reporting message failure"
+                    )
+                    // Report like any other load error. That dispatches MessageLoadingFailed,
+                    // which dismisses the message and runs the normal teardown — releaseResources()
+                    // cannot be called from here, it bails out while the view is still attached.
+                    listener?.error()
```
Shahroz16, line 301: Could we clean up the dead WebView explicitly before returning true? Android treats the view passed to onRenderProcessGone as unusable and asks the host to remove it and clear its references. Right now we report through the normal dismissal path, which later calls stopLoading/removeJavascriptInterface on that same view, and the modal path never reaches releaseResources. A small renderer-gone cleanup path that cancels the timer and color-scheme job, removes the child, and clears webView without using the normal WebView operations would make the true return value match the platform contract. It would also be good for this test to assert that cleanup, rather than only the listener call.

mahmoud-elmorabea: Done — and I checked the part I was most sceptical of, which turned out to be the important bit. `releaseResources()` has exactly one caller, `InlineInAppMessageViewController.kt:218`, so **the modal path never reaches it at all**. My original comment claimed the dismissal path would handle teardown; that was wrong for modals.

Added `releaseCrashedWebView()`, called before we report and before returning true. It cancels the timer and the colour-scheme job, detaches the JS interface, removes the view from its parent and destroys it — each step guarded on its own, since the renderer is already gone and any of them can throw. It also deliberately runs while the view is still attached, which is the state a crash leaves us in and precisely what `releaseResources()` refuses to do.

Nulling `webView` there has a second benefit: the later `stopLoading()` / `releaseResources()` calls on the dismissal path become no-ops instead of driving a dead view.

Test added as you suggested — asserts the child is removed and `destroy()` was called, not just the listener call. Verified it fails without the cleanup call.

Shahroz16 (approved):
The renderer teardown now matches the platform lifecycle, and the new test covers removal and destruction. Looks good.

PR #827, opened 2026-08-30
chore: classify and log in-app message failures internally

mahmoud-elmorabea:
## MBL-2310 · Android stack — step 3 of 4

| Step | PR | |
| --- | --- | --- |
| 2 | #826 — `chore:` report WebView render process termination as an in-app failure |  |
| **3** | **#827 — `chore:` classify and log in-app message failures internally** | **← you are here** |
| 4 | #828 — `fix:` surface the failure reason on `errorWithMessage` | cuts the Android release |

---
Gives every in-app failure a reason and keeps the detail the SDK was throwing away. Internal only — `apiDump` is unchanged.

Stacked on #826. The Android mirror of ios#1237. Groundwork for MBL-2310; the next PR puts the reason on `errorWithMessage`.

### Why

Every failure site in `EngineWebView` collapsed into a bare `listener?.error()`, so why a message failed survived only in whatever each site happened to log. The richest sources were dropped outright:

- **The renderer's own `errorMessage`.** The 3.0 bundle sends `{ target, errorMessage }` with the engine `error` event — e.g. `Unable to find "step-2" in payload.` The bridge discarded the whole parameter map.
- **Everything the four `WebViewClient` error callbacks hand us**: `WebResourceError` code and description, the HTTP status from `onReceivedHttpError`, and `SslError.primaryError`.

### Change

An internal `InAppMessageError` (`reason`, `detail`, `code`) with five coarse reasons — `NETWORK`, `TIMEOUT`, `RENDER_FAILED`, `WEB_VIEW_CRASHED`, `INTERNAL_ERROR` — matching iOS exactly.

All sites now route through one `reportFailure` choke point that classifies and logs before notifying the listener, so the next PR changes the hand-off in one place.

`InAppMessageViewController.routeError` is classified too, though renderer 3.0 never emits `routeError` — noted in a comment so nobody counts it as live coverage.

### The internal seam

`EngineWebViewListener.error()` takes no arguments and **is public API** (`messaginginapp.api:201`), as is `EngineWebViewInterface` itself. So the bridge cannot hand the parsed detail over through the listener yet. It goes through an `internal var onEngineError` instead — deliberately temporary, and it disappears when the listener can take the error in the `fix:` PR.

Same constraint iOS hit with `EngineWebDelegate`, so both platforms split the same way.

### Verification

`EngineWebViewInterfaceErrorTest` drives real JSON through `postMessage` — the actual bridge entry point — and covers all four `errorMessage`/`target` combinations plus log formatting.

- `:messaginginapp:testDebugUnitTest` green, 6 new tests.
- `:messaginginapp:lintDebug` green.
- `./gradlew apiDump` produces **no diff** — every new declaration is `internal` or `private`.

### DIFF
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
@@ -4,11 +4,9 @@ import android.annotation.SuppressLint
 import android.content.Context
 import android.content.res.Configuration
 import android.graphics.Color
-import android.net.http.SslError
 import android.os.Build
 import android.util.AttributeSet
 import android.webkit.RenderProcessGoneDetail
-import android.webkit.SslErrorHandler
 import android.webkit.WebResourceError
 import android.webkit.WebResourceRequest
 import android.webkit.WebResourceResponse
@@ -257,6 +255,15 @@ internal class EngineWebView @JvmOverloads constructor(
                     )
                 }
 
+                // Deliberately no onReceivedSslError override. The platform default already
+                // cancels the request, which is the behaviour we want and the only part that
+                // matters for safety — we never continue past a certificate error. Overriding it
+                // replaced that default with our own cancel() and reported every failure as a
+                // message-level NETWORK error, including certificate failures on subresources,
+                // which dismissed messages that would have rendered. The callback carries no
+                // WebResourceRequest, so there is no reliable way to tell the document from a
+                // subresource. A certificate failure on the document aborts the main-frame load
+                // and surfaces through onReceivedError below, which is already frame-gated.
                 override fun onReceivedHttpError(
                     view: WebView?,
                     request: WebResourceRequest?,
@@ -293,25 +300,6 @@ internal class EngineWebView @JvmOverloads constructor(
                             reason = InAppMessageErrorReason.NETWORK,
                             detail = if (hasDetail) error?.description?.toString() else null,
                             code = if (hasDetail) error?.errorCode else null
-                        )
-                    )
-                }
-
-                override fun onReceivedSslError(
-                    view: WebView?,
-                    handler: SslErrorHandler?,
-                    error: SslError?
-                ) {
-                    // Overriding this removed the platform default's cancel(), so the handler has
-                    // to be resolved here or the request stays suspended. Always cancel: the
-                    // renderer is first-party HTTPS and we never continue past a certificate error.
-                    handler?.cancel()
-
-                    reportFailure(
-                        InAppMessageError(
-                            reason = InAppMessageErrorReason.NETWORK,
-                            detail = "SSL error loading the renderer",
-                            code = error?.primaryError
                         )
                     )
                 }
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
@@ -4,9 +4,11 @@ import android.annotation.SuppressLint
 import android.content.Context
 import android.content.res.Configuration
 import android.graphics.Color
+import android.net.http.SslError
 import android.os.Build
 import android.util.AttributeSet
 import android.webkit.RenderProcessGoneDetail
+import android.webkit.SslErrorHandler
 import android.webkit.WebResourceError
 import android.webkit.WebResourceRequest
 import android.webkit.WebResourceResponse
@@ -47,6 +49,14 @@ internal class EngineWebView @JvmOverloads constructor(
     private val logger = SDKComponent.logger
     private var lastResolvedColorScheme: String? = null
     private var colorSchemeJob: Job? = null
+
+    /**
+     * The renderer document this view loaded.
+     *
+     * `onReceivedSslError` gives no [WebResourceRequest], so this is the only way to tell a
+     * certificate failure on the message itself from one on an image or font it pulls in.
+     */
+    private var documentUrl: String? = null
 
     private val inAppMessagingManager = SDKComponent.inAppMessagingManager
 
@@ -191,8 +201,8 @@ internal class EngineWebView @JvmOverloads constructor(
         elapsedTimer.start("Engine render for message: ${configuration.messageId}")
         val messageData = mapOf("options" to configuration)
         val jsonString = Gson().toJson(messageData)
-        val messageUrl =
-            "${state.environment.getGistRendererUrl()}/index.html"
+        val messageUrl = "${state.environment.getGistRendererUrl()}/index.html"
+        documentUrl = messageUrl
         logger.debug("Rendering message with URL: $messageUrl")
         webView?.let {
             it.settings.javaScriptEnabled = true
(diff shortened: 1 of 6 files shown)


messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
```diff
@@ -244,31 +248,58 @@ internal class EngineWebView @JvmOverloads constructor(
...
+                    )
                 }
 
                 override fun onReceivedHttpError(
                     view: WebView?,
                     request: WebResourceRequest?,
                     errorResponse: WebResourceResponse?
                 ) {
-                    listener?.error()
+                    reportFailure(
```
Shahroz16, line 268: Can we ignore these request-scoped errors when request?.isForMainFrame == false? Android invokes both onReceivedHttpError and the request-based onReceivedError for images, fonts, iframes, and other subresources. Without the gate, a missing image can be surfaced as a message-level NETWORK 404 and dismiss an otherwise renderable message. The deprecated overload remains main-frame-only on API 21-22, so this also makes behavior differ by OS version. I would add one main-frame and one subresource test around both request-based callbacks.

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
```diff
@@ -244,31 +248,58 @@ internal class EngineWebView @JvmOverloads constructor(
...
+                    )
                 }
 
                 override fun onReceivedSslError(
                     view: WebView?,
                     handler: SslErrorHandler?,
                     error: SslError?
                 ) {
-                    listener?.error()
+                    reportFailure(
```
Shahroz16, line 310: Could we call handler?.cancel() before reporting the failure? Once we override onReceivedSslError, Android expects us to resolve the handler with cancel or proceed; leaving it unresolved can keep the request suspended. We should always cancel here and never continue past a certificate error.

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
mahmoud-elmorabea: Confirmed, and I agree with the analysis — the request-based `onReceivedError` and `onReceivedHttpError` both fire for subresources, so a missing image can surface as a message-level `NETWORK` failure and dismiss a renderable message. The API 21-22 divergence via the main-frame-only deprecated overload is real too.

I have left it out of this stack though, because it predates the work: on `origin/main` all four callbacks already call `listener?.error()` with no gate, so the dismiss-on-broken-asset behaviour is what ships today. This PR only changes what we log alongside it. Gating on `isForMainFrame` would change rendering behaviour for live messages with broken assets, which felt like more than an in-app-error-reason ticket should carry silently.

Happy to file it as its own ticket with your analysis, or fold it in here if you would rather it ship together — your call.

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
mahmoud-elmorabea: Confirmed — no `cancel()` or `proceed()`, so the handler is left unresolved. Overriding removed the default implementation's `cancel()`, and we should never proceed past a certificate error.

Same reasoning as the sibling comment: this predates the stack, the override exists unchanged on `origin/main`, and this PR only alters the classification we log. I would rather not slip a TLS-handling change into a `chore:` PR that reviewers are reading as logging-only.

I would group it with the `isForMainFrame` gate in one follow-up ticket — both are `WebViewClient` contract issues in the same block. Say the word and I will file it, or pull both in here.

Shahroz16 (changes requested):
After rechecking the complete stack and an independent Opus pass, I think the two existing WebViewClient threads should be fixed here rather than moved to a follow-up. This PR turns those callbacks into public NETWORK failures, and both fixes are small and testable: ignore subresource errors, cancel the SSL handler, and cover the callback behavior. The rest of this PR looks good.

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
mahmoud-elmorabea: Done — your argument for doing it here is better than mine for deferring it. This PR is what promotes these callbacks into a publicly reported `NETWORK` reason, so leaving it would ship a wrong, documented cause to the host rather than just the silent dismissal that exists today.

Both request-based callbacks now return early unless `request.isForMainFrame`. That also lines API 23+ up with 21-22, where the deprecated overload is main-frame-only already.

Tests as you asked, one main-frame and one subresource case for each. Negative control: with the gates removed, both subresource tests fail.

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
mahmoud-elmorabea: Done. `handler?.cancel()` before reporting, never `proceed()` — the renderer is first-party HTTPS, so there is no case where continuing past a certificate error is right.

Covered by a test asserting `cancel()` is called exactly once and `proceed()` never; it fails without the call.

Shahroz16: One more edge case remains in this callback: Android invokes `onReceivedSslError` while loading any resource, not only the main frame. We now ignore HTTP and request errors for subresources, but an image or font certificate failure would still surface `NETWORK` and dismiss the whole message. Could we keep cancelling every handler but report the failure only for the renderer document, and add a subresource test?

mahmoud-elmorabea: Good catch — and it pushed us somewhere better than the gate. Rather than adding a third frame check, I have **removed the `onReceivedSslError` override entirely**.

The reasoning: the platform default already calls `handler.cancel()`, which is the only part that matters for safety — we never continue past a certificate error either way. Overriding it replaced that default with our own cancel and, as you spotted, reported every certificate failure as a message-level `NETWORK` error including subresources. Dropping the override fixes both problems at once and makes the class smaller instead of larger.

It also avoids the awkward part of gating this one. Unlike the two request-based callbacks there is no `WebResourceRequest` here, so the only way to identify the document would be comparing `SslError.getUrl()` against the renderer URL — a string heuristic in security-adjacent code, and one whose realistic failure mode (redirects, query strings) is a silent non-match.

A certificate failure on the document aborts the main-frame load and should surface through `onReceivedError`, which is already frame-gated, as `NETWORK` with `ERROR_FAILED_SSL_HANDSHAKE`.

**Two things worth your eyes on this.**

The cost is real: we lose `SslError.primaryError`, which distinguished untrusted from expired from hostname mismatch. That is the most precise cert diagnostic we could have offered, and this ticket is about precision. I think the trade is right, but it is a trade.

And the load-bearing assumption is that an SSL-aborted main-frame load does reach `onReceivedError` at all — Chromium suppresses that callback for some aborted loads. If it does not fire, a genuine document certificate failure degrades to `TIMEOUT` after 5s, which is misleading. I have not verified that on a device yet and did not want to claim it. It is on the device-test list, and if it turns out not to fire I will come back and implement the gate as you originally described.

Test kept and inverted: it now pins that we never `proceed()`, and that a certificate error reports nothing of ours.

Shahroz16: The documented Android split is narrower than the assumption here: `onReceivedSslError` handles recoverable certificate errors, while `onReceivedError(... ERROR_FAILED_SSL_HANDSHAKE)` is only promised for non-recoverable SSL failures. With this override removed, a recoverable certificate error on the renderer document is canceled by default but never reaches our listener, so it becomes `TIMEOUT` after 5s. Could we keep the override, always cancel, and distinguish the current main-frame URL before reporting `NETWORK`? If URL matching is the available signal, a redirected-main-frame test would pin the tradeoff better than relying on a callback Android does not promise. https://developer.android.com/reference/android/webkit/WebViewClient#onReceivedSslError(android.webkit.WebView,android.webkit.SslErrorHandler,android.net.http.SslError)

mahmoud-elmorabea: You are right and I was wrong — thank you for the documentation link, that settles the exact thing I said I could not verify.

The recoverable/non-recoverable split is the part I missed. `ERROR_FAILED_SSL_HANDSHAKE` is only promised for non-recoverable failures, and the common cases — expired, untrusted, hostname mismatch — are recoverable, so they arrive at `onReceivedSslError` and nowhere else. Dropping the override turned exactly those into a 5s `TIMEOUT`. That is worse than what we started with.

Restored, with the two concerns split as you suggested:

- **Cancellation is unconditional**, every resource, so the handler is always resolved and we never continue past a certificate error.
- **Reporting is gated** on the failing URL matching the renderer document, which the view now remembers.

Three tests: document (refused **and** reported), subresource (refused, not reported), and the redirected-main-frame case you asked for. That last one pins the tradeoff honestly — after a redirect the URL no longer matches, so the error is refused but not reported and surfaces later as a timeout. The failure mode is silence rather than a wrong dismissal, which is the safer of the two directions, and it is now written down rather than assumed.

Shahroz16 (approved):
The SSL path now always cancels the handler, reports only failures on the renderer document, and covers document, subresource, and redirect behavior. I rechecked the complete stack at this head and the focused WebView/error/controller tests pass. This PR looks good.

PR #828, opened 2026-08-30
fix: surface the failure reason on in-app error callback

mahmoud-elmorabea:
## MBL-2310 · Android stack — step 4 of 4

| Step | PR | |
| --- | --- | --- |
| 2 | #826 — `chore:` report WebView render process termination as an in-app failure |  |
| 3 | #827 — `chore:` classify and log in-app message failures internally |  |
| **4** | **#828 — `fix:` surface the failure reason on `errorWithMessage`** | **← you are here** |

---
Puts the reason an in-app message failed onto `errorWithMessage`. Closes the MBL-2310 work on Android.

Stacked on #827. Paired with ios#1237's follow-up.

### Why

`errorWithMessage` delivered only `messageId` and `deliveryId`, so an integrator learned that a message failed but never why. The reason existed on device in debug logs only, and nothing is recorded server-side, so diagnosing meant shipping a debug-logging build.

### API

```kotlin
enum class InAppMessageErrorReason { NETWORK, TIMEOUT, RENDER_FAILED, WEB_VIEW_CRASHED, INTERNAL_ERROR }

data class InAppMessageError(
    val reason: InAppMessageErrorReason,
    val detail: String? = null,
    val code: Int? = null
)

fun errorWithMessage(message: InAppMessage, error: InAppMessageError) = errorWithMessage(message)
```

Reasons are coarse on purpose: each maps to a different thing an integrator would actually do — check connectivity, look at a slow renderer, report the message to us, or file an SDK bug. `detail` carries the renderer's own `errorMessage` or the WebView's error description; it is diagnostic text for logs, not for parsing.

### Additive, including for Java

The new overload has a default body, so existing listeners keep compiling **and keep receiving failures** — the default forwards to the reason-less callback.

This is what #823 was for. The API dump shows the payoff:

```
public abstract interface class io/customer/messaginginapp/type/InAppEventListener {
    public abstract fun errorWithMessage (…InAppMessage;)V
    public fun errorWithMessage (…InAppMessage;…InAppMessageError;)V   <- not abstract
}
public final class io/customer/messaginginapp/type/InAppEventListener$DefaultImpls { … }
```

`public fun` rather than `public abstract fun` is the real Java default method, so Java implementers are not forced to override it. `DefaultImpls` is retained, so already-compiled Kotlin still links.

### Verification

**The Java canary is the headline.** `samples/java_layout` implements `InAppEventListener` in Java and is **unchanged in this PR — 0 files touched.** It compiles green against the SDK consumed as a **published artifact**:

```
IS_DEVELOPMENT=true ./gradlew publishToMavenLocal
./gradlew :samples:java_layout:assembleDebug -PcioSDKVersion=local --refresh-dependencies
```

Confirmed with 0 `:messaginginapp:` source tasks, so it really resolved the aar rather than building from source.

- `samples/kotlin_compose` updated to demo the new callback, logging and tracking `reason`, `detail`, and `code`. It keeps the reason-less override too, since that member is still required.
- `:messaginginapp:testDebugUnitTest` and `:messaginginapp:lintDebug` green.
- The temporary `onEngineError` seam added in #827 is removed — the listener carries the error directly now.
- New test asserts the SDK uses the classified overload and **never** the reason-less one, plus a case pinning `INTERNAL_ERROR` for an unclassified failure.

### DIFF
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
@@ -56,6 +56,12 @@ internal class EngineWebView @JvmOverloads constructor(
      */
     private var documentUrl: String? = null
 
+    /**
+     * Set once the first failure is reported. An engine renders one message, so it fails at most
+     * once.
+     */
+    private var hasReportedFailure = false
+
     private val inAppMessagingManager = SDKComponent.inAppMessagingManager
 
     private val state: InAppMessagingState
@@ -472,6 +478,15 @@ internal class EngineWebView @JvmOverloads constructor(
      * Single exit for every failure in this view: classify, then notify the listener.
      */
     private fun reportFailure(error: InAppMessageError) {
+        // First failure wins, and the latch — not the cancel — is what makes that deterministic.
+        // The bootstrap TimerTask runs independently of the WebView callbacks, so cancelling the
+        // timer alone still races with a task that has already started. Without this the host could
+        // be told NETWORK and then TIMEOUT about the same message, the second overwriting the real
+        // cause. Mirrors the same guard on iOS.
+        if (hasReportedFailure) return
+        hasReportedFailure = true
+        cleanupTimer()
+
         val listener = this.listener
         if (listener == null) {
             // Nothing downstream will see this one, so it has to be logged here. The timeout timer
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
@@ -472,11 +472,16 @@ internal class EngineWebView @JvmOverloads constructor(
      * Single exit for every failure in this view: classify, then notify the listener.
      */
     private fun reportFailure(error: InAppMessageError) {
-        // Not logged here any more. In the previous step this log was the only place the
-        // classification surfaced, because the listener could not carry it. Now that it can, the
-        // controller logs the same failure with the message id and route attached, so logging in
-        // both places would double up on every failure.
-        listener?.error(error)
+        val listener = this.listener
+        if (listener == null) {
+            // Nothing downstream will see this one, so it has to be logged here. The timeout timer
+            // can still fire after the view is detached and the listener cleared.
+            logger.error("In-app message failed with no listener attached: ${error.describeForLogs()}")
+            return
+        }
+        // Deliberately not logged on this path: the controller logs the same failure straight after,
+        // with the message id and route attached, so logging here too would double up on every one.
+        listener.error(error)
     }
 
     /**
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
@@ -43,9 +43,7 @@ internal class EngineWebView @JvmOverloads constructor(
     private var timerTask: TimerTask? = null
     private var webView: WebView? = null
     private var elapsedTimer: ElapsedTimer = ElapsedTimer()
-    private val engineWebViewInterface = EngineWebViewInterface(this).apply {
-        onEngineError = { error -> reportFailure(error) }
-    }
+    private val engineWebViewInterface = EngineWebViewInterface(this)
     private val logger = SDKComponent.logger
     private var lastResolvedColorScheme: String? = null
     private var colorSchemeJob: Job? = null
@@ -439,6 +437,10 @@ internal class EngineWebView @JvmOverloads constructor(
         listener?.error()
     }
 
+    override fun error(error: InAppMessageError) {
+        reportFailure(error)
+    }
+
     /**
      * Tears down a WebView whose render process has died.
      *
@@ -468,13 +470,10 @@ internal class EngineWebView @JvmOverloads constructor(
 
     /**
      * Single exit for every failure in this view: classify, log, then notify the listener.
-     *
-     * [EngineWebViewListener.error] takes no arguments and is public API, so the classified error
-     * reaches the logs but not the host yet.
      */
     private fun reportFailure(error: InAppMessageError) {
         logger.error("In-app message failed: ${error.describeForLogs()}")
-        listener?.error()
+        listener?.error(error)
     }
 
     /**
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
@@ -469,10 +469,13 @@ internal class EngineWebView @JvmOverloads constructor(
     }
 
     /**
-     * Single exit for every failure in this view: classify, log, then notify the listener.
+     * Single exit for every failure in this view: classify, then notify the listener.
      */
     private fun reportFailure(error: InAppMessageError) {
-        logger.error("In-app message failed: ${error.describeForLogs()}")
+        // Not logged here any more. In the previous step this log was the only place the
+        // classification surfaced, because the listener could not carry it. Now that it can, the
+        // controller logs the same failure with the message id and route attached, so logging in
+        // both places would double up on every failure.
         listener?.error(error)
     }
 
--- a/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
+++ b/messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
@@ -31,6 +31,7 @@ import io.customer.messaginginapp.ui.bridge.EngineWebViewDelegate
 import io.customer.sdk.core.di.SDKComponent
 import java.util.Timer
 import java.util.TimerTask
+import java.util.concurrent.atomic.AtomicBoolean
 import kotlinx.coroutines.Job
 
 internal class EngineWebView @JvmOverloads constructor(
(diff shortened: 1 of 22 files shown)


messaginginapp/src/main/java/io/customer/messaginginapp/type/InAppMessageError.kt
```diff
@@ -7,7 +7,7 @@ package io.customer.messaginginapp.type
  * it — check connectivity, look at a slow renderer, report the message content to us, or file an
  * SDK bug. Finer detail belongs in [InAppMessageError.detail].
  */
-internal enum class InAppMessageErrorReason {
+enum class InAppMessageErrorReason {
```
Shahroz16, line 14: Since this is a new public enum, could we tell integrators to keep a fallback branch when switching on it? Adding a reason later will otherwise break Kotlin callers that use an exhaustive when expression. The same guidance on iOS would also make the wrapper contract easier to evolve.

messaginginapp/src/main/java/io/customer/messaginginapp/type/InAppMessageError.kt
```diff
@@ -33,13 +33,17 @@ internal enum class InAppMessageErrorReason {
  * @param code the underlying platform error code where the failing layer had one, e.g. a
  * [android.webkit.WebResourceError] code or an HTTP status. Null when there was no numeric code.
  */
-internal data class InAppMessageError(
+data class InAppMessageError(
```
Shahroz16, line 42: Could we make the safety guidance for detail a little stronger before this becomes public? Some values come directly from the renderer, so I would say it is for local diagnostics and should not be forwarded verbatim to analytics. The sample currently tracks error-detail, which encourages exactly that pattern.

messaginginapp/src/main/java/io/customer/messaginginapp/type/InAppMessageError.kt
mahmoud-elmorabea: Done — the KDoc now says to keep an `else ->` branch, and why: more reasons may be added and an exhaustive `when` will stop compiling when they are.

Matching guidance added on iOS in customerio/customerio-ios#1238, with one difference worth knowing about. There I documented a plain `default:` rather than `@unknown default:`, because the iOS SDK ships as source without library evolution, so the compiler treats the enum as frozen at the integrator's build and `@unknown default:` is the wrong tool. Net effect for wrapper authors is the same on both platforms: always keep a fallback branch.

messaginginapp/src/main/java/io/customer/messaginginapp/type/InAppMessageError.kt
mahmoud-elmorabea: Done, and the point about the sample landed — it was demonstrating exactly the pattern the guidance warns against.

`detail` is now documented as local diagnostics only: log it, don't parse it, and don't forward it verbatim to analytics or crash reporting, with the reminder to branch on `reason` instead. `kotlin_compose` now logs the detail and tracks only `error-reason` and `error-code`. Same change to APN-UIKit on the iOS side.

Shahroz16 (approved):
The enum-growth guidance and diagnostic-detail handling are clear now, and the sample no longer forwards raw detail. This PR looks good. The parent #827 still has the two WebViewClient fixes before the full stack is ready.

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
```diff
@@ -436,14 +438,19 @@ internal class EngineWebView @JvmOverloads constructor(
     }
 
     /**
-     * Single exit for every failure in this view: classify, log, then notify the listener.
-     *
-     * [EngineWebViewListener.error] takes no arguments and is public API, so the classified error
-     * reaches the logs but not the host yet.
+     * Single exit for every failure in this view: classify, then notify the listener.
      */
     private fun reportFailure(error: InAppMessageError) {
```
Shahroz16, line 486: Could we make the first classified failure terminal here? `reportFailure` leaves the bootstrap timer armed, and the `TimerTask` runs independently of WebView callbacks. If teardown is delayed, the same engine can report the real `NETWORK` failure and then a second `TIMEOUT`; cancelling the timer alone can also race with a task that has already started. A small first-failure latch plus `cleanupTimer()` here, with a test that sends two different failures and asserts only the first reason reaches the listener, would make the callback deterministic.

Shahroz16 (changes requested):
One final issue remains in the current leaf: a classified failure is not terminal, so the bootstrap timer can deliver a second timeout. The rest of this PR still looks good.

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
mahmoud-elmorabea: Fixed — and this one is squarely on me. I added exactly this latch to iOS in the sibling PR, wrote the commit message explaining why an engine should fail at most once, and then did not carry it across. Parity is the point of these paired stacks.

`reportFailure` now latches the first failure and calls `cleanupTimer()`. Your point about the race is the reason the ordering matters: the latch is the guard and the cancel is belt-and-braces, because cancelling alone still loses to a `TimerTask` that has already started.

Test sends two different failures — a main-frame `onReceivedError`, then `onRenderProcessGone` — and captures what reaches the listener: exactly one call, reason `NETWORK`. Negative control passes, it fails without the latch.

Bugbot flagged the same thing independently on this commit; both are covered by this change.

Shahroz16: Could we make the check-and-set atomic as well? The timer runs on its own thread, and renderer errors arrive through the JS bridge, so both can read false before either writes true. I reproduced duplicate listener callbacks from concurrent timeout and renderer failures. AtomicBoolean.compareAndSet(false, true), or serializing failures onto one thread, would complete the first-failure guarantee.

Shahroz16 (commented):
The sequential failure case is fixed. One concurrent failure case remains in the existing thread.

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
mahmoud-elmorabea: Fixed in `06a4dcaf`, and you were right on both counts — including one you did not name.

The check-and-set is now `AtomicBoolean.compareAndSet(false, true)`. The race you reproduced is real: `TimerTask` runs on the timer's own thread, renderer errors arrive on the JS bridge thread, and the `WebViewClient` callbacks are on the UI thread, so three callers could interleave on a plain read-then-write.

The worse half was the missing memory barrier. The field was a non-volatile `var`, so there was no guarantee the write ever became visible to the other threads — the latch could have failed without any tight interleaving at all. `compareAndSet` closes both in one move, which is why I took it over serializing onto one thread.

Worth recording why this got through: the latch was ported from iOS, where it is genuinely safe — `Timer` fires on the run loop that scheduled it and the `WKWebView` delegate callbacks are main-thread, so only one thread touches it there. I mirrored the code without re-checking the threading model underneath it. Parity of implementation, not of correctness.

Test contends eight threads on one engine across forty fresh instances and asserts the host hears exactly one failure. Negative control passes — it fails against the plain `Boolean`, passes against `AtomicBoolean`. Full module suite: 542 tests, 0 failures.

Shahroz16: Thanks, the atomic latch closes the concurrent-failure window. The concurrent and sequential failure tests pass, along with the callback propagation tests; existing listeners still receive the fallback callback.

Shahroz16 (approved):
The concurrent failure fix is verified. All 135 focused in-app tests and the module build pass, with the legacy listener fallback preserved.

messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
```diff
@@ -439,6 +449,10 @@ internal class EngineWebView @JvmOverloads constructor(
         listener?.error()
```
sydneycollins-cio, line 449: Two paths bypass the first-failure latch: error() here and the routeError chain. Both dormant today, but if routeError revives, the host hears RENDER_FAILED then TIMEOUT for one message. Routing both through reportFailure would close it. Fine as a follow-up.

messaginginapp/src/main/java/io/customer/messaginginapp/type/InAppMessageError.kt
```diff
@@ -29,17 +33,23 @@ internal enum class InAppMessageErrorReason {
...
  * @param detail human-readable detail from the layer that failed — a WebView error description, or
- * the message the renderer itself reported. Free-form and unstable: log it, don't parse it.
+ * the message the renderer itself reported. Free-form, unstable, and partly renderer-supplied, so
+ * treat it as **local diagnostics only**: write it to your logs, don't parse it, and don't forward
+ * it verbatim to analytics or crash reporting. Branch on [reason] instead.
  * @param code the underlying platform error code where the failing layer had one, e.g. a
  * [android.webkit.WebResourceError] code or an HTTP status. Null when there was no numeric code.
  */
-internal data class InAppMessageError(
+data class InAppMessageError(
```
sydneycollins-cio, line 42: Public data class: adding a field later breaks binary compat via constructor/copy/componentN. Worth a conscious call before release: accept that, or make it a plain class now.

messaginginapp/src/test/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebViewConcurrentFailureTest.kt
```diff
@@ -0,0 +1,117 @@
...
+        override fun routeLoaded(route: String) = Unit
+        override fun sizeChanged(width: Double, height: Double) = Unit
+    }
+
+    @Test
+    fun reportFailure_givenSimultaneousFailures_expectExactlyOneReachesHost() {
+        val context = ApplicationProvider.getApplicationContext<Context>()
+        // A single pass can win the race by luck, so repeat over fresh engines and assert on the
+        // whole run. Against a plain read-then-write flag this fails well inside the first few.
+        val engines = 40
```
sydneycollins-cio, line 73: Nit: 40 engines x 8 threads is heavy for one latch. Candidate to shrink if CI time creeps.

sydneycollins-cio (approved):
LGTM.

Non-blocking: RN/Flutter bridges implement InAppEventListener and will silently keep using the reason-less callback. Is there a tracked follow-up to adopt the new reason callback in both wrappers so the failure reason reaches JS/Dart?

PR #830, opened 2026-08-31
chore(sample): persist SDK logs to an NDJSON file in java_layout

mahmoud-elmorabea:
[MBL-2322: Persist SDK logs to an NDJSON file in the sample apps](https://linear.app/customerio/issue/MBL-2322/persist-sdk-logs-to-an-ndjson-file-in-the-sample-apps)

Android half of MBL-2322. iOS is [customerio-ios#1242](https://github.com/customerio/customerio-ios/pull/1242).

Sample app only — no SDK files touched, `make validate-public-api` clean.

Field drives currently leave almost no analysable record: Logcat is a ring buffer that a multi-hour drive overruns. This installs a log dispatcher that mirrors every SDK record into `cio-diagnostics` under the app's external files directory as NDJSON, with a device-state snapshot on each record.

The output is the input for the rest of the geofence testing strategy — see [Testing Trips Without Taking Them](https://claude.ai/code/artifact/9a7cda78-cf55-46fb-b452-ff13b338a56c) for the overview and [Testing Without Driving](https://claude.ai/code/artifact/7394e40e-e61e-4cac-9542-cf5336da76f0) for the engineering detail. Full spec in `geofence-logging-pr-plan.md`.

**Schema-identical to the iOS sink**, deliberately. The same off-device tooling reads both, and a per-platform dialect would double every analysis script.

## What it does

- Installs `setLogDispatcher` as the **first** statement of `Application.onCreate`. Every cold background wake runs `onCreate` first and reaches SDK code within milliseconds.
- **Forwards to Logcat first.** The SDK dispatches as `logDispatcher?.invoke(...) ?: <logcat>`, so a dispatcher that does not forward silently empties Logcat for everyone else using this app.
- Envelope per record: `v seq ts mono pid up src tag lvl msg dev`. `elapsedRealtimeNanos` for `mono` — it keeps counting through deep sleep, which is exactly the interval a backgrounded phone spends between geofence wakes.
- `[Tag]` prefix lifted into its own field; `msg` otherwise holds the complete original string, including the ` || key=value` tail MBL-2323 adds.
- File header carries `boot` (derived, comparable with iOS) **and** `bootCount` from `Settings.Global.BOOT_COUNT`, which is exact.
- Device state from **OS-pushed broadcasts only — no timers, no polling, never a Wi-Fi scan.** A repeating wakeup pulls the phone out of Doze and improves the very behaviour we are measuring. The one read-on-demand value is the standby bucket, which has no broadcast; it is a local binder call that schedules nothing.
- Rotates **daily, not per launch** — background relaunches would otherwise shatter one drive across a dozen files. Retention 14 files / 100 MB, never wiped on launch.
- Three ways off the device: share sheet on the Location Test screen (FileProvider scoped to the diagnostics directory only), `adb pull` with no `run-as`, and MTP over USB.

Parsing is deliberately **not** on-device. One parser, off-device, fixable after the fact and re-runnable over files already captured.

## Two behaviours worth flagging in review

**SDK log level is forced to `DEBUG`.** The SDK filters by level *before* the dispatcher runs, and `CustomerIO.initialize` re-applies the configured level, so a stored level of ERROR would produce an empty file after a three-hour drive. This overrides the Settings screen's log level; the Location Test screen says so on screen.

**`getExternalFilesDir` rather than internal storage.** No permission needed and it survives `adb pull` without `run-as` — deliberately relaxed, because this is a sample app whose whole job is producing diagnostics.

## Verified on an emulator (API 35)

386 records across two launches in one file, every line valid JSON. `Geofence`, `Init`, `SSE`, `Polling`, `Push`, `CIO-Inbox` tags all parsed; 151 debug / 19 info. Append-across-relaunch confirmed by two `session.start` records with distinct `seq=1` in the same file, so process death stays visible inline. `fg` tracks a home-key press correctly.

```json
{"v":1,"ev":"file.open","ts":"2026-08-31T20:07:58.779+04:00","boot":"2026-08-30T15:49:58.244+04:00",
 "bootCount":50,"device":{"model":"Google sdk_gphone64_arm64","os":"Android","osVersion":"15 (API 35)"},
 "app":{"id":"io.customer.android.sample.java_layout","version":"1.0","build":"1"},
 "sdk":{"version":"4.20.2"},"tz":"Asia/Dubai"}
```

One fix came out of running it: the first ~26 records of every session carried `batt=-1` because `ACTION_BATTERY_CHANGED` is sticky and lands just after registration. Now seeded with a null-receiver read at startup, which returns the current value immediately and costs no wakeup.

## Known limitation, recorded not solved

The SDK passes only level and message to the dispatcher, never the `Throwable` (`LoggerImpl.logIfMatchesCriteria`). Stack traces on error records therefore cannot reach the file. Noted in the code.

---

## Review fixes folded in

These were raised against this PR but had been committed on the stacked enrichment branch. They now live here, with the code they change.

- **The session header repeats on every open**, not only when the file is new. A same-day reboot or relaunch reuses the file, resets `elapsedRealtime` and changes `bootCount`, so everything appended afterwards was correlated to the wrong boot. Retention still runs only when a file was actually created — the failure paths return before `rolloverAt` is set, so putting it there would re-enter on every append.
- **Retention now runs on the midnight rollover**, not only at process start.
- **Size eviction no longer keeps an older file after deleting a newer one.** With 70 MB / 40 MB / 5 MB newest-first it kept 70, deleted 40, then kept 5. A latch now removes the first file that exceeds the remaining budget and everything older.
- **The no-op screen receiver is gone.** `update("screen") { it }` left the snapshot unchanged, so the callback never fired and `Snapshot` had no screen field either way.
- **`standbyBucket()` is cached with a 60 s TTL.** It was making a `UsageStatsManager` binder call for every record while the sink lock was held, which under Doze could serialize SDK logging behind IPC — perturbing the very path being measured.
- **The battery sentinel** reports `null` rather than a magic value when the level is unknown.
- **The app's own records use the same `||` delimiter as the SDK's tail.** `session.start` and `device.state` carried `key=value` inline in the prose, so the off-device parser — which splits on the last `||` and accepts the remainder only if it is entirely `key=value` — never yielded their fields.
- **Dropped an `@OptIn(InternalCustomerIOApi::class)`** for a marker nothing in the file needs any more.

Known and not addressed here: an open failure is not retried until the next midnight rollover; `application/x-ndjson` filters out some share targets; and the pure `prune` / `splitTag` / `json` helpers have no unit tests, which is how the eviction ordering bug survived.

### DIFF
--- a/samples/java_layout/src/main/java/io/customer/android/sample/java_layout/diagnostics/DiagnosticDeviceState.kt
+++ b/samples/java_layout/src/main/java/io/customer/android/sample/java_layout/diagnostics/DiagnosticDeviceState.kt
@@ -1,6 +1,5 @@
 package io.customer.android.sample.java_layout.diagnostics
 
-import android.app.Activity
 import android.app.Application
 import android.app.usage.UsageStatsManager
 import android.content.BroadcastReceiver
@@ -11,9 +10,11 @@ import android.net.ConnectivityManager
 import android.net.Network
 import android.net.NetworkCapabilities
 import android.os.Build
-import android.os.Bundle
 import android.os.PowerManager
 import android.os.SystemClock
+import androidx.lifecycle.DefaultLifecycleObserver
+import androidx.lifecycle.LifecycleOwner
+import androidx.lifecycle.ProcessLifecycleOwner
 
 /**
  * The `dev` block attached to every record.
@@ -33,7 +34,6 @@ internal class DiagnosticDeviceState(private val application: Application) {
     private val lock = Any()
     private var snapshot = Snapshot()
     private var onChange: ((String) -> Unit)? = null
-    private var startedActivities = 0
 
     @Volatile
     private var bucketCache = "unknown"
@@ -139,27 +139,19 @@ internal class DiagnosticDeviceState(private val application: Application) {
     }
 
     /**
-     * Foreground state from activity callbacks rather than `ProcessLifecycleOwner`, so the sink
-     * needs no extra dependency and behaves the same on every supported API level.
+     * Foreground state from `ProcessLifecycleOwner`, which is what the SDK itself observes.
+     *
+     * Counting started activities looks equivalent and is not: the count drops to zero and back
+     * during a configuration change, so a rotation produced a spurious background/foreground pair
+     * in the trace. ProcessLifecycleOwner debounces exactly that.
      */
     private fun registerLifecycleCallbacks() {
-        application.registerActivityLifecycleCallbacks(object : Application.ActivityLifecycleCallbacks {
-            override fun onActivityStarted(activity: Activity) {
-                startedActivities += 1
-                update("fg") { it.copy(foreground = startedActivities > 0) }
+        ProcessLifecycleOwner.get().lifecycle.addObserver(
+            object : DefaultLifecycleObserver {
+                override fun onStart(owner: LifecycleOwner) = update("fg") { it.copy(foreground = true) }
+                override fun onStop(owner: LifecycleOwner) = update("fg") { it.copy(foreground = false) }
             }
-
-            override fun onActivityStopped(activity: Activity) {
-                startedActivities = (startedActivities - 1).coerceAtLeast(0)
-                update("fg") { it.copy(foreground = startedActivities > 0) }
-            }
-
-            override fun onActivityCreated(activity: Activity, savedInstanceState: Bundle?) = Unit
-            override fun onActivityResumed(activity: Activity) = Unit
-            override fun onActivityPaused(activity: Activity) = Unit
-            override fun onActivitySaveInstanceState(activity: Activity, outState: Bundle) = Unit
-            override fun onActivityDestroyed(activity: Activity) = Unit
-        })
+        )
     }
 
     private fun registerThermalListener() {
--- a/samples/java_layout/src/main/java/io/customer/android/sample/java_layout/diagnostics/DiagnosticDeviceState.kt
+++ b/samples/java_layout/src/main/java/io/customer/android/sample/java_layout/diagnostics/DiagnosticDeviceState.kt
@@ -13,26 +13,15 @@ import android.net.NetworkCapabilities
 import android.os.Build
 import android.os.Bundle
 import android.os.PowerManager
+import android.os.SystemClock
 
 /**
  * The `dev` block attached to every record.
- *
- * Two sessions over the same route are only comparable if you know what state the phone was in for
- * each. Doze, standby bucket, power-save mode, thermal pressure and whether Wi-Fi was up all change
- * geofence behaviour materially — geofence monitoring is a *network-location* feature, so "was
- * there Wi-Fi" is frequently the whole explanation for a session that behaved unlike its neighbour.
- *
- * **Everything here is pushed by the OS. Nothing is polled.** No timer, no periodic sampler, and
- * never a Wi-Fi scan. A repeating wakeup is a wakeup the device would not otherwise have taken: it
- * pulls the phone out of Doze and improves the very behaviour we are trying to measure. An observer
- * that changes what it observes is worse than no observer.
- *
- * The one exception is the standby bucket, which has no broadcast and is read on demand — a purely
- * local query against `UsageStatsManager` that schedules nothing.
  */
 internal class DiagnosticDeviceState(private val application: Application) {
     private data class Snapshot(
-        val batteryLevel: Float = -1f,
+        /** `null` until a battery broadcast has been seen; never a sentinel. */
+        val batteryLevel: Float? = null,
         val charging: Boolean = false,
         val powerSave: Boolean = false,
         val idle: Boolean = false,
@@ -45,6 +34,12 @@ internal class DiagnosticDeviceState(private val application: Application) {
     private var snapshot = Snapshot()
     private var onChange: ((String) -> Unit)? = null
     private var startedActivities = 0
+
+    @Volatile
+    private var bucketCache = "unknown"
+
+    @Volatile
+    private var bucketReadAt = 0L
 
     private val powerManager: PowerManager? =
         application.getSystemService(Context.POWER_SERVICE) as? PowerManager
@@ -69,7 +64,9 @@ internal class DiagnosticDeviceState(private val application: Application) {
         val current = synchronized(lock) { snapshot }
         return buildString(160) {
             append('{')
-            append("\"batt\":").append(String.format(java.util.Locale.US, "%.2f", current.batteryLevel))
+            append("\"batt\":").append(
+                current.batteryLevel?.let { String.format(java.util.Locale.US, "%.2f", it) } ?: "null"
+            )
             append(",\"charging\":").append(current.charging)
             append(",\"powerSave\":").append(current.powerSave)
             append(",\"idle\":").append(current.idle)
@@ -92,7 +89,7 @@ internal class DiagnosticDeviceState(private val application: Application) {
                         val scale = intent.getIntExtra("scale", -1)
                         val status = intent.getIntExtra("status", -1)
                         current.copy(
-                            batteryLevel = if (level >= 0 && scale > 0) level.toFloat() / scale else -1f,
+                            batteryLevel = if (level >= 0 && scale > 0) level.toFloat() / scale else null,
                             charging = status == android.os.BatteryManager.BATTERY_STATUS_CHARGING ||
                                 status == android.os.BatteryManager.BATTERY_STATUS_FULL
                         )
@@ -101,8 +98,6 @@ internal class DiagnosticDeviceState(private val application: Application) {
                         update("powerSave") { it.copy(powerSave = powerManager?.isPowerSaveMode == true) }
                     PowerManager.ACTION_DEVICE_IDLE_MODE_CHANGED ->
                         update("idle") { it.copy(idle = isDeviceIdle()) }
-                    Intent.ACTION_SCREEN_ON, Intent.ACTION_SCREEN_OFF ->
-                        update("screen") { it }
                 }
             }
         }
@@ -111,8 +106,6 @@ internal class DiagnosticDeviceState(private val application: Application) {
             // Sticky broadcast: registering delivers the current value immediately, without
             // costing a wakeup or a poll.
             addAction(Intent.ACTION_BATTERY_CHANGED)
-            addAction(Intent.ACTION_SCREEN_ON)
-            addAction(Intent.ACTION_SCREEN_OFF)
             if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
                 addAction(PowerManager.ACTION_POWER_SAVE_MODE_CHANGED)
             }
@@ -193,7 +186,7 @@ internal class DiagnosticDeviceState(private val application: Application) {
 
         synchronized(lock) {
             snapshot = snapshot.copy(
-                batteryLevel = if (level >= 0 && scale > 0) level.toFloat() / scale else -1f,
+                batteryLevel = if (level >= 0 && scale > 0) level.toFloat() / scale else null,
                 charging = status == android.os.BatteryManager.BATTERY_STATUS_CHARGING ||
                     status == android.os.BatteryManager.BATTERY_STATUS_FULL,
                 powerSave = powerManager?.isPowerSaveMode == true,
@@ -226,13 +219,18 @@ internal class DiagnosticDeviceState(private val application: Application) {
 
     /**
      * Which app-standby bucket the OS has us in.
-     *
-     * No broadcast exists for this, so it is read when a record is written rather than cached from
-     * a change notification. The read is a local binder call that schedules nothing — the rule this
-     * class follows is "cause no wakeups", not "never call a system service".
      */
     private fun standbyBucket(): String {
         if (Build.VERSION.SDK_INT < Build.VERSION_CODES.P) return "unsupported"
+        val now = SystemClock.elapsedRealtime()
+        if (bucketReadAt != 0L && now - bucketReadAt < BUCKET_TTL_MS) return bucketCache
+        val value = readStandbyBucket()
+        bucketCache = value
+        bucketReadAt = now
+        return value
+    }
+
+    private fun readStandbyBucket(): String {
         return runCatching {
             val manager = application.getSystemService(Context.USAGE_STATS_SERVICE) as? UsageStatsManager
             when (manager?.appStandbyBucket) {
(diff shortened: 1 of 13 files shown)


samples/java_layout/src/main/java/io/customer/android/sample/java_layout/diagnostics/DiagnosticLogWriter.kt
```diff
@@ -0,0 +1,162 @@
...
+        val isNew = !file.exists()
+
+        stream = runCatching { FileOutputStream(file, /* append = */ true) }.getOrNull()
+        if (stream == null) return
+
+        currentFile = file
+        rolloverAt = startOfNextDay(now)
+        writesSinceExistenceCheck = 0
+
+        if (isNew && header.isNotEmpty()) write(header)
```
Shahroz16: Could we write the header on every process session, not only when today's file is new? A same-day reboot reuses the file, resets `elapsedRealtime`, and changes `bootCount`, but the only boot metadata remains from the first boot. Everything appended afterward is correlated to the wrong boot/build. Repeating the NDJSON `file.open` header when the file is reopened would make each session self-describing.

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/diagnostics/DiagnosticDeviceState.kt
```diff
@@ -0,0 +1,285 @@
...
+                            charging = status == android.os.BatteryManager.BATTERY_STATUS_CHARGING ||
+                                status == android.os.BatteryManager.BATTERY_STATUS_FULL
+                        )
+                    }
+                    PowerManager.ACTION_POWER_SAVE_MODE_CHANGED ->
+                        update("powerSave") { it.copy(powerSave = powerManager?.isPowerSaveMode == true) }
+                    PowerManager.ACTION_DEVICE_IDLE_MODE_CHANGED ->
+                        update("idle") { it.copy(idle = isDeviceIdle()) }
+                    Intent.ACTION_SCREEN_ON, Intent.ACTION_SCREEN_OFF ->
+                        update("screen") { it }
```
Shahroz16: These broadcasts currently do no observable work: `update("screen") { it }` leaves the snapshot unchanged, so `update()` never invokes the callback, and `Snapshot` has no screen field. Could we either store and emit `screenOn`, or remove these two actions and the branch?

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/diagnostics/DiagnosticLogWriter.kt
```diff
@@ -0,0 +1,162 @@
...
+    private fun closeCurrentFile() {
+        runCatching { stream?.close() }
+        stream = null
+    }
+
+    fun append(line: String) {
+        synchronized(lock) {
+            val now = System.currentTimeMillis()
+            if (now >= rolloverAt) {
+                openCurrentFile(now)
```
Shahroz16, line 88: Retention is pruned only from `open()`. A process that stays alive across midnight keeps creating daily files here without applying the 14-file/100 MB policy until the next process start. Could we run `prune()` after opening the rollover file?

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/diagnostics/DiagnosticLogWriter.kt
```diff
@@ -0,0 +1,162 @@
...
+     */
+    private fun prune() {
+        var kept = 0
+        var total = 0L
+        for (file in files()) {
+            val withinCount = kept < MAX_FILES
+            val withinSize = total + file.length() <= MAX_TOTAL_BYTES
+            // The newest file is always kept, however large — it may be the drive nobody has
+            // pulled off the device yet.
+            if (kept == 0 || (withinCount && withinSize)) {
```
Shahroz16: This can delete a newer file and then keep an older one. For example, with 70 MB, 40 MB, and 5 MB newest-first, it keeps 70, deletes 40, then keeps 5. That conflicts with "evict oldest" and creates holes in the retained history. Could we latch eviction once the first file exceeds the remaining budget, so that file and everything older are removed?

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/diagnostics/DiagnosticDeviceState.kt
```diff
@@ -0,0 +1,285 @@
...
+     */
+    fun snapshotJson(): String {
+        val current = synchronized(lock) { snapshot }
+        return buildString(160) {
+            append('{')
+            append("\"batt\":").append(String.format(java.util.Locale.US, "%.2f", current.batteryLevel))
+            append(",\"charging\":").append(current.charging)
+            append(",\"powerSave\":").append(current.powerSave)
+            append(",\"idle\":").append(current.idle)
+            append(",\"bucket\":").append(DiagnosticEnvelope.json(standbyBucket()))
```
Shahroz16, line 73: `snapshotJson()` says it only reads cached values, but `standbyBucket()` does a `UsageStatsManager` binder call for every log record, while the global sink lock is held by the caller. Under Doze or system load this can serialize SDK logging behind IPC and perturb the geofence path being measured. Could we cache the bucket with a short TTL or read it outside the sink lock?

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/diagnostics/DiagnosticLogWriter.kt
mahmoud-elmorabea: Fixed. The header now writes on every `open()` rather than only when the file is new, so a same-day relaunch re-describes itself. Retention stays behind `isNew` — the failure paths return before `rolloverAt` is set, so pruning outside that guard would re-enter on every append.

Verified on a Galaxy S24: three sessions in one file, each with its own header and pid.

One gap left, which you may have a view on: a mid-process midnight rollover rewrites the header string captured at process start, so its `ts` is stale even though `boot`/`bootCount`/versions are still correct. Making it a provider would close that.

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/diagnostics/DiagnosticDeviceState.kt
mahmoud-elmorabea: Removed — both actions and the branch. You were right that nothing observable happened.

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/diagnostics/DiagnosticLogWriter.kt
mahmoud-elmorabea: Fixed. `append()` re-enters `openCurrentFile` at the rollover, and the new day's file is `isNew`, so retention runs there now.

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/diagnostics/DiagnosticLogWriter.kt
mahmoud-elmorabea: Fixed with the latch you described: once a file exceeds the remaining budget, that file and everything older go. The newest file is still kept whatever its size.

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/diagnostics/DiagnosticDeviceState.kt
mahmoud-elmorabea: Fixed — 60s TTL cache, so the binder call no longer happens per record under the sink lock.

While confirming this on hardware I also picked up the battery sentinel you flagged separately: `batt` now reports a real level on a device (0.69 on a Galaxy S24, 0.75 on an iPhone 15) and `null` where the platform has no battery, instead of the old -1.0.

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/diagnostics/DiagnosticLog.kt
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    /**
+     * `getExternalFilesDir` rather than internal storage: no permission is required, the directory
+     * survives `adb pull` without `run-as`, and it is visible over MTP — three independent ways to
+     * get a drive off the phone instead of one.
+     */
+    @JvmStatic
+    fun directory(application: Application): File =
+        File(application.getExternalFilesDir(null) ?: application.filesDir, DIRECTORY_NAME)
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Valid — fixed in b9c1a295.

Accurate about the missing root. One correction on the impact: `share()` already wraps `getUriForFile` in `runCatching`, so the outcome is the "Could not share the diagnostic log" toast rather than an exception, and logging and `adb pull` are unaffected. A quiet failure rather than a loud one — which is what makes it worth fixing, not what makes it urgent, since `getExternalFilesDir` returns null only with external storage unmounted and primary storage is emulated on any modern device.

Added a `files-path` root beside the external one, with a distinct `name` since FileProvider requires unique names.

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/diagnostics/DiagnosticEnvelope.kt
```diff
@@ -0,0 +1,152 @@
...
+
+    private val formatter = ThreadLocal.withInitial {
+        SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS", Locale.US)
+    }
+
+    /**
+     * ISO 8601 with milliseconds and a local UTC offset.
+     */
+    fun iso8601(millis: Long): String {
+        val base = formatter.get()!!.format(Date(millis))
```
Shahroz16, line 114: The formatter is created with the default timezone once, while the offset below is read from `TimeZone.getDefault()` on every record. If the device changes timezone while this process remains alive, the wall-clock text stays in the old zone but gets paired with the new offset, so the timestamp represents the wrong instant. `dayFormat` in the writer has the same stale-zone behavior for filenames. Could we derive the text and offset from one current timezone value per write, update the filename formatter too, and add a test that changes `TimeZone.setDefault()` between writes?

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/diagnostics/DiagnosticLogWriter.kt
```diff
@@ -0,0 +1,163 @@
...
+            val withinCount = kept < MAX_FILES
+            val withinSize = total + file.length() <= MAX_TOTAL_BYTES
+            // The newest file is always kept, however large — it may be the drive nobody has
+            // pulled off the device yet.
+            if (!budgetSpent && (kept == 0 || (withinCount && withinSize))) {
+                kept += 1
+                total += file.length()
+            } else {
+                budgetSpent = true
+                file.delete()
```
mrehan27: **P2** — this can unlink the file the writer holds open. Given a future-dated `cio-diag-` file over the 100 MB budget (clock set forward, or a capture copied back over MTP), `files()` sorts it first, `kept == 0` keeps it, and today's new 0-byte file then fails `withinSize` and is deleted. The stream stays writable, so up to 255 records go nowhere until the existence check reopens. The iOS twin guards this with `if url.path != currentPath`.

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/diagnostics/DiagnosticLogWriter.kt
```diff
@@ -0,0 +1,163 @@
...
+    fun files(): List<File> =
+        (directory.listFiles { file -> file.name.startsWith(FILE_PREFIX) } ?: emptyArray())
+            // Names are `cio-diag-YYYY-MM-DD`, so lexical order is chronological order and no
+            // filesystem timestamp has to be trusted.
+            .sortedByDescending { it.name }
+
+    /**
+     * Enforces the retention policy by evicting the **oldest** files.
+     */
+    private fun prune() {
```
mrehan27, line 147: **P2** — `prune`, `splitTag`, `json` and `iso8601` are pure and need no device, but have no tests, and `prune` alone has now produced three ordering bugs found only in review: the eviction latch, retention on rollover, and the open-file delete above. A few cases over a temp dir would have caught all three and would pin the retention contract going forward.

mrehan27 (approved):
Reviewed alongside customerio-android#831 and the iOS pair. Schema parity with iOS holds, and last round's fixes all landed — the envelope now reads one timezone for both halves of the timestamp.

Two items inline: `prune()` can unlink the file it just opened, which the iOS twin guards against and this doesn't, and the pure helpers still have no tests.

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/diagnostics/DiagnosticEnvelope.kt
Shahroz16: The formatter half is fixed, but `rolloverAt` is still computed only when the file opens. After an eastward timezone change, `append()` keeps comparing against midnight in the old zone and continues writing to the previous day file until that boundary. Could we store the zone ID used for the current file and reopen or recalculate when `TimeZone.getDefault().id` changes? A New York to London test would cover the field-drive case.

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/diagnostics/DiagnosticLogWriter.kt
```diff
@@ -0,0 +1,184 @@
...
+
+    private fun closeCurrentFile() {
+        runCatching { stream?.close() }
+        stream = null
+    }
+
+    fun append(line: String) {
+        synchronized(lock) {
+            val now = System.currentTimeMillis()
+            if (now >= rolloverAt || TimeZone.getDefault().id != openZoneId) {
```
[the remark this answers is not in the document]
Shahroz16: I agree this can lose the rest of a field drive. After the failed reopen, `stream` is null but `openZoneId` and `rolloverAt` still describe the old file; if the device returns to that zone, neither reopen condition fires and every write becomes a no-op. Could we retry whenever `stream == null`, or clear the open-file state on failure, and add a focused failed-reopen test?

Shahroz16 (approved):
Looks good for the tester workflow. The session headers, retention and timezone fixes are in place. I’m treating the failed-reopen plus timezone-revert case as optional sample-app hardening, consistent with the iOS discussion.

samples/java_layout/src/main/java/io/customer/android/sample/java_layout/diagnostics/DiagnosticLogWriter.kt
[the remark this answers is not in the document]
Shahroz16: Given the same tester-only scope discussed on the iOS PR, I’m happy to leave this as optional hardening. It needs a failed reopen followed by a timezone switch back, and doesn’t affect SDK delivery.

PR #831, opened 2026-08-31
chore(geofence): attach structured key=value diagnostics to geofence logs

mahmoud-elmorabea:
[MBL-2323: Enrich geofence logs with a structured key=value tail](https://linear.app/customerio/issue/MBL-2323/enrich-geofence-logs-with-a-structured-keyvalue-tail)

iOS half is [customerio-ios#1243](https://github.com/customerio/customerio-ios/pull/1243).

**Stacked on #830** — targets `mbl-2322-android-log-sink`, retargets to `main` when that merges. #830 captures what the SDK says; this makes the SDK say enough. All sample-app changes live in #830; this PR is `geofence/` only.

## What a customer build sees

**No diagnostic value is ever computed or emitted unless the gate is on.** Coordinates, fix accuracy, region identifiers, rankings, durations — all of it sits behind one check at the single place a tail is constructed, so a field added later needs no privacy review of its own. The expensive reads are gated too: deserializing the cached region list to log a count would otherwise land on every background wake.

Two things this does **not** claim, both corrected from an earlier revision of this description:

- **Output is not byte-identical to `main`.** Every log line that existed before is unchanged, but this PR adds roughly nine new prose lines that emit whether or not the gate is on — a sync being skipped, a module waking, a registration being applied. They are ungated on purpose: they are useful in a normal debug log and carry no value the tail exists to protect. The contract test asserts the *prose half is identical between the two gate states*, which is the property that actually matters.
- **One pre-existing line changed meaning.** `logSyncSucceeded` now counts what is *monitored* rather than what was *requested* — if a removal fails, the orphaned region is still registered with the OS and is now counted. `GeofenceRepositoryTest` pins the new semantics.

## The gate

A manifest `meta-data` entry in the host app, read once through `PackageManager`:

```xml
<meta-data android:name="io.customer.geofence.diagnostics" android:value="true" />
```

Deliberately not an API. Anything reachable from application code — even behind an opt-in annotation — is something a customer app can switch on, and the tail is the main thing keeping coordinates out of a production log. A manifest entry cannot be called, cannot appear in autocomplete, and cannot be reached by a wrapper SDK. Our sample app sets it; nothing else does. `apiCheck` produces no diff because nothing public changed.

*(Caveat worth stating: `logSyncSkippedInvalidLocation` already prints coordinates ungated on `main`. This PR does not make that worse, but the gate is not the only path a coordinate can reach a log today.)*

## What the tail carries when it is on

```
Geofence sync skipped (app-launch): no location available || ev=sync.skipped io=out why=no_location ctx=app_launch
```

`ev=` is the stable machine key (prose is what gets reworded; `ev` is the contract) and `io=in|out|obs` classifies each record for replay. Delivery records are namespaced under `delivery.` so replay can drop the family: what replay checks is that the SDK *emitted* a transition, not that it reached the backend. 34 distinct events.

Silent components given a voice: process start and launch reason — two separate records, `module.init` for what initialized the module and `module.wake` for what actually woke it, deliberately not collapsed because initialization runs first with no launch evidence and the wake arrives milliseconds later carrying the only fact worth having — plus registered region identifiers, the N-of-M ranking decision, movement-trigger geometry, API fetch outcome and duration, storage load, and cooldown remaining on suppression.

Values are sanitized against the format's own separators — whitespace, `=`, `,`, `:`, `|` — because region ids are workspace-authored and one of those characters would otherwise split a field.

## Android has better transition data than iOS

`GeofencingEvent.triggeringLocation` is a **real fix the OS computed for this crossing** — accuracy, speed, bearing, and the mock flag. CoreLocation supplies no position with a geofence event at all, so iOS falls back to its own cached fix. `fixsrc` records the difference (`os_trigger` here, `cached` there), which matters the first time someone compares overshoot across platforms.

| Key | Why |
|---|---|
| `fixsrc` | `os_trigger` / `cached` / `fresh_request` / `none`. |
| `acc`, `vacc` | Accuracy, attached to the thing being measured. |
| `age` | From `elapsedRealtimeNanos`, not `getTime()` — the wall clock steps under NTP, the monotonic one cannot and counts through deep sleep. |
| `sim` | `Location.isMock`. Marks a fix from `adb emu geo fix` or a route driver, so a bench corpus and a drive corpus cannot silently merge. |

Per-transition position is deliberately **not** repeated on `transition.emitted`: it lives on the `os.callback.received` record for the whole broadcast, which carries the ids to join on. Threading the `Location` down to that call site would mean widening `dispatchTransition`, which has **78 test references**, for information already recorded.

## Registration reporting differs from iOS, on purpose

iOS emits one `registration.diff` with add/remove/keep counts; this emits separate `registration.added` and `registration.removed` records, because the underlying OS APIs differ. Both platforms then agree on `registration.applied n=`, with the movement trigger excluded from the count. The parser accepts both shapes.

## Test impact: near zero, against a predicted 20+

The plan expected the prose-to-token conversion on `logSyncSkipped` to be "the single largest breakage point" at 10 references (actually 20 here). It broke nothing:

- **Tokens are derived from the existing prose**, not replaced by an enum, so every `reason: String` signature is untouched. The cost is that a reworded sentence would silently change what analysis groups on — so the contract test **pins all six known `logSyncSkipped` tokens**, converting that into a failing test.
- **New methods instead of widening verified ones.** Registered identifiers go in a new `logRegionsRegisteredIds` rather than into `logSyncSucceeded`, whose five `verify` blocks would all have needed updating. `mockk(relaxed = true)` absorbs new methods free.

`logTransitionSuppressed` gained a parameter and has zero test references; `logSyncSucceeded` gained a defaulted `elapsedMillis`, which its `verify` blocks now pass `any()` for.

## Contract test

`GeofenceLogTailTest` invokes all 58 logger entry points and asserts: with the gate **off**, no record carries a delimiter, a machine key, or any diagnostic value; with it **on**, each record carries `ev=`, a valid `io=`, and its expected keys, and no value contains whitespace. It also asserts the prose half is **identical between the two gate states** — enabling diagnostics must append, never rewrite.

## Verification

**419 geofence + core tests pass.** `ktlint --android` clean, `apiCheck` clean.

Captured end to end through #830's sink on an emulator (API 35), driven from New York to London:

```
Geofence module initialized (app_start) || ev=module.init io=obs launch=app_start
Geofence module woken (boot_restore)    || ev=module.wake io=obs launch=boot_restore
Fetched 50 nearby geofence(s) from the server || ev=api.fetch.result io=in ok=true n=50 ms=1907
Geofence sync succeeded: 19 regions registered + 1 movement trigger || ev=sync.completed io=out n=19 mvmt=true ms=1946
Geofence '4033': device already inside a newly-registered fence — synthesizing ENTER … || ev=transition.synthesized io=out id=4033 t=enter why=initial_enter_inside
Geofence '4033' ENTER: queued for at-least-once delivery … || ev=transition.emitted io=out id=4033 t=enter
Geofence '4033' ENTER: delivered via WorkManager (direct HTTP) … || ev=delivery.sent io=out id=4033 t=enter via=work_manager
```

The `module.init` / `module.wake` split earns itself in the first two lines: initialization ran 32 ms before the wake that carried the actual launch reason. The full registration and delivery path is visible in one file. A browsable rendering of the session, alongside the iOS one, is [here](https://claude.ai/code/artifact/1b1a110b-1e41-4d4e-a8e8-f0b272cc6e9b).

One thing the capture surfaced and this PR does **not** address: the same enter reaches `delivery.sent` several times for one region id. At-least-once delivery permits duplicates, so it may be correct, but the rate is worth a look.


geofence/src/main/kotlin/io/customer/geofence/GeofenceLogger.kt
```diff
@@ -1,131 +1,523 @@
...
+
+    /**
+     * Which regions the OS is actually monitoring, by identifier.
+     *
+     * Counts alone cannot answer the first question anyone asks of a drive that missed a crossing —
+     * "was this geofence even being monitored when I drove through it?" — so the identifiers travel
+     * too. A separate method rather than widening [logSyncSucceeded], whose five `verify` blocks
+     * would all need updating for no benefit.
+     */
+    fun logRegionsRegisteredIds(ids: List<String>, movementTriggerId: String?) {
```
Shahroz16, line 64: The gate only removes the structured tail; this and the other new diagnostic-only methods still emit entirely new records when `CioDiagnostics.enabled` is false. That contradicts the documented and tested "byte-identical to before instrumentation" guarantee, and the test cannot catch it because it calls logger methods directly rather than observing call sites. Could we either gate these methods or call sites as a whole, or narrow the documented guarantee to tail content only?

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
```diff
@@ -183,8 +183,14 @@ internal class GeofenceRepositoryImpl(
...
 
     /** Cache holds regions but none are registered with the OS (e.g. regs lost on sign-out) → re-register. */
-    private fun hasUnregisteredCache(): Boolean =
-        store.getCachedRegions().isNotEmpty() && store.getRegisteredIds().isEmpty()
+    private fun hasUnregisteredCache(): Boolean {
+        val cached = store.getCachedRegions()
+        val registered = store.getRegisteredIds()
+        // Logged here because this is where a cold start first asks what it has to work from —
+        // "was there anything cached at all" is otherwise guesswork when reading a drive.
+        logger.logStorageLoaded(regionCount = cached.size, hasAnchor = registered.isNotEmpty())
```
Shahroz16: Two issues here: this log only runs from the fourth `refreshAction` branch, so stale, ranking-stale, sign-in, and boot-restore paths never emit the `storage.loaded` record it is meant to explain. Also, `hasAnchor` is derived from registered IDs rather than `getLastApiFetchLocation()`, and is necessarily false whenever `hasUnregisteredCache()` returns true. Could we emit this at the actual load or restore entry points and compute the anchor from the stored location?

Shahroz16: This fixes the identify and launch branches, and the anchor value is now correct. Boot restore still calls `restoreFromCache()` without going through `refreshAction()`, though, so it still cannot emit `storage.loaded`. Could we emit the same record from the restore path as well?

geofence/src/main/kotlin/io/customer/geofence/GeofenceLogger.kt
mahmoud-elmorabea: You were right, and it was only half addressed when you last looked — the code comments were scrubbed but the PR description still made the claim. That description is now rewritten.

The guarantee is narrowed to what is actually true: the prose half is identical between the two gate states, which is what the contract test asserts. Roughly nine new prose lines do emit ungated; that is deliberate, since they are useful in a normal debug log and carry no value the tail exists to protect.

Separately, every `GeofenceDiagnostics.isEnabled` check now lives inside `GeofenceLogger`/`GeofenceLogTail` — none in production logic. `tryAcquire` returns the remaining cooldown it already computed rather than having the emitter branch on diagnostics and re-read the store.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
mahmoud-elmorabea: Fixed. One note on the anchor: I used `getLastMovementTriggerLocation()` rather than `getLastApiFetchLocation()`. It is the restore anchor and matches what iOS reports, so `anchor=` means the same thing on both platforms — but say if you would rather it track the fetch location and I will change both.

mahmoud-elmorabea: Fixed. `restoreFromCache()` emits its own `storage.loaded`, since it never enters `refreshAction()`. The count is passed as a lambda and the record gates itself inside the logger, so with diagnostics off the cached region list is never deserialized on that path.

Confirmed on a Galaxy S24: `module.init launch=app_start` at seq 15, then `module.wake launch=boot_restore` at seq 21.

geofence/src/main/kotlin/io/customer/geofence/GeofenceLogTail.kt
```diff
@@ -0,0 +1,194 @@
...
+        io: GeofenceLogIo,
+        fields: List<Pair<String, String?>> = emptyList()
+    ): String {
+        // The single gate for every diagnostic value. Prose is emitted by the caller either way.
+        if (!GeofenceDiagnostics.isEnabled) return ""
+
+        val parts = StringBuilder(DELIMITER).append("ev=").append(ev).append(" io=").append(io.wire)
+        for ((key, value) in fields) {
+            if (value == null) continue
+            parts.append(' ').append(key).append('=').append(foldWhitespace(value))
```
Shahroz16, line 49: `tail()` only folds whitespace for finished values, so plain workspace-authored IDs passed as `id` or `key` can still contain `=`, `,`, `:` or `|` and change the parser structure. `sanitize()` protects list and ranked members, but these single-token fields do not call it. Could we sanitize token fields at their call sites, while leaving deliberately structured values alone, and add a case such as `store,north` or `a=b`?

mrehan27 (approved):
Reviewed alongside customerio-android#830 and the iOS pair. Nothing above P3 from me.

The failing `instrumentation-test (kotlin_compose, 31)` is not from this PR: the same test and exception failed on `mbl-2310-android-expose-error-reason` nine minutes earlier, and #830 — this PR's base — passes that job.

Also checked the new gate for crash safety on a background wake: `SDKComponent.android()` and `getApplicationInfo` are both inside `runCatching`, so an early read returns null instead of throwing.

Shahroz16 (approved):
Looks good. The token escaping, restore logging and diagnostics gate now match the documented contract. The remaining Compose instrumentation failure has the same in-app SDK initialization stack on unrelated #828, so I don’t see it as a blocker for this change.

PR #832, opened 2026-09-01
chore(geofence): qualify a fix before letting it judge containment

mrehan27:
Part of: [MBL-2319](https://linear.app/customerio/issue/MBL-2319/improve-geofence-transition-reliability)

## Stack

- [#824 trust geometry only from live fixes](https://github.com/customerio/customerio-android/pull/824)
- 👉 [#832 qualify a fix before letting it judge containment](https://github.com/customerio/customerio-android/pull/832)
- [#833 polygon wire contract and geometry foundation](https://github.com/customerio/customerio-android/pull/833)

## Problem

#824 decided *which passes* hold a real fix. It did not ask how old that fix is.
The live path uses `getCurrentLocation`, which may answer from cache, and the
fix's timestamp was discarded before its coordinates seeded containment, retired
a reported-ENTER mark or synthesized an initial ENTER. A host-supplied
`Location` can be arbitrarily old and went through the same path. Raised by
@Shahroz16 on #824.

## Change

**A stale fix runs no pass.** `onLocationAcquired` is the single freshness gate: a
fix older than `MAX_LIVE_FIX_AGE_MS` (5 min) keeps the pending sync intent and
triggers nothing, and the repository trusts every fix that reaches
`refreshFromLiveFix` as live. A stamp ahead of now counts as stale too: our own
fixes are monotonic and cannot produce one, so it can only be an untrustworthy
host-supplied time. One gate rather than two, so a fix at the edge of the limit
cannot pass the first check and fail a later one after the intent is spent.

Age is measured on `SystemClock.elapsedRealtime`, so a wall-clock correction
cannot age a fresh fix. The path is `Location.elapsedRealtimeNanos` →
`LocationSnapshot` → `LocationTracker` → `Event.LocationFixAcquired` →
`GeofenceServices`, with `clock.elapsedRealtime()` on the other side.

**`setLastKnownLocation(Location)` forwards the fix time.** It falls back to
`Location.time` mapped onto the monotonic base when `elapsedRealtimeNanos` is
unset, since host-built Locations often stamp only wall clock. The `(lat, lng)`
overload carries no time and stays a bare assertion.

**No accuracy margin.** An earlier revision of this PR gated seeding, synthesis
and mark retirement on the reported horizontal accuracy. All three are gone:

- On seeding and synthesis, requiring the error circle to fit made any fence
  smaller than the fix unjudgable, silencing the initial-ENTER backstop
  indoors, where this path reports 100 m accuracy. Raised by @Shahroz16.
- On retirement, keeping a mark while the device was "not certainly outside"
  kept it when the device was genuinely outside an edited fence. GMS then
  registers the new circle with nothing to fire, and the redundant-ENTER guard
  drops the next real arrival because the mark survived. That reversed
  397ede94 in #824, which had widened this retirement on purpose.

Containment and retirement are both the exact point check from #824.
Accuracy no longer travels through the location module or the event.

## Public API

None. `core/api/core.api` is byte-identical to the base. `LocationAcquired`
keeps its published `(latitude, longitude)` shape; the fix time rides a separate
`LocationFixAcquired` marked `@InternalCustomerIOApi`, which `nonPublicMarkers`
excludes from the dump. Both are published per fix.

## Behaviour notes

The gate accepts the cache ages we measured: `getCurrentLocation` returned the
same fix for up to 124 s indoors, and those still judge containment. A bound
tight enough to demote them would silence the indoor backstop on age alone.
What the gate rejects is the tail beyond anything measured, chiefly a
host-supplied `Location`.

A stale fix keeps the pending sync intent rather than spending it. Spending it on a fix that
cannot judge containment would leave nothing to drive the pass that seeds containment and fires
the initial-ENTER backstop. A later fix discharges it, and if none arrives, foreground entry falls
through to the awaiting-location retry.

## Tests

geofence 440, location 65, core 103, all passing. ktlint, apiCheck and Android
Lint (core, geofence, location) green. Direct tests cover the age-gate
boundary, the future-stamp demotion, the host wall-clock fallback, the
services→repository hop and the exact point checks on seeding and retirement.

### DIFF
--- a/core/src/main/kotlin/io/customer/sdk/communication/Event.kt
+++ b/core/src/main/kotlin/io/customer/sdk/communication/Event.kt
@@ -57,7 +57,7 @@ sealed class Event {
      * [horizontalAccuracyMeters] and [fixTimeMillis] describe the fix itself. Both are null when the
      * source did not report them, typically a host-supplied location.
      */
-    data class LocationAcquired(
+    data class LocationAcquired @JvmOverloads constructor(
         val latitude: Double,
         val longitude: Double,
         val horizontalAccuracyMeters: Double? = null,
--- a/core/src/main/kotlin/io/customer/sdk/communication/Event.kt
+++ b/core/src/main/kotlin/io/customer/sdk/communication/Event.kt
@@ -61,15 +61,14 @@ sealed class Event {
     ) : Event()
 
     /**
-     * The same fix as [LocationAcquired], plus what it can be trusted to resolve, for subscribers
-     * that judge geometry. Null fields mean the source reported none.
-     * [fixElapsedRealtimeMillis] is monotonic since boot, so a clock correction cannot age a fix.
+     * The same fix as [LocationAcquired], plus when it was taken, for subscribers that judge
+     * geometry. [fixElapsedRealtimeMillis] is monotonic since boot, so a clock correction cannot
+     * age a fix; null means the source reported no time.
      */
     @InternalCustomerIOApi
     data class LocationFixAcquired(
         val latitude: Double,
         val longitude: Double,
-        val horizontalAccuracyMeters: Double? = null,
         val fixElapsedRealtimeMillis: Long? = null
     ) : Event()
 
--- a/core/src/main/kotlin/io/customer/sdk/communication/Event.kt
+++ b/core/src/main/kotlin/io/customer/sdk/communication/Event.kt
@@ -53,10 +53,15 @@ sealed class Event {
     /**
      * Published by the location module on every fresh location fix. Other modules
      * subscribe to react to location updates without depending on its internals.
+     *
+     * [horizontalAccuracyMeters] and [fixTimeMillis] describe the fix itself. Both are null when the
+     * source did not report them, typically a host-supplied location.
      */
     data class LocationAcquired(
         val latitude: Double,
-        val longitude: Double
+        val longitude: Double,
+        val horizontalAccuracyMeters: Double? = null,
+        val fixTimeMillis: Long? = null
     ) : Event()
 
     class DeleteDeviceTokenEvent : Event()
--- a/core/src/main/kotlin/io/customer/sdk/communication/Event.kt
+++ b/core/src/main/kotlin/io/customer/sdk/communication/Event.kt
@@ -1,5 +1,6 @@
 package io.customer.sdk.communication
 
+import io.customer.base.internal.InternalCustomerIOApi
 import io.customer.sdk.events.Metric
 import java.util.Date
 import java.util.UUID
@@ -53,15 +54,23 @@ sealed class Event {
     /**
      * Published by the location module on every fresh location fix. Other modules
      * subscribe to react to location updates without depending on its internals.
-     *
-     * [horizontalAccuracyMeters] and [fixTimeMillis] describe the fix itself. Both are null when the
-     * source did not report them, typically a host-supplied location.
      */
-    data class LocationAcquired @JvmOverloads constructor(
+    data class LocationAcquired(
+        val latitude: Double,
+        val longitude: Double
+    ) : Event()
+
+    /**
+     * The same fix as [LocationAcquired], plus what it can be trusted to resolve, for subscribers
+     * that judge geometry. Null fields mean the source reported none.
+     * [fixElapsedRealtimeMillis] is monotonic since boot, so a clock correction cannot age a fix.
+     */
+    @InternalCustomerIOApi
+    data class LocationFixAcquired(
         val latitude: Double,
         val longitude: Double,
         val horizontalAccuracyMeters: Double? = null,
-        val fixTimeMillis: Long? = null
+        val fixElapsedRealtimeMillis: Long? = null
     ) : Event()
 
     class DeleteDeviceTokenEvent : Event()
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceFixQuality.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceFixQuality.kt
@@ -1,0 +1,28 @@
+package io.customer.geofence
+
+/**
+ * What a fix can be trusted to resolve, carried alongside the coordinates it describes. Either
+ * field is null when the source did not report it.
+ */
+internal data class GeofenceFixQuality(
+    val accuracyMeters: Double? = null,
+    val fixTimeMillis: Long? = null
+) {
+    /**
+     * How far the device may be from where this fix puts it. Unknown accuracy means no margin: a
+     * host supplying its own location is asserting a position, not measuring one.
+     */
+    val containmentMarginMeters: Double
+        get() = accuracyMeters?.takeIf { it.isFinite() && it > 0.0 } ?: 0.0
+
+    /** Whether the fix is recent enough to still describe where the device is. */
+    fun isFresh(nowMillis: Long): Boolean {
+        val takenAt = fixTimeMillis ?: return true
+        // A future stamp is a clock disagreement, not evidence of freshness.
+        return (nowMillis - takenAt) in 0..GeofenceConstants.MAX_LIVE_FIX_AGE_MS
+    }
+
+    internal companion object {
+        val UNKNOWN = GeofenceFixQuality()
+    }
+}
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceFixQuality.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceFixQuality.kt
@@ -8,12 +8,13 @@ internal data class GeofenceFixQuality(
     val fixElapsedRealtimeMillis: Long? = null
 ) {
     /**
-     * Whether the fix still describes where the device is. Both sides monotonic since boot, so a
-     * stamp ahead of now can only be skew in a supplied value and is treated as current.
+     * Whether the fix still describes where the device is. Both sides monotonic since boot, so our
+     * own fixes can never be stamped ahead of now; one that is came from a host-supplied time that
+     * cannot be trusted to judge geometry, and it ranks and prunes like an anchor instead.
      */
     fun isFresh(nowElapsedRealtimeMillis: Long): Boolean {
         val takenAt = fixElapsedRealtimeMillis ?: return true
-        return nowElapsedRealtimeMillis - takenAt <= GeofenceConstants.MAX_LIVE_FIX_AGE_MS
+        return (nowElapsedRealtimeMillis - takenAt) in 0..GeofenceConstants.MAX_LIVE_FIX_AGE_MS
     }
 
     internal companion object {
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceFixQuality.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceFixQuality.kt
@@ -1,24 +1,19 @@
 package io.customer.geofence
 
 /**
- * What a fix can be trusted to resolve, carried alongside the coordinates it describes. Either
- * field is null when the source did not report it.
+ * When a fix was taken, carried alongside the coordinates it describes. Null when the source did
+ * not report it.
  */
 internal data class GeofenceFixQuality(
-    val accuracyMeters: Double? = null,
     val fixElapsedRealtimeMillis: Long? = null
 ) {
     /**
-     * How far the device may be from where this fix puts it. Unknown accuracy means no margin: a
-     * host supplying its own location is asserting a position, not measuring one.
+     * Whether the fix still describes where the device is. Both sides monotonic since boot, so a
+     * stamp ahead of now can only be skew in a supplied value and is treated as current.
      */
-    val containmentMarginMeters: Double
-        get() = accuracyMeters?.takeIf { it.isFinite() && it > 0.0 } ?: 0.0
-
-    /** Whether the fix still describes where the device is. Both sides monotonic since boot. */
     fun isFresh(nowElapsedRealtimeMillis: Long): Boolean {
         val takenAt = fixElapsedRealtimeMillis ?: return true
-        return (nowElapsedRealtimeMillis - takenAt) in 0..GeofenceConstants.MAX_LIVE_FIX_AGE_MS
+        return nowElapsedRealtimeMillis - takenAt <= GeofenceConstants.MAX_LIVE_FIX_AGE_MS
     }
 
     internal companion object {
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceFixQuality.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceFixQuality.kt
@@ -10,7 +10,10 @@ internal data class GeofenceFixQuality(
     /**
      * Whether the fix still describes where the device is. Both sides monotonic since boot, so our
      * own fixes can never be stamped ahead of now; one that is came from a host-supplied time that
-     * cannot be trusted to judge geometry, and it ranks and prunes like an anchor instead.
+     * cannot be trusted to judge geometry.
+     *
+     * A fix that fails this drives no pass at all: [io.customer.geofence.GeofenceServices] declines
+     * it and leaves the live-fix intent armed, so a later fix still gets to seed containment.
      */
     fun isFresh(nowElapsedRealtimeMillis: Long): Boolean {
         val takenAt = fixElapsedRealtimeMillis ?: return true
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceFixQuality.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceFixQuality.kt
@@ -6,7 +6,7 @@ package io.customer.geofence
  */
 internal data class GeofenceFixQuality(
     val accuracyMeters: Double? = null,
-    val fixTimeMillis: Long? = null
+    val fixElapsedRealtimeMillis: Long? = null
 ) {
     /**
      * How far the device may be from where this fix puts it. Unknown accuracy means no margin: a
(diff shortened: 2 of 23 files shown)


geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
```diff
@@ -438,6 +496,10 @@ internal class GeofenceRepositoryImpl(
         // maxMonitoringDistance, or the distance-capped /nearest returned none here — so an EXIT
         // re-ranks/re-fetches as the device travels. Only maxBusinessGeofences = 0 means "feature off".
         val monitoringEnabled = config.maxBusinessGeofences > 0
+        // A fix locates the device to within its accuracy, not to a point, so "certainly inside"
+        // and "certainly outside" leave an ambiguous band that must neither seed nor retire.
+        val margin = quality.containmentMarginMeters
```
Shahroz16: Could we avoid using the raw low-power accuracy as a hard veto for both containment seeding and synthetic ENTER? The SDK requests `LocationGranularity.DEFAULT`, which maps to `PRIORITY_LOW_POWER` and can return city-level accuracy. A device inside a 150 m fence with an 800 m fix can never satisfy `distance + margin <= radius`, so we neither seed nor fire the fallback that exists when GMS drops `INITIAL_TRIGGER_ENTER`. I think we either need a more accurate geofence fix, or keep the exact point check for inside/synthesis and use the margin only before retiring outside state.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
```diff
@@ -480,16 +545,19 @@ internal class GeofenceRepositoryImpl(
...
-                    val movedAwayIds = reRegisteredIds - insideIds
+                    val insideNow = nearest.filter {
+                        it.distanceTo(latitude, longitude) + margin <= it.radius
+                    }.map { it.id }.toSet()
+                    val outsideNow = nearest.filter {
+                        it.distanceTo(latitude, longitude) - margin > it.radius
+                    }.map { it.id }.toSet()
+                    // Without a record the later genuine EXIT looks unentered and gets dropped, but
+                    // an anchor would seed a place the device left days ago — so it judges nothing.
+                    val insideIds = insideNow.takeIf { fixSource.trustsGeometry }
```
Shahroz16, line 522: When every fence lands in the ambiguous band, `insideNow` is empty but this still passes a non-null empty set for LIVE/MOVEMENT. The store treats that as a decisive "inside nothing" reading and ends the no-record grace. If GMS also misses the initial ENTER, the later EXIT is then dropped as never entered. Could we pass `null` when the fix cannot decide any candidate, and reserve `emptySet()` for a genuinely decisive outside reading?

location/src/main/kotlin/io/customer/location/LocationTracker.kt
```diff
@@ -100,7 +100,12 @@ internal class LocationTracker(
      * Processes an incoming location: caches in memory, persists
      * coordinates, and attempts to send a location track event.
      */
-    fun onLocationReceived(latitude: Double, longitude: Double) {
+    fun onLocationReceived(
```
Shahroz16, line 103: Could we also preserve the quality from `setLastKnownLocation(Location)`? That overload still delegates to the lat/lng overload, so this receives null accuracy/time and the geofence path treats a potentially old, coarse Android `Location` as exact and current. The coordinate-only overload can remain an asserted point, but the `Location` overload should pass `hasAccuracy()/accuracy` and `time`.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
mrehan27: Taken as written, in 91e3f462 — containment is an exact point check again and the margin now applies only before retiring a record.

You are right that it was a veto. Widening the inside test by the fix's own accuracy makes any fence smaller than the fix unjudgable at every distance, so the backstop that exists because GMS drops INITIAL_TRIGGER_ENTER would never fire on this path. Retirement is the one direction where the margin can only remove false confidence, so that is where it stays.

Naming the trade so it is on record rather than assumed: a coarse fix can still seed and synthesize from an exact point check, which is the fabrication risk you originally raised. The age gate still covers the cached-fix half of it — measured ages reach 124s, and those demote to ANCHOR.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
mrehan27: Correct, and this was the worse of the two — I introduced a third state and folded it into an existing one.

#824 made `inside` nullable for exactly two meanings: null is "could not judge", emptySet is "judged, inside nothing". The margin added "judged, decided nothing" and passed it as emptySet, so the store wrote the key, ended the no-record grace and armed the unmatched-EXIT guard on a fix that decided nothing — dropping the later genuine EXIT, which is the failure the nullable parameter existed to prevent.

Fixed by the same commit rather than separately: with the exact inside check restored, an empty set is once again a real reading, so the third state no longer exists and nothing needs threading through `reconcileEnteredIds`. Mutating the call site to stop ending the grace now fails two tests.

location/src/main/kotlin/io/customer/location/LocationTracker.kt
mrehan27: Agreed and done in 91e3f462. I had lumped the two overloads together and only reasoned about the coordinate one.

`setLastKnownLocation(lat, lng)` genuinely has nothing to report, so it stays a bare assertion. The `Location` overload does carry both, and discarding them meant a host fix that is coarse or minutes old was treated as exact and current. It now forwards `accuracy` when `hasAccuracy()` and `time` when it is set, with a test for each direction; dropping either field fails.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
mrehan27: Correcting one claim in my reply above: "measured ages reach 124s, and those demote to ANCHOR" is wrong — `MAX_LIVE_FIX_AGE_MS` is 5 minutes, so every cache age we measured still judges containment. That is deliberate (a bound tight enough to demote them would silence the indoor backstop on age alone), but it means the age gate only rejects the tail beyond anything measured — chiefly host-supplied `Location`s, which can be arbitrarily old. The description now states this accurately. If you'd rather the measured cache ages demote too, that is a one-constant change with the indoor-backstop cost named above.

geofence/src/main/kotlin/io/customer/geofence/GeofenceFixQuality.kt
```diff
@@ -0,0 +1,28 @@
...
+     * host supplying its own location is asserting a position, not measuring one.
+     */
+    val containmentMarginMeters: Double
+        get() = accuracyMeters?.takeIf { it.isFinite() && it > 0.0 } ?: 0.0
+
+    /** Whether the fix is recent enough to still describe where the device is. */
+    fun isFresh(nowMillis: Long): Boolean {
+        val takenAt = fixTimeMillis ?: return true
+        // A future stamp is a clock disagreement, not evidence of freshness.
+        return (nowMillis - takenAt) in 0..GeofenceConstants.MAX_LIVE_FIX_AGE_MS
```
Shahroz16: Could we carry a monotonic fix timestamp or allow a small negative skew here? `fixTimeMillis` comes from `Location.time`, while `nowMillis` is `System.currentTimeMillis()`. Android documents wall-clock location times as subject to clock jumps. Any future-stamped fix, even from a small skew, is demoted until the clocks catch up, disabling containment seeding and the initial-ENTER backstop. `Location.elapsedRealtimeNanos` would make this age check reliable.

core/src/main/kotlin/io/customer/sdk/communication/Event.kt
```diff
@@ -53,10 +53,15 @@ sealed class Event {
     /**
      * Published by the location module on every fresh location fix. Other modules
      * subscribe to react to location updates without depending on its internals.
+     *
+     * [horizontalAccuracyMeters] and [fixTimeMillis] describe the fix itself. Both are null when the
+     * source did not report them, typically a host-supplied location.
      */
-    data class LocationAcquired(
+    data class LocationAcquired @JvmOverloads constructor(
```
Shahroz16: `@JvmOverloads` restores the old constructors, but the data-class `copy(DD)` and old `copy$default` signatures are still removed in `core.api`. A precompiled consumer calling `LocationAcquired.copy(...)` can therefore get `NoSuchMethodError`. Could we keep the original public data-class shape and carry fix quality separately, or otherwise preserve those methods?

geofence/src/main/kotlin/io/customer/geofence/GeofenceFixQuality.kt
mrehan27: Good catch — fixed in 9ca7a838. The whole chain is monotonic now rather than just the comparison: `Location.elapsedRealtimeNanos` → `LocationSnapshot` → `LocationTracker` → the event → `isFresh`, with `clock.elapsedRealtime()` on the other side. A clock jump can no longer age out a fresh fix. This also matches what `PolygonRouteProcessor` already does.

core/src/main/kotlin/io/customer/sdk/communication/Event.kt
mrehan27: You were right, and `@JvmOverloads` was the wrong fix — fixed in 9ca7a838. `LocationAcquired` is back to its published `(latitude, longitude)` shape, so `copy`/`copy$default` are no longer removed. Fix quality moved to a separate `LocationFixAcquired`, marked `@InternalCustomerIOApi` — since that is in `nonPublicMarkers`, `core/api/core.api` is now byte-identical to the base.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
mrehan27: Dropped the margin from retirement as well (f4e36db8). Keeping the mark while the device is "not certainly outside" also keeps it when the device is genuinely outside the edited circle but inside the error band. GMS then registers the new circle with nothing to fire, the next real ENTER is dropped as redundant because the mark survived, and the EXIT after it is delivered because `entered_ids` still carries the fence. That's the swallowed arrival the redundant-ENTER guard exists to prevent. 397ede94 had already widened this retirement for anchor passes; the margin narrowed it again, which is how the swallow returned. Checked against the emitter and store, not reasoned: `isRedundantEnter` sits ahead of the cooldown and `claimExit` succeeds on the carried record.

What the margin bought was suppressing a second ENTER for a device inside both circles when a noisy fix said otherwise, and the cooldown already suppresses that unless the device dwelled past `duplicateEventsExpiry` (1 h default). Both need a geometry edit to arise at all, and a missed arrival is the worse of the two. With no consumer left, accuracy is gone from the event, tracker and receiver; #832 is now the age gate only, and a fix stamped ahead of now still demotes (0edb8ec1). 428 / 65 / 103 tests, apiCheck and lint green.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
```diff
@@ -103,15 +110,30 @@ internal class GeofenceRepositoryImpl(
...
+        // Judged with the epoch above, before the slot wait: both belong to the fix, not the pass.
+        val now = clock.elapsedRealtime()
+        val fixSource = if (quality.isFresh(now)) {
+            FixSource.LIVE
+        } else {
+            logger.logStaleFixDemoted(
+                now - (quality.fixElapsedRealtimeMillis ?: now),
+                GeofenceConstants.MAX_LIVE_FIX_AGE_MS
+            )
+            FixSource.ANCHOR
```
Shahroz16: Could we re-arm the pending live-fix intent when this demotes the fix? `onLocationAcquired` has already cleared both flags before quality is judged. In MANUAL mode, foreground refresh is disabled and `isAwaitingLocation()` is now false, so the next fresh host location is ignored. That leaves the stale anchor as the only pass, with no containment seed or initial-ENTER backstop. A stale-then-fresh MANUAL test would capture this.

mrehan27: Confirmed and fixed in 2b0e3124. Verified the mechanism rather than reasoning about it: `onLocationAcquired` CAS'd both flags to false before `refreshFromLiveFix` judged age, so a stale fix spent the intent and was then demoted — and with foreground retry off in MANUAL mode, `isAwaitingLocation()` was already false when the next host fix arrived.

The intent is now spent only on a fix fresh enough to judge containment; a stale one still drives the anchored pass but leaves the flags armed. Added the stale-then-fresh MANUAL test you suggested, and confirmed it fails if the consume is made unconditional again.

mrehan27: Fixed properly in 37e69db7. A stale fix now keeps the intent and runs nothing, rather than spending it on a pass that is demoted anyway; a later fix discharges it, and in MANUAL mode foreground entry falls through to `retrySyncAwaitingLocation`, which acts on the armed intent. Skipping the pass rather than re-arming around it is what keeps the one-shot guard intact and avoids `triggerSync` clearing the flag — the two problems with my first attempt. Verified with tests on both flag paths (host-refresh and no-location skip), a repeated-stale-fix storm test, and stale-then-fresh; removing the guard or letting it fall through to `triggerSync` each fail all four.

mrehan27: Commit references above are stale after the rebases. The stale-fix behaviour is now `61836ad1`: a fix older than the limit keeps the intent and runs nothing. `53fdfcc6` then made `onLocationAcquired` the only freshness check, removing the repository's second one, so a fix at the edge of the limit can no longer pass the first check and be demoted after the intent is spent. Boundary tests live in `GeofenceServicesTest`; 436 geofence tests, 0 failures.

geofence/src/main/kotlin/io/customer/geofence/GeofenceFixQuality.kt
```diff
@@ -0,0 +1,23 @@
...
+ * not report it.
+ */
+internal data class GeofenceFixQuality(
+    val fixElapsedRealtimeMillis: Long? = null
+) {
+    /**
+     * Whether the fix still describes where the device is. Both sides monotonic since boot, so our
+     * own fixes can never be stamped ahead of now; one that is came from a host-supplied time that
+     * cannot be trusted to judge geometry, and it ranks and prunes like an anchor instead.
+     */
```
Shahroz16, line 17: Could we update this last sentence to match the service behavior? An untrusted fix is ignored and keeps the live-fix intent armed; it doesn't rank or prune as an anchor anymore.

location/src/test/java/io/customer/location/LocationOrchestratorTest.kt
```diff
@@ -66,8 +66,8 @@ class LocationOrchestratorTest {
         orchestrator(LocationTrackingMode.OFF).requestLocation(LocationRequestIntent(tracked = false))
 
         coVerify { provider.requestLocation(LocationGranularity.DEFAULT) }
-        verify { tracker.onLocationReceivedWithoutTracking(37.7749, -122.4194) }
-        verify(exactly = 0) { tracker.onLocationReceived(any(), any()) }
+        verify { tracker.onLocationReceivedWithoutTracking(37.7749, -122.4194, any()) }
```
Shahroz16: Could we pin the actual fix time here instead of `any()`? This silent path feeds the geofence freshness gate, where `null` is treated as fresh. A known snapshot value here, plus a small provider test for the nanos-to-millis conversion, would keep the production handoff covered.

Shahroz16 (approved):
The stale-fix lifecycle issue is addressed. I left two small non-blocking follow-ups for the handoff coverage and KDoc.

geofence/src/main/kotlin/io/customer/geofence/GeofenceFixQuality.kt
mrehan27: Fixed in `5ab3e048` — the doc was left behind by the gate moving to the services layer.

It now says a fix that fails `isFresh` drives no pass at all: `GeofenceServices` declines it and leaves the live-fix intent armed, so a later fix still gets to seed containment. The stale sentence about ranking and pruning as an anchor is gone.

location/src/test/java/io/customer/location/LocationOrchestratorTest.kt
mrehan27: Good catch, and the gap was wider than the `any()` — fixed in `5ab3e048`.

The fixture never set a fix time either, so the whole silent path could have stopped carrying a timestamp with nothing failing. That fails **open**: null reads as current at the freshness gate, so the regression would have opened it rather than closed it. All four call sites now assert a specific value that the fixture supplies.

On the provider test: the conversion had no coverage because it sat inside the GMS success callback, unreachable without a live client. I extracted it to `Location.toSnapshot()` — flagging that as the one production change here, since it is a testability extraction rather than a behaviour fix. It is covered for scale (90 s → 90,000 ms), truncation rather than rounding, and the platform's zero-means-unset, which must stay null rather than becoming 0 ms and aging the fix from boot.

Mutation-checked, four for four: dropping the age in the mapping, passing nanos through unscaled, treating 0 as a real timestamp, and dropping the age on the way to the tracker each fail a named test. That last one is the regression you were pointing at, and it now fails four orchestrator tests instead of none.

location/src/main/kotlin/io/customer/location/LocationServicesImpl.kt
```diff
@@ -96,3 +108,13 @@ internal class LocationServicesImpl(
...
+
+/**
+ * A host-built [Location] often stamps only wall-clock [Location.time]. Fall back to it, mapped onto
+ * the monotonic base, so an old host fix is still aged rather than trusted as current.
+ */
+private fun Location.fixElapsedRealtimeMillis(): Long? =
+    elapsedRealtimeNanos.takeIf { it > 0L }?.let { it / NANOS_PER_MILLI }
+        ?: time.takeIf { it > 0L }?.let { SystemClock.elapsedRealtime() - (System.currentTimeMillis() - it) }
+
+private const val NANOS_PER_MILLI = 1_000_000L
```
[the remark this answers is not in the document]
mrehan27: Fair observation, but the two are deliberately not the same function, and sharing them whole would be a regression rather than a cleanup.

`LocationServicesImpl.fixElapsedRealtimeMillis()` takes a **host-supplied** `Location`, which commonly stamps only wall-clock `time`, so it falls back to that mapped onto the monotonic base. `FusedLocationProvider.toSnapshot()` maps a `Location` from GMS `getCurrentLocation`, which always sets `elapsedRealtimeNanos`. Null there is a real signal — a fix GMS couldn't stamp — and giving that path a wall-clock fallback would trust `Location.time` from a source that has no reason to need it, masking the anomaly instead of reporting it.

What genuinely overlaps is one expression and the constant. Leaving it for now: this is a duplicated expression rather than duplicated behaviour, both copies sit in the same module a few files apart, and it predates this PR (the provider's copy was inline in the GMS callback until it was extracted here for testability). Not worth another CI cycle on an approved change, but worth folding into the next PR that touches this module.

PR #833, opened 2026-09-02
chore(geofence): add polygon wire contract and geometry foundation

mrehan27:

## Stack

- [#824 trust geometry only from live fixes](https://github.com/customerio/customerio-android/pull/824)
- [#832 qualify a fix before letting it judge containment](https://github.com/customerio/customerio-android/pull/832)
- 👉 [#833 polygon wire contract and geometry foundation](https://github.com/customerio/customerio-android/pull/833)

First of four polygon PRs; the remaining three are not open yet.

## What this is

@Shahroz16's polygon foundation, rebased onto the settled base with **authorship preserved** —
his three commits taken as written, plus one scope commit of mine on top:

- `feat: add polygon geofence geometry foundation (MBL-2287)`
- `fix: gate polygon runtime behind explicit support`
- `fix: align polygon foundation with v1 contract`
- `chore(geofence): defer the polygon evaluator to the PR that runs it`

It adds the wire contract (`shape`, `geometry`, `enclosing_circle`), `PolygonGeometry` (ray-cast
containment + per-edge distance on an equirectangular projection), the wake-circle validator, and
`PolygonSupport` so nothing polygon-shaped runs unless the capability is declared. No behaviour
change for circle-only workspaces: absent `capabilities` means circles only.

## Scope

Foundation only — decode, geometry and gating. Nothing registers or evaluates a polygon yet; that
arrives with wake-scoped monitoring. Reviewing this as a self-contained unit is the point of the
split.

`PolygonRouteProcessor` and its two collaborators (`PolygonAccuracyEvaluator`,
`PolygonTransitionStateMachine`) were in an earlier revision of this PR and have been moved to the
monitoring PR: nothing here calls them, so shipping them would have put 336 lines of unreachable
code in a release. Their route-replay tests moved with them. Everything remaining has a caller —
`PolygonGeometry` has six.

## Known gaps, closing in the parity PR

Cross-SDK parity was re-run against iOS's regenerated fixtures today (30 samples, 5 shapes). This
slice is at **23/30**, and all three gaps are already understood and fixed later in the stack:

- **Antimeridian.** `from()` rejects rings crossing +-180, and the ray cast compares raw longitudes
  so containment inverts there. Verified fix in hand: wrap the ray cast's longitudes relative to the
  query point, as `localOffsetMeters` already does for distance, and drop the reject. Takes it to
  **30/30**; the covering fixture (`square400_dateline`) is new from iOS today.
- **Concave rings** rejected by `require(isConvex())` — the spec never required convexity.
- **Duplicate consecutive vertices** rejected rather than collapsed.

They are grouped in the parity PR rather than split across the stack, because they share the
fixtures that pin them.

## Review round

Four findings from @Shahroz16, all valid and all fixed here. Three are the same defect wearing
different clothes: a single malformed polygon took more than its own record with it.

**Nested wire objects rejected the whole response.** `geometry` and `enclosing_circle` decode
before `toDomainRegions()` reaches its per-record try/catch, so a required field there failed the
payload — otherwise-valid circles included. They are nullable now, exactly as `metadata` already
was for the same reason, and a missing field drops only its own record.

**One bad stored vertex wiped the whole cache.** `PolygonCoordinate` validated its range in an
`init` block, which is the deserializer's constructor: the throw landed mid-`List<GeofenceRegion>`,
and `readJson()` then dropped `cached_regions` entirely. The check moved to `PolygonGeometry.from`,
so a ring that no longer validates is rejected at use and the valid regions around it survive.

**The wake-circle check rejected geometry the backend had admitted.** Containment allowed a fixed
1 m tolerance, but a sphere overstates distance by a *fraction* of it, so a vertex sitting exactly
on a valid 1 km circle measured long and dropped the polygon — and the gap widened with the radius.
The tolerance is proportional now (0.5%, comfortably over the ~0.34% by which Earth's radius varies
about the mean). Over-admitting is the safe direction here: the 1 km platform margin absorbs it,
while under-admitting loses the region outright.

**`base_radius_m` is preserved rather than overwritten.** The padded GMS radius used to be the only
radius that survived, which would leave the monitoring PR unable to rank against the backend's own
wake circle. Both are carried now. This is additive — a `baseRadiusMeters` field — rather than the
suggested "keep base in `radius`, pad at registration", because polygons are filtered out of
registration in this PR, so padding logic there would land with no caller.

## Tests

489 geofence / 103 core / 65 location tests, 0 failures. ktlint, apiCheck and `lintDebug` clean.
No public API change.

Geometry keeps direct coverage after the move — concave rejection, boundary points, the GeoJSON
closing vertex, antimeridian, self-intersection, zero area and pole rejection — and inverting the
ray-cast parity fails two of those tests.

Each fix above was mutation-tested: reverting it in isolation fails a named test (the flat
tolerance, each required wire field, the `init` range check, and the dropped base radius). Worth
recording that the first pass of this was wrong — a lethal control showed the harness was fine,
so three mutations were probing the wrong line, and the boundary test placed its vertex with
111,320 m/degree while the validator measures with 111,195, landing it 1 m *inside* the circle it
claimed to sit on. Both corrected.

One change is deliberately not covered: the earth-radius constant moves from 6,400,000 to
`PolygonGeometry`'s 6,371,000, and no test can distinguish them, because the proportional tolerance
scales with the same factor the radius does. It is kept as a consistency fix — the module was
measuring the same geometry two ways, and rounding the radius *up* is the wrong direction for a
containment check — not as a behavioural one.


geofence/src/main/kotlin/io/customer/geofence/api/GeofenceApiResponse.kt
```diff
@@ -80,19 +93,45 @@ internal data class GeofenceApiRegion(
     val metadata: JsonElement? = null
 )
 
+@Serializable
+internal data class GeofenceApiGeometry(
+    @SerialName("type")
+    val type: String,
+    @SerialName("coordinates")
+    val coordinates: JsonElement
```
Shahroz16: Could we make the fields inside `geometry` and `enclosing_circle` tolerant too? These objects are decoded before `toDomainRegions()` reaches its per-record try/catch. A single partial polygon, for example a geometry missing `coordinates`, rejects the entire response along with otherwise valid circles. That defeats the "one bad region costs itself" contract here. Nullable or defaulted nested fields would let `toDomain()` drop only the affected record.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonWakeCircle.kt
```diff
@@ -0,0 +1,77 @@
...
+        val centralAngle = 2.0 * atan2(sqrt(clamped), sqrt(1.0 - clamped))
+        return MAXIMUM_EARTH_RADIUS_METERS * centralAngle
+    }
+
+    internal companion object {
+        // Prototype policy value. Field calibration freezes the versioned production margin.
+        const val PLATFORM_WAKE_MARGIN_METERS = 1_000.0
+        const val MAXIMUM_TRIGGER_RADIUS_METERS = 100_000.0
+        const val CONTAINMENT_TOLERANCE_METERS = 1.0
+        private const val MAXIMUM_EARTH_RADIUS_METERS = 6_400_000.0
```
Shahroz16: Can we use the same WGS84 distance model as the backend here, or give this check a proportional tolerance? Multiplying by a 6,400 km sphere overestimates north/south distance near the equator by about 1%. A vertex exactly on a backend-valid 1 km circle is measured at roughly 1,010 m here, so the fixed 1 m tolerance rejects it and drops the polygon. The mismatch grows with radius.

geofence/src/main/kotlin/io/customer/geofence/api/GeofenceApiResponse.kt
```diff
@@ -164,6 +247,93 @@ internal fun GeofenceApiRegion.toDomain(): GeofenceRegion? {
...
+        logger.logPolygonDropped(id, "backend-provided enclosing circle is invalid or does not contain the polygon")
+        return null
+    }
+    return GeofenceRegion(
+        id = id,
+        name = name,
+        externalId = externalId,
+        latitude = trigger.center.latitude,
+        longitude = trigger.center.longitude,
+        radius = trigger.radiusMeters,
```
Shahroz16, line 297: Should we preserve `base_radius_m` separately instead of replacing it with the padded GMS radius here? V1 slot ranking uses the backend base wake circle, but after this assignment the only retained radius is base + 1,000 and the filter ranks against the exact polygon boundary. The monitoring PR cannot select the intended wake-circle slots from this model. Keeping the base radius in `GeofenceRegion` and applying padding at registration would preserve both meanings.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonGeometry.kt
```diff
@@ -0,0 +1,251 @@
...
+
+@Serializable
+internal data class PolygonCoordinate(
+    @SerialName("latitude")
+    val latitude: Double,
+    @SerialName("longitude")
+    val longitude: Double
+) {
+    init {
+        require(latitude.isFinite() && latitude in -90.0..90.0) { "invalid latitude" }
```
Shahroz16: Could we keep range validation out of the serializable constructor, or make cached-region decoding item-scoped? One out-of-range stored coordinate throws while decoding the whole `List<GeofenceRegion>`, and `readJson()` then removes `cached_regions`. That makes one corrupt polygon take every valid cached circle with it until a fetch succeeds. Letting `PolygonGeometry.fromOrNull()` reject only the bad ring would match the per-region fail-closed behavior.

geofence/src/main/kotlin/io/customer/geofence/api/GeofenceApiResponse.kt
mrehan27: Fixed in `630c8676`.

Both nested objects are nullable throughout now — `type`, `coordinates`, and all three `enclosing_circle` fields — so a partial polygon drops only its own record and the circles around it still map. That puts them on the same footing as `metadata`, which was already tolerant for exactly this reason; the comment there now reads as the rule rather than a one-off.

Three tests pin it: geometry missing `coordinates`, an empty `geometry` object, and `enclosing_circle` without `base_radius_m`, each alongside a valid circle that must survive. Making any of those fields required again fails them.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonWakeCircle.kt
mrehan27: You're right, and the fix is the proportional tolerance — `630c8676`.

A sphere overstates distance by a fraction of it, so a fixed 1 m was always going to fail somewhere: your 1 km example measured ~1004 m and dropped the polygon, and the gap widens with radius. Tolerance is now `max(1 m, baseRadius × 0.5%)`. The 0.5% sits comfortably above the ~0.34% by which Earth's radius varies about the mean, so it covers any latitude. Over-admitting is the safe direction — the 1 km platform margin absorbs it, whereas under-admitting loses the region outright.

I also moved the radius constant from 6,400,000 to `PolygonGeometry`'s 6,371,000, but I want to be straight that **no test covers that half and none can**: the tolerance scales by the same factor the radius does, so the two are indistinguishable at every radius. It's kept as a consistency fix — the module was measuring the same geometry two different ways, and rounding the radius *up* is backwards for a containment check — not as a behavioural one. Say if you'd rather I drop it.

A test places a vertex at 1,004 m against a 1,000 m circle and expects acceptance; restoring the flat tolerance fails it. A second at 1,050 m still rejects, so it stays a tolerance.

geofence/src/main/kotlin/io/customer/geofence/api/GeofenceApiResponse.kt
mrehan27: Agreed on the problem, and fixed in `630c8676` — but I took the other half of your suggestion, so flag it if you disagree.

You're right that the padded radius was the only one surviving, which would leave the monitoring PR ranking against a circle the backend never sent. Rather than keeping the base in `radius` and padding at registration, I added `baseRadiusMeters` to `GeofenceRegion` and left `radius` as the value GMS actually registers.

The reason is that polygons are filtered out of the registered set in this PR (`GeofenceDistanceFilter` drops them unless `isPolygonMonitoringEnabled`), so padding-at-registration would land here with no caller — which is exactly the unreachable code I stripped out of this PR earlier. The additive field gives the monitoring PR both meanings without shipping a code path nothing runs. If you'd rather have the semantics your way, it's a better home in the monitoring PR where registration actually happens.

Removing the field from the mapping fails a test asserting the polygon carries base 100 m and registered 1,100 m.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonGeometry.kt
mrehan27: Fixed in `630c8676`, and it was worse than one bad ring — the `init` block is the deserializer's constructor, so the throw landed mid-`List<GeofenceRegion>` and `readJson()` dropped `cached_regions` wholesale. Every valid cached circle went with it until a fetch succeeded.

Range checking moved to `PolygonGeometry.from`, so `fromOrNull` isolates it the way it already isolates every other ring rule, and the coordinate stays a plain serializable value. `PolygonCoordinate.fromOrNull(lat, lng)` covers the two wire call sites that were doing `runCatching` around the constructor.

Pinned by a store test that writes a cached list with one out-of-range vertex next to a valid circle: the circle survives, and the bad region's `polygonGeometryOrNull()` still returns null, so it's rejected at use rather than by taking the cache down. Putting the `init` block back fails it.

Shahroz16 (approved):
The four issues from the earlier pass are addressed. The wire parsing, wake-circle tolerance, base-radius preservation, and cache isolation now hold up.

geofence/src/main/kotlin/io/customer/geofence/api/GeofenceApiResponse.kt
Shahroz16: This split makes sense. One thing for the monitoring PR: please have the distance filter rank and cap polygons against `baseRadiusMeters` rather than the ring boundary before enabling support. Otherwise a region can be evicted while the device is still inside the backend wake circle. Not blocking this disabled foundation.

PR #837, opened 2026-09-03
chore(geofence): stage transitions durably before they are delivered

mrehan27:

## Stack

- [#835 polygon transition evaluation core](https://github.com/customerio/customerio-android/pull/835)
- [#836 session ownership and atomic transition staging](https://github.com/customerio/customerio-android/pull/836)
- 👉 [#837 stage transitions durably before delivery](https://github.com/customerio/customerio-android/pull/837)
- [#838 polygon location engine](https://github.com/customerio/customerio-android/pull/838)

Three more branches are ready locally and not open yet: the approach monitor + service
controller, the wiring that enables polygon monitoring, and the geofence instrumentation CI
job. Holding them until this half is reviewed.

## What this is

A geofence callback used to acquire a cooldown slot, then persist, then schedule. Three problems
follow from that order, and this changes it:

- a failed outbox write **lost the crossing** — the cooldown was rolled back, but nothing retried
- a delivery that landed after a process death could **commit containment for a user who had since
  signed out**, because nothing checked whose session observed it
- two crossings of the same fence in the same second **collided on the pending key**

Transitions are now staged in the region store first, delivered through one ordered outbox chain,
and committed only against the session generation and geometry revision that observed them. The
cooldown filter splits into `isAllowed` + `record`, so there is nothing to roll back, and the
pending key uses the transition id rather than a timestamp.

`GeofenceBusinessTransitionProcessor` is the new seam: the broadcast receiver hands it an edge and
it owns stage → deliver → commit. Touches `:core`'s pending-delivery store and `:messagingpush`
because the outbox and its flusher are shared.

## Circles only

No polygon behaviour here. This is the delivery path every already-shipped circle geofence uses, so
it is the riskiest PR in the stack and deliberately isolated.

## Tests

992 tests across geofence, core and messagingpush, 0 failures. ktlint, `lintDebug` and `apiCheck`
clean. No public API change.

## Review round

Three fixes from @Shahroz16's review, each reproduced before being changed.

**Routing was never armed.** `saveRoutableRegisteredIds` had no production caller. `beginUserSession`
writes an explicit empty routable set and `getRoutableRegisteredIds` stops falling back once that key
exists, so the first callback after an identify classified every live ID as unknown and removed it
from GMS. Registration now arms routing for the same IDs it saved, guarded by the user-state
generation.

**The session opened too late.** Arming on registration was not enough on its own: `beginUserSession`
had one production caller, the broadcast receiver, and an identify refresh SKIPs on a fresh last-sync
stamp without registering. The first callback then opened the session, wrote the empty set, read it
back in the same dispatch, and pulled every fence. The session now opens at identify, which clears
the last-sync stamp so that refresh takes the remote path and arms routing before any transition can
arrive. The receiver keeps its call for cold starts.

**A refused payload blocked the queue.** Every non-2xx reached the worker as a bare `IOException`, so
a 400 or 401 classified as retryable and its row was resent forever — and because this PR gives every
geofence event one ordered chain draining the oldest row, that blocked all later ENTERs and EXITs.
Responses now carry their status (`HttpRequestFailure`, an `IOException` subclass, so existing
predicates are unaffected) and a terminal one drops the row and continues the drain. Deliberately
narrow: an unexpected exception still keeps the row and retries.

**Store cleanup.** Eleven Android Lint warnings this PR introduced in `GeofenceRegionStore` are gone —
edit sites that discard the result use the KTX `edit(commit = true)` form; three keep an explicit
editor because they return whether the write reached disk, which the KTX form can't express. Also
drops `clearPendingPolygonApproachBatches` (no caller anywhere in the stack) and makes
`hasActiveUserSession` private, since only that file calls it.

**Identify during an in-flight refresh.** An identify landing while another refresh held the slot was
dropped as a duplicate. That pass belongs to the previous session, so it can only refuse to arm
routing for a generation it no longer serves — and `beginUserSession` has already cleared routing, so
nothing was left pending and the next callback removed every live fence. The slot now records which
session its holder serves: a caller sharing that generation still drops (identify and app-launch fire
together on the same anchor), a newer session waits, keeping anchor rather than live-fix semantics.

**Live fences removed while routing was unarmed.** Switching the receiver's orphan guard to the
routable set introduced a window of its own: an identify clears routing and only the completing
refresh re-arms it, so a callback in between saw every live fence as an orphan and removed it from
GMS — and the refresh that followed found those IDs in `registeredIds` with matching params, skipped
them as unchanged, and never re-added them. The guard now splits the unroutable IDs: ones the
registration bookkeeping still claims are dropped but left registered, and only IDs it no longer
claims are evicted.

The movement trigger is exempt and still drives its refresh. It carries no business meaning — routing
it only re-fetches — and it is the only refresh path that runs without the app being opened, so
dropping its EXIT would leave an unarmed session with nothing to re-arm it until the next launch.
That also closes the 30s slot-wait residual: a refused arm now recovers on the next trigger EXIT
rather than waiting for a launch.

**Movement pass didn't stamp the slot.** `refresh` and `refreshFromLiveFix` record which session owns
the refresh slot; `handleMovement` didn't, so while a movement pass held it the generation read as
`NO_SESSION`. A refresh serving that same session couldn't recognise the holder as its own, missed the
duplicate drop, and polled the full slot wait instead. No data loss — it waited and then skipped as
fresh — but the three slot holders now behave the same way.

**A retired fence kept reporting after its removal failed.** The receiver drops a transition for a
fence the backend deleted by looking up its retained definition, but nothing ever wrote one, so that
branch could never be taken. A deleted fence whose OS removal failed stays registered and routable,
and kept emitting business events until a later removal succeeded. The stale definitions are now
persisted when removal fails and cleared when it succeeds. They are merged with what is already
retained rather than rebuilt from the catalog, because the first failure is the last pass that still
sees the deleted fence in the catalog: rebuilding from it alone dropped the tombstone on the very
first retry.

**Session ownership, reworked rather than patched again.** @Shahroz16 found that adopting a legacy
install's persisted session moved the user-state generation, which invalidated a launch refresh
already in flight: that pass saved its registered IDs but skipped the catalog write it gates on the
generation, so a new fence fired with no name and no geoset IDs. Removing the bump fixed that and
broke something else, which he also found: the adopt branch cannot tell "adopting the persisted
user" from "a first session naming someone else", and the bump had been the only thing making a real
switch visible.

Patching it a third time was the wrong move, so the branch is gone. An unowned session opens as a
switch. The switch already kept registered IDs, retained regions and the registration stamps, so live
fences stay registered and keep reporting; they are unarmed until a refresh arms them, and the
movement trigger is exempt from the unarmed drop, so an upgrade with no network still recovers by
re-ranking from the cached catalog. Launch now opens the session before its refresh, as identify
already did, so the switch precedes the pass that has to arm routing for it.

**Launch could undo a concurrent identify.** @Shahroz16 paused launch between reading the identified
user and calling `beginUserSession`, identified a second user in that gap, then let launch resume: it
reopened the older user, clearing the routing the identify's refresh had just armed, and the next
callback was dropped with nothing queued to re-arm. Launch now uses `beginUserSessionIfAbsent`, which
does the owner check and the write under the lock that already serialises this. Identify and the
receiver keep `beginUserSession`, because they are the paths that legitimately redefine a session;
launch only ever needed to establish one.

The cost: launch no longer reconciles a stale owner against the secure user. If a process dies between
an identify's secure write and the geofence handler, the next launch runs under the older owner until
the replayed `UserChangedEvent` switches it, and a callback in that window is dropped as unarmed.

Routing no longer falls back to the registered set when the key is absent. Every session writes the
key, so the fallback was unreachable, and it encoded the policy this change deletes.

That left an offline upgrade unable to recover on its own, which is a regression this PR would have
introduced rather than a cost worth stating: registered fences, empty routing, and no completing pass
to arm them. So a failed remote pass now re-ranks from the cached catalog, the way a failed movement
pass already does. It is deliberately narrow — only when the pass failed, routing is empty *and* a
cache exists — so a session that is already routing never re-registers from a stale cache on a failed
fetch, and `refresh` still returns the remote failure either way.

What this costs, stated rather than buried. On an upgraded install the first OS callback is dropped
as unarmed where adoption would have forwarded it, and one duplicate ENTER is possible after a reboot
because the emitted-ENTER marks are cleared. Both are once per install, and both are the safe
direction: dropped rather than attributed to the wrong user.

A re-rank registers whatever the cache holds, including a fence the backend has deleted since the
last successful fetch. That is not new: every LOCAL pass and every movement re-arm already work that
way, because the cache is authoritative until a REMOTE succeeds, and the next successful one removes
it and tombstones it if the OS removal fails.

Verified: 614 geofence tests here, each new guard mutation-checked so it kills exactly one test and
nothing else — restoring the adopt branch, restoring the routing fallback, deleting the offline
re-rank, and widening its unarmed guard. The upgrade sequence is covered by unit tests against the
real store rather than a stub; none of it has been exercised on a device.
Validated on device as well as in tests: after an identify, 10 fences register, `routable_registered_ids`
matches `registered_ids`, a real ENTER routes and delivers, and no ID is dropped as unknown. The
identify-during-refresh and unarmed-window races are not reachable from the sample app, which only
exposes logout/login (a full reset); those rest on unit tests.

### DIFF
--- a/core/src/main/kotlin/io/customer/sdk/data/store/PendingDeliveryStore.kt
+++ b/core/src/main/kotlin/io/customer/sdk/data/store/PendingDeliveryStore.kt
@@ -74,7 +74,10 @@ class PendingDeliveryStore<T : PendingDeliveryStore.PendingDeliveryEntry>(
     fun appendAll(entries: List<T>): Boolean {
         if (entries.isEmpty()) return true
         return lock.withLock {
-            val all = readAll().toMutableList()
+            // A caller recovering a staged outbox attempt may append the same stable keys again.
+            // Replace those rows atomically instead of duplicating one physical delivery.
+            val incomingKeys = entries.mapTo(mutableSetOf(), PendingDeliveryEntry::key)
+            val all = readAll().filterNot { it.key in incomingKeys }.toMutableList()
             all.addAll(entries)
             while (all.size > maxEntries) {
                 all.removeAt(0)
@@ -93,12 +96,17 @@ class PendingDeliveryStore<T : PendingDeliveryStore.PendingDeliveryEntry>(
      * Remove the entry whose [PendingDeliveryEntry.key] equals [key]. No-op
      * if no such entry exists. Skips the write when no entry was actually
      * removed so a transient read failure cannot wipe the file.
+     *
+     * @return `true` when the store no longer holds [key] — it was removed, or was already absent.
+     * `false` only when the entry is present and the removing write failed, so it is still queued.
+     * A caller that removes to mark work *done* must check this: treating a failed removal as done
+     * lets it re-read the same head entry and repeat the work without bound.
      */
-    fun remove(key: String) {
-        lock.withLock {
+    fun remove(key: String): Boolean {
+        return lock.withLock {
             val entries = readAll()
             val filtered = entries.filterNot { it.key == key }
-            if (filtered.size == entries.size) return@withLock
+            if (filtered.size == entries.size) return@withLock true
             writeAll(filtered)
         }
     }
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
@@ -6,13 +6,12 @@ import android.content.Intent
 import androidx.annotation.VisibleForTesting
 import com.google.android.gms.location.Geofence
 import com.google.android.gms.location.GeofencingEvent
+import io.customer.geofence.di.geofenceBusinessTransitionProcessor
 import io.customer.geofence.di.geofenceLogger
 import io.customer.geofence.di.geofenceManager
 import io.customer.geofence.di.geofenceRegionStore
 import io.customer.geofence.di.geofenceServices
-import io.customer.geofence.di.geofenceTransitionEmitter
 import io.customer.sdk.communication.Event
-import io.customer.sdk.core.di.AndroidSDKComponent
 import io.customer.sdk.core.di.SDKComponent
 import io.customer.sdk.core.di.clock
 import io.customer.sdk.core.di.setupAndroidComponent
@@ -21,8 +20,6 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.sync.Mutex
-import kotlinx.coroutines.sync.withLock
 import kotlinx.coroutines.withTimeoutOrNull
 
 /** Receives OS geofence transition callbacks and dispatches them to the SDK. */
@@ -97,10 +94,20 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
         val timestamp = SDKComponent.clock.currentTimeSeconds()
         val dispatchStartUptimeMs = SDKComponent.clock.elapsedRealtime()
         val androidComponent = SDKComponent.android()
+        // Cold-start callbacks can beat the posted launch initialization. Establish the persisted
+        // secure user's session synchronously; legacy installs without an owner are migrated by the
+        // store without discarding their already-live OS registrations.
+        androidComponent.secureUserStore.getUserId()?.takeIf { it.isNotEmpty() }?.let {
+            androidComponent.geofenceRegionStore.beginUserSession(it)
+        }
+        // A previous callback can have staged a transition before the file outbox was writable.
+        // Recover it before interpreting this edge so an ENTER followed by EXIT stays ordered.
+        androidComponent.geofenceBusinessTransitionProcessor.recoverPendingTransitions()
         // Defense-in-depth against orphans (failed clearAll, app-data wipe, SDK
         // ID-format changes): events for unregistered IDs are dropped and the OS-side
         // registration is removed so it stops firing.
-        val registeredIds = androidComponent.geofenceRegionStore.getRegisteredIds()
+        val userStateGeneration = androidComponent.geofenceRegionStore.userStateGeneration()
+        val registeredIds = androidComponent.geofenceRegionStore.getRoutableRegisteredIds()
         val (knownIds, unknownIds) = triggeringGeofenceIds.partition { it in registeredIds }
         if (unknownIds.isNotEmpty()) {
             unknownIds.forEach { logger.logTransitionDroppedUnknownId(it) }
@@ -121,6 +128,15 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
                 return@forEach
             }
 
+            val region = androidComponent.geofenceRegionStore.getCachedRegion(geofenceId)
+            if (region == null && androidComponent.geofenceRegionStore.getRegisteredRegion(geofenceId) != null) {
+                // The server removed this region, but an earlier GMS removal failed. Its retained
+                // definition is a routing tombstone only; never manufacture business activity for
+                // a fence that no longer exists. Every orphan callback also retries OS cleanup.
+                logger.logTransitionDroppedRetiredId(geofenceId)
+                androidComponent.geofenceManager.removeGeofencesByIds(listOf(geofenceId))
+                return@forEach
+            }
             val transition = when (gmsTransitionType) {
                 Geofence.GEOFENCE_TRANSITION_ENTER -> Event.GeofenceTransition.ENTER
                 Geofence.GEOFENCE_TRANSITION_EXIT -> Event.GeofenceTransition.EXIT
@@ -130,17 +146,15 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
                 }
             }
 
-            // Each broadcast runs on its own scope, so two transitions for one fence would
-            // otherwise interleave read-decide-persist and lose a containment record.
-            transitionMutex.withLock {
-                handleBusinessTransition(
-                    geofenceId = geofenceId,
-                    transition = transition,
-                    timestamp = timestamp,
-                    androidComponent = androidComponent,
-                    logger = logger
-                )
-            }
+            androidComponent.geofenceBusinessTransitionProcessor.process(
+                geofenceId = geofenceId,
+                transition = transition,
+                timestampSeconds = timestamp,
+                enforceConfiguredTransition = region != null,
+                expectedRegionRevision = region?.transitionRevision(),
+                expectedUserStateGeneration = userStateGeneration,
+                requireRegistered = true
+            )
         }
 
         // Hold the goAsync window open until the refresh lands so the OS doesn't kill a
@@ -155,55 +169,6 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
         }
     }
 
-    private suspend fun handleBusinessTransition(
-        geofenceId: String,
-        transition: Event.GeofenceTransition,
-        timestamp: Long,
-        androidComponent: AndroidSDKComponent,
-        logger: GeofenceLogger
-    ) {
-        // Ahead of the identity checks below: being inside a fence is a physical fact, independent of
-        // whether the transition is deliverable.
-        val store = androidComponent.geofenceRegionStore
-        val cachedRegion = store.getCachedRegion(geofenceId)
-        when (transition) {
-            Event.GeofenceTransition.ENTER -> store.recordEntered(geofenceId)
-            // An unmatched EXIT is a GMS reconciliation artifact, not a crossing — delivering it
-            // fires EXIT campaigns at people who were never there. The two guards after claimExit
-            // cover the cases where an absent record proves nothing: an upgraded install with no set
-            // yet, and a region that never monitored ENTER (a missing cache row counts as unknown).
-            Event.GeofenceTransition.EXIT -> if (
-                !store.claimExit(geofenceId) &&
-                store.hasContainmentRecord() &&
-                cachedRegion?.transitionTypes?.contains(GeofenceTransitionType.ENTER) == true
-            ) {
-                logger.logExitDroppedNeverEntered(geofenceId)
-                return
-            }
-        }
-
-        // Snapshot userId so a sign-out + sign-in before delivery can't reattribute this
-        // transition. Empty userId is treated as "not identified" per `isUserIdentified`.
-        val userId = androidComponent.secureUserStore.getUserId()?.takeIf { it.isNotEmpty() }
-        // Identified-only: the backend rejects anonymous geofence tracks, so drop before spending a
-        // cooldown slot or persisting a row neither channel could send.
-        if (userId == null) {
-            logger.logTransitionDroppedAnonymous(geofenceId, transition.name)
-            return
-        }
-
-        androidComponent.geofenceTransitionEmitter.emit(
-            geofenceId = geofenceId,
-            transition = transition,
-            userId = userId,
-            timestampSeconds = timestamp,
-            geofenceName = cachedRegion?.name,
-            metadata = cachedRegion?.metadata ?: emptyMap(),
-            geosetIds = cachedRegion?.geosetIds ?: emptyList(),
-            monitorsExit = cachedRegion?.transitionTypes?.contains(GeofenceTransitionType.EXIT) == true
-        )
-    }
-
     private fun transitionName(gmsTransitionType: Int): String = when (gmsTransitionType) {
         Geofence.GEOFENCE_TRANSITION_ENTER -> "ENTER"
         Geofence.GEOFENCE_TRANSITION_EXIT -> "EXIT"
@@ -215,8 +180,5 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
         // goAsync grants ~10s before the OS considers the receiver blocked; total budget for
         // one dispatch (persistence + GMS awaits + movement-refresh wait), with headroom.
         private const val DISPATCH_WAIT_BUDGET_MS = 8_000L
-
-        // Process-wide: a receiver instance lives for one broadcast, so the lock has to outlive it.
-        private val transitionMutex = Mutex()
     }
 }
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
@@ -107,13 +107,32 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
         // ID-format changes): events for unregistered IDs are dropped and the OS-side
         // registration is removed so it stops firing.
         val userStateGeneration = androidComponent.geofenceRegionStore.userStateGeneration()
-        val registeredIds = androidComponent.geofenceRegionStore.getRoutableRegisteredIds()
-        val (knownIds, unknownIds) = triggeringGeofenceIds.partition { it in registeredIds }
-        if (unknownIds.isNotEmpty()) {
-            unknownIds.forEach { logger.logTransitionDroppedUnknownId(it) }
-            // Result ignored — a failed removal self-heals on the next orphan event.
-            androidComponent.geofenceManager.removeGeofencesByIds(unknownIds)
+        val routableIds = androidComponent.geofenceRegionStore.getRoutableRegisteredIds()
+        val (routableTriggeringIds, unroutableIds) = triggeringGeofenceIds.partition { it in routableIds }
+        // An identify clears routing and only the completing refresh re-arms it, so between the two
+        // a live registration is unroutable without being an orphan. Removing it there strands it:
+        // the refresh that follows still sees the id in registeredIds with matching params, skips it
+        // as unchanged, and never re-adds it to the OS. Only evict what the bookkeeping no longer
+        // claims, and drop the rest — routing is what authorizes a business event.
+        val unarmedTriggerIds = if (unroutableIds.isEmpty()) {
+            emptyList()
+        } else {
+            val registeredIds = androidComponent.geofenceRegionStore.getRegisteredIds()
+            val (unarmedIds, orphanIds) = unroutableIds.partition { it in registeredIds }
+            orphanIds.forEach { logger.logTransitionDroppedUnknownId(it) }
+            if (orphanIds.isNotEmpty()) {
+                // Result ignored — a failed removal self-heals on the next orphan event.
+                androidComponent.geofenceManager.removeGeofencesByIds(orphanIds)
+            }
+            // The movement trigger is SDK-owned and carries no business meaning — routing it only
+            // re-fetches. It is also the only refresh path that runs without the app being opened,
+            // so dropping its EXIT here would leave an unarmed session with nothing to re-arm it
+            // until the next launch. handleMovement waits for the slot and runs as the current user.
+            val (triggerIds, businessIds) = unarmedIds.partition { it == GeofenceConstants.MOVEMENT_TRIGGER_ID }
+            businessIds.forEach { logger.logTransitionDroppedUnarmedId(it) }
+            triggerIds
         }
+        val knownIds = routableTriggeringIds + unarmedTriggerIds
 
         var movementRefreshJob: Job? = null
         knownIds.forEach { geofenceId ->
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
@@ -610,9 +610,6 @@ internal class GeofenceRepositoryImpl(
                 logger.logSyncSkipped("user changed during refresh")
                 return@withLock Result.success(Unit)
             }
-            // Snapshot for the routing arm below. beginUserSession bumps this under its own lock, so
-            // an identify landing after this point invalidates the write rather than racing it.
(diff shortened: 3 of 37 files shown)


core/src/main/kotlin/io/customer/sdk/data/store/PendingDeliveryStore.kt
```diff
@@ -74,7 +74,10 @@ class PendingDeliveryStore<T : PendingDeliveryStore.PendingDeliveryEntry>(
     fun appendAll(entries: List<T>): Boolean {
         if (entries.isEmpty()) return true
         return lock.withLock {
-            val all = readAll().toMutableList()
+            // A caller recovering a staged outbox attempt may append the same stable keys again.
+            // Replace those rows atomically instead of duplicating one physical delivery.
+            val incomingKeys = entries.mapTo(mutableSetOf(), PendingDeliveryEntry::key)
+            val all = readAll().filterNot { it.key in incomingKeys }.toMutableList()
```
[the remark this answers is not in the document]
mrehan27: Tried to build this interleaving and it cannot exist. Emission drains older staged attempts before appending its own (`recoverPendingTransitionsLocked()` at the top of the emit path), so the queue can never hold the later edge alone while an earlier one is still staged — staging an ENTER whose append failed, then processing an EXIT that succeeds, leaves the outbox as `[ENTER, EXIT]`, not `[EXIT]`.

Added `recoverPendingTransitions_givenExitAlreadyQueuedWhileOlderEnterIsStillStaged_expectEnterFirst` to pin it; my own setup assertion failed on the first run because the ENTER was already restored ahead of the EXIT. Removing the drain-first call fails that test.

geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
```diff
@@ -97,10 +94,20 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
...
+        }
+        // A previous callback can have staged a transition before the file outbox was writable.
+        // Recover it before interpreting this edge so an ENTER followed by EXIT stays ordered.
+        androidComponent.geofenceBusinessTransitionProcessor.recoverPendingTransitions()
         // Defense-in-depth against orphans (failed clearAll, app-data wipe, SDK
         // ID-format changes): events for unregistered IDs are dropped and the OS-side
         // registration is removed so it stops firing.
-        val registeredIds = androidComponent.geofenceRegionStore.getRegisteredIds()
+        val userStateGeneration = androidComponent.geofenceRegionStore.userStateGeneration()
+        val registeredIds = androidComponent.geofenceRegionStore.getRoutableRegisteredIds()
```
Shahroz16: Could we wire the routable set back into the successful registration path before this lands? On a direct A-to-B identify, beginUserSession(B) writes an explicit empty set, but production only calls saveRegisteredIds(...); saveRoutableRegisteredIds(...) has no caller. The first B callback therefore classifies every live ID as unknown, removes it from GMS, and the explicit empty key keeps later callbacks blocked too. Saving the successful idsToSave as routable with a user-generation check, plus an A-to-B regression test, would close that loop.

geofence/src/main/kotlin/io/customer/geofence/worker/GeofenceEventWorker.kt
```diff
@@ -78,53 +72,70 @@ internal class GeofenceEventWorker(
...
+                    logger.logEventDeliveredButNotRemoved(entry.geofenceId, entry.transition.name)
+                    return Result.retry()
+                }
+                is PendingDeliveryResult.Retryable -> {
+                    logger.logEventDeliveryRetryable(
+                        entry.geofenceId,
+                        entry.transition.name,
+                        outcome.cause?.message
+                    )
+                    return Result.retry()
```
Shahroz16, line 124: Could we give a permanently rejected head row an escape path? CustomerIOHttpClient wraps every non-2xx response, including 400 and 401, in IOException, so this returns Result.retry() indefinitely. Because every geofence event now shares one ordered chain and each worker drains the oldest row, one bad payload prevents all later ENTER and EXIT events from being delivered. A typed HTTP-status classification and a bounded dead-letter or drop policy for terminal responses would preserve ordering without wedging the queue.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
```diff
@@ -509,6 +512,13 @@ internal class GeofenceRepositoryImpl(
                         newIds + staleIds
                     }
                     store.saveRegisteredIds(idsToSave)
+                    // Routing is armed separately: beginUserSession writes an explicit empty routable
+                    // set, and getRoutableRegisteredIds stops falling back to the registered set once
+                    // that key exists. Without this write the first callback after an identify
+                    // classifies every live ID as unknown and removes it from the OS.
+                    if (!store.saveRoutableRegisteredIdsIfCurrent(idsToSave, userStateGeneration)) {
+                        logger.logSyncSkipped("user changed before routing could be armed")
+                    }
```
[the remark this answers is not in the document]
mrehan27: Confirmed and fixed in 1919d8eb. Real, and the part that makes it bite is that the refusal was supposed to be self-healing and wasn't.

`onRegistered` on the remote path calls `setLastSyncTimestamp`, so a refused arm still stamped the sync fresh. The new user's own refresh — triggered by the same identify that caused the refusal — then read that stamp, took the SKIP branch in `refreshAction` without registering, and never armed routing. Its first callback classified every live fence as unknown and removed it. So the window wasn't "empty until the next registering refresh"; the stale pass actively suppressed that refresh.

Fix: a refused arm skips `onRegistered` entirely, leaving the stamp stale so the new user's refresh takes the remote path, re-registers, and arms routing under its own generation.

One thing worth flagging from the fix: two existing tests started failing, because the relaxed mock answers `false` for the new store call — which is the identify-landed-mid-pass branch. Every test that hadn't thought about it was silently skipping `onRegistered`. The suite now defaults that call to `true`, so only tests that mean to exercise a refusal do.

Verified: reverting the guard so the stamp happens regardless kills `refresh_givenUserChangedBeforeRoutingWasArmed_expectSyncNotStampedFresh`, and there's a control asserting the stamp still happens on the ordinary path. core + geofence 690/690, ktlint + lintDebug + apiCheck clean.

geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
mrehan27: Confirmed and fixed, in two parts — the second one because wiring it into registration alone wasn't enough.

`saveRoutableRegisteredIds` had no production caller, exactly as you describe. Registration now arms routing for the same ids it saved, guarded by the user-state generation so a pass that began before an identify can't re-arm the departing user's fences (3fffb345's parent, 5f546fb6).

The part your comment led me to that I'd have missed: arming on registration only fixes the callbacks *after* the damaging one. `beginUserSession` had a single production caller — the broadcast receiver — and an identify refresh returns SKIP on a fresh last-sync stamp without registering at all. So the first callback opened the session, wrote the empty routable set, read it back in the same dispatch, and removed every live fence. The session now opens at identify (`GeofenceServicesImpl.onUserIdentified`), which clears the last-sync stamp so that refresh takes the remote path and arms routing before any transition can arrive. The receiver keeps its call for cold starts.

A third fix came out of the second: a refused arm still ran `onRegistered`, which stamps the sync fresh — so the new user's own refresh would SKIP on our timestamp and never arm routing. It now skips the stamp, leaving that refresh on the remote path (1919d8eb).

Tests: an A-to-B store test, a `coVerifyOrder` test pinning that the session opens before the refresh, and one asserting the sync isn't stamped when the arm is refused — each verified by reverting the fix and watching the named test fail.

geofence/src/main/kotlin/io/customer/geofence/worker/GeofenceEventWorker.kt
mrehan27: Confirmed and fixed in 5f546fb6, with one deliberate narrowing.

Verified the chain end to end: `CustomerIOHttpClient` wraps every non-2xx in `IOException`, `sendRemoveOnSuccess` defaults `isRetryable` to `{ it is IOException }`, so a 400 classifies as retryable — and even the `Failed` branch returned `Result.retry()`. With this PR's single ordered chain draining the oldest row, one refused payload blocks every later ENTER and EXIT.

Responses now carry their status via `HttpRequestFailure(statusCode, responseBody)`. It extends `IOException` so existing `it is IOException` predicates elsewhere (push, in-app) classify it exactly as before; only callers that opt into reading the status behave differently. Retryable set is 408, 429 and 5xx. A terminal status drops the head row and continues the drain, the same disposal the anonymous-row case above it already uses.

The narrowing: only a terminal HTTP status drops. An unexpected exception still keeps the row and retries, preserving `doWork_givenNonIOException_expectRetryScheduledAndChainPreserved` — a bad state or serialization slip may well succeed next time, and only an HTTP rejection is known to be permanent.

One thing you should know rather than find later: this was explicitly deferred to a post-release follow-up in the 2026-07-20 triage, and I implemented it here without checking that record. Rehan's call is to keep it in this PR rather than extract it, since five PRs are stacked on top.

geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
Shahroz16: Could we also retain the identify refresh when another refresh is in progress? I reproduced A waiting in GMS registration, then identifying B: B clears routing, but `refresh()` returns because A owns the slot. A then refuses to arm the new generation, leaving routing empty with no retry pending; the next callback removes the live IDs. The sequential A-to-B control passes, but the overlap test fails. Queuing a refresh for the current session, while preserving anchor rather than live-fix semantics, would close the remaining gap.

Shahroz16 (commented):
The terminal HTTP handling and normal A-to-B routing fixes check out. All 460 focused geofence, core and push tests passed. The added overlapping-identify regression still fails as described in the existing routing thread, so I’m holding approval here and on dependent #838. This branch also currently needs its base conflict resolved.

geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
mrehan27: Fixed. `refresh()` now waits for the slot instead of dropping when the generation differs, so B's pass arms the new session rather than returning early — anchor semantics kept, still not a live fix.

Pinned by `refresh_givenIdentifyWhileAnotherRefreshHoldsTheSlot_expectRoutingArmedForTheNewSession`, which parks A inside `replaceGeofences`, bumps the generation to 8, and asserts `armed == [7L, 8L]`: A refuses as stale, B arms.

geofence/src/main/kotlin/io/customer/geofence/worker/GeofenceEventTracker.kt
```diff
@@ -87,7 +87,7 @@ internal class AsyncGeofenceEventTracker(
         }
         CoroutineScope(dispatcher.background).launch {
             try {
-                pendingStore.sendRemoveOnSuccess(entry) {
+                pendingStore.sendRemoveOnSuccess(entry, ::isRetryableDeliveryFailure) {
```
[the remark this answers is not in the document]
mrehan27: Correct that the result is discarded, but the recovery described as missing does exist: the entry stays in the store and `GeofenceLifecycleObserver` flushes it at `AT_LEAST_ONCE` on every foreground, re-publishing and re-removing it, deduped backend-side on `transitionId`.

Nothing queues behind it either — the async path sends per entry rather than draining the head, so the ordered-head framing applies to the worker, which only runs when WorkManager is available (i.e. when this fallback is not in use).

The behaviour also predates this PR: the only change here is the `isRetryableDeliveryFailure` argument. Previously `remove()`'s Boolean was ignored and the result reported `Delivered` regardless — this PR just named the outcome.

geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
Shahroz16: Could we also distinguish temporarily unarmed IDs from true orphans here? I reproduced a callback while B's identify refresh was waiting on HTTP: routing was empty, so the receiver removed the live business fence. The completed refresh then kept its ID as unchanged and never re-added it. Keeping registered-but-unarmed IDs in the OS while dropping their transitions, or updating the registration bookkeeping when removing them, would close this window.

Shahroz16 (commented):
The overlapping-identify regression passes now. One routing-window issue remains in the existing thread.

geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
mrehan27: Fixed. The orphan guard now splits the unroutable ids against `getRegisteredIds()`: ones the registration bookkeeping still claims are dropped but left registered, and only ids it no longer claims are evicted. Checking the base confirmed this window was introduced here rather than pre-existing — the receiver used to partition on `getRegisteredIds()`, so clearing the routable key had no effect on it until this PR.

The movement trigger is exempt and still drives its refresh. It is in `registeredIds` too, so it would otherwise have been dropped in the same window — and it is the only refresh path that runs without the app being opened, so an unarmed session would have had nothing to re-arm it until the next launch. That also makes the 30s slot-wait residual genuinely self-healing.

Covered by `dispatchTransition_givenRegisteredIdWhileRoutingUnarmed_expectDroppedButOsRegistrationKept`, `dispatchTransition_givenUnarmedAndOrphanIds_expectOnlyOrphanRemoved` and `dispatchTransition_givenMovementTriggerExitWhileRoutingUnarmed_expectRefreshStillDriven`, each mutation-tested.

Not reproducible from the sample app — logout/login takes a full `reset()`, so the direct A-to-B identify this needs is not reachable there. It rests on the unit tests.

Shahroz16: Thanks, the registered-but-unarmed case now passes. Could we also establish the persisted session before a launch refresh, or avoid invalidating it when migration adopts the same user? I reproduced a launch without `identify`, with an existing user/registration but no session-owner key. A callback during the GMS registration await bumps the generation, so the refresh saves the new registered IDs but skips the new catalog. The next callback for a new fence is then forwarded without its configured name or geoset IDs. Establishing the session before launch passes the same regression.

Shahroz16 (commented):
The registered-but-unarmed fix passes the original regression. One launch-time session migration issue remains in the existing discussion, so I am holding approval.

geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
mrehan27: Confirmed and fixed in 1f705119. The migration branch adopts the persisted user, which is not a session change, but it bumped the generation anyway and so invalidated the in-flight launch refresh. That pass still saved its registered ids, then skipped the catalog write it gates on the generation. Routing falls back to the registered ids while the routing key is absent, which is exactly the upgrade state, so the newly registered fence was routable with no cached row and fired with no name and no geoset ids.

The generation now stays put on adoption. Pinned by `beginUserSession_givenLegacyInstallAdoptedMidRefresh_expectInFlightWriteStillArmsRouting`, and mutation-checked by restoring the bump, which kills that test and nothing else. I took this over establishing the session before the launch refresh because the spurious bump is the defect itself, and leaving it in place would keep invalidating passes for a user change that never happened.

Sweeping for the same shape turned up one more instance: `routingArmed` is decided several store writes before `onRegistered` runs, so an identify landing in that gap stamps the new session's cache fresh off this pass. Its own refresh then skips and leaves routing empty until a movement EXIT re-arms it. That fix is in #846 rather than here, so this PR stays on what you raised.

Shahroz16: Thanks, the same-user migration regression now passes. Could we establish the persisted session before starting launch work so this branch cannot also absorb a real user switch? With no session-owner key, I parked A's launch in GMS registration and identified B. Adoption kept generation 0, so B's refresh was dropped as a duplicate. When A's registration then failed, no B refresh ran and the old fence set remained. Both switch cases pass on the previous commit and when A's session is established first.

Shahroz16 (commented):
The same-user migration fix is verified. A real user switch during that legacy launch window now loses the incoming user's refresh; details are in the existing discussion. I am holding approval for that case.

geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
mrehan27: Confirmed, and you were right that this branch was the problem rather than the bump. Rather than patch it a third time I removed it.

Establishing the session before launch work is in, mirroring what identify already did (`edc7a51d`). But that alone was not enough: `onAppLaunch` runs from a coroutine posted to the main thread, so it reads the secure store late too and could adopt the wrong user by the same route. The adopt branch could not tell "adopting the persisted user" from "a first session naming someone else", and no ordering fix removes that.

So the branch is gone (`35117459`). An unowned session opens as a switch. That branch already kept registered ids, retained regions and the registration stamps, so live fences stay registered and keep reporting, unarmed until a pass arms them, which is the state the registered-but-unarmed guard from the earlier round already handles. Routing also no longer falls back to the registered set when the key is absent: every session now writes the key, so the fallback was unreachable and it encoded the policy this replaces.

That opened one real gap, which is fixed here rather than noted (`637ca8ad`): if the launch refresh is the pass that should arm routing and it fails, nothing routes and nothing recovers, because only a completing pass arms it. A failed remote pass now re-ranks from the cached catalog, as a failed movement pass already does, and only when routing is empty and a cache exists, so a session that is already routing never re-registers from a stale cache on a failed fetch.

What it costs, so it is on the record: on an upgraded install the first OS callback is dropped as unarmed where adoption would have forwarded it, and one duplicate ENTER is possible after a reboot because the emitted-ENTER marks are cleared. Both once per install, both the safe direction.

609 geofence tests, 987 across the three modules, 0 failures. Each new guard is mutation-checked to kill exactly one test: restoring the adopt branch, restoring the routing fallback, deleting the offline re-rank, and widening its unarmed guard. Two existing tests were passing for the wrong reason and are fixed here too, including `refresh_givenApiFailure_expectFailurePropagatedAndNoPersistOrRegister`, whose claim held only because its stubbed store returned an empty catalog. None of this has been exercised on a device; the upgrade sequence is covered against the real store rather than a stub.

Shahroz16: Thanks, the legacy switch and offline recovery cases now pass. Could we also fence the launch session-open against a newer identify? I paused launch after its A read, identified B, then resumed beginUserSession(A). B's refresh completed with secure user B but session owner A. The next business callback reopened B, cleared routing and was dropped, with no refresh queued. Preventing that stale launch from overwriting the newer session, plus a regression for this interleaving, would close the window.

Shahroz16 (commented):
The earlier legacy-switch and offline-recovery cases now pass. I reproduced a separate launch-versus-identify race that leaves routing cleared after a successful refresh; the details are in the existing session discussion. I am holding approval for that case.

geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
mrehan27: Fixed in `0faf8aa3`. Launch now uses `beginUserSessionIfAbsent`, which does the owner check and the write together under `enteredLock`, so a stale launch read cannot reopen a user that an identify has since replaced. Identify and the receiver keep `beginUserSession`: those are the paths that legitimately redefine a session, and launch only ever needed to establish one.

Covered by `onAppLaunch_givenAnIdentifyLandsBetweenTheReadAndTheStore_expectTheNewerSessionKept`, which runs your interleaving against the real store and services, plus three store tests. Mutation-checked: dropping the absent guard and reverting the call site each kill it, and one mutant that survived the first pass (`generation > 0` in place of the owner check) is now killed too.

Cost, on the record: launch no longer reconciles a stale owner against the secure user, so if a process dies between an identify's secure write and the geofence handler, the next launch runs under the older owner until the replayed `UserChangedEvent` switches it. One callback in that window is dropped as unarmed.

The same read-then-call shape is in `onUserIdentified` and the receiver's dispatch, and I left both. There the write is the identity signal, so "open only if absent" would be wrong, and closing it properly means resolving the identity inside the store under its lock, which changes the interface and all three call sites. The window is a deschedule between two adjacent statements with no suspension point, and the consequence is one dropped callback that the next pass re-arms. Say the word if you would rather we did it.

614 geofence tests, 992 across the three modules, 0 failures. Not exercised on a device.

Shahroz16: Thanks, this closes the stale-launch overwrite. I reran the A-read, identify-B, resume-A regression against the real services/store and verified that B's routing survives the next callback. The initialization path also republishes UserChangedEvent, so the persisted-owner case you noted is repaired during the same startup. I am happy to approve this fix with that temporary startup window documented here.

Shahroz16 (approved):
The launch-versus-identify regression and earlier migration cases now pass. I reran 264 focused tests and the geofence build on this head. This is ready for the feature branch; #846 still needs to land before the stack goes to main.

PR #839, opened 2026-09-03
chore(geofence): trust the backend's polygon validation

mrehan27:

## Stack

- 👉 [#839 trust the backend's polygon validation](https://github.com/customerio/customerio-android/pull/839)
- [#840 collapse repeated positions when canonicalising a ring](https://github.com/customerio/customerio-android/pull/840)
- [#841 support polygons that cross the antimeridian](https://github.com/customerio/customerio-android/pull/841)

Mirrors the same removal @Shahroz16 is making on iOS, from @mahmoud's review point: the backend
already admits or rejects a polygon before it reaches an SDK, so re-implementing its rules on the
device only creates ways for iOS, Android and the backend to disagree about the same payload — and
puts any rule change behind an SDK release.

## Dropped

- per-vertex coordinate range
- the 500-vertex cap
- the covering-circle coverage test (every vertex within `base_radius_m` + slack)
- convexity
- the 85° latitude cap

Convexity is the clearest of them: the backend admits concave rings, our ray cast handles them, and
we rejected them anyway. The latitude cap is a second commit because it wasn't on iOS's list — its
only stated reason was to keep a coordinate-linear ring inside the coverage check being removed
here, and it held polygons to a stricter rule than the rest of the module, where a circle at the
pole was always registerable. iOS then re-baselined the shared vectors to accept that case, which
settled it.

With coverage gone the wake-circle validator no longer looks at the ring at all, so it loses its
`geometry` parameter and its haversine helper with it.

## Kept

What the region needs to exist: a `position` with at least two elements, a ring the geometry kernel
can use (three distinct vertices after unclosing, no self-intersection, non-zero area), and a wake
radius the platform can register.

## Parity

Against the re-baselined `test-vectors-nearest-payload-polygon-v2.json` (22 payload cases):

| | score |
|---|---|
| `feature/geofence-polygons` | 16/22 |
| after the four removals | 19/22 |
| after the latitude cap | 20/22 |
| with #840 on top | 22/22 |

Worth recording that we were never at 25/25 on the old 25-case set either — 21/25 was the standing
number before any of this.

## Tests

487 geofence tests, 0 failures. ktlint, `lintDebug` and `apiCheck` clean. No public API change.

Seven tests pinned the removed rules; each is rewritten to state the new contract rather than
deleted, and the concave one now also asserts the ray cast answers correctly inside the reflex
notch — that being the actual risk of accepting concave rings.

Rewriting the near-pole test turned up that its fixture spans 205° of longitude, so it was really an
antimeridian case: it still fails on the antimeridian guard, which nothing here touches. Its
replacement isolates latitude.

## Effect on the open stack

Independent of #835–#838, which share this base. When this lands they rebase onto it, and #838 needs
its coordinate-range guard reinstated inline in `toPolygonLocationFix`: that one validates an
OS-reported fix rather than a backend payload, so it survives this change.

## Review round

@Shahroz16 asked to keep the platform-validity check on the enclosing circle's **centre**, and he was
right — it is not a geometry rule the backend owns. Removing the coverage check took away the only
thing that incidentally rejected an out-of-range centre, and `registerBatch` builds the whole
business batch in one `map`, so `Geofence.Builder` throwing on one bad record fails **every** fence
in that sync and rolls the movement trigger back with it. Checking the centre isolates the record.
iOS keeps the equivalent check (`CLLocationCoordinate2DIsValid`), so this stays at parity.

`PolygonShapeAdversarialIntegrationTest` asserted the V1 envelope rejects a concave L — the rule this
PR removes. Rewritten to assert the shape is accepted and the notch it wraps around still answers
OUTSIDE.

**Positions that are not numbers.** Positions decode as `JsonElement`, so the literal `"NaN"` parses to
a `Double` and passes every geometry check — each one compares, and all comparisons against NaN are
false. The record registered and failed later in the strict cache encoder, interrupting the catalog
update after the sync had run. Non-finite positions are now dropped at the wire boundary. This is not
one of the coordinate-range rules the backend owns; it is whether the value is a number at all.

### DIFF
--- a/geofence/src/main/kotlin/io/customer/geofence/api/GeofenceApiResponse.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/api/GeofenceApiResponse.kt
@@ -275,13 +275,13 @@ private fun GeofenceApiRegion.toPolygonRegionOrNull(
         return null
     }
     val baseRadiusMeters = enclosingCircle.baseRadiusMeters
-    val wakeCenter = PolygonCoordinate.fromOrNull(enclosingCircle.latitude, enclosingCircle.longitude)
-    val trigger = if (wakeCenter == null || baseRadiusMeters == null) {
+    val wakeLatitude = enclosingCircle.latitude
+    val wakeLongitude = enclosingCircle.longitude
+    val trigger = if (wakeLatitude == null || wakeLongitude == null || baseRadiusMeters == null) {
         null
     } else {
         PolygonWakeCircleValidator().prepareOrNull(
-            geometry = polygon,
-            wakeCircle = PolygonWakeCircle(wakeCenter, baseRadiusMeters)
+            PolygonWakeCircle(PolygonCoordinate(wakeLatitude, wakeLongitude), baseRadiusMeters)
         )
     }
     if (trigger == null) {
@@ -333,7 +333,7 @@ private fun JsonElement.toPolygonVerticesOrNull(): List<PolygonCoordinate>? {
         if (position.size < 2) return null
         val longitude = (position[0] as? JsonPrimitive)?.doubleOrNull ?: return null
         val latitude = (position[1] as? JsonPrimitive)?.doubleOrNull ?: return null
-        PolygonCoordinate.fromOrNull(latitude, longitude) ?: return null
+        PolygonCoordinate(latitude, longitude)
     }
 }
 
--- a/geofence/src/main/kotlin/io/customer/geofence/api/GeofenceApiResponse.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/api/GeofenceApiResponse.kt
@@ -277,7 +277,14 @@ private fun GeofenceApiRegion.toPolygonRegionOrNull(
     val baseRadiusMeters = enclosingCircle.baseRadiusMeters
     val wakeLatitude = enclosingCircle.latitude
     val wakeLongitude = enclosingCircle.longitude
-    val trigger = if (wakeLatitude == null || wakeLongitude == null || baseRadiusMeters == null) {
+    val trigger = if (
+        wakeLatitude == null || wakeLongitude == null || baseRadiusMeters == null ||
+        // Not a geometry rule the backend owns — a platform precondition. Geofence.Builder throws on
+        // an out-of-range centre, and registration builds the whole business batch in one map, so a
+        // single bad record fails every fence in the sync and rolls the movement trigger back with
+        // it. Checking here isolates the record instead.
+        !LocationCoordinates.isValid(wakeLatitude, wakeLongitude)
+    ) {
         null
     } else {
         PolygonWakeCircleValidator().prepareOrNull(
--- a/geofence/src/main/kotlin/io/customer/geofence/api/GeofenceApiResponse.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/api/GeofenceApiResponse.kt
@@ -340,6 +340,11 @@ private fun JsonElement.toPolygonVerticesOrNull(): List<PolygonCoordinate>? {
         if (position.size < 2) return null
         val longitude = (position[0] as? JsonPrimitive)?.doubleOrNull ?: return null
         val latitude = (position[1] as? JsonPrimitive)?.doubleOrNull ?: return null
+        // Not a range rule — the backend owns those. A position is decoded as a JsonElement, so the
+        // literal "NaN" parses to a Double and then passes every geometry check, because each of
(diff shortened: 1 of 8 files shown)


geofence/src/main/kotlin/io/customer/geofence/api/GeofenceApiResponse.kt
```diff
@@ -275,13 +275,13 @@ private fun GeofenceApiRegion.toPolygonRegionOrNull(
         return null
     }
     val baseRadiusMeters = enclosingCircle.baseRadiusMeters
-    val wakeCenter = PolygonCoordinate.fromOrNull(enclosingCircle.latitude, enclosingCircle.longitude)
-    val trigger = if (wakeCenter == null || baseRadiusMeters == null) {
+    val wakeLatitude = enclosingCircle.latitude
+    val wakeLongitude = enclosingCircle.longitude
+    val trigger = if (wakeLatitude == null || wakeLongitude == null || baseRadiusMeters == null) {
```
Shahroz16: Could we keep the platform-validity check for the wake-circle center? Trusting the backend for polygon admission makes sense, but GMS still requires latitude and longitude to be in range. A center such as 91,0 now survives mapping, Geofence.Builder throws while building the shared business batch, and replaceGeofences rolls back the movement trigger, so one malformed record disables every fence for that sync. LocationCoordinates.isValid(wakeLatitude, wakeLongitude) here would isolate the record without reimplementing polygon geometry rules.

mrehan27: Agreed and fixed in 926090db — you're right that this isn't a geometry rule the backend owns.

Traced the blast radius before changing it: `registerBatch` builds the request with `regions.map { it.toGmsGeofence() }`, so one out-of-range centre throws inside that map, fails the whole batch, and `replaceGeofencesInternal` then rolls the movement trigger back. So it's every fence in the sync plus the safety net, not just the bad record.

Worth noting it's a regression this PR introduced rather than a pre-existing gap: the coverage check I removed was the only thing incidentally rejecting a bad centre, and nothing replaced it. `LocationCoordinates.isValid(wakeLatitude, wakeLongitude)` in `toPolygonRegionOrNull` now isolates the record, which also matches what iOS keeps (`CLLocationCoordinate2DIsValid` on the circle centre, vertices unchecked) — so the two SDKs still agree on this payload.

Test asserts a 91-latitude centre drops only that record and leaves the sibling circle registered; reverting the check kills it.

geofence/src/main/kotlin/io/customer/geofence/api/GeofenceApiResponse.kt
```diff
@@ -333,7 +340,7 @@ private fun JsonElement.toPolygonVerticesOrNull(): List<PolygonCoordinate>? {
         if (position.size < 2) return null
         val longitude = (position[0] as? JsonPrimitive)?.doubleOrNull ?: return null
         val latitude = (position[1] as? JsonPrimitive)?.doubleOrNull ?: return null
-        PolygonCoordinate.fromOrNull(latitude, longitude) ?: return null
+        PolygonCoordinate(latitude, longitude)
```
Shahroz16, line 348: Could we keep a finite-number check here without bringing back the backend’s coordinate-range rules? These positions are decoded as `JsonElement`, so `[0, "NaN"]` survives `doubleOrNull`. I reproduced `[[[0,0],[1,0],[1,1],[0,"NaN"],[0,0]]]` being accepted, then throwing in the strict cache encoder. Polygon monitoring is still gated off, but once enabled this record would interrupt the catalog update after registration. Checking both values with `isFinite()` and testing this wire path would keep the bad record isolated.

Shahroz16 (commented):
The enclosing-circle center check addresses the earlier batch-registration issue. I found one remaining numeric-usability issue below, so I’m holding approval here and on the dependent #840/#841 stack.

geofence/src/main/kotlin/io/customer/geofence/api/GeofenceApiResponse.kt
mrehan27: Fixed. Added an `isFinite()` check on both values in `toPolygonVerticesOrNull`, so a non-finite position drops the record without reintroducing the backend's range rules.

Covered by `parseAndMap_givenPolygonPositionThatIsNotFinite_expectRecordDroppedNotWholeSync` using your payload — the polygon is dropped and the circle still syncs.

Shahroz16 (approved):
The finite-coordinate guard closes my remaining concern. I rechecked the malformed-input cases and polygon tests; this is good to go.

PR #841, opened 2026-09-04
chore(geofence): support polygons that cross the antimeridian

mrehan27:
## Summary

A polygon whose ring crosses the antimeridian was rejected at decode, so a fence iOS delivers was silently dropped on Android. iOS supports these deliberately (`unwrapLongitudes`, iOS #1249) and enforces `square400_dateline` in CI, so this is Android's gap to close.

The fix is the frame, not a special case. The ring is unwrapped once at construction into a contiguous line, and every query — containment, boundary distance, and admission — is compared on that same line. Removing the decode guard alone would have been worse than the bug: the ray cast compared raw longitudes, which **inverts** containment across the seam.

## Validation

Reverted each part in isolation to confirm the tests catch it:

| Reverted | Tests that fail |
| --- | --- |
| decode guard restored | 5 |
| ray cast → raw longitudes | 2 (inside and outside) |
| `isOnSegment` → raw longitudes | 1 (BOUNDARY) |
| queries mapped to the point, not the ring line | 2 (antipodal) |
| closure detected on raw equality | 1 |
| span bound removed | 1 |

Cross-SDK geometry vectors **30/30**, up from 23/30, against iOS's current fixtures. Geofence suite 498/498; ktlint, `lintDebug` and `apiCheck` clean.

## Review rounds

Four defects found after the first commit, all reproduced before being fixed:

- **Antipodal queries** (Bugbot). Mapping each longitude onto the query point splits the ring whenever the wrap boundary falls between two vertices. A device at lon 180 read as inside a fence at lon ±1, and `edgeDistanceToOrNull` turns that into `0f` — ranking it ahead of every genuinely nearby fence. The distance half of this predates the PR.
- **Admission left on raw longitudes.** 4,765 of 18,688 generated *simple* seam-crossing rings were rejected as self-intersecting; whether a fence hit it depended only on whether the payload spelled the seam vertex `180` or `-180`.
- **Closure with the opposite seam sign.** `[-180 … 180]` never looked closed, so its zero-length edge was rejected — the same spelling-dependent rejection this stack exists to end.
- **Rings winding around a pole.** Simple and non-degenerate, so nothing rejected them, then evaluated against a closing chord most of the way round the earth: a point deep inside the encircled cap read as OUTSIDE.

## Stack

1. [#839](https://github.com/customerio/customerio-android/pull/839): trust the backend's polygon validation
2. [#840](https://github.com/customerio/customerio-android/pull/840): collapse repeated positions when canonicalising a ring
3. **#841:** antimeridian support

**Canonicalisation compares source positions.** Unwrapping sums a short arc per vertex, so a closed
ring in decimal coordinates left its closing position ~1.14e-13 from the one it repeats. Exact equality
on the accumulated line never saw the closure, and the sliver edge that survived read as a
self-intersection — while the same ring written without its closing position was accepted. The
comparison now runs on the source longitudes through `normalizeLongitude`, which is what keeps
`180` and `-180` equal, without inheriting the drift.

Review against `mbl-2287-collapse-duplicate-vertices`.

Linear: [MBL-2287](https://linear.app/customerio/issue/MBL-2287/android-add-polygon-wire-contract-and-geometry-foundation)

### DIFF
--- a/geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonGeometry.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonGeometry.kt
@@ -26,46 +26,74 @@ internal enum class PolygonPointRelation {
 internal class PolygonGeometry private constructor(
     val vertices: List<PolygonCoordinate>
 ) {
+    /**
+     * Ring longitudes made contiguous: each vertex advances from the previous one by the short arc,
+     * so a ring crossing the antimeridian runs 179.5 -> 180.5 rather than 179.5 -> -179.5.
+     *
+     * Every query is mapped onto this same line by [onRingLine] before it is compared. Wrapping each
+     * longitude independently against the query point instead would break the ring apart whenever the
+     * wrap boundary fell between two of its vertices — a point roughly antipodal to the ring — turning
+     * a two-degree seam edge into a 358-degree chord and reporting far-away points as inside.
+     */
+    private val ringLongitudes: DoubleArray = DoubleArray(vertices.size).also { unwrapped ->
+        unwrapped[0] = vertices[0].longitude
+        for (index in 1 until vertices.size) {
+            unwrapped[index] = unwrapped[index - 1] +
+                normalizeLongitude(vertices[index].longitude - vertices[index - 1].longitude)
+        }
+    }
+
     fun relationTo(point: PolygonCoordinate): PolygonPointRelation {
+        val pointLongitude = onRingLine(point.longitude)
         var inside = false
-        var previous = vertices.last()
+        var previousIndex = vertices.lastIndex
 
-        for (current in vertices) {
-            if (point.isOnSegment(previous, current)) return PolygonPointRelation.BOUNDARY
+        for (index in vertices.indices) {
+            if (isOnSegment(point.latitude, pointLongitude, previousIndex, index)) {
+                return PolygonPointRelation.BOUNDARY
+            }
 
-            // Longitudes are cast relative to the query point and wrapped, the same trick
-            // localOffsetMeters already uses for distance. A ring crossing the antimeridian holds raw
-            // longitudes on both sides of +/-180, and comparing those directly inverts containment for
-            // every point across the seam rather than merely misplacing the boundary.
-            val currentLongitude = normalizeLongitude(current.longitude - point.longitude)
-            val previousLongitude = normalizeLongitude(previous.longitude - point.longitude)
-            val intersects = (current.latitude > point.latitude) != (previous.latitude > point.latitude) &&
-                0.0 < (previousLongitude - currentLongitude) *
-                (point.latitude - current.latitude) /
-                (previous.latitude - current.latitude) + currentLongitude
+            val currentLatitude = vertices[index].latitude
+            val previousLatitude = vertices[previousIndex].latitude
+            val currentLongitude = ringLongitudes[index]
+            val previousLongitude = ringLongitudes[previousIndex]
+            val intersects = (currentLatitude > point.latitude) != (previousLatitude > point.latitude) &&
+                pointLongitude < (previousLongitude - currentLongitude) *
+                (point.latitude - currentLatitude) /
+                (previousLatitude - currentLatitude) + currentLongitude
             if (intersects) inside = !inside
-            previous = current
+            previousIndex = index
         }
 
         return if (inside) PolygonPointRelation.INSIDE else PolygonPointRelation.OUTSIDE
     }
 
     fun boundaryDistanceMeters(point: PolygonCoordinate): Double {
+        val pointLongitude = onRingLine(point.longitude)
         var minimumDistance = Double.MAX_VALUE
-        var previous = vertices.last()
-        for (current in vertices) {
-            minimumDistance = min(minimumDistance, point.distanceToSegmentMeters(previous, current))
-            previous = current
+        var previousIndex = vertices.lastIndex
+        for (index in vertices.indices) {
+            minimumDistance = min(
+                minimumDistance,
+                distanceToSegmentMeters(point.latitude, pointLongitude, previousIndex, index)
+            )
+            previousIndex = index
         }
         return minimumDistance
     }
 
-    private fun PolygonCoordinate.distanceToSegmentMeters(
-        first: PolygonCoordinate,
-        second: PolygonCoordinate
+    /** [longitude] expressed on [ringLongitudes]' line, so ring and query share one frame. */
+    private fun onRingLine(longitude: Double): Double =
+        ringLongitudes[0] + normalizeLongitude(longitude - ringLongitudes[0])
+
+    private fun distanceToSegmentMeters(
+        pointLatitude: Double,
+        pointLongitude: Double,
+        firstIndex: Int,
+        secondIndex: Int
     ): Double {
-        val firstOffset = localOffsetMeters(first)
-        val secondOffset = localOffsetMeters(second)
+        val firstOffset = localOffsetMeters(pointLatitude, pointLongitude, firstIndex)
+        val secondOffset = localOffsetMeters(pointLatitude, pointLongitude, secondIndex)
         val segmentX = secondOffset.first - firstOffset.first
         val segmentY = secondOffset.second - firstOffset.second
         val lengthSquared = segmentX * segmentX + segmentY * segmentY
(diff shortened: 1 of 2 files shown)


geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonGeometry.kt
```diff
@@ -33,10 +33,16 @@ internal class PolygonGeometry private constructor(
...
+            // longitudes on both sides of +/-180, and comparing those directly inverts containment for
+            // every point across the seam rather than merely misplacing the boundary.
+            val currentLongitude = normalizeLongitude(current.longitude - point.longitude)
+            val previousLongitude = normalizeLongitude(previous.longitude - point.longitude)
             val intersects = (current.latitude > point.latitude) != (previous.latitude > point.latitude) &&
-                point.longitude < (previous.longitude - current.longitude) *
+                0.0 < (previousLongitude - currentLongitude) *
                 (point.latitude - current.latitude) /
-                (previous.latitude - current.latitude) + current.longitude
+                (previous.latitude - current.latitude) + currentLongitude
```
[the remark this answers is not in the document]
mrehan27: Confirmed and fixed in efd4578d — this was a regression I introduced, and the differential is clean.

Reproduced with an ordinary ring on the prime meridian (lon -1..+1), querying along the equator. On my commit, and on its parent for comparison:

```
              this PR (8c19b0c5)      parent
lon   90.0    OUTSIDE   9,895,657 m   OUTSIDE   9,895,657 m
lon  179.0    OUTSIDE  19,791,079 m   OUTSIDE  19,791,079 m
lon  180.0    INSIDE       55,597 m   OUTSIDE      55,597 m
lon -179.0    INSIDE       55,597 m   OUTSIDE      55,597 m
```

So a device half a world away read as inside the fence. It matters because `edgeDistanceToOrNull` maps "not outside" to `0f`, which ranks that fence ahead of every genuinely nearby one and lets it survive the distance cap.

Note the last column: the 55 km distance was already wrong before this PR. Wrapping each longitude independently against the query point splits the ring apart whenever the wrap boundary falls between two of its vertices, and the distance path had that same split. Fixing only the relation would have left it.

The fix is the frame, not a special case: the ring is unwrapped once at construction into a contiguous line, and every query is mapped onto that same line before anything is compared. Ring and point now always share one frame, so a seam edge stays two degrees wide instead of becoming a 358-degree chord.

Verified rather than argued — each control run in isolation:
- reverting to raw longitudes kills the 3 seam tests
- the committed 8c19b0c5 file against the new tests kills both antipodal tests
- geofence suite 493/493, cross-SDK geometry vectors 30/30, ktlint + lintDebug + apiCheck clean

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonGeometry.kt
```diff
@@ -22,44 +22,78 @@ internal enum class PolygonPointRelation {
...
+     * wrap boundary fell between two of its vertices — a point roughly antipodal to the ring — turning
+     * a two-degree seam edge into a 358-degree chord and reporting far-away points as inside.
+     */
+    private val ringLongitudes: DoubleArray = DoubleArray(vertices.size).also { unwrapped ->
+        unwrapped[0] = vertices[0].longitude
+        for (index in 1 until vertices.size) {
+            unwrapped[index] = unwrapped[index - 1] +
+                normalizeLongitude(vertices[index].longitude - vertices[index - 1].longitude)
+        }
+    }
```
Shahroz16: Could we run the structural checks in this same unwrapped longitude frame? ringLongitudes fixes relation and distance, but hasNonCollinearVertices() and hasSelfIntersection() below still use raw plus-or-minus 180 degree longitudes. For example, the valid seam-crossing ring [(179,-2), (179.5,-2), (180,-1), (179,-1)] in (longitude, latitude) order is simple when unwrapped, but raw validation turns the 179.5 to -180 edge into a 359.5 degree chord and rejects it as self-intersecting. Reusing the unwrapped coordinates in those checks would make admission agree with runtime geometry.

mrehan27: Confirmed and fixed in 5863322f. Your example is real — it just needed writing the way a payload would.

As given (seam vertex `180`) it is accepted, because raw longitudes 179 / 179.5 / 180 / 179 never wrap. Written `-180` — equally legal GeoJSON for the same point — it is rejected, exactly as you described.

To find out whether that was your one vector or a class, I generated simple seam-crossing rings and checked them against an independent in-test simplicity oracle:

```
                          before      after
simple rings straddling the seam:
  tried 18,688            4,765       0      rejected
prime-meridian control (18,688)  0     0      rejected
seam-crossing bowtie             REJECTED    REJECTED
```

So a quarter of simple rings across the seam were being refused, and whether a given fence hit it depended only on which of the two legal spellings the payload used. The bowtie line is the control — the guard isn't a blanket exemption for seam rings.

Fix is the one you suggested: admission now runs on the same unwrapped ring the evaluator uses, so what `from` accepts and what it later evaluates are the same shape. I also moved the zero-length-edge check onto that frame, since 180 and -180 are equal there and were not before.

This was my incomplete fix rather than a pre-existing gap — I moved `relationTo` and the distance path onto the ring line in efd4578d and left the structural checks behind, so it belongs in this PR. Suite 495/495, cross-SDK geometry vectors 30/30, ktlint + lintDebug + apiCheck clean.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonGeometry.kt
```diff
@@ -74,71 +105,106 @@ internal class PolygonGeometry private constructor(
...
+            require(vertices.isNotEmpty()) { "polygon requires at least one position" }
+
+            // Canonicalisation happens on the unwrapped line, so 180 and -180 are recognised as one
+            // position. A ring closed with the opposite sign to the one it opened with — both legal
+            // GeoJSON — would otherwise survive unclosing and then be rejected for the zero-length
+            // edge that closure left behind.
+            val longitudes = unwrapLongitudes(vertices)
+            fun samePosition(first: Int, second: Int): Boolean =
+                vertices[first].latitude == vertices[second].latitude &&
+                    longitudes[first] == longitudes[second]
```
Shahroz16: Could we compare the source positions while preserving `180`/`-180` equivalence, rather than require exact equality after accumulating unwrapped longitudes? In longitude/latitude order, `[(-122.1,37), (-122.1002,37), (-122.1003,37.0001), (-122.1,37)]` is rejected as self-intersecting, while the same ring without its closing position is accepted. Unwrapping leaves the closing longitude about `1.14e-13` from the first, so it survives as a spurious edge. A decimal-coordinate closed-ring regression would cover this.

Shahroz16 (commented):
The antimeridian query checks look good, including the negative-longitude case flagged by the bot. I reproduced a separate closed-ring regression below, so I can’t approve this yet.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonGeometry.kt
mrehan27: Fixed — and this was my regression from the previous round. Closure and duplicate collapse now compare the source positions with `normalizeLongitude` applied to the difference, so `180`/`-180` stay equivalent without inheriting the unwrap drift.

Covered by `from_whenClosedRingUsesDecimalCoordinates_thenCanonicalizesClosingVertex`, using your exact ring.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonGeometry.kt
```diff
@@ -22,44 +22,75 @@ internal enum class PolygonPointRelation {
...
         }
         return minimumDistance
     }
 
-    private fun PolygonCoordinate.distanceToSegmentMeters(
-        first: PolygonCoordinate,
-        second: PolygonCoordinate
+    /** [longitude] expressed on [ringLongitudes]' line, so ring and query share one frame. */
+    private fun onRingLine(longitude: Double): Double =
+        ringLongitudes[0] + normalizeLongitude(longitude - ringLongitudes[0])
```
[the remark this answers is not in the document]
mrehan27: Could not reproduce. The ring is unwrapped once into a contiguous line and every query is mapped onto that same line via `onRingLine`, so there is no per-point wrap left to overshoot.

Searched for a counterexample across ~24k randomized seam probes here, and another ~20k run independently in a separate checkout — no divergence in either. Leaving as-is.

Shahroz16 (approved):
The decimal closed-ring regression is fixed. I rechecked the closure and antimeridian cases, including negative-longitude queries; this is good to go.

PR #842, opened 2026-09-05
chore(geofence): record the fetched fence catalog, and stamp each log file with its own header

mahmoud-elmorabea:
Two independent diagnostics follow-ups, both small, both found while using the sink on a real drive.

### 1. Record the fetched fence catalog

A capture named fences only by opaque id. Nobody reading a log could tell which geoset a crossing belonged to, and a replay could not place the circles at all — the scenario transform emitted `fixture.api.fetch body:null` because the geometry is nowhere in the log.

The response already carries `name`, `latitude`, `longitude`, `radius` and `geosetIds`, and we discarded all of it. Each successful fetch now emits one record per fence, behind the existing build-time gate:

```
ev=fence.cataloged io=in id=11125 name=Momo_Dubai_Test gs=4471 lat=25.10967 lon=55.18406 rad=150 tt=enter,exit
```

Re-fetching the geometry from the workspace at replay time is **not** an equivalent alternative: fences move, so a drive replayed months later would silently run against today's circles rather than the ones it actually crossed — and a capture from a customer has no workspace to query.

Names are sanitized like any other value; a workspace-authored name can contain spaces, commas and `=`, each of which would break the parser's split. `gs` and `tt` join the composed-key set alongside `ids`, since their commas are structure rather than untrusted input.

Gated **whole** rather than gated-tail, and deliberately absent from the `invocations` table that asserts the gated-tail contract. These records carry no prose worth emitting on their own, so with diagnostics off they must not exist at all; putting the detail in the prose to satisfy that table would leak coordinates with the gate off. Their own tests pin the same guarantees.

### 2. Rebuild the file header per file

The header was built once at process start and the writer re-wrote that same string on every open, so a process outliving a day stamped each rolled-over file with the original session's timestamp. Observed on a device — four consecutive daily files, every one reading `ts=2026-09-02T23:53:44.454`.

Re-writing on open is still correct for the case it was added for (a same-day relaunch reuses the file and needs a fresh header, or later records correlate to the wrong boot). Passing a provider keeps that and makes the timestamp describe the file it heads.

### Verification

- 428 geofence unit tests pass, up from 423
- `apiCheck`, ktlint, and `:samples:java_layout:assembleDebug` clean
- End to end: injecting the new records into the real Dubai drive capture turns `fixture.api.fetch body:null` into the full geometry for all fetches

No public API change — `GeofenceLogger` is internal and the new parameter is defaulted.

### DIFF
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceLogger.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceLogger.kt
@@ -557,7 +557,11 @@ internal class GeofenceLogger(private val logger: Logger) {
     }
 
     /** Outcome of a nearby-geofence fetch. An **input**: replay feeds the response back. */
-    fun logApiFetchResult(returnedCount: Int, elapsedMillis: Long?) {
+    fun logApiFetchResult(
+        returnedCount: Int,
+        elapsedMillis: Long?,
+        regions: List<GeofenceRegion> = emptyList()
+    ) {
         logger.debug(
             "Fetched $returnedCount nearby geofence(s) from the server" +
                 tail(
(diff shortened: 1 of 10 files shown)


geofence/src/main/kotlin/io/customer/geofence/GeofenceLogger.kt
```diff
@@ -571,6 +622,43 @@ internal class GeofenceLogger(private val logger: Logger) {
...
+                "Geofence '${region.id}' catalogued" +
+                    tail(
+                        "fence.cataloged",
+                        GeofenceLogIo.INPUT,
+                        listOf(
+                            "id" to region.id,
+                            // Sanitized like any other value: a workspace-authored name can contain
+                            // spaces, commas and `=`, all of which would break the parser's split.
+                            "name" to region.name,
+                            "gs" to composedList(region.geosetIds),
```
mrehan27, line 652: **P2** — `composedList` doesn't sanitize its elements (unlike `list`), and `gs` being in `COMPOSED_KEYS` means only whitespace gets folded — so a server-supplied `geosetId` containing `=`, `,`, `:` or `|` breaks the tail's split. iOS uses `GeofenceLog.list(...)` here; `list(region.geosetIds)` matches it and still keeps the commas, same as `ids`.

mrehan27 (approved):
Cross-repo with customerio-ios#1253. The rename and the move below the durable write are both right — logging an acceptance that a rolled-back write could retract was the real problem.

One item inline on the new catalog record.

PR #843, opened 2026-09-05
chore(sample): filter noisy modules out of the diagnostics sink

mahmoud-elmorabea:
Stops non-geofence testing filling devices. Branched off `main`, independent of #842.

### Why

The sink captured every record the SDK emits. Measured across seven real captures: one **idle** simulator run was 23,136 records and 12.9 MB, of which **22,544 (94%) were in-app polling**, against nine geofence records — 0.03%. It grew at 28 KB/s doing nothing.

### The predicate

A **deny-list**, not an allow-list — a module we haven't met yet keeps being recorded rather than silently vanishing. Two rules guard against the deny-list being wrong:

1. **Errors are never dropped.** A filter must not be why a failure went unseen.
2. **A location keep-list runs *before* the deny rules.** It costs nothing today — every record it saves is already kept — but it means the next person adding a deny pattern cannot clip a geofence record by accident.

Tag rules alone would have fixed iOS and barely dented Android: `Logger.debug(msg, tag = null)` leaves the tag optional and **59–68% of Android records carry none**, that bucket being the in-app redux store. Hence the untagged prefix rules — including *both* shapes of the CDP envelope, since iOS logs it bare and Android prefixes it with prose. Matching only one meant the same payload was dropped on one platform and kept on the other, which is how the single outbound location body survived by accident.

### `dev` on change

The device-state snapshot was 19–31% of every file and rode every record, while changing rarely — 947 surviving records in the corpus carried 37 distinct snapshots. It now writes on change, on the first record of each file, and on a heartbeat. The heartbeat is **wall-clock**: `elapsedRealtime` resets per process and one file spans several.

### Honesty about what was removed

The active deny set is written into the file header. Without it a filtered file is silently lossy — nothing separates "in-app was quiet" from "in-app was removed".

Schema **1 → 2**, because `dev` stopped being mandatory. No parser exists in any repo yet, so the bump is free now and wouldn't be later.

### Measured result

Over all eight captures including the 2026-09-05 Dubai drive:

| | before | after |
|---|---|---|
| records | 26,646 | **896** (3.4%) |
| bytes | 14.3 MB | **361 KB** (2.5%) |
| location-bearing records lost | — | **0** of 235 |
| distinct geofence event kinds in the Dubai drive | 19 | **19** |

Per-capture retention ranges 10–48%; the 0.4% simulator figure is a pure in-app stress run and shouldn't anchor expectations.

### Toggle

A switch on the Location screen turns filtering off for when the bug *is* in a module the deny-list removes. The flip is itself recorded, so a capture says which way it was set.

### On tests

`samples/java_layout` has no unit test source set, and by decision it will not gain one — sample apps are not unit tested on either platform. The predicate is validated instead by replaying all eight real captures in `geofence-log-samples/`, which is how the numbers above were produced.

Worth knowing when reading them: that harness is a parallel Python implementation of the same predicate, so it can drift from the Kotlin. The two are short and side-by-side, but they are not the same code.

### DIFF
--- a/samples/java_layout/src/main/java/io/customer/android/sample/java_layout/diagnostics/DiagnosticEnvelope.kt
+++ b/samples/java_layout/src/main/java/io/customer/android/sample/java_layout/diagnostics/DiagnosticEnvelope.kt
@@ -27,7 +27,7 @@ internal object DiagnosticEnvelope {
         tag: String?,
         level: CioLogLevel,
         message: String,
-        deviceState: String
+        deviceState: String?
     ): String = buildString(message.length + 256) {
         append('{')
         append("\"v\":").append(DiagnosticLog.SCHEMA_VERSION)
@@ -40,7 +40,8 @@ internal object DiagnosticEnvelope {
         if (tag != null) append(",\"tag\":").append(json(tag))
         append(",\"lvl\":").append(json(level.wire()))
         append(",\"msg\":").append(json(message))
-        append(",\"dev\":").append(deviceState)
+        // Omitted when unchanged since the last record that carried it (schema 2).
+        if (deviceState != null) append(",\"dev\":").append(deviceState)
         append('}')
     }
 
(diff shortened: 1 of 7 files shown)


mrehan27 (approved):
Cross-repo with customerio-ios#1254. Deny-list over allow-list, errors never dropped, and location vocabulary checked first is the right shape — and the filter sits after the Logcat forward, so console output is untouched.

The item on the iOS twin applies here too: the first record after a midnight rotation can omit `dev`, because the cadence reset runs inside `append()` after the record was already built. Details on customerio-ios#1254.

PR #844, opened 2026-09-08
chore(geofence): improve geofence API response decoding

mrehan27:
Part of: [MBL-2388](https://linear.app/customerio/issue/MBL-2388/android-improve-geofence-api-response-decoding)

## What this is

Two decoding defects found by the iOS team during a cross-platform parity audit. Independent of
the MBL-2288 stack: this branch sits directly on `feature/geofence-polygons`.

**A padded shape discriminator dropped the record.** `toDomain` matched on `shape?.lowercase()`
with no trim, so `" circle "` missed the circle branch and fell through to the unsupported-shape
path. It now trims before matching, and a value that is empty afterwards routes the way a missing
field does — an empty string names no shape, so failing closed as if it named an unknown one was
wrong. iOS hit the same bug and fixed it on their side last week.

Two observable consequences worth stating. A response whose records all carried a blank shape used
to drop every record, throw "all N regions dropped", fail the refresh and preserve live state;
those records now map as circles and the refresh succeeds. The error-level unsupported-geometry log
for blank shapes also disappears. That safety net was incidental rather than chosen, and blank is
absent, so the new behaviour is the intended one.

Nothing the backend did not describe can register: the circle branch still drops a record carrying
`geometry` or `enclosing_circle`, and `toCircleRegionOrNull` still requires a valid centre and a
positive radius.

**A drop reason claimed a check that does not exist.** When a polygon's enclosing circle was
rejected the log read "backend-provided enclosing circle is invalid or does not contain the
polygon". There is no containment check on that path: `PolygonWakeCircleValidator` derives no
enclosing circle and re-checks no coverage, it only verifies the centre is in range and the radius
is finite, positive and inside the OS limit. It now names what is actually checked, covering all
four ways that branch is reached including an out-of-range centre.

Note for review order: #838 carries the same stale line, so whichever lands second reconciles it
and should take this wording.

## Tests

938 tests across geofence, core and messagingpush, 0 failures. ktlint, `lintDebug` and `apiCheck`
clean. No public API change.

Mutation-tested: reverting the trim fails exactly the padded and blank cases and nothing else.
Covers a padded circle, a padded polygon, a blank shape, a blank shape carrying geometry (which
must stay dropped), and an unknown shape asserting the raw server value reaches the log.

The 11 `lintDebug` warnings on this branch are all pre-existing in `GeofenceRegionStore.kt`, none
in the file this touches. Their cleanup lives in #837.

### DIFF
--- a/geofence/src/main/kotlin/io/customer/geofence/api/GeofenceApiResponse.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/api/GeofenceApiResponse.kt
@@ -202,7 +202,10 @@ internal fun GeofenceApiRegion.toDomain(
     polygonSupport: PolygonSupport = PolygonSupport.Disabled
 ): GeofenceRegion? {
     if (id.isBlank()) return null
-    return when (shape?.lowercase()) {
+    // Surrounding whitespace is not a shape the backend meant to name, and neither is an empty
+    // string: both are the field carrying nothing, so they route as if it were absent rather than
+    // dropping the record as an unsupported shape.
+    return when (shape?.trim()?.lowercase()?.takeIf(String::isNotEmpty)) {
         null, CIRCLE_SHAPE -> {
             if (geometry != null || enclosingCircle != null) {
                 SDKComponent.geofenceLogger.logPolygonDropped(id, "shape discriminator is missing or inconsistent")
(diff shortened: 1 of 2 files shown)


Shahroz16 (approved):
The padded and blank shape handling looks good, and the geometry inconsistency guard still holds. All 66 response-mapping tests and the geofence module build pass. No remaining concern in this PR.

PR #845, opened 2026-09-08
fix(deps): update Faraday security pin to 1.10.6

Shahroz16:
Updates the Fastlane build dependency from Faraday 1.10.5 to 1.10.6, the patched 1.x release for [CVE-2026-54297](https://rubysec.com/advisories/CVE-2026-54297/). Regenerated the lockfile conservatively; no other dependency versions changed.

Verified a frozen bundle install, exact Faraday 1.10.6 resolution, and Fastlane 2.237.0 loading and CLI startup with Ruby 3.4.4. Android formatting/lint and `:core:assembleDebug` also passed.

This changes Ruby build tooling only. Mobile runtime behavior and deployment targets are unchanged.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,7 @@ source "https://rubygems.org"
 
 # Security pins for transitive deps (MBL-1694)
 gem 'webrick', '1.8.2'
-gem 'faraday', '1.10.5'
+gem 'faraday', '1.10.6'
 gem 'rexml', '3.3.3'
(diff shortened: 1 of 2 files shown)


mrehan27 (approved):
Dependency change looks right — `Gemfile` plus both lockfile sections, faraday the only gem that moved, and it's the repo's only Gemfile.

`fix(deps):` will cut a patch release on merge; customerio/customerio-ios#1257 used `chore(deps):` for the same fix.

PR #846, opened 2026-09-08
chore(geofence): stamp cache freshness only for the session that fetched it

mrehan27:

## Stack

- [#837 stage transitions durably before delivery](https://github.com/customerio/customerio-android/pull/837)
- 👉 this PR, and [#838 polygon location engine](https://github.com/customerio/customerio-android/pull/838), both branch from #837

## What this is

Found while sweeping for the class of bug behind the launch-time session finding on #837, and kept
separate so that review does not reopen for it.

`registerNearestAndPersist` decides `routingArmed` at `saveRoutableRegisteredIdsIfCurrent`, then runs
six more store writes before `onRegistered` stamps the cache fresh. An identify landing in that gap
takes the departing pass's fetch as the new session's freshness. That session's own refresh then sees
a fresh cache, skips, and leaves routing empty until a movement EXIT re-arms it, which is the state
#837 exists to prevent.

The pass now offers the generation it snapshotted, and the store writes the anchor and the sync
timestamp only while that generation is still current, under the same lock `beginUserSession` takes,
so the check and the write cannot be separated.

`saveLastApiFetchLocation` and `setLastSyncTimestamp` are private now. Neither had a production caller
left, and leaving them on the interface meant the next caller reaching for the obvious name reopens
the window.

## Tests

617 geofence tests, 0 failures. `lintDebug` 0 issues, `apiCheck` and ktlint clean, no compiler
warnings.

Two mutations, each killing exactly one test and nothing else: dropping the generation check kills the
store refusal test, and re-reading the generation instead of passing the snapshot kills the plumbing
test.

Worth knowing: no test drives the real store through the repository, so the production skip is pinned
in two halves, the store refusal and the offered generation, rather than end to end.

### DIFF
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
@@ -407,11 +407,21 @@ internal class GeofenceRepositoryImpl(
                     // Cache + anchor + timestamp only on remote fetch; Tier A reuses them.
                     // Skip the config save when backend didn't ship one this response —
                     // a null parse must not clobber a previously cached value.
-                    onRegistered = {
+                    onRegistered = { userStateGeneration ->
                         store.saveCachedRegions(regions)
                         parsedConfig?.let { store.saveCachedConfig(it) }
-                        store.saveLastApiFetchLocation(GeofenceLocation(latitude, longitude))
-                        store.setLastSyncTimestamp(clock.currentTimeMillis())
+                        // Offered with the pass's own generation, not a re-read: an identify
+                        // landing after routing was armed would otherwise stamp the new session's
+                        // cache fresh off this pass, and its refresh would then SKIP and leave
+                        // routing empty until a movement EXIT re-armed it.
+                        val stamped = store.saveApiFetchStateIfCurrent(
+                            location = GeofenceLocation(latitude, longitude),
+                            syncTimestamp = clock.currentTimeMillis(),
+                            expectedUserStateGeneration = userStateGeneration
+                        )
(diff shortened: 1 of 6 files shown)


Shahroz16 (commented):
The freshness guard works in the real repository/store regression: identifying after routing is armed leaves the new session stale, and its own refresh fetches and rearms routing. The same test fails on the base. I found no additional issue in this patch, but the [legacy user-switch case in #837](https://github.com/customerio/customerio-android/pull/837#discussion_r3933767979) also reproduces here because that switch leaves the generation unchanged. I am holding approval for the inherited issue.

mrehan27:
Rebased onto #837's session-ownership rework and force-pushed. One content change beyond the rebase: #837 added a test that called `saveLastApiFetchLocation` and `setLastSyncTimestamp`, which this PR makes private, so that new caller is migrated to `saveApiFetchStateIfCurrent` here. The rebase compiled clean but the test source did not, so it is worth knowing the change is not purely mechanical.

612 geofence tests, 0 failures, lint 0 issues, apiCheck and ktlint clean. Nothing else in this PR changed.

Shahroz16 (commented):
The freshness fix still passes the real repository/store regressions on this head. The earlier legacy-switch issue is fixed, but the [launch-versus-identify race in #837](https://github.com/customerio/customerio-android/pull/837#discussion_r3963963700) also reproduces here: the first callback clears routing after a successful refresh. I am holding approval for that inherited issue.

Shahroz16 (approved):
The inherited launch race is fixed, and the session-scoped freshness guard still passes the real repository/store regressions. All 224 focused tests and the geofence build passed on this head.

PR #849, opened 2026-09-09
refactor(geofence): move the decisions out of the OS wrappers

mahmoud-elmorabea:
Android counterpart to customerio-ios#1258. Trimmed to what the recorded-drive replay harness actually drives.

The SDK's decisions lived inside the classes that wrap the OS, so substituting the OS in a test substituted the decisions too. This moves the line a test substitutes along onto the line where the OS starts. No behaviour change intended.

### Seams

- `GeofenceCrossingPipeline` + `GeofenceCrossing`: the receiver's routing (orphan drop and self-heal, movement routing, containment bookkeeping, unmatched-exit guard, anonymous drop) behind a crossing in SDK terms. The `BroadcastReceiver` keeps what needs an OS: the intent, the error channel, the `goAsync` window.
- `GeofenceRegistrar`: an interface over the four `GeofencingClient` calls, with the DI graph and repository typed on it. `GeofenceManager` implements it; its registration body is unchanged from `main`.
- `GeofenceForegroundCoordinator`: the foreground gates and the awaiting-a-fix retry, out of the `ProcessLifecycleOwner` callback. Identify and app-launch take their anchor from it too.

### Logging contract

Every geofence record states `io=in|out|obs`. Outputs are exactly `registration.applied`, `transition.accepted` and `module.reset`; decisions about inputs (drops, suppressions, synthesised enters) are observations. New inputs: `identity.changed`, `location.fix` on arrival, `permission.changed` on foreground entry. `module.reset` carries the same key as iOS.

### Not here

The replay harness and its scenario tests. Geofence unit tests: 441 passed, read from the JUnit XML.

Fixes MBL-2309


mahmoud-elmorabea:
Superseded by the stack on `testing/geofencing-replay-harness`: #861 (log io contract) → #862 (registrar interface) → #863 (crossing pipeline) → #864 (foreground coordinator). Same content, split by idea and with narrative comments stripped.

PR #851, opened 2026-09-09
chore(geofence): adopt main's geofence log diagnostics

mrehan27:

## Why this exists

`feature/geofence-polygons` is 17 commits behind `main`, and the merge conflicts in 8 files / 498 lines. Both sides rewrote the same geofence delivery path: main added structured `key=value` log tails, this branch replaced the cooldown claim/rollback with a split `isAllowed` + `record`.

Resolving that inside one direct-push merge would put a lot of judgement somewhere nobody reviews it. So main's content lands here as reviewed PRs first.

Correction, added after the merge: this stack did not shrink the merge (37 hunks became 44). What it bought is that every production conflict resolved to ours, and main's diagnostics arrived already reviewed. Without #854 the merge would have silently auto-applied main's `logRankEvaluated` calling `edgeDistanceTo`, which does not exist on this branch.

- 👉 **this PR** — the diagnostics: `GeofenceDiagnostics`, `GeofenceLogTail`, `GeofenceLogger`
- next — `isAllowed` returns remaining seconds, so `cooldownRemainingSeconds` has a source
- next — drop blank geoset ids, matching main and iOS

## What this is

`GeofenceDiagnostics.kt`, `GeofenceLogTail.kt` and `GeofenceLogTailTest.kt` are taken verbatim from main. `GeofenceLogger.kt` is main's file plus the 11 log methods that exist only on this branch.

Verified rather than assumed: the three new files are byte-identical to main, the logger diff against main is **+59 / −0** with zero removed lines, and both sides carry **63** `tail(` sites. All 71 methods across both parents are present, with bodies and doc comments intact.

## Two of main's methods described behaviour this branch deleted

Taking main's version wholesale imported prose, and in one case a machine field, that contradicts the code here.

**`logPersistFailed`** claimed it "rolled back cooldown so a later crossing can retry". There is no cooldown rollback on this branch. Kept this branch's wording, with main's tail.

**`logEventDeliveryFailed`** claimed HTTP delivery "will not retry", and emitted `retry=false`. On this branch a refused row is dropped only when the cause is a terminal `HttpRequestFailure` **and** `store.remove` succeeds; a non-terminal cause or a failed removal keeps the row and returns `Result.retry()`. So a constant is wrong in one direction or the other. It now takes `willRetry`, computed from the same expression that decides disposal, so the field and the behaviour cannot drift.

`logUnknownTransition` gained a required `geofenceId` on main. Its one call site is updated.

## What is deliberately not here

**14 of main's log methods have no production caller yet** (`logModuleInitialized`, `logCallbackReceived`, `logRegionsRegisteredIds`, `logTransitionAccepted` and others). Their callers live in the files the later pieces touch, so a diagnostics-enabled log will not yet mark a launch or a callback. Expected, not a bug.

**The 11 methods carried from this branch emit no tail.** Writing new diagnostic schemas is a change worth designing, not something to bury in a merge-prep PR. Three of them (`logTransitionDroppedUnarmedId`, `logTransitionDroppedRetiredId`, `logGmsCallTimedOut`) leave a replay gap, since a callback ending in an unarmed or retired drop produces no `out` record; those are worth doing when the delivery path is reconciled.

## Tests

1047 across geofence, core and messagingpush, 0 failures. `lintDebug` 0 issues, `apiCheck` and ktlint clean, no new compiler warnings.

`GeofenceLogTailTest` asserts field *names* only, so nothing pinned the new `retry` value. `GeofenceEventWorkerTest` now captures the emitted log, and three mutants each kill exactly one test: hardcoding `retry` false, hardcoding it true, and dropping the row regardless of whether the removal succeeded.

Not exercised on a device.

### DIFF
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
@@ -160,7 +160,7 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
                 Geofence.GEOFENCE_TRANSITION_ENTER -> Event.GeofenceTransition.ENTER
                 Geofence.GEOFENCE_TRANSITION_EXIT -> Event.GeofenceTransition.EXIT
                 else -> {
-                    logger.logUnknownTransition(gmsTransitionType)
+                    logger.logUnknownTransition(geofenceId, gmsTransitionType)
                     return@forEach
                 }
             }
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceDiagnostics.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceDiagnostics.kt
@@ -1,0 +1,51 @@
+package io.customer.geofence
+
+import android.content.pm.PackageManager
+import io.customer.sdk.core.di.SDKComponent
+
+/**
+ * Whether the SDK emits the diagnostic tail.
+ *
+ * Read from the host app's manifest, deliberately not from an API. Anything reachable — even
+ * behind an opt-in annotation — is something a customer app can switch on, and the tail is the
+ * only thing keeping coordinates out of a production log.
+ *
+ * Enable with, in the app's `AndroidManifest.xml`:
+ * ```xml
+ * <meta-data android:name="io.customer.geofence.diagnostics" android:value="true" />
+ * ```
+ */
+internal object GeofenceDiagnostics {
+    const val MANIFEST_KEY = "io.customer.geofence.diagnostics"
+
+    @Volatile
+    private var override: Boolean? = null
+
+    @Volatile
+    private var cached: Boolean? = null
+
+    val isEnabled: Boolean
+        get() {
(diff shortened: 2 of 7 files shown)


Shahroz16 (commented):
The diagnostics port looks sound, and the focused tests pass on the completed stack. Could we rerun the [Android smoke check](https://github.com/customerio/customerio-android/actions/runs/34395900073/job/102615458891) before approval? Its artifact shows Maestro's device server dying during viewHierarchy after the first launch went offline, rather than an SDK assertion failure. I am holding approval for that red check, not asking for a code change.

mrehan27:
@Shahroz16 re-ran it and it's green, 50/50 now.

It was the emulator, not the code: the log shows the device going offline during the first app launch and the single retry failing behind it. #852, #853 and #854 all contain this exact commit (77252dbf) and all passed Android smoke, so the code can't have been the cause.

Ready for another look.

Shahroz16 (approved):
Thanks for rerunning it. Android smoke passed on the same reviewed commit, and all 50 checks are now green. That clears the only remaining hold on this diagnostics port.

PR #852, opened 2026-09-09
chore(geofence): report the cooldown remainder instead of a boolean

mrehan27:

## Stack

- 👉 **this PR** — the cooldown reports its remainder
- next — drop blank geoset ids

## What this is

main's suppression log carries `cooldownRemainingSeconds`. That field has no source on this branch: #837 replaced the cooldown's claim-and-rollback with an `isAllowed` check plus a separate `record` write, and the check returned a `Boolean`.

`suppressedForSeconds` returns `null` to proceed, otherwise the seconds left on the window. The check already computes that figure, so returning it costs no second store read on a background wake.

**Not named `tryAcquire` like main.** Nothing is acquired here, and `record()` remains the separate write with nothing to roll back. Reusing main's name would re-import the confusion #837 removed.

## The boundary is unchanged

Old: proceed when `last == null || (now - last) >= cooldownMs`. New: `null` when there is no previous emit, or when `elapsed >= cooldownMs`. Same decision at `elapsed == cooldownMs`, still covered by `cooldown_givenPreviousEmitExactlyAtCooldownBoundary_expectAllowedAndRecorded`.

## Tests

1048 across geofence, core and messagingpush, 0 failures. `lintDebug` 0 issues, `apiCheck` clean, ktlint clean.

About 35 stubs moved from `returns true/false` to `returns null/12.0`, which is where an inverted sense would hide, so two assertions now pin the value rather than its nullness: `suppressedForSeconds_givenHalfTheWindowElapsed_expectTheRemainderInSeconds`, and an emitter check that the reported figure reaches `logTransitionSuppressed`. Replacing the remainder with a constant kills exactly the first.

Two test names said `expectTrue`/`expectFalse` and now describe the returned remainder instead.

Not exercised on a device.

### DIFF
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceCooldownFilter.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceCooldownFilter.kt
@@ -16,20 +16,23 @@ internal class GeofenceCooldownFilter(
     private val clock: Clock
 ) {
     /**
-     * Atomically checks the cooldown and records the emit if allowed. Returns true if the
-     * caller should proceed to emit, false if the transition is within the cooldown window.
+     * `null` when the caller should proceed to emit, otherwise the seconds still left on the
+     * window. The check already computes that figure, so returning it rather than recomputing
+     * keeps reporting a suppression free of a second store read on a background wake.
+     *
+     * Checks only. [record] is the separate write, so there is nothing to roll back.
      */
     @Synchronized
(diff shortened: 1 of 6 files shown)


Shahroz16 (approved):
The remaining-seconds result preserves the cooldown boundary and recovery bypass. The focused cooldown/emitter tests pass on the completed stack, and this head's checks are green.

PR #853, opened 2026-09-09
chore(geofence): drop blank geoset ids from the transition fan-out

mrehan27:

## Stack

- [#852 report the cooldown remainder instead of a boolean](https://github.com/customerio/customerio-android/pull/852)
- 👉 **this PR** — drop blank geoset ids

## What this is

The one **behaviour** change in the whole merge, which is why it is on its own rather than folded into a diagnostics PR.

A catalog row carrying an empty geoset id persisted an entry with an empty `geosetId` and reported two events where iOS reports one. Blanks are now dropped alongside duplicates, matching main and iOS.

Dropping every id falls through to the existing no-geosets path, so one null-geoset event is still emitted rather than none.

Matches main exactly, including that `isNotEmpty()` is not `isBlank()`: an id of `" "` is still kept on both platforms. Deliberately not "improved" here, since divergence is the thing this sequence is removing.

## Tests

1050 across geofence, core and messagingpush, 0 failures. `lintDebug` 0 issues, `apiCheck` and ktlint clean.

Two new cases, and removing the filter kills exactly those two: `emit_givenBlankGeosetAlongsideReal_expectBlankDropped` and `emit_givenEveryGeosetBlank_expectOneNullGeosetEvent`.

Not exercised on a device.

### DIFF
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceTransitionEmitter.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceTransitionEmitter.kt
@@ -206,9 +206,11 @@ internal class GeofenceTransitionEmitter(
             // One transitionId shared across the per-geoset fan-out.
             val transitionId = UUID.randomUUID().toString()
             val name = geofenceName?.takeIf { it.isNotEmpty() }
-            // One event per geoset; no geosets → one null-geoset event. Distinct so a repeated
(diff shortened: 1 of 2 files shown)


Shahroz16 (approved):
The empty-geoset filter preserves the real IDs and the no-geoset fallback. The focused fan-out and recovery tests pass on the completed stack, and this head's checks are green.

PR #854, opened 2026-09-09
chore(geofence): port main's remaining geofence diagnostics

mrehan27:

## Stack

- [#852 report the cooldown remainder instead of a boolean](https://github.com/customerio/customerio-android/pull/852)
- [#853 drop blank geoset ids from the transition fan-out](https://github.com/customerio/customerio-android/pull/853)
- 👉 **this PR** — port the remaining diagnostics

## What this is

#851 brought in 14 log methods with no callers. This wires up the 12 that have a place here:

| Where | Records |
|---|---|
| `ModuleGeofence` | app-start entry point |
| `GeofenceBootReceiver` | boot-restore wake |
| `GeofenceBroadcastReceiver` | the OS callback, plus two unparseable-intent cases |
| `GeofenceRepository` | what storage held, what the API returned, how ranking resolved, what ended up registered |
| `GeofenceTransitionEmitter` | an accepted transition, below the durable write |

The receiver's record is placed before any routing decision and before the `Location` is narrowed to two doubles, since that is the only point where the OS's own triggering fix still exists.

## Two adaptations, not straight copies

**Edge distances.** main calls `edgeDistanceTo`, which this branch replaced with the nullable `edgeDistanceToOrNull` for polygons. A polygon whose ring never arrived now reports no distance, matching what ranking already does with it, rather than one derived from its coarse circle. The ring is re-derived rather than reused from ranking, which costs nothing unless diagnostics are enabled.

**API failures.** `logApiFetchFailed` replaces `logSyncFailed` on the fetch branch only. A response-mapping failure still reports as a sync failure, as it does here today.

## One behaviour change worth seeing

`logSyncSucceeded` now counts `idsToSave` minus the movement trigger, rather than `nearest`. When a stale fence's removal fails, the OS is still monitoring it, so it is included.

That flips one existing expectation from 1 to 2, in `refresh_givenAddSucceedsButStaleRemovalFails_expectUnremovedStalePreserved` — a test whose own comment already says the unremoved id is preserved and will be retried. The old number described what ranking selected; the new one describes what is actually monitored.

## Still without callers, deliberately

`logEventInvalidInput` and `logEventWorkerEntryMissing` describe a worker that looks up a single row by key. This branch drains the oldest row in a loop instead, so those paths do not exist here. `logReceiverSkipped` has no production caller on main either.

## Tests

1051 across geofence, core and messagingpush, 0 failures. `lintDebug` 0 issues, `apiCheck` and ktlint clean, no new compiler warnings.

Not exercised on a device.

### DIFF
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceBootReceiver.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceBootReceiver.kt
@@ -32,6 +32,7 @@ class GeofenceBootReceiver : BroadcastReceiver() {
         try {
             SDKComponent.setupAndroidComponent(context = context)
             val scope = SDKComponent.scopeProvider.geofenceScope
+            logger.logModuleWoke(GeofenceLaunchReason.BOOT_RESTORE)
             logger.logSyncTriggered(REASON_BOOT_RESTORE)
             scope.launch {
                 try {
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
@@ -61,16 +61,32 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
 
     @VisibleForTesting
     internal suspend fun handleGeofencingEvent(geofencingEvent: GeofencingEvent?) {
-        if (geofencingEvent == null) return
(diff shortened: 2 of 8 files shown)


Shahroz16 (approved):
The diagnostic records now follow the actual persistence and retry outcomes, and expensive fields remain gated off when diagnostics are disabled. The geofence build and all 246 focused tests passed on this head; CI is green.

PR #855, opened 2026-09-10
chore(geofence): drop a dead log method and align the test logger override

mrehan27:

Post-merge cleanup after `main` landed on `feature/geofence-polygons` (71d3f314).

- `logTransitionEmitting` lost its only caller in #854 when the accepted record moved below the durable write. main never had the method, so the merge could not bring it back.
- The receiver test overrode `GeofenceLogger`; `Logger` is what main's version and the rest of the suite override, and it reaches the same singleton.
- Our worker drains the oldest row rather than a keyed one, so main's already-delivered record lost its call site and an empty-queue wake returned silently. Restored, emitted only when the wake found nothing: a full drain exits through the same branch.

Heads-up for reviewers: `why=queue_empty` is a new token. It does not exist on main or iOS. The old `already_delivered` named the foreground flush, but the common cause is a sibling node in the same chain — the emitter enqueues one node per entry and the worker drains every row, so a geoset fan-out has the first node deliver both.

1058 tests, ktlint, apiCheck and lint all clean.

### DIFF
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceLogger.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceLogger.kt
@@ -822,13 +822,13 @@ internal class GeofenceLogger(private val logger: Logger) {
         )
     }
 
-    fun logEventWorkerEntryMissing(key: String? = null) {
+    fun logEventWorkerEntryMissing() {
         logger.debug(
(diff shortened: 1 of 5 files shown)


Shahroz16 (approved):
The removed method has no callers, and accepted-transition logging still happens after the durable write. The Logger override reaches the same geofence logger through DI. The geofence build and all 50 receiver tests pass, and the independent review is complete. Looks ready.

PR #856, opened 2026-09-11
chore(core): keep the pending delivery queue when one row or one read fails

mrehan27:
Part of: [MBL-2413](https://linear.app/customerio/issue/MBL-2413/mobile-make-pending-geofence-delivery-resilient-to-corrupt-rows)

`readAll` decoded the file as one list, so a single undecodable row was caught as "treating as empty" and the next `append` rewrote the file without the rows still queued. Rows now decode individually and a bad one is dropped and counted.

A failed read reports unreadable rather than empty, and no mutating operation writes over it. `remove` and `claim` used to return success after such a read, marking a delivery done while its row was still on disk. `contains()` replaces the presence checks in the claim helpers, which had the same conflation.

Pre-existing since #715 and present on main, so messagingpush is exposed the same way via `PendingPushDeliveryMetric`.

Residual: a file that is not valid JSON at all still drops everything, which needs filesystem corruption given the atomic rename. The geofence worker still maps unreadable to empty at its own `loadAll`, so it reports `queue_empty` there — a null-aware `loadAll` is a follow-up.

2043 tests across all modules, ktlint, apiCheck and lint clean.

### DIFF
--- a/core/src/main/kotlin/io/customer/sdk/data/store/PendingDeliveryClaim.kt
+++ b/core/src/main/kotlin/io/customer/sdk/data/store/PendingDeliveryClaim.kt
@@ -65,7 +65,7 @@ suspend fun <T : PendingDeliveryStore.PendingDeliveryEntry> PendingDeliveryStore
         // A false claim is either another channel taking the entry (now gone) or the removing write
         // failing (still present). Only the former was delivered elsewhere; a still-present entry was
         // claimed by no one, so retry rather than reporting success.
-        return if (get(entry.key) == null) {
+        return if (contains(entry.key) == false) {
             PendingDeliveryResult.AlreadyClaimed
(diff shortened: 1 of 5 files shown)


Shahroz16 (approved):
The mixed-row and unreadable-file paths preserve queued rows as intended. The focused store, claim, geofence and push tests passed, and both queue regressions failed on the base commit and passed here.

Could we also switch the stale `get(deliveryId)` stub in `PushDeliveryMetricsWorkerTest.doWork_givenEntryAlreadyDelivered_expectSkipWithoutTracking` to `contains(deliveryId) returns false`? The test passes today, but it now relies on the relaxed mock default. This is test cleanup, not an approval hold. The separate worker retry work remains covered by MBL-2419.

PR #857, opened 2026-09-11
chore(geofence): catalogue a polygon as a polygon

mrehan27:
Part of: [MBL-2418](https://linear.app/customerio/issue/MBL-2418/mobile-add-structured-diagnostics-for-polygon-geofences)

## Stack
- 👉 [#857 catalogue a polygon as a polygon](https://github.com/customerio/customerio-android/pull/857)
- [#858 give the polygon records a machine-readable tail](https://github.com/customerio/customerio-android/pull/858)
- [#859 catalogue a fence before it can be dropped](https://github.com/customerio/customerio-android/pull/859)

The fence catalog emitted `lat`/`lon`/`rad` straight off the region. For a polygon those describe the registered trigger circle, and that radius is the backend's enclosing circle plus our 1000m platform margin — so every polygon catalogued as a placeable circle a kilometre too big, with nothing marking it as not the fence. A replay placed it and got a confident wrong region rather than a visible gap.

`sh` names the shape, `nv` and `ring` carry the geometry, and `rad` now reports the backend's radius so the field means the same thing on both platforms. A polygon with no backend radius omits `rad` rather than substituting the padded one.

`ring` joins the composed keys: sanitizing folds its commas and runs every coordinate into one token that parses as garbage instead of failing.

`nv` counts the canonical ring — duplicates collapsed, closing vertex dropped — matching what membership is computed against, and stays authoritative against the 25-entry cap so a consumer can detect truncation and refuse.

Field names and semantics agreed with iOS (#1263).

2039 tests across all modules, ktlint, apiCheck and lint clean.

### DIFF
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceLogTail.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceLogTail.kt
@@ -69,7 +69,7 @@ internal object GeofenceLogTail {
      * The only values that compose the format's separators on purpose. Everything else is treated
      * as an untrusted token: geofence identifiers are workspace-authored and can hold anything.
      */
-    private val COMPOSED_KEYS = setOf("ranked", "evicted", "ids", "gs", "tt")
+    private val COMPOSED_KEYS = setOf("ranked", "evicted", "ids", "gs", "tt", "ring")
 
     /**
      * Applied to every finished value. Only whitespace, which is what separates one `key=value`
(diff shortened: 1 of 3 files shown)


Shahroz16 (approved):
The catalog keeps the backend's unpadded radius and a canonical vertex count, with detectable ring truncation. The focused JVM tests passed. The pre-drop catalog coverage is already implemented in #859, so I am not asking for a backport here.

PR #858, opened 2026-09-11
chore(geofence): give the polygon records a machine-readable tail

mrehan27:
Part of: [MBL-2418](https://linear.app/customerio/issue/MBL-2418/mobile-add-structured-diagnostics-for-polygon-geofences)

## Stack
- [#857 catalogue a polygon as a polygon](https://github.com/customerio/customerio-android/pull/857)
- 👉 [#858 give the polygon records a machine-readable tail](https://github.com/customerio/customerio-android/pull/858)
- [#859 catalogue a fence before it can be dropped](https://github.com/customerio/customerio-android/pull/859)

Five polygon records emitted prose only. A capture could not tell a polygon that was dropped from one that was registered and never fired, which is the question these records exist to answer.

`registration.rejected` for the three decode-time drops, `rank.excluded` for a region left out of a ranking pass, `polygon.undecided` for a fix that cannot settle containment. A refused fix deliberately does not use `os.callback.dropped`: that key is what makes received-vs-dropped countable, and a location update has no receipt to net against.

Reasons move from prose strings to enums carrying `(wire, detail)`. The prose is unchanged, and the token is fixed so rewording a sentence cannot move what analysis groups on. Tokens agreed with iOS; `runtime_unsupported`, `fix_too_old` and `rank.excluded` are Android-only and marked as such.

`rank.excluded` is separate from `registration.rejected` because one is per fetch and the other repeats per ranking pass on the same fence. The per-pass `rank.evaluated` already lists the fence under `evicted`; this record adds the reason, which a shared id list cannot carry.

Tests now pin `io` and the method name per row, so every record is covered rather than the five touched here. That found `logApiFetchFailed`, which emits a tail and had no test row at all.

690 geofence tests, ktlint, apiCheck and lint clean.

### DIFF
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceDistanceFilter.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceDistanceFilter.kt
@@ -66,12 +66,12 @@ internal class GeofenceDistanceFilter(
         longitude: Double
     ): Float? {
         if (region.isPolygon && !polygonSupport.isPolygonMonitoringEnabled) {
-            logger.logPolygonRegionNotRanked(region.id, "polygon monitoring is not enabled in this build")
+            logger.logPolygonRegionNotRanked(region.id, PolygonNotRankedReason.RUNTIME_UNSUPPORTED)
             return null
         }
         val distance = region.edgeDistanceToOrNull(latitude, longitude, cachedGeometry(region))
         if (distance == null) {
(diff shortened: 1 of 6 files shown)


Shahroz16 (approved):
The reason enums keep the existing decisions intact, and the focused log-tail and API tests passed. The structured fields remain behind the diagnostics gate.

PR #859, opened 2026-09-11
chore(geofence): catalogue a fence before it can be dropped

mrehan27:
Part of: [MBL-2419](https://linear.app/customerio/issue/MBL-2419/android-improve-geofence-diagnostics-and-queue-failure-handling)

## Stack
- [#857 catalogue a polygon as a polygon](https://github.com/customerio/customerio-android/pull/857)
- [#858 give the polygon records a machine-readable tail](https://github.com/customerio/customerio-android/pull/858)
- 👉 [#859 catalogue a fence before it can be dropped](https://github.com/customerio/customerio-android/pull/859)

The fence catalog ran on mapped regions, so a record validation rejected produced no row and the capture silently lacked the fence that failed, which is the case the catalog exists to diagnose. A response where every record dropped produced nothing at all: the mapper throws, and the fetch record sat after it.

Now built from the wire. `GeofenceCatalogEntry` describes a record as the server sent it, `toCatalogEntries()` lives in the api package so the logger never sees the wire type, and the fetch record moves above the mapping.

`sh` carries the claim rather than a verdict: an absent discriminator reads as `circle` exactly as the mapper reads it, and an unsupported value comes through verbatim. A record claiming no shape while carrying polygon geometry catalogs as `sh=circle` **and** keeps its ring, so the capture shows a mis-described record instead of a plain circle. `registration.rejected why=undescribed_shape` names the conflict.

For every record the old mapper accepted the emitted fields are unchanged, verified field by field: a polygon's `lat`/`lon` were already the enclosing-circle centre, `rad` the same wire radius, `nv`/`ring` the same canonical vertices.

`resolveTransitionTypes` reported unknown values from inside its own mapping, so the catalog resolving the same list would have double-reported every one. Split into a logging wrapper and a pure resolver, pinned by a test.

The catalog parameter is a lambda evaluated after the diagnostics gate, so a build with diagnostics off no longer decodes and validates every polygon ring a second time per fetch.

2045 tests across all modules, ktlint, apiCheck and lint clean.

### DIFF
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceCatalogEntry.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceCatalogEntry.kt
@@ -1,0 +1,26 @@
+package io.customer.geofence
+
+import io.customer.geofence.polygon.PolygonCoordinate
+
+/**
+ * One fetched record as the server described it, built before validation decides whether it can be
+ * monitored.
+ *
+ * Deliberately not the mapped region: a malformed record is the one a capture most needs named, and
+ * it never becomes a region. Fields are nullable because this is the wire's account of a fence
+ * rather than a usable one, so a row carries whatever survived and omits the rest.
+ */
+internal data class GeofenceCatalogEntry(
+    val id: String,
+    val name: String?,
+    val geosetIds: List<String>,
+    /** The shape the server claimed, which for an unsupported value is that value verbatim. */
+    val shape: String,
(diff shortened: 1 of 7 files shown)


Shahroz16 (approved):
The catalog now runs before region validation, so rejected records remain visible, and the diagnostics-off path stays lazy. The repository, API and log-tail tests passed. This covers the catalog portion of MBL-2419; the queue retry follow-up remains separate.

PR #860, opened 2026-09-11
chore(geofence): stop reading an unreadable delivery queue as an empty one

mrehan27:
Part of: [MBL-2419](https://linear.app/customerio/issue/MBL-2419/android-improve-geofence-diagnostics-and-queue-failure-handling)

`loadAll()` returns an empty list for a file it could not read. So the event worker emitted `why=queue_empty` — a record asserting an earlier node drained the queue, which was never established — and returned success while the rows sat on disk with nothing coming back for them. The foreground flush had the same defect one layer up, reporting a snapshot of zero for a queue it never read.

`loadAllOrNull()` draws the distinction the store already draws for a single key in `contains()`. The worker retries on it, capped at 5 attempts against `runAttemptCount`, then gives up the wake. Nothing is ever written over, so giving up costs only that wake: the next transition enqueues a fresh worker and the flush retries independently. `Result.failure()` is safe here where the delivery paths below it avoid it, because every dependent in the chain would fail the same read and the rows behind are untouched; the comment says so at the site, since the file argues the opposite case seventy lines down.

The flush reports `onUnreadable()` rather than a false snapshot. Skipping was right as the action but wrong as the trace: the flush is the one path that runs without a new transition, so staying silent meant a queue that went unreadable while the device sat idle left no record at all — the same defect this ticket names, moved up a layer.

Both paths file `ev=queue.unreadable why=read_failed`, with `via` separating a worker wake from a foreground flush. That token is iOS's, agreed cross-platform and already on their `main`.

Also folded in: the test cleanup Shahroz asked for on #856, which merged before it was addressed. `PushDeliveryMetricsWorkerTest` stubbed `get(deliveryId)`, but `claimSendRestore` now asks `contains()`, so the stub was dead and the test passed on a relaxed mock's `Boolean?` default of false.

2063 tests across all modules, ktlint, apiCheck and lint clean.

### DIFF
--- a/geofence/src/main/kotlin/io/customer/geofence/worker/GeofenceEventWorker.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/worker/GeofenceEventWorker.kt
@@ -81,14 +81,9 @@ internal class GeofenceEventWorker(
         var sawEntry = false
         while (true) {
             val queued = store.loadAllOrNull() ?: run {
-                // Unreadable, which is not empty: the rows are still on disk and nothing has been
-                // sent. Reporting success here would say the queue drained.
-                //
-                // Result.failure() is safe here where the delivery paths below avoid it: a terminal
-                // node cancels the chain's dependents, but every one of them would fail this same
-                // read, and the rows they would have sent are untouched on disk for the next
-                // transition or the foreground flush. Retrying past the cap only spins on a file
-                // that no number of attempts recovers.
+                // Unreadable is not empty: nothing was sent, so success would claim a drain.
+                // failure() is safe here where the paths below avoid it — every dependent would
+                // fail this same read, and the rows behind are untouched on disk.
                 val willRetry = runAttemptCount < GeofenceConstants.MAX_UNREADABLE_QUEUE_ATTEMPTS
                 logger.logEventWorkerQueueUnreadable(runAttemptCount, willRetry)
                 return if (willRetry) Result.retry() else Result.failure()
--- a/geofence/src/main/kotlin/io/customer/geofence/worker/GeofenceEventWorker.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/worker/GeofenceEventWorker.kt
@@ -8,6 +8,7 @@ import androidx.work.NetworkType
 import androidx.work.OneTimeWorkRequestBuilder
 import androidx.work.WorkerParameters
 import androidx.work.await
+import io.customer.geofence.GeofenceConstants
 import io.customer.geofence.di.geofenceEventTracker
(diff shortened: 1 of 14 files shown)


geofence/src/main/kotlin/io/customer/geofence/worker/GeofenceEventWorker.kt
```diff
@@ -79,7 +80,15 @@ internal class GeofenceEventWorker(
         val store = SDKComponent.android().pendingGeofenceDeliveryStore
         var sawEntry = false
         while (true) {
-            val entry = store.loadAll().firstOrNull() ?: run {
+            val queued = store.loadAllOrNull() ?: run {
+                // Unreadable is not empty: nothing was sent, so success would claim a drain.
+                // failure() is safe here where the paths below avoid it — every dependent would
+                // fail this same read, and the rows behind are untouched on disk.
+                val willRetry = runAttemptCount < GeofenceConstants.MAX_UNREADABLE_QUEUE_ATTEMPTS
```
Shahroz16: Could we make the total-attempt and whole-chain cutoff explicit in the comment and add a WorkManager regression test? `runAttemptCount` includes delivery retries, so the first unreadable read after five network retries gives up immediately. Already-appended workers fail too, even if the file becomes readable before they would run; recovery needs a later enqueue or foreground flush. The rows survive and a fresh chain recovers them, so I’m treating the bounded policy as intentional and this as a non-blocking clarification, not a request for endless retries.

Shahroz16 (approved):
Unreadable queues now stay intact without being reported as empty, and the foreground path exits before handoff. The geofence build and 142 focused tests passed, and CI is green. I also checked terminal-chain failure and fresh-enqueue recovery with a local WorkManager probe. Left one non-blocking suggestion to document and test the retry cutoff's full scope.

geofence/src/main/kotlin/io/customer/geofence/worker/GeofenceEventWorker.kt
mrehan27: You are right, and it was worse than a missing comment: the constant claimed to count unreadable reads and was compared against `runAttemptCount`, which is the whole request. Four of the five retry sites are delivery failures, so they spend the same budget.

Renamed it to `MAX_WORKER_RUN_ATTEMPTS` and the comment now states the shared budget and why giving up early is safe, which is your point about rows surviving for a later enqueue or flush.

Added a test for the case you named: attempts already spent on delivery retries, first unreadable read gives up. Mutation checked, forcing `willRetry = true` kills both cap tests.

Left the per-cause counter alone. That would be a behaviour change, not a rename, and the bounded policy is intentional as you read it.

PR #861, opened 2026-09-11
chore(geofence): classify diagnostic records as input, output or observation

mahmoud-elmorabea:
Every geofence log record now states whether it is data entering the SDK, a decision leaving it, or an internal step. Only three records are decisions worth asserting: `registration.applied`, `transition.accepted` and `module.reset`; everything else that claimed to be an output is now an observation, and a test pins the set per event so drift is a deliberate act. Three inputs that previously had no record gain one: `identity.changed`, `location.fix` on arrival, and `permission.changed` on foreground entry. `location.fix` writes one debug line per host fix, with the machine tail gated behind diagnostics, as on iOS. A reset whose OS clear fails now records `module.reset ok=false`, so the branch that keeps state for the next refresh is an outcome rather than silence; the sign-out-started record moves from `module.reset why=sign_out` to `info why=sign_out_reset_started`, so the graded key means one thing. Logging only; no behaviour change.

Part of MBL-2309

### DIFF
--- a/geofence/src/main/kotlin/io/customer/geofence/GeofenceLifecycleObserver.kt
+++ b/geofence/src/main/kotlin/io/customer/geofence/GeofenceLifecycleObserver.kt
@@ -26,13 +26,31 @@ internal class GeofenceLifecycleObserver(
     private val deliveryFlusher: PendingDeliveryFlusher<PendingGeofenceDelivery>,
     private val eventBus: EventBus,
     private val regionStore: GeofenceRegionStore,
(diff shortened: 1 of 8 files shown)


Shahroz16 (approved):
The input/decision/observation split preserves the existing geofence behavior, and the permission, reset and log-tail checks passed (128 focused tests). The remaining red Compose instrumentation job crashes in unchanged in-app networking after SDK teardown, so it still needs a green rerun before merge.

PR #862, opened 2026-09-11
refactor(geofence): put geofence registration behind a registrar interface

mahmoud-elmorabea:
`GeofenceRegistrar` names the four calls the SDK makes to Play Services to register, restore, remove and clear regions. `GeofenceManager` implements it and its registration body is unchanged; the DI graph and the repository are typed on the interface so a test can substitute the OS side without carrying the batch policy along. The receiver test overrides the interface instead of the class, which is what the graph now resolves. The `replaceGeofences` override drops its `= emptySet()` default because Kotlin forbids defaults on overrides; the interface carries it and callers resolve it there. No behaviour change.

Part of MBL-2309


Shahroz16 (approved):
The registrar interface and DI override keys line up with the production callers, with registration behavior preserved. The geofence build and 158 focused manager, repository and receiver tests passed.

PR #864, opened 2026-09-11
refactor(geofence): move the foreground decisions out of ModuleGeofence

mahmoud-elmorabea:
The foreground gates, the awaiting-a-fix retry, the anchor choice and the automatic fix acquisition lived in a closure built inside the module's lifecycle wiring, which nothing could construct without a `ProcessLifecycleOwner`. `GeofenceForegroundCoordinator` holds them now. Identify and app launch take their anchor from the same object, so the three paths cannot drift apart again. The module tests for these rules move to the coordinator, plus five for the combined foreground path. No behaviour change; the anchor consolidation is the part to read closest.

Part of MBL-2309


Shahroz16 (approved):
The shared anchor precedence is preserved, and the foreground path now explicitly moves the identity read off the main thread. The geofence build and all 17 coordinator tests passed, including the host-refresh and identity-read-failure cases.

PR #865, opened 2026-09-14
chore(geofence): fetch the nearest geofences from the v2 endpoint

mrehan27:
`apiHost` embeds the API version (`cdp.customer.io/v1`) and `/track` owns that version, so the geofence endpoint could not move to v2 by changing the path alone. That would compose `/v1/v2/geofences/nearest`.

A request now carries its own `apiVersion`. The client splits whatever `apiHost` holds, defaults to that, replaces it when the request names one, and rebuilds. Requests that name none are unchanged.

Replacing rather than appending matters: a host already moved to v3 composes `/v2`, not `/v3/v2`. Splitting a trailing `/v\d+` rather than trimming a literal `"/v1"` is what makes that work. A host whose name contains a version (`v1.example.com`) has no segment to split, and a host carrying none still gets asked at the version the endpoint exists at.

## Verified on device

Not just in tests. A Pixel 6 running this build fetched from v2, catalogued all 9 fences and registered the 5 circles:

```
ev=api.fetch.result io=in ok=true n=9 ms=3698
```

That run also settled an open question: v2 returns polygons without a `capabilities` field in the request body. This build has no polygon runtime, so it sent only latitude and longitude, and all 4 polygons still came back with geometry and enclosing circles.

## Tests

A real v2 capture is decoded through the shipping path: numeric `id` and `geoset_ids` coerce to strings, all four polygon rings build, absent `transition_types` falls back to ENTER|EXIT, and `distance_m` is ignored.

The swap is asserted on the composed URL, not just the helper. Mutating `composeUrl` to ignore `apiVersion` kills 5 tests; trimming a hardcoded `"/v1"` instead of splitting kills the v3 case; dropping the empty-segment handling kills the doubled-slash case.

Known gap: the one line that calls `composeUrl` is not covered. Passing `apiVersion = null` there survives all 1889 tests. `CustomerIOHttpClientImpl` reads its dependencies from `SDKComponent.android()` and opens a real connection, so covering it needs HTTP test infrastructure that does not exist yet. Out of scope here rather than forgotten.

Part of: https://linear.app/customerio/issue/MBL-2424/mobile-fetch-nearest-geofences-from-the-v2-endpoint


geofence/src/main/kotlin/io/customer/geofence/api/GeofenceApiService.kt
```diff
@@ -31,7 +31,8 @@ internal class GeofenceApiServiceImpl(
             path = ENDPOINT_PATH,
             method = HttpMethod.POST,
             headers = mapOf("Content-Type" to "application/json"),
-            body = body
+            body = body,
+            apiVersion = ENDPOINT_API_VERSION
```
Shahroz16, line 35: Could we keep this paired with #870 for rollout and retain `apiVersion = v2` when resolving their shared request-construction changes? The backend gates polygons on the workspace flag, not `capabilities`, so an all-polygon V2 catalog currently reaches the disabled mapper and fails the refresh. #870 supplies the enabled runtime that handles that catalog; this is an integration dependency, not a request to duplicate its implementation here.

Shahroz16 (approved):
The request-level version reaches the composed URL while /track stays on the configured version. The backend V2 route and response shape match, and all 23 focused routing/API/payload tests passed. Left an integration note for the runtime follow-up; no separate mapping fix is needed in this PR.

geofence/src/main/kotlin/io/customer/geofence/api/GeofenceApiService.kt
mrehan27: Agreed on the ordering: this should not reach a workspace with polygons enabled before #870 is in. Treating that as a release-sequencing constraint on the feature branch rather than a merge blocker. The backend half, that polygons are gated on the workspace flag rather than `capabilities`, is not verified on our side yet.

PR #866, opened 2026-09-14
chore(geofence): make the polygon evaluator thresholds plain constants

mrehan27:
`PolygonEvaluatorConfig` looked injectable but nothing ever constructed it with non-default values, in production or in tests. Its seven `require` bounds only ever validated the defaults against themselves. The fields are now constants where they are used and the dead validation is gone.

Two other fixes in the same area:

- `MAXIMUM_ACCEPTED_ACCURACY_METERS` (50, gates shrinking the movement trigger) and `maximumAcceptedAccuracyMeters` (200, gates judging containment) shared a name four times apart. Now `MAX_TRIGGER_FIX_ACCURACY_METERS`, `MAX_EVALUATED_FIX_ACCURACY_METERS` and `MAX_DECISIVE_FIX_ACCURACY_METERS`. The two 50s stay separate on purpose: whether that equality is intentional is not established from source, so calibration can move either alone.
- `EARTH_RADIUS_METERS` was defined twice. Now one, owned by `PolygonGeometry` where the projection lives. iOS uses the same `6371000.0`, and this value decides whether tight-fitting polygons validate, so two copies free to drift would quietly change which fences are monitored.

## Tests

Mutating the 200 ceiling or the 0.4 confidence threshold left the suite green, so neither was pinned. The existing coarse-fix test looked like it covered the ceiling but passed on geometry instead: at 250 m accuracy the disk swamps that small fence whether the ceiling is 200 or 20.

Added a wider fence and a 199/201 pair where only the ceiling can separate them. Both mutants now die, in both directions.

Part of: https://linear.app/customerio/issue/MBL-2415/mobile-define-geofence-timing-and-geometry-constants


Shahroz16 (approved):
The extracted constants preserve the previous numeric values, with separate accuracy limits and the shared projection radius. The geofence build and 43 focused evaluator, geometry and movement-policy tests passed.

PR #867, opened 2026-09-14
chore(geofence): give the last five records a machine-readable tail

mrehan27:
These five were the only log methods a capture could not parse, and the tail test carried them on a declared exception list. That list is now empty, so a record added without a tail fails rather than being declared away. That list existed because five polygon records once shipped with no tail at all.

| Method | Tail |
|---|---|
| `logContainmentJudged` | `ev=containment.judged io=out n=` |
| `logEventDeliveredButNotRemoved` | `ev=delivery.sent io=out ok=true retry=true why=not_removed` |
| `logGmsCallTimedOut` | `ev=os.error io=in ok=false op= why=timeout` |
| `logTransitionDroppedRetiredId` | `ev=transition.dropped io=out why=retired_id` |
| `logTransitionDroppedUnarmedId` | `ev=transition.dropped io=out why=routing_unarmed` |

`logGmsCallTimedOut` is `io=in` on purpose. GMS going silent is the environment's, and the SDK only reacts to it, so replay should feed it back rather than compare it.

`ev=containment.judged` is a new key and the vocabulary guard caught it, which is what it is for. It is Android-only: none of these five records exists on iOS today, and neither does `os.error` itself (iOS has `os.monitor.failed` and `os.stream.failed`), so that divergence predates this PR. Nothing here is agreed cross-platform.

## The table could not tell two of these apart

The last two share `transition.dropped` and differ only by `why`, and the table checked that a key was present, not what it held. Swapping `routing_unarmed` for `retired_id` passed every assertion, so a consumer's only discriminator between them was untested.

Rows can now pin exact values, and the reasons that discriminate are pinned. That mutant now dies.

Not fixed here: `why` values stay unpinned on the pre-existing rows. Pinning the whole table is worth doing but would bury this diff.

Part of: https://linear.app/customerio/issue/MBL-2419/android-improve-geofence-diagnostics-and-queue-failure-handling


Shahroz16 (approved):
All five records now have parseable tails, and the exact reason-value checks distinguish the retired/unarmed cases. The build and 35 log-tail tests passed. When this joins #861, its narrower output/observation classification will also need to be carried through for these new rows.

PR #868, opened 2026-09-14
chore(geofence): pin the transition enum name where it is persisted

mrehan27:
`Event.GeofenceTransition` has no `@SerialName`, so its constant name is the literal value in two persisted places:

- the queued delivery row, `PendingGeofenceDelivery.transition`
- the cooldown preferences key, `cio_cooldown_<user>:<fence>:ENTER`

Moving that name orphans both on upgrade: every queued delivery stops decoding, and every fence emits one suppressed event again. No test failed on either.

The existing round-trip test cannot catch it, because it re-encodes with whatever the new spelling is. So one test decodes a literal row as an older version wrote it, and one reads the cooldown keys back as written.

Tests only, no production change.

## Verified

Annotating the enum `@Serializable` with `@SerialName("enter")`/`("exit")` kills only the delivery test. Spelling the cooldown key `name.lowercase()` kills only the cooldown test. Full geofence suite 708/0/0, ktlint clean, `lintDebug` 0 issues.

An encode-side assertion was dropped on review: its only unique catch was `@SerialName` plus a `@JsonNames` alias for the old spelling, which reads both old and new rows and so does not break the guarantee this pins.

## Stack

- [#867 give the last five records a machine-readable tail](https://github.com/customerio/customerio-android/pull/867)
- 👉 this PR

Part of: https://linear.app/customerio/issue/MBL-2419/android-improve-geofence-diagnostics-and-queue-failure-handling


Shahroz16 (approved):
The literal legacy row and actual SharedPreferences keys pin the persisted spelling independently of a same-version round trip. The build and all 30 focused delivery/cooldown tests passed.

PR #869, opened 2026-09-14
chore(geofence): add the responsive polygon runtime

mrehan27:
Adds the runtime that lets a polygon be monitored rather than only decoded: an approach monitor over the fused location stream, a receiver and worker for its batches, and the service controller that owns per-polygon fine monitoring.

Nothing turns it on. `PolygonSupport` still defaults to `Disabled`. The PR on top enables it.

Not split further: the receiver reaches the controller and the controller takes the monitor as a constructor param, so they are one dependency cycle.

## Fixed in review

Sessions:
- A named `stop` tore down a session it did not own, and a permission refusal left the deadline armed.
- Adopting an unowned session cleared both location anchors, so the first reboot after an upgrade registered nothing.
- App launch, boot restore and the OS callback read the identified user and then opened a session with that value, so an identify landing in between was undone. The read now happens under the lock that opens the session.
- The movement-trigger write after the GMS await was not generation scoped.

Registrations:
- Re-centring the movement trigger removed before adding, so a failed add left no trigger at all.
- A same-generation re-arm cancelled the request it was about to make.
- Removal paths used `FLAG_UPDATE_CURRENT`, minting a `PendingIntent` that never existed.
- `getActivePolygonIds()` was read outside `controllerLock`.

Two things are deliberate and now say so in code: the approach receiver stays outside the kill switch because its registration outlives the process, and recovery does not re-arm because the active set outlives the session.

The three approach records now carry a machine-readable tail.

## Verified

797 tests, 0 failures. Lint 0, ktlint and apiCheck clean. Each new guard dies to its own mutant.

## Stack

- 👉 this PR


geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonApproachMonitor.kt
```diff
@@ -0,0 +1,340 @@
...
+        private const val MAXIMUM_RETRY_MS = 300_000L
+
+        val LOCATION_REQUEST: LocationRequest = LocationRequest.Builder(
+            Priority.PRIORITY_BALANCED_POWER_ACCURACY,
+            UPDATE_INTERVAL_MS
+        )
+            .setMinUpdateIntervalMillis(FASTEST_UPDATE_INTERVAL_MS)
+            .setMinUpdateDistanceMeters(MINIMUM_DISPLACEMENT_METERS)
+            .setWaitForAccurateLocation(false)
+            .build()
```
Shahroz16, line 321: Could we put the remaining session budget on the `LocationRequest` itself with `setDurationMillis`, including retries and cold-process adoption? The request currently has an infinite duration, while the two-minute timer dies with our process. This PendingIntent request survives a system kill, so without another qualifying callback it can remain registered past the promised bound. I reproduced the infinite duration in the actual request object; a remaining-duration regression test would cover this.

geofence/src/main/kotlin/io/customer/geofence/GeofenceBootReceiver.kt
```diff
@@ -59,6 +60,9 @@ class GeofenceBootReceiver : BroadcastReceiver() {
             SDKComponent.geofenceLogger.logSyncSkippedNoPermission(REASON_BOOT_RESTORE)
             return
         }
+        android.secureUserStore.getUserId()?.takeIf { it.isNotEmpty() }?.let {
+            android.polygonGeofenceServiceController.beginUserSession(it)
```
Shahroz16: Could we preserve the restore anchor when opening an ownerless session here? For an upgrade cache without the new owner key, `beginUserSession` removes both location anchors before `restoreFromCache` reads them, so boot restore returns success without registering anything. I reproduced that sequence with the real store and repository; the same cache restores before this extra call. A boot-specific adoption path could retain the anchor while still clearing user-scoped transition state.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonGeofenceServiceController.kt
```diff
@@ -0,0 +1,482 @@
...
+        val trigger = GeofenceRegion(
+            id = GeofenceConstants.MOVEMENT_TRIGGER_ID,
+            latitude = location.latitude,
+            longitude = location.longitude,
+            radius = safeRadius,
+            transitionTypes = listOf(GeofenceTransitionType.EXIT)
+        )
+        val result = manager.replaceMovementTrigger(trigger)
+        if (result.isFailure) return null
+        store.saveLastMovementTriggerLocation(GeofenceLocation(location.latitude, location.longitude))
```
Shahroz16: Could we revalidate the user generation before publishing the movement-trigger location, and coordinate this operation with reset? If sign-out completes while `replaceMovementTrigger` is awaiting GMS, this continuation writes the departing user's location back after reset cleared it. The stop call below is generation-scoped, but this store write is not. A reset-during-registration regression reproduces the stale write.

Shahroz16 (commented):
I checked this with the #870 wiring and completed the independent review. The existing identify/stop concern is addressed by that follow-up, so I am not treating it as missing here. I am holding approval for the OS request lifetime and the boot/reset state issues below, each reproduced in focused tests. The existing 166 focused tests pass, but do not cover these counterexamples.

geofence/src/main/kotlin/io/customer/geofence/GeofenceManager.kt
```diff
@@ -68,6 +68,24 @@ internal class GeofenceManager(
...
+    @RequiresPermission(Manifest.permission.ACCESS_FINE_LOCATION)
+    suspend fun replaceMovementTrigger(region: GeofenceRegion): Result<Unit> {
+        require(region.id == GeofenceConstants.MOVEMENT_TRIGGER_ID) {
+            "movement trigger must use the reserved request id"
+        }
+        if (!permissionChecker.hasRequiredLocationPermissions()) {
+            logger.logMissingPermission("ACCESS_FINE_LOCATION")
+            return Result.failure(SecurityException("Required location permissions not granted"))
+        }
+        val removal = removeGeofencesByIds(listOf(GeofenceConstants.MOVEMENT_TRIGGER_ID))
```
mahmoud-elmorabea: Removing before adding leaves no movement trigger at all when the add fails.

Every other path in this file re-registers the trigger by calling `addGeofences` with the same `requestId` and letting GMS replace it in place — `replaceGeofences` does exactly that, prepending `buildMovementTrigger(...)` with no prior remove. The explicit remove here opens a window: if `registerBatch` fails or hits `GMS_CALL_TIMEOUT` (5s), the trigger is gone, `updateMovementTriggerFromAcceptedFix` returns `null`, and nothing re-arms it. The approach session then hits its 2-minute bound and the device is deaf until an app foreground or an unrelated business-fence EXIT.

It also doubles the worst-case latency to 10s, which matters on the broadcast path (`PolygonApproachReceiver`, and the movement-EXIT call in #870).

Could this just be `registerBatch` on its own?

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonApproachMonitor.kt
```diff
@@ -0,0 +1,340 @@
...
+                expectedUserStateGeneration,
+                sessionDeadlineElapsedRealtimeMs
+            ).also {
+                activePendingIntent = it
+            }
+            removalRetryJobs.remove(current)?.cancel()
+            removalRetryAttempts.remove(current)
+            previous to current
+        }
+        registration.first?.let(::removeUpdates)
```
mahmoud-elmorabea: A same-generation re-arm double-registers and logs a `polygon.approach.stopped` for a session that never stopped.

When `start()` runs again for the same generation after the deadline expired, the guard above doesn't early-return, and `pendingIntent()` returns an *equal* PendingIntent — identity is the generation in the data URI, as the companion doc says. So `previous == current`, and this line removes the very request `requestUpdates(registration.second, ...)` is about to make.

The removal's success listener then finds `desired && activePendingIntent == pendingIntent`, emits `logPolygonApproachMonitoringStopped()`, and calls `requestUpdates` a second time. Net: two `requestLocationUpdates` per arm, plus a false `polygon.approach.stopped` tail that a replay grading on `polygon.approach.*` would read as a real teardown.

Skipping the removal when `previous == current` would cover it.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonGeofenceServiceController.kt
```diff
@@ -0,0 +1,482 @@
...
+                return@synchronized false
+            }
+            if (store.getActivePolygonIds().isEmpty()) engine.stop()
+            true
+        }
+        if (!sessionIsCurrent) return PolygonSamplingDecision.STALE
+        val safelyPassive = lastAcceptedLocation?.let {
+            updateMovementTriggerFromAcceptedFix(it, expectedUserStateGeneration)
+        } != null
+        return if (safelyPassive || store.getActivePolygonIds().isEmpty()) {
```
mahmoud-elmorabea: This `getActivePolygonIds()` read is outside `controllerLock` — every other read of it in this class is inside `synchronized(controllerLock)`.

A concurrent `reconcileRegisteredPolygons` or `deactivate` landing between the guarded `sessionIsCurrent` block above and this line flips the answer: either the session is told CONTINUE with nothing active (battery spent on nothing), or STOP with a polygon still active (its EXIT is never observed). Cheap to fold into the `sessionIsCurrent` block.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonGeofenceServiceController.kt
```diff
@@ -0,0 +1,482 @@
...
+        store.retainCoarseInsidePolygonIds(ids)
+        removed.forEach(engine::deactivate)
+        lastCoarseTransitionElapsedNanos.keys.retainAll(ids)
+        if (store.getActivePolygonIds().isEmpty() || !hasMatchingIdentifiedUserLocked()) {
+            approachMonitor.stop(store.userStateGeneration())
+        }
+        if (store.getActivePolygonIds().isEmpty()) engine.stop()
+    }
+
+    fun recover() = synchronized(controllerLock) {
```
mahmoud-elmorabea, line 208: `recover()` only ever tears down — it stops the monitor when the active set is empty, but does nothing when polygons *are* active.

After process death mid-approach, the active set survives in the store but nothing resumes sampling until the next coarse callback arrives. Is that deliberate (bounded sessions shouldn't outlive their process), or should recovery re-arm for the surviving active set?

geofence/src/main/AndroidManifest.xml
```diff
@@ -17,6 +18,11 @@
             android:enabled="false"
             android:exported="false" />
 
+        <!-- Receives the displacement-gated polygon approach fixes registered with GMS. -->
+        <receiver
+            android:name="io.customer.geofence.polygon.PolygonApproachReceiver"
```
mahmoud-elmorabea, line 30: This is the only geofence receiver `GeofenceReceiverToggle` doesn't manage — the other two ship `android:enabled="false"` and get toggled with the module, this one is always enabled, so it sits outside the kill switch.

Probably intentional (you need it live to receive the teardown fixes for an already-registered PendingIntent), but the comment above doesn't say so and the asymmetry reads as an oversight.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonApproachMonitor.kt
```diff
@@ -0,0 +1,340 @@
...
+            desired = false
+            userStateGeneration = null
+            sessionDeadlineElapsedRealtimeMs = null
+            registrationRetryAttempt = 0
+            registrationRetryJob?.cancel()
+            registrationRetryJob = null
+            sessionTimeoutJob?.cancel()
+            sessionTimeoutJob = null
+            activePendingIntent.also { activePendingIntent = null }
+                ?: generationToRemove?.let {
```
mahmoud-elmorabea: Nit: `pendingIntent()` uses `FLAG_UPDATE_CURRENT`, which *creates* the PendingIntent when none exists — so a removal for a generation that was never registered leaves a fresh PendingIntent behind rather than being a no-op. Same in `removeStaleGeneration`.

`FLAG_NO_CREATE` (and skipping the removal when it returns null) matches what this path is actually asking.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonApproachMonitor.kt
```diff
@@ -0,0 +1,340 @@
...
+        if (cause is SecurityException) {
+            synchronized(lock) {
+                if (userStateGeneration == requestGeneration && activePendingIntent == pendingIntent) {
+                    desired = false
+                    userStateGeneration = null
+                    activePendingIntent = null
+                    // Nothing is registered now, so leaving this armed would later "stop" a request
+                    // that never existed and log a removal that did not happen.
+                    sessionTimeoutJob?.cancel()
+                    sessionTimeoutJob = null
```
mahmoud-elmorabea, line 220: Nit: this branch clears `desired`, `userStateGeneration`, `activePendingIntent` and the timeout job, but not `sessionDeadlineElapsedRealtimeMs` — the one field left inconsistent after a fail-closed teardown.

Harmless today because every reader gates on `desired` first, but it's an invariant a later change could trip over.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonApproachMonitor.kt
```diff
@@ -0,0 +1,340 @@
...
+                }
+        } catch (e: RuntimeException) {
+            retryRemoval(pendingIntent, e)
+        }
+    }
+
+    private fun retryRemoval(pendingIntent: PendingIntent, cause: Throwable) {
+        logger.logPolygonApproachRequestFailed(cause.message, operation = "remove_updates")
+        synchronized(lock) {
+            if (desired && activePendingIntent == pendingIntent) return
```
mahmoud-elmorabea: Nit: returning here leaves any existing `removalRetryAttempts` / `removalRetryJobs` entry for this PendingIntent in place — nothing else removes it on this path.

Since the key is a per-generation PendingIntent, that's one stranded map entry per user generation for the life of the process. Small, but the maps are the only unbounded state in the class.

mahmoud-elmorabea (approved):

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonApproachMonitor.kt
mrehan27: Fixed. The request carries the remaining session budget via `setDurationMillis`, read at request time so a retry or a post-removal restart continues the session rather than renewing it. `start_givenAPartlySpentSession_expectOnlyTheRemainingBudgetOnTheRequest` pins it and dies to a full-budget mutant.

geofence/src/main/kotlin/io/customer/geofence/GeofenceBootReceiver.kt
mrehan27: Confirmed, and it was a regression against the base rather than a latent bug: the base opened no session here, so restore ran against intact anchors. Adopting an unowned session now keeps both anchors. The sync stamp is still cleared, because retaining it throttles the pass that arms routing, which three existing tests caught.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonGeofenceServiceController.kt
mrehan27: Fixed. The store gained `saveLastMovementTriggerLocationIfCurrent`, mirroring the API-fetch anchor, and the write is skipped once the session has moved on. Covered by a reset-during-registration test plus a positive control, since a guard that always refused would pass the negative on its own.

geofence/src/main/kotlin/io/customer/geofence/GeofenceManager.kt
mrehan27: Fixed, and good catch. Re-centring now re-adds with the same requestId and nothing is removed first, so a failed add can no longer leave the device with no trigger. Our emulator measurement supports the in-place replace: GMS emits no EXIT on a same-id re-add.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonApproachMonitor.kt
mrehan27: Confirmed and fixed. The removal is skipped when `previous == current`. A test reproduced it exactly as described: three `requestLocationUpdates` for two arms, plus the false `polygon.approach.stopped`.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonGeofenceServiceController.kt
mrehan27: Fixed, wrapped in `controllerLock` rather than folded into the block above, so it still reads after the await instead of before it. No dedicated test: it is a structural fix and a race test here would be flaky, so the existing decision tests are the only cover.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonGeofenceServiceController.kt
mrehan27: Deliberate, and now commented. The active set means polygons whose circle contains the device, which outlives the two-minute session, so re-arming here would open a fresh session on every foreground for someone sitting inside a polygon. Nothing is lost by not doing it: the request outlives the process, delivery resumes on its own, and the bound is re-adopted from the deadline carried in the intent extras.

geofence/src/main/AndroidManifest.xml
mrehan27: Intentional, and the comment now says so. Its registration is a PendingIntent held by GMS that outlives both the process and the kill switch, so disabling this one would strand a live request with nothing able to receive the delivery that tears it down.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonApproachMonitor.kt
mrehan27: Fixed. Removal paths use a `FLAG_NO_CREATE` variant and do nothing when it returns null, here and in `removeStaleGeneration`. The cold-process test was itself relying on create-if-absent to fake a previous process, so it now registers first, and a new test covers the never-registered case.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonApproachMonitor.kt
mrehan27: Fixed, the deadline is cleared with the rest of that teardown.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonApproachMonitor.kt
mrehan27: Fixed, the retry attempt and job entries for that PendingIntent are dropped on this path.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonGeofenceServiceController.kt
Shahroz16: Thanks, the lock and guarded anchor write address the earlier concerns. Could we also recheck the generation after the await and prevent an old continuation from replacing a newer sampling session? In a controller/receiver test, identify and a new polygon wake land while recentering is suspended. The anchor write is correctly refused, but this branch sees the new user's active polygon and returns CONTINUE. The receiver then calls `start` for the old generation, which can replace the new approach request. A controlled delayed-registration test reproduces this without relying on thread timing; the decision and session adoption both need to remain tied to the current generation.

Shahroz16 (commented):
The duration limit, ownerless restore and guarded anchor fixes check out, including focused regression and mutation checks. I completed the Claude Code Opus adversarial pass and am still holding approval for the [post-await generation issue](https://github.com/customerio/customerio-android/pull/869#discussion_r4015791573), reproduced with the controller and receiver while a new user session becomes active. This runtime path becomes live through #870. The existing 244 focused tests pass, but do not cover that interleaving.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonGeofenceServiceController.kt
mrehan27: Fixed in `476add4e`. You are right that the decision was not tied to the generation. The final block took `controllerLock` but checked only the active set, while every other read in that pass checks the generation too, and the trigger update just above it awaits GMS unlocked. It now returns STALE when the generation moved, so an old continuation cannot answer CONTINUE and have the receiver start a session for one that has already ended.

Test reproduces it without thread timing: the registration flips the generation while suspended, and the pass returns STALE. Reverting the check fails it.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonGeofenceServiceController.kt
```diff
@@ -0,0 +1,482 @@
...
+    fun beginUserSession(userId: String) = synchronized(controllerLock) {
+        val changed = store.activeUserSessionId() != userId
+        val previousGeneration = store.userStateGeneration()
+        store.beginUserSession(userId)
+        if (changed) {
+            engine.stop()
+            approachMonitor.stop(previousGeneration)
+            lastCoarseTransitionElapsedNanos.clear()
+        }
+    }
```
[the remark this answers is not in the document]
mrehan27: Reopening this rather than leaving it resolved, because it is deferred and not fixed.

`GeofenceServices.onUserIdentified` does write to the region store directly, so an identify bumps the generation without going through the controller's `openSessionLocked`, and neither `engine.stop()` nor `approachMonitor.stop(previousGeneration)` runs. An independent review reached the same finding.

Not fixed here because `GeofenceServicesImpl` has no reference to the controller, so closing it means a new constructor dependency and DI wiring, which is too much for a PR already in review. The exposure is bounded: the approach request carries `setDurationMillis`, so an orphaned one expires at the OS rather than running until something else clears it.

It goes in the follow-up PR with the other deferred items.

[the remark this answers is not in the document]
mrehan27: Correcting my previous reply: I posted it before tracing the caller, and it is wrong.

`onUserIdentified` has exactly one caller, `ModuleGeofence`, and the line directly above it already calls `polygonGeofenceServiceController.beginUserSession(userId)`. That goes through `openSessionLocked`, so `engine.stop()` and `approachMonitor.stop(previousGeneration)` do run. The `regionStore.beginUserSession` inside `onUserIdentified` then no-ops, because `beginUserSessionLocked` returns early when the owner already matches.

So it is a redundant call, not a bypass, and there is no orphaned engine or approach request. On this PR alone it cannot arise at all, since `PolygonSupport` is `Disabled` and the engine never starts.

Nothing deferred here after all, and nothing for the follow-up PR.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonGeofenceServiceController.kt
Shahroz16: Thanks, the new guard closes the delayed-registration case, and I verified that reverting it fails the regression. I am no longer holding approval for that issue. One smaller follow-up remains: if the old receiver is preempted after CONTINUE, a new identify, registration and polygon wake can complete before its `start` resumes and replaces the newer request. A controlled-preemption test shows that call sequence with the real controller and monitor, but there is no suspension point in this gap and device incidence is unverified, so I am treating it as non-blocking. Could we keep adoption tied to the current generation in a follow-up? The old request's timeout removes it but does not restore the newer request.

Shahroz16 (approved):
The post-await generation guard fixes the delayed-registration counterexample. All 66 focused tests pass; removing the guard fails its regression and restoring it passes. Claude Code Opus review and the focused challenge are complete. I am approving this fix, with the narrower decision-to-adoption race explicitly retained as a non-blocking follow-up in the existing thread.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonGeofenceServiceController.kt
mrehan27: Noted for the follow-up: keeping adoption tied to the current generation, so a `start` resuming after preemption cannot replace a newer request. Agreed it is non-blocking, and agreed the old request's timeout removes it without restoring the newer one.

Shahroz16: Thanks, #873 closes the newer-request displacement described here. The controlled-preemption probe now keeps the newer request, and removing the guard reproduces the loss. Could we keep the sign-out-only case as a separate follow-up? If sign-out stops generation 8 before the paused receiver resumes start(8), the equal-generation restart can still arm a bounded request for the signed-out user. That is pre-existing, its callbacks are rejected by the generation checks, and device incidence is unverified, so it does not hold up #873.

Shahroz16: Thanks, #878 addresses the sign-out-only follow-up too. With the real store, controller, receiver and monitor in the controlled-preemption probe, sign-out now leaves only the original request and its teardown; the paused receiver cannot re-arm it. Removing the store guard reproduces that extra request. Same-user restart and newer-user registration controls still pass. This remains modeled scheduling, not a device reproduction.

PR #870, opened 2026-09-14
chore(geofence): enable responsive polygon monitoring

mrehan27:
Turns the runtime on. `PolygonSupport.Enabled` becomes the production opt-in, so polygons are requested, mapped, ranked, registered as their wake circle, and routed into the service controller. Ranking gains an OS-slot ceiling and pinning, so a polygon with an outstanding EXIT cannot be evicted by a discovery cap.

## Behaviour worth calling out

Routing arms on the newly registered ids rather than everything saved, so an id kept only to retry a failed OS removal is a tombstone and cannot generate events.

`unchangedRegisteredIds` now also requires routable membership. That is what makes an unroutable id get re-added rather than skipped, and a test pins it.

The request no longer sends a `capabilities` field. It came from the original spike, not an API contract: a live v2 fetch returns every polygon with geometry when it is absent, and iOS has never sent it.

## The receiver keeps an unroutable registration

A callback can arrive for an id that is registered but not routable, either mid-identify or as a cleanup-only registration after a handoff. Those are byte-identical in the store, so the receiver cannot tell them apart, and it does not need to: after any session open routing is empty, so the next refresh re-adds every requested fence.

Evicting would be wrong. Business fences register with `INITIAL_TRIGGER_ENTER` and routing arms only after the add, so a fence the refresh just added can report ENTER while routing still reads empty. Evicting there removes a fence the same pass then publishes as registered and routable, and every later refresh reads it as unchanged. Silently gone until reboot.

## Known, not changed here

One movement EXIT registers the trigger twice, once by the controller and once by the refresh that follows. Redundant, not a defect: same centre, one extra round trip. The controller's call covers the paths where the refresh does not register at all.

## Verified

835 tests, 0 failures. Lint 0, ktlint and apiCheck clean.

A refresh test asserting a second same-session pass is serialized was dropped: the generation guard covers it and a passing test already asserts the drop.

## Stack

- 👉 this PR


geofence/src/main/kotlin/io/customer/geofence/ModuleGeofence.kt
```diff
@@ -210,8 +219,10 @@ class ModuleGeofence @JvmOverloads constructor(
             val launchScope = SDKComponent.scopeProvider.geofenceScope
             launchScope.launch {
                 try {
+                    sdkAndroid.geofenceTransitionEmitter.recoverPendingTransitions()
                     val existingUserId = sdkAndroid.secureUserStore.getUserId()
                     if (!existingUserId.isNullOrEmpty()) {
+                        sdkAndroid.polygonGeofenceServiceController.beginUserSession(existingUserId)
```
Shahroz16: Could this launch path use an absent-only session open, like `onAppLaunch` already does? If launch reads user A and identify B lands before this call, `beginUserSession(A)` reopens the old owner and clears B's routing. The subsequent `onAppLaunch` reads B but leaves that existing owner alone, so B's refresh cannot publish. Extending the interleaving test to include this newly added controller call reproduces it.

geofence/src/main/kotlin/io/customer/geofence/GeofenceDistanceFilter.kt
```diff
@@ -39,16 +49,75 @@ internal class GeofenceDistanceFilter(
...
         pruneGeometryCache(regions)
-        return regions
+        // A polygon is registered as its wake circle, and the ring it ranks by can sit far inside
+        // that circle. Ranking distance alone would therefore evict a polygon the OS is monitoring
+        // right now, and its outstanding EXIT would never arrive. Derived from this fix rather than
+        // read from the activation set: the device can be inside the circle before any coarse ENTER
+        // is observed — a first registration, an OS state wipe, a re-registration that cleared it.
+        val allPinnedIds = pinnedIds + regions
+            .filter { it.isPolygon && it.distanceTo(latitude, longitude) <= it.radius }
+            .mapTo(mutableSetOf(), GeofenceRegion::id)
```
Shahroz16: Is bypassing the discovery cap intentional for polygons that have never been registered? With `max=1` and no explicit pins, two unregistered polygons whose enlarged wake circles contain the fix displace a circle the device is actually inside, and both polygons are selected. I reproduced that result. The first-registration comment suggests a deliberate policy, but there is no outstanding polygon EXIT to preserve yet. Could we clarify that exception, or restrict automatic pins to existing registrations?

Shahroz16 (commented):
The identify/reset wiring addresses the earlier teardown concern in #869. I am still holding approval for the inherited runtime issues reviewed there and the launch/identify race below. The 480 focused tests pass; the additional interleaving probe exposes the missing case. I also left a separate discovery-policy question, which is not the basis for withholding approval.

geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
```diff
@@ -157,7 +170,15 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
                 // ENTER fires on every re-registration and boot-restore can fire
                 // EXIT. Only EXIT drives a refresh.
                 if (gmsTransitionType == Geofence.GEOFENCE_TRANSITION_EXIT) {
-                    movementRefreshJob = androidComponent.geofenceServices.onMovementTriggerExit(latitude, longitude)
+                    val polygonMovementRadius = androidComponent.polygonGeofenceServiceController.onMovementTriggerExit(
```
mahmoud-elmorabea: This now runs synchronously inside the broadcast window, ahead of the refresh that used to start immediately.

`onMovementTriggerExit` can reach `updateMovementTriggerFromAcceptedFix` → `replaceMovementTrigger`, which is two awaited GMS calls at `GMS_CALL_TIMEOUT = 5.seconds` each — up to 10s before `geofenceServices.onMovementTriggerExit` below is even called. A background broadcast gets roughly 10s.

`GeofenceServices.onMovementTriggerExit`'s own doc says the movement trigger "usually fires with the app backgrounded, where the process is fair game for the OS the moment the receiver finishes" — this puts the entire budget in front of the work that comment exists to protect. Could the polygon evaluation move inside the refresh job (or hand the radius over asynchronously) so the job is created first?

Same shape applies to `PolygonApproachReceiver` when the WorkManager enqueue fails and it falls back to `handleLocations` on the broadcast path.

geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
```diff
@@ -157,7 +170,15 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
...
+                        triggeringLocation = triggeringLocation,
+                        expectedUserStateGeneration = userStateGeneration
+                    )
+                    movementRefreshJob = androidComponent.geofenceServices.onMovementTriggerExit(
```
mahmoud-elmorabea, line 173: The movement trigger ends up registered twice per movement EXIT: the controller above re-centres it via `replaceMovementTrigger`, then `handleMovement` → `registerNearestAndPersist` prepends `buildMovementTrigger(...)` and registers it again at the same centre.

Two `addGeofences` round trips and two `registration.applied` tails for one EXIT. If the fast re-arm is deliberate (don't stay deaf while the async refresh runs), worth saying so in a comment — otherwise it's free latency inside the window flagged above, and it changes what the replay corpus sees for a single crossing.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
```diff
@@ -492,11 +538,40 @@ internal class GeofenceRepositoryImpl(
...
         }
+        if (!osStateWiped) {
+            val registeredBusinessIds = store.getRegisteredIds() - GeofenceConstants.MOVEMENT_TRIGGER_ID
+            val requestedBusinessIds = regions.asSequence()
+                .map(GeofenceRegion::id)
+                .filter { it != GeofenceConstants.MOVEMENT_TRIGGER_ID }
+                .toSet()
+            val additions = requestedBusinessIds - existingBusinessIds
+            val stale = (registeredBusinessIds - requestedBusinessIds).sorted()
+            val projectedPeak = registeredBusinessIds.size + 1 + additions.size
```
mahmoud-elmorabea: `projectedPeak` double-counts fences that are registered *and* changed.

`additions` is `requestedBusinessIds - existingBusinessIds`, where `existingBusinessIds` is the *unchanged* set — so a fence that is already registered but whose geometry changed lands in both `registeredBusinessIds.size` and `additions.size`. Re-adding it replaces its slot rather than consuming a new one, so the peak is over-estimated and stale fences get pre-removed that didn't need to be.

Safe direction, but it makes the 100-slot arithmetic wrong and the pre-removal count hard to reason about. `(additions - registeredBusinessIds).size` is the real growth.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
```diff
@@ -506,11 +581,14 @@ internal class GeofenceRepositoryImpl(
...
     private fun unchangedRegisteredIds(regions: List<GeofenceRegion>): Set<String> {
         val registeredBusinessIds = store.getRegisteredIds() - GeofenceConstants.MOVEMENT_TRIGGER_ID
+        val routableBusinessIds = store.getRoutableRegisteredIds() - GeofenceConstants.MOVEMENT_TRIGGER_ID
         val cachedById = store.getCachedRegions().associateBy { it.id }
         return regions
             .filter { region ->
                 val cached = cachedById[region.id]
-                region.id in registeredBusinessIds && cached?.equalsForRegistration(region) == true
+                region.id in registeredBusinessIds &&
+                    region.id in routableBusinessIds &&
```
mahmoud-elmorabea, line 603: Worth a deliberate answer on the battery cost this creates: every identify now opens a fine-location session for every polygon the device is sitting in.

`beginUserSession` writes an explicit empty routable set, so with this added condition *nothing* counts as unchanged after an identify. The whole nearest set is re-added with `INITIAL_TRIGGER_ENTER`, GMS re-reports ENTER for each fence the device is inside, and the new polygon branch in `GeofenceBroadcastReceiver` routes those to `activate()` → `approachMonitor.start()` — two minutes of balanced-power sampling per polygon, with no movement at all.

Correct by construction (a fresh session has to re-evaluate containment), but for an app that identifies often it's a recurring cost that nothing in the change flags.

geofence/src/main/kotlin/io/customer/geofence/api/GeofenceApiService.kt
```diff
@@ -51,5 +59,7 @@ private data class GeofenceNearestRequest(
     @SerialName("latitude")
     val latitude: Double,
     @SerialName("longitude")
-    val longitude: Double
+    val longitude: Double,
+    @SerialName("capabilities")
+    val capabilities: List<String>
```
mahmoud-elmorabea, line 64: `capabilities` has no default, so kotlinx serialises it unconditionally — a polygon-disabled build now sends `"capabilities": []` on every `POST /geofences/nearest` rather than omitting the field.

Has the backend confirmed that an empty array and an absent field mean the same thing? If "empty means omit" is the intended contract, this wants a default plus `@EncodeDefault(EncodeDefault.Mode.NEVER)` so the disabled path keeps the old wire shape.

mahmoud-elmorabea (approved):

geofence/src/main/kotlin/io/customer/geofence/ModuleGeofence.kt
mrehan27: Confirmed, though not fixed with an absent-only open. `secureUserStore` is written before the identify event is published, so after a process death the owner can genuinely differ from the secure user, and `IfAbsent` would never repair that: every polygon path refuses and every refresh ends cleanup-only. The read is handed to the store instead and runs under the lock that opens the session, so a concurrent identify is either already visible here or lands afterwards. A threaded test reproduces your race and kills an outside-the-lock mutant.

geofence/src/main/kotlin/io/customer/geofence/GeofenceDistanceFilter.kt
mrehan27: You were right, and the consequence is worse than stated: the displaced circle's ENTER is lost outright, because initial-enter synthesis only runs over what ranking returns. Automatic pins now require registration, supplied by the repository, so the filter pins only what its caller asks it to.

geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
mrehan27: Real defect, fixed. The radius is handed over as a suspending provider and resolved inside the refresh once it holds the slot, so the job is created first and the GMS time is charged to it. A test makes the evaluation hang and asserts the job was still created. Your note about `PolygonApproachReceiver`'s WorkManager fallback holds, but it has no job to create, so there is nothing to reorder there.

geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
mrehan27: Confirmed, and left alone deliberately. It is redundant work rather than a defect: same centre, one extra round trip, nothing lost. Dropping the controller's call would also strand the trigger on the paths where the refresh does not register at all, such as monitoring disabled or an early return, so it is not just deleting a line. Noted in the description instead of carrying a ticket for it.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
mrehan27: Confirmed and fixed to `(additions - registeredBusinessIds).size`. A new test pins that a pass with no net growth registers before it removes; the existing genuine-growth test still passes as the control.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
mrehan27: Left as is, with a comment at `activate` recording the cost. It is smaller than it looks: the attached fix is evaluated first and a decisive one ends the session immediately, so the two minutes only land on an ambiguous re-ENTER. It is also not suppressible without weakening the guarantee, since a polygon cannot recover containment from registration synthesis the way a circle does, so suppressing would leave a newly identified user standing inside a ring with no ENTER until they move.

geofence/src/main/kotlin/io/customer/geofence/api/GeofenceApiService.kt
mrehan27: Good question, and leaving it unchanged pending a backend answer. A server gating on membership treats absent and empty identically, so changing it either way without the contract is a guess. Once confirmed, a default plus omit is a one-line change.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
```diff
@@ -492,11 +543,44 @@ internal class GeofenceRepositoryImpl(
...
+                )
+                store.saveRetainedRegisteredRegions(
+                    store.getRetainedRegisteredRegions().filterNot { it.id in preRemove }
+                )
+                val remainingPolygonIds = remainingIds.mapNotNull { id ->
+                    store.getRegisteredRegion(id)?.takeIf(GeofenceRegion::isPolygon)?.id
+                }.toSet()
+                polygonController?.reconcileRegisteredPolygons(remainingPolygonIds)
+            }
+        }
```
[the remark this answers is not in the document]
mrehan27: Not a defect, but it found a real gap in the comments and a commit adds one.

By the time that block writes, `removeGeofencesByIds` has already succeeded, so GMS no longer holds those fences. The store's job is to mirror what the OS holds, so recording the removal is correct regardless of how the add goes. Deferring it to the success path would be the bug: the store would keep claiming fences that are gone, the next pass would count them in `registeredBusinessIds`, over-estimate the peak, and try to remove them again.

The comment quoted is the one on the post-add stale cleanup, which governs a different block. Pre-removal is a deliberate exception to it, forced by GMS rejecting a whole batch at quota, and nothing here said so. That sentence is now in the code.

Nothing is lost on the failure path either: the ids are stale by definition, the success path would have removed them anyway, and the next pass sees a consistent store, a smaller stale set and an add that now fits.

geofence/src/main/kotlin/io/customer/geofence/api/GeofenceApiService.kt
mrehan27: Removed the field instead of defaulting it.

It came from the original polygon spike, not from an API contract. A live v2 fetch returns every polygon with geometry and enclosing circles when `capabilities` is absent, and iOS has never sent it. So there is no contract to match here, and the empty-versus-absent question goes away with the field.

`PolygonSupport` keeps only the thing it actually decides: whether a decoded polygon may be monitored.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
```diff
@@ -227,9 +238,10 @@ internal class GeofenceRepositoryImpl(
     private fun movedBeyondFetchRadius(distanceFromAnchor: Float, config: GeofenceConfig): Boolean =
         distanceFromAnchor >= config.remoteFetchRefreshTriggerRadius
 
-    /** Cache holds regions but none are registered with the OS (e.g. regs lost on sign-out) → re-register. */
+    /** Cache holds regions but OS or current-session routing state is absent → re-register. */
     private fun hasUnregisteredCache(cachedRegions: List<GeofenceRegion>): Boolean =
-        cachedRegions.isNotEmpty() && store.getRegisteredIds().isEmpty()
```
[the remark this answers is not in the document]
mrehan27: The radius mismatch is real. `safeRadiusMeters` only shrinks, the trigger registers with that value, and `isRankingStale` still compares against `config.localRefreshTriggerRadius`.

The path named here cannot reach it though. Identify and launch anchor the pass at `getLastMovementTriggerLocation()`, which is the same value `distanceFromLastRegistration` is measured from, so that distance is 0 by construction and the guard is dead on those paths at any radius. Only the foreground live-fix refresh can make it true.

Fixed here: the `movement.registered` record logged the config radius instead of the one actually registered, so a capture would report 1000m for a trigger registered at 150m.

Not fixed here: the staleness check itself. The registered radius is never persisted, so there is nothing to compare against after a process death. That needs a store change and goes in its own PR, along with the fact that the guard cannot fire on the paths its comment claims.

geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
Shahroz16: The provider change fixes the movement-only path. Could we also create that refresh before awaiting polygon handlers in a mixed EXIT batch? In a receiver test with two polygon IDs ahead of the movement ID and each handler delayed by 4.5 seconds, the refresh is created at 9 seconds, after the receiver's 8-second wait budget is spent. Processing the movement ID first fixes the regression, with all 57 receiver tests passing. Otherwise the receiver can finish without waiting for the refresh, leaving background work exposed to process termination. These are simulated GMS waits, not a device-kill reproduction.

Shahroz16 (commented):
The launch identity check, registered-only pinning and movement-provider changes check out; 476 focused test executions pass. I completed the Claude Code Opus adversarial pass and approved #872's radius correction, so that does not need another fix here. Approval is still held for the inherited [sampling-session issue in #869](https://github.com/customerio/customerio-android/pull/869#discussion_r4015791573) and the [mixed EXIT batch ordering](https://github.com/customerio/customerio-android/pull/870#discussion_r4015791947). Both have targeted test counterexamples; neither is a claim of reproduced device termination or production frequency.

geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
mrehan27: Fixed in `ad3d1db3`. The movement trigger now sorts to the front of the batch whatever order GMS delivered it in. The polygon waits still spend the budget, but the refresh job exists before they start and gets whatever is left, so the receiver holds the window open for it.

Test puts a polygon ahead of the trigger with a delayed handler and asserts the refresh is created first. Reverting the sort fails it.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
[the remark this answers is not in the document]
mrehan27: Update: the staleness half landed in #872, which is approved. It persists the registered radius beside the centre and compares against that, so the case described here is closed rather than still open.

geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
```diff
@@ -147,15 +160,30 @@ class GeofenceBroadcastReceiver : BroadcastReceiver() {
             businessIds.forEach { logger.logTransitionDroppedUnarmedId(it) }
             triggerIds
         }
-        val knownIds = routableTriggeringIds + unarmedTriggerIds
+        // Movement trigger first, whatever order GMS delivered. The polygon handlers below await
+        // GMS, so a mixed batch with polygons ahead of the trigger spends the dispatch budget
+        // before the refresh job exists, and the receiver then finishes without waiting for it.
+        val knownIds = (routableTriggeringIds + unarmedTriggerIds)
+            .sortedByDescending { it == GeofenceConstants.MOVEMENT_TRIGGER_ID }
```
Shahroz16: Could we preserve circle EXITs already in this batch before the refresh can retire them, while still starting it before the polygon waits? For a `polygon, circle, movement` EXIT batch, this sort starts the refresh, then awaits the polygon handler. If that refresh evicts the circle under the monitoring-distance cap, its later `process(requireRegistered = true)` silently drops the EXIT that GMS already delivered. With a fix about 350m from a 100m circle and a valid 100m distance cap, the real ranker evicts it; the receiver/outbox test passes in the old order and fails with this sort. Refresh publication and GMS timing are modeled, not device-reproduced. Processing circles, then movement, then polygons is one way to retain both guarantees.

Shahroz16 (commented):
The sort fixes refresh creation before polygon waits, and the 104 focused receiver/services/live-path tests pass. Claude Code Opus found a separate consequence of moving the refresh ahead of pending circle events; I verified the EXIT loss in a focused receiver/processor/outbox test with the real distance filter and modeled refresh publication. I am holding approval for that case below. The upstream #869 fix and #872 radius correction are approved, so neither needs duplicate implementation here.

geofence/src/main/kotlin/io/customer/geofence/GeofenceBroadcastReceiver.kt
mrehan27: Good catch, and it was a regression I introduced. Fixed in `e69c08ff`.

Your ordering: circles, then the movement trigger, then polygons. Circles first so an EXIT the OS already delivered is processed before the refresh can retire the fence, the trigger next so its job exists before the polygon handlers await GMS, polygons last. The sort is stable, so GMS order survives within each group.

Test asserts the full sequence on a `polygon, circle, movement` batch. It kills two mutants: raw GMS order, and the movement-first-only sort I had before, so it pins the regression rather than just the original concern.

One cost worth naming: the sort needs the catalogue to know which ids are polygons, so the dispatch now reads cached regions once up front. The loop already read per id, so it is one deserialisation, not a new per-id cost.

Shahroz16: Thanks, this closes the EXIT-loss case. I reran it with the real receiver, processor, distance filter and file outbox: the final order retains the circle EXIT, while restoring movement-first loses it. Raw GMS order still fails the refresh-before-polygon check. All 60 receiver tests, including the added review probes, pass with the final code. The refresh publication and GMS waits are modeled, not device timing.

Shahroz16 (approved):
The circle EXIT now reaches the outbox before this batch starts its refresh, and the refresh still starts before polygon waits. All 105 focused tests pass. I also checked the stronger outbox regression and both ordering mutations, and completed the Claude Code Opus review. The reported blocker is addressed; the existing non-blocking #869 adoption follow-up remains separate.

PR #872, opened 2026-09-15
chore(geofence): judge staleness against the trigger that was registered

mrehan27:
A polygon can shrink the movement trigger to a fifth of the configured radius. The staleness check still compared the device against the configured one, so a device 300m from a 150m trigger read as fresh and the pass skipped.

That is not just a stale ranking. The OS holds a fired trigger in the outside state, so skipping without re-centring leaves no background wake at all until an app open, reboot or update.

## What changed

The registered radius is persisted beside the centre, in the same edit so the pair cannot disagree after a partial write, and cleared with it. `isRankingStale` compares against that.

An install with no radius recorded falls back to the configured value, so upgrades behave exactly as before. A test pins that.

Also corrects the guard's comment. It claimed `refresh()` catches an EXIT missed while the process was dead. Identify, launch and foreground retry all anchor the pass at the stored centre, so their distance is 0 by construction. Only a live fix can make the guard true, which in `AUTOMATIC` mode is the silent fix taken on every foreground entry.

## Verified

839 tests, 0 failures. Lint 0, ktlint and apiCheck clean. Reverting the comparison fails the new test.

## Stack

- 👉 this PR


geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
```diff
@@ -829,7 +838,10 @@ internal class GeofenceRepositoryImpl(
                     // re-center close to their real position. Clear only when nothing is registered
                     // (kill switch) — the trigger, and thus its location, is gone.
                     if (monitoringEnabled) {
-                        store.saveLastMovementTriggerLocation(GeofenceLocation(latitude, longitude))
+                        store.saveLastMovementTriggerLocation(
+                            GeofenceLocation(latitude, longitude),
+                            movementTriggerRadiusMeters ?: config.localRefreshTriggerRadius
```
Shahroz16, line 843: Could we also assert the persisted radius in the existing adaptive-radius test? It checks the registered and logged `725f`, but the store write is matched with `any()`. This code passes the correct value today; pinning it would catch a later change that accidentally saves the configured radius and restores the staleness bug. Non-blocking for this fix.

Shahroz16 (approved):
The live-fix path now compares against the radius actually registered, with the configured-radius fallback preserved for older stored state. All 391 focused tests pass. Reverting the comparison fails the shrunk-trigger regression while the legacy control still passes; restoring it passes both. This approves the radius correction, not rollout of the unresolved runtime issues in #869/#870. The red Compose job is the unchanged in-app teardown crash and still needs a green rerun.

geofence/src/main/kotlin/io/customer/geofence/GeofenceRepository.kt
mrehan27: Done in `72634e85`, pinned to `725f`. That `any()` was mine: a bulk edit fixing the signature change substituted matchers across the mock-based tests and weakened this one.

mrehan27:
On the red Compose job: it passed on the re-run, so it was the usual in-app teardown flake rather than anything from this stack.

Audited the `AndroidSDKComponent` usage in geofence anyway, since a cold-process entry point reaching it before setup would look the same. All 15 `SDKComponent.android()` call sites are covered: the three receivers and two workers each call `setupAndroidComponent` before any android access, and the one pre-setup read (`geofenceLogger` in the boot receiver) resolves only core dependencies. `GeofenceDiagnostics.readManifest` is the one unguarded-looking read and it is deliberate: it returns false without caching when the context is not reachable yet, so a later read after init still gets the real answer.

Shahroz16 (approved):
Thanks for pinning the persisted radius. The 235 focused repository/store tests pass, and changing the write back to the configured radius fails this assertion while restoring the actual radius passes it. Claude Code Opus review is complete and this head's CI is green. This approves the radius correction; the upstream runtime reviews remain separate.

PR #873, opened 2026-09-15
chore(geofence): refuse to arm an approach session that has already ended

mrehan27:
`stop()` already declines to tear down a session it does not name. `start()` had no equivalent rule, so a caller preempted after deciding to continue could resume holding the generation it captured before suspending, arm it, and replace the request a newer session armed while it was away. The displaced request is not restored when the stale one times out.

Raised by @Shahroz16 on #869 as a non-blocking follow-up.

## What changed

`start()` refuses any generation below the highest it has armed, since generations only move forward.

The high-water mark is tracked separately from the live generation because `stop()` clears the live one. Comparing against the live field alone would let a stale caller through once a session had ended, which is the same bug one step later.

## Verified

841 tests, 0 failures. Lint 0, ktlint and apiCheck clean. Removing the guard fails both new tests: the older-generation case and the after-a-stop case.

Reachability is not device-reproduced. The window is the coroutine resuming after `processApproachLocations`, and the cost of being wrong is a replaced request rather than a crash.

## Stack

- [#872 judge staleness against the trigger that was registered](https://github.com/customerio/customerio-android/pull/872)
- 👉 this PR


Shahroz16 (approved):
Thanks, this closes the newer-request displacement from #869, including after the newer request has stopped, while preserving same-generation restarts. The build and all 59 focused tests pass, including the controlled-preemption probe; removing the guard reproduces the displacement and fails both new regressions. Claude Code Opus completed the independent review. This is JVM/Robolectric evidence, not device timing; the pre-existing sign-out-only re-arm remains a separate non-blocking follow-up.

PR #875, opened 2026-09-15
fix(in-app): keep EngineWebView subscribed to the lifecycle owner it is currently under

### LINKED ISSUE #874: Inline message stops receiving taps after its host Fragment view is recreated (EngineWebView never re-subscribes to the new lifecycle owner)
**Environment:** messaging-in-app 4.18.2 (the same code is on `main`), via customerio-reactnative 6.5.2, React Native 0.86 (Fabric), react-native-screens 4.25.2 native stack, androidx.fragment 1.8.9, Android 16 (Pixel 9a).

**Steps**
1. Host `InlineInAppMessageView` (through the React Native `InlineInAppMessageView`) on a screen of a `@react-navigation/native-stack` navigator and let an inline message embed.
2. Push any other screen on top of it.
3. Go back.
4. Tap the message's action button or its close button.

**Expected:** the tap is handled (action dispatched / message dismissed).

**Actual:** nothing happens. No `Received event from WebView` line, no `tap`, no click metric. The message is still rendered, its JavaScript handlers are intact and `message.performAction(...)` returns without error. A message embedded *after* the navigation works; the damage is per `EngineWebView` instance and lasts until re-embed or process restart. Backgrounding/foregrounding the app does **not** cause it.

**Cause**

`EngineWebView.setup()` calls `viewLifecycleOwner?.addObserver(this)` exactly once, on whatever `ViewTreeLifecycleOwner` it finds — under a native stack that is the *view* lifecycle owner of the react-native-screens `ScreenFragment` hosting the screen, not the Activity.

When a screen is pushed, react-native-screens removes the Fragment below the top (`ScreenStack.onUpdate`). That Fragment's view lifecycle goes `ON_PAUSE` → `EngineWebView.onLifecyclePaused()` → `EngineWebViewInterface.detach()` sets `isAttachedToWebView = false` and removes the JavaScript interface, and is then destroyed. On pop the Fragment gets a **new** view lifecycle owner, but the same `EngineWebView` instance survives (React Native owns the view and react-native-screens re-parents it), and it is still registered on the destroyed owner. `ON_RESUME` never arrives, `isAttachedToWebView` stays `false`, and `EngineWebViewInterface.postMessage` drops every event at its first line:

```kotlin
if (!isAttachedToWebView) return
```

`BaseInlineInAppMessageView` already re-registers *its* observer in `onAttachedToWindow` (#611 / MBL-1329), which is why the store side survives return navigation — `EngineWebView` was never given the same treatment.

Log signature of a broken round trip (SDK at debug level): a single `EngineWebView onLifecyclePaused` at the push and no `EngineWebView onLifecycleResumed` after the pop.

**Suggested fix**

In `EngineWebView`, override `onAttachedToWindow` to `viewLifecycleOwner?.addObserver(this)` (and `onDetachedFromWindow` to remove it), or re-attach the interface on window attach. `LifecycleRegistry.addObserver` replays the owner's current state, so a RESUMED owner immediately re-attaches the interface.

**Workaround we verified** (in the React Native wrapper's `ReactInlineInAppMessageView`, applied as a patch):

```kotlin
override fun onAttachedToWindow() {
    super.onAttachedToWindow()
    val lifecycle = findViewTreeLifecycleOwner()?.lifecycle ?: return
    for (i in 0 until childCount) (getChildAt(i) as? LifecycleObserver)?.let(lifecycle::addObserver)
}

override fun onDetachedFromWindow() {
    findViewTreeLifecycleOwner()?.lifecycle?.let { l ->
        for (i in 0 until childCount) (getChildAt(i) as? LifecycleObserver)?.let(l::removeObserver)
    }
    super.onDetachedFromWindow()
}
```

With it, the pop logs `EngineWebView onLifecycleResumed` and taps reach native again on the same device and build.

ShanavasPS:
Fixes #874

### Problem

`EngineWebView` subscribes to a lifecycle owner exactly once, in `setup()`, and never records which one. Under a Fragment that owner is the Fragment's *view* lifecycle, which is replaced whenever the Fragment's view is destroyed and recreated. Hosts that keep the same `EngineWebView` instance alive across that — React Native does this on every native-stack push/pop through react-native-screens — see the old owner pause the engine (which detaches the JavaScript interface) and then die. The new owner never delivers `ON_RESUME` because nothing subscribed to it, so `EngineWebViewInterface.postMessage` drops every renderer event at its first line (`if (!isAttachedToWebView) return`), taps and close included, until the message is re-embedded or the process restarts. The message keeps rendering, so nothing looks wrong.

`BaseInlineInAppMessageView` already re-registers *its* observer in `onAttachedToWindow` (#611); this gives `EngineWebView` the same treatment.

### Changes

- `EngineWebView` now holds the lifecycle it observes (`observedLifecycle`), null while there is no message to keep the bridge attached for.
  - `setup()` subscribes through `observeCurrentLifecycle()`.
  - `onAttachedToWindow` moves the subscription to the owner currently reachable through the view tree if it differs. `LifecycleRegistry.addObserver` replays the owner's state, so a resumed owner re-attaches the interface immediately; subscribing to the same owner twice is a no-op.
  - `onDetachedFromWindow` drops the observer but keeps the reference, which records that there is a message to re-subscribe for.
  - `stopLoading()` and `releaseResources()` clear it, so a stopped engine is never re-subscribed by a later re-attach. They also now unsubscribe from the lifecycle actually observed instead of whichever one is current — after a Fragment swap that was a different owner, so the old registration leaked. `releaseResources()` previously did not unsubscribe at all.
- `EngineWebViewLifecycleTest` (Robolectric), six cases: owner replaced around a detach/re-attach; `setup()` run under an already-destroyed owner attaches once shown under a live one; a stopped engine stays unsubscribed across a re-attach; an engine without a message never subscribes; `stopLoading()` after an owner swap unsubscribes from the owner actually observed; `releaseResources()` unsubscribes. The interface is observed through Robolectric's WebView shadow, subscriptions through `LifecycleRegistry.observerCount`.

### Behaviour note

Temporary detaches under a live owner — a RecyclerView item scrolled off-screen, for example — drop the subscription and get it back on re-attach. The replayed `ON_RESUME` calls `onLifecycleResumed()`, whose `addJavascriptInterface` is idempotent and cheap, and while off-screen the engine receives no renderer events, so the missed `ON_PAUSE` has no effect. This mirrors what `BaseInlineInAppMessageView` already does with its own observer. Modal messages are unaffected: their engine lives in `GistModalActivity`, whose lifecycle owner is the Activity.

### Testing

- `./gradlew :messaginginapp:testDebugUnitTest --tests "*EngineWebViewLifecycleTest" --tests "*EngineWebViewSettingsTest"` — green locally. Run against the previous behaviour, the re-subscribe cases fail without the `onAttachedToWindow` move (3 of 6) and the teardown cases fail with the old "unsubscribe from whatever owner is current" code (2 of 6).
- `make lint`, `make validate-public-api` — clean locally (internal class, no API change).
- The same re-subscribe, applied from the React Native wrapper's host view, was verified on a Pixel 9a / Android 16: after a push/pop the SDK logs `EngineWebView onLifecycleResumed` and taps reach `tap()` again; without it the round trip logs only the pause and every tap is dropped.

https://claude.ai/code/session_01YUCnpw4wXV4oaVNp7Y7MMr


messaginginapp/src/main/java/io/customer/messaginginapp/gist/presentation/engine/EngineWebView.kt
```diff
@@ -90,6 +103,56 @@ internal class EngineWebView @JvmOverloads constructor(
...
+    override fun onAttachedToWindow() {
+        super.onAttachedToWindow()
+        if (observedLifecycle != null) {
+            observeCurrentLifecycle()
+        }
+    }
+
+    override fun onDetachedFromWindow() {
+        // Keep observedLifecycle: it records that there is a message to re-subscribe for.
+        observedLifecycle?.removeObserver(this)
```
Shahroz16, line 132: Could we pause the active load before dropping this lifecycle subscription? If a loading inline view leaves the window while its owner is still resumed and the app then backgrounds, the engine now misses ON_PAUSE and its five-second timeout dismisses the message in the background. I reproduced that callback path with Robolectric; it stays paused with the previous code. Pausing the observed engine on detach, with a detach-then-background regression test, would preserve the existing behavior.

Shahroz16 (commented):
The owner-recreation path now delivers taps, and the stopped-engine guard behaves correctly. I'm holding approval for the detach-then-background timeout regression noted inline.

PR #878, opened 2026-09-15
chore(geofence): tie the approach session to the generation the store says is live

mrehan27:
Sign-out ends a user generation, but `PolygonApproachMonitor` could not see that. `stop` clears the live generation and sign-out arms nothing newer, so a receiver preempted between deciding to sample and calling `start` resumed with the old generation and opened a fine-location session for the signed-out user.

`start` now refuses any generation the store does not report as live. The read is under the monitor's lock, so sign-out either bumps the generation before the check, or names the request armed here and removes it.

This replaces `highestArmedUserStateGeneration` from #873. The store subsumes it: generations only move forward, and every caller's generation comes from the store. It also covers a case that field could not, since it was null in a fresh process, which is exactly where a stale PendingIntent extra arrives.

Trade-off: the monitor's lock can now block behind an encrypted-prefs commit held under the store's lock. That critical section already made a binder call, so this widens an existing hold rather than adding a new one.

## Verified

846 tests, 0 failures. ktlint, apiCheck and Android Lint clean.

Mutation: deleting the guard fails 3 tests, including both added by #873. Inverting the comparison fails 10, including the positive control.

Reported by @Shahroz16 on #869.

## Stack

- 👉 this PR
- [#879 record a fix that was evaluated but could not be judged](https://github.com/customerio/customerio-android/pull/879)


Shahroz16 (approved):
Thanks, this closes the sign-out-only re-arm left after #873 while preserving same-user restarts and newer-user requests. The build, 60 focused tests and three controlled-preemption probes pass; removing the store guard reproduces both stale-request paths. Claude Code Opus completed the independent review. This is JVM/Robolectric evidence, not a device-timing claim.

PR #879, opened 2026-09-15
chore(geofence): record a fix that was evaluated but could not be judged

mrehan27:
A polygon fix that arrived, was usable, and was evaluated but could not separate inside from outside produced no record at all, so a capture cannot tell it from a fix that never arrived. Tomorrow's calibration drive measures exactly that case.

The evaluator now reports why it could not judge: `accuracy_too_low`, `on_boundary`, `within_accuracy`. The processor emits `ev=polygon.undecided` with `id`, `sh`, `why`, `edge` and `acc`, alongside the existing fix-level producer of the same event.

`AMBIGUOUS` covered four situations, and one is a decisive fix that agrees with the committed state. That happens on nearly every fix for a device sitting inside a polygon, so it reports no reason and writes no record. Worst realistic volume is indoors, where every fix exceeds the 50 m decisive ceiling: roughly 24 records per 2 minute session with 3 active polygons.

`edge` is a minimum distance to the ring, so it does not say which side the fix was on. The committed state is derivable from the surrounding transition records.

## Verified

849 tests, 0 failures. ktlint, apiCheck and Android Lint clean.

Mutation: making the agreeing branch report a reason fails the anti-flood test, deleting the record fails both positive tests, and renaming a wire token fails two pinning tests. That last one was a real gap found in review: the tokens are shared with iOS and nothing pinned them.

## Stack

- [#878 tie the approach session to the generation the store says is live](https://github.com/customerio/customerio-android/pull/878)
- 👉 this PR


Shahroz16 (approved):
The new records distinguish genuinely undecidable fixes from ordinary state agreement without changing transition evidence. The build and 91 focused tests pass, along with 22,500 baseline comparisons, actual-output checks and the three mutation checks. Claude Code Opus completed the independent review. For capture analysis, records with an id are per fence, while the existing id-less records are per fix, so raw event totals are not directly comparable. Device captures and the off-device parser were not exercised.

PR #880, opened 2026-09-15
chore(geofence): sign the polygon undecided edge and record the fix age

mrehan27:

## Stack

- 👉 [#880 sign the polygon undecided edge and record the fix age](https://github.com/customerio/customerio-android/pull/880)

#879 added `ev=polygon.undecided` but left two things out that the calibration work needs.

`edge` was an unsigned minimum distance to the ring, so a row could not tell a fix refused 26 m inside the polygon from one refused 26 m outside it. The evaluator already computed the relation and discarded the sign. It is now positive inside, negative outside, matching iOS.

The decisions themselves still compare the unsigned distance. The sign is attached only on the way out to the record, so this changes no behaviour.

`age` joins it so a captured trace can be replayed against a staleness rule offline. The computation is shared with the existing `age` on fix-quality records rather than duplicated.

## Verified

852 tests, 0 failures. ktlint, apiCheck and Android Lint clean.

Mutation: stripping the sign fails the outside-edge test; dropping `age` fails both the field test and the log-tail pin.


Shahroz16 (approved):
Verified that the signed edge only changes diagnostics, and the age comes from the actual fix's monotonic timestamp. The module builds and all 94 focused tests pass; CI is green. Looks good.

PR #882, opened 2026-09-15
chore(geofence): judge a polygon arrival by the fix, its departure by the margin

mrehan27:

## Stack

- 👉 this PR

## The problem

The decisive rule required `edge > accuracy + margin` in both directions. Three of our four real polygons are 24-42 m inradius, and at median background accuracy no interior point of those satisfies it. Those fences could never report an arrival from a single fix — not at any session timing, not at any approach speed. The two PRs below this one improve when and how often we sample; neither helps, because the geometry refuses first.

## The change

Arrival asks only which side of the ring the fix is on. Departure keeps the clearance margin, widened from 10 m to 20 m.

The asymmetry is the point. A missed arrival is recovered only if the device moves again: the movement trigger's EXIT and any approach session opened nearby both re-activate the polygon and evaluate that fix, so one polygon's wake can recover a neighbour's missed arrival. Stand still after a missed wake and nothing re-derives the visit, because polygons are excluded from initial-ENTER synthesis. That stationary case is the one this trades for, and it is the main use case. A spurious arrival costs a transition that the next decisive fix corrects. Departure is the opposite — ending a visit early on a noisy fix reports something that did not happen.

## Corroboration expires

A pending arrival is evidence about the visit happening now. The confirmation counted agreeing fixes and stored no time, and stopping the two-minute approach request does not clear it for a polygon that stays active, so a marginal fix on one pass and another on a wake five minutes later completed an arrival neither pass observed. Walking past a shop twice is not a visit.

Retired once the gap between fixes exceeds a minute. The session samples every 15 s, so a corroborating fix is normally the next one. Reported by Shahroz with a reproduction; his five-minute gap is the test, and a one-sample control holds the other side so expiry cannot quietly disable corroboration instead of bounding it.

## What this costs

A device just outside a ring whose fix lands inside will report a visit it did not make. That is the accepted direction, and the widened departure margin keeps it from flickering back out immediately.

**The catchment concern is addressed in the second commit.** The first commit removed the clearance margin, which left nothing bounding accuracy against the boundary — any fix whose reported centre landed inside committed a visit. Measured on our own cached rings at the median background accuracy, the exterior band such a fix could have come from is 9,600-10,400 m² per retail fence, up to 1.8x the venue's own area. That band is the pavement. (`customerio-ios-1d` raised this and computed it independently; our two figures for Rainbow agree to 0.06%.)

Arrival now asks for corroboration only when the fix could have come from either side:

| | rule |
|---|---|
| uncertainty does not reach the ring | decides alone — it cannot have been taken from outside |
| uncertainty reaches the ring | needs a second agreeing fix |

A fix too coarse to judge breaks the run rather than counting toward it, so two inside readings minutes apart cannot corroborate each other.

Effect at the median accuracy of 19.5 m:

| fence | instant | needs 2 | single-fix catchment |
|---|---|---|---|
| Bundu Khan | 39% | 61% | 10,443 m² → 0 |
| Rainbow | 32% | 68% | 9,976 m² → 1 m² |
| Tim Hortons | 4% | 96% | 9,631 m² → 0 |
| Safari Homes | 88% | 12% | 56,804 m² → 0 |

The cost is one sampling interval, about 15 s, for most of a retail interior. Not a lost visit.

Worth noting what this does *not* rule out: a passer-by producing two consecutive inside-reading fixes. That needs two independent errors in the same direction across ~15-30 s of walking, which is much rarer but is not computable from geometry — it needs the field capture.

The accuracy ceiling stays at 50 m. Indoor fixes measure around 100 m and carry no containment information about a 24 m fence, so raising it needs the real accuracy distribution rather than an argument from geometry. That is the one number here still waiting on a measurement.

## Verified

870 tests, 0 failures. ktlint, apiCheck and Android Lint clean. 37 production lines.

Mutation, both directions: restoring the symmetric rule fails four arrival tests; dropping the departure margin fails five departure tests. No test covers both.

Two tests that asserted the old guarantee were rewritten rather than adjusted — `givenPolygonSmallerThanTypicalAccuracy_expectNoEnterEverCommitted` and `givenAccuracyCircleStraddlingTheRing_expectNoTransition`. Both encoded "a small or marginal fix never enters", which is precisely what changes. The half of the first one that still holds, that arrival is judged against the ring and never the enclosing trigger circle, is now its own test.


geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonAccuracyEvaluator.kt
```diff
@@ -121,18 +130,21 @@ internal class PolygonAccuracyEvaluator {
...
         val boundaryDistanceMeters = geometry.boundaryDistanceMeters(sample.coordinate)
-        val requiredDistance = sample.horizontalAccuracyMeters + DECISIVE_BOUNDARY_MARGIN_METERS
-        if (boundaryDistanceMeters <= requiredDistance) {
-            return undecided(
-                PolygonUndecidedReason.WITHIN_ACCURACY,
-                signedBoundaryDistance(boundaryDistanceMeters, relation)
-            )
-        }
         return when {
             committedState == PolygonCommittedState.OUTSIDE && relation == PolygonPointRelation.INSIDE ->
```
Shahroz16, line 139: For a calibration follow-up, could we record the signed edge, accuracy and age for accepted arrivals too? This path now produces a transition without the evidence fields, and matching-state near-boundary fixes no longer produce undecided rows. Manual ground-truth notes can still tell us whether an arrival was false, but the SDK capture alone won't show which fix quality led to it. This doesn't need to hold up the experiment.

Shahroz16 (approved):
Approving this evaluator change for the calibration experiment, not as validation of the production false-arrival policy. The module builds and all 63 focused tests pass. The stack still needs the mixed-polygon fix in #881 before landing.

One detail for the calibration notes: cooldown is per direction. A false ENTER does not delay its correcting EXIT, but it can suppress a later real ENTER for the remaining window. Could we correct that distinction in the description and update the engine's old symmetric-rule documentation?

mrehan27:
Good call on the evidence fields, and the underlying concern turned out to be larger than a logging gap — thanks for scoping the approval the way you did.

Removing the clearance margin left nothing bounding accuracy against the boundary. Measured on our own cached rings at the median background accuracy, the exterior band a single inside-reading fix could have been taken from is 9,600-10,400 m2 per retail fence, up to 1.8x the venue's own area. That band is the pavement, so a large share of reported visits would have been people walking past.

The second commit fixes it. A fix whose uncertainty does not reach the ring still decides alone; one that could have come from either side now needs a second agreeing fix, and a fix too coarse to judge breaks the run rather than counting toward it. The single-fix band drops to under 1 m2 on all four fences. The cost is one sampling interval for most of a retail interior.

On the record itself: I have not added evidence fields to accepted arrivals in this PR, because the change above altered what is worth recording. An accepted arrival now has a second dimension - whether it decided alone or was corroborated - and that belongs in the record alongside the signed edge, accuracy and age. Happy to do it as the next commit here if you would rather it landed before the field capture, since without it the trace says why fixes were refused and nothing about why they were accepted.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonAccuracyEvaluator.kt
mrehan27: Planned as the next commit, not in this one. It will record signed edge, accuracy and age for accepted arrivals, split by whether the fix decided alone or needed corroboration, and also for fixes the accuracy gate discards. A log of only what passed the gate cannot tell us where to move the gate.

Separately, please re-read the asymmetry paragraph before you merge. I had written that a missed arrival has nothing to re-derive it, and that was too strong. onMovementTriggerExit re-activates every polygon within range of the triggering fix and evaluates it, and any approach session opened nearby does the same, so one polygon's wake can recover a neighbour's missed arrival. What is genuinely unrecoverable is the stationary case: arrive, miss the wake, stay put. The trade still points the same way because that is the main use case, but you approved on a broader claim than was true.

Shahroz16 (commented):
Thanks for correcting the recovery explanation: a later movement or nearby wake can recover an arrival; the stationary case is the limitation. The evidence records in #884 also address the logging follow-up. I'm keeping this scoped to calibration, but holding approval on the current revision for the confirmation-lifetime issue below.

For the calibration notes, Android's accuracy radius is a 68% confidence estimate, not a hard bound, so clearing that circle doesn't prove a fix could not have come from outside. The geometric catchment calculation should stay separate from a measured false-arrival rate. The engine's old symmetric-rule class documentation also still needs updating.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonRouteProcessor.kt
```diff
@@ -79,14 +88,33 @@ internal class PolygonRouteProcessor(
...
+                    // passer-by on the pavement cannot produce one. A marginal fix can, so it
+                    // needs a second agreeing fix — which a real visit supplies on the next
+                    // sample and someone walking past usually does not.
+                    PolygonEvidence.ENTER -> when {
+                        !result.requiresCorroboration -> {
+                            arrivalConfirmations.clear(fence.id)
+                            PolygonTransition.ENTER
+                        }
+                        else -> arrivalConfirmations.evaluate(fence.id, committedState, evidence)
+                            ?: return@mapNotNull null
```
Shahroz16, line 113: Could we expire pending arrival confirmations when the sampling window ends, or bound their age using the fix timestamps? The confirmation stores only a count, and stopping the two-minute approach request doesn't clear it for a polygon that remains active. I reproduced a marginal fix followed by another wake five minutes later: the second marginal fix commits ENTER using the old observation. That can combine separate passes by a shop rather than corroborate the current visit. Expiring the old confirmation makes the controller/engine regression pass; a test spanning an expired window and a new wake would cover this.

mrehan27: Good catch, and the reproduction made it quick. Fixed rather than deferred, since it lets two separate passes at a shop complete an arrival neither of them observed.

Went with bounding the age rather than hooking the session end: it holds even when a polygon stays active across wakes, which is exactly the case you reproduced, and it is testable without the controller. A pending arrival is retired once the gap between fixes exceeds a minute, against a 15 s sampling interval, so a real corroborating fix is unaffected.

Your five-minute gap is one of the two tests. The other is a one-sample control, because expiry that never fires and expiry that always fires both make the first test pass, and the second of those would silently turn corroboration off. Mutation: removing the expiry fails the first, an unbounded gap fails the first, a zero gap fails the control.

mrehan27:
bugbot run

Shahroz16 (approved):
Approving this revision for the calibration experiment. The confirmation-lifetime issue is fixed, the module builds, and all 67 focused tests pass. The completed Claude Code Opus pass found no new blocker. The earlier documentation notes still apply, and this is not validation of the production false-arrival rate.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonRouteProcessor.kt
Shahroz16: Thanks, the one-minute bound addresses this. I reran the controller/engine case with a new wake five minutes later, and it no longer combines the old fix. A fresh fix 15 seconds after that still completes the arrival. Removing the expiry reproduces the failure, and restoring it makes the test pass again. The 60-second boundary and replay checks also pass in the local JVM tests.

PR #883, opened 2026-09-16
chore(geofence): stop seeding polygon containment from the wake circle

mrehan27:

## Stack

- 👉 this PR

## Found by running it

Setting up today's emulator validation, three polygons emitted EXIT from 2.5 km away, **before any of them had ever been entered**:

```
14:37:20 ev=os.callback.received ids=cio_movement_trigger t=exit
14:37:20 ev=transition.accepted id=18 t=exit n=1
14:37:20 ev=transition.accepted id=16 t=exit n=1
14:37:20 ev=transition.accepted id=17 t=exit n=1
14:38:09 ev=os.callback.received ids=18,16 t=enter      <- the first real approach, 49s later
```

`reconcileContainment` seeds a region as entered whenever the fix falls inside its registered radius. For a polygon that radius is the wake circle, not the ring — a kilometre against a 40 m shop. A foreground fix a few hundred metres from the centres had marked all three as visited.

## Why it matters more than one wrong event

The phantom EXIT spends that fence's hour-long duplicate-event cooldown, so **the next genuine visit to the same shop is silently suppressed**. In a field capture that reads as "the SDK didn't fire", which is the wrong conclusion.

The same record feeds four other readers, so the one line fixes more than the phantom event: the polygon was pinned through the discovery caps, re-activated on coarse EXIT, handed to the evaluator as committed INSIDE so the next fix was judged a departure, and used to shrink the movement trigger.

## The fix

Exclude polygons from the seed, exactly as the sibling seed in `registerNearestAndPersist` already does and for the stated reason — a polygon's containment belongs to the accuracy-aware evaluator, and a sync fix carries none of that evidence.

Circles are unchanged and still seed as before. The fail-open argument in the existing comment holds for a circle, where the registered radius *is* the fence.

`reconcileEnteredIds` is additive (`carried + stillInside`), so a genuinely entered polygon is not dropped by being left out of the seed.

## Verified

869 tests, 0 failures. ktlint, apiCheck and Android Lint clean. One production line.

Mutation: restoring the unfiltered seed fails the new test specifically.

Pre-existing, not introduced by this stack — but the stack makes it reachable more often, and it would have corrupted today's field capture.

## Upgrading

An install that already ran the old code carries polygon ids in `entered_ids`, and the carry-forward in `reconcileEnteredIds` keeps them while the fence stays registered. Each reports one EXIT it never entered, once, on the next decisive outside fix. Pre-release installs only, and a sign-in or sign-out clears it.


Shahroz16 (approved):
Verified that a wake circle can no longer seed a polygon visit, while circles still seed and genuinely entered polygons remain carried forward. The module builds and all 236 focused tests pass; restoring the old predicate fails the regression. This fix is ready. The stack still needs the confirmation-lifetime issue in #882 addressed before landing.

PR #884, opened 2026-09-16
chore(geofence): record the evidence behind a polygon decision, not only a refusal

mrehan27:

## Stack

- 👉 this PR

## The problem

`polygon.undecided` records every fix the margins turned away. Nothing records the ones they let through, so a capture shows one side of the distribution the margins sit in.

That matters for the drive. `MAX_DECISIVE_FIX_ACCURACY_METERS` is 50.0 and no measurement supports that number yet. A log of refusals alone cannot say where to move it: it shows which fixes were turned away and none of the ones that were accepted.

## The change

A `polygon.decided` record on every claimed transition, with the same fields its refusal counterpart carries plus one more:

```
ev=polygon.decided id=16 sh=polygon t=enter edge=41.2 acc=5.0 age=2.5 cor=false
```

`cor` separates an arrival whose accuracy circle never reached the ring from one that needed a second agreeing fix, which is the split #882 introduced and the one the false-arrival policy is calibrated against.

The signed edge now rides on decided results too, so accepted and refused fixes are directly comparable. It was previously null outside the undecided paths.

A fix discarded as too old also carries its accuracy now. It is a fix the session received and never evaluated, which is exactly the population a gate is calibrated against.

## And the fixes that changed nothing

A third record, `polygon.unchanged`, for a fix that was decisive and agreed with what we already believe. It carries membership, signed edge, accuracy and age.

Those are most of the fixes. In today's emulator walk-in, ten were evaluated and one left a record. Refusals and transitions are the two ends; the ordinary good fix in between was invisible, and that is the population a margin is calibrated against.

Bounded by the sampling session, so roughly eight rows per fence per wake rather than a stream. iOS already records this as `polygon.verdict` plus `polygon.undelivered why=no_change`.

## Not in this PR

No thresholds move. This is the instrumentation that lets the drive decide whether any of them should.

## Verified

875 tests, 0 failures. ktlint, apiCheck and Android Lint clean.

Mutation: dropping the record fails all three new tests; hardcoding "decided alone" fails only the corroboration test; unsigning the edge fails only the departure test.

Review found three more guards that survived mutation, plus a KDoc that was false for the confirmed evidence policy, which returned its transitions with no signed edge at all. Fixed in the second commit; all three now fail a named test. One of them survived because `any()` in a mockk verify also matches the nulls the defaults supply.

The log vocabulary contract test rejected `polygon.decided` until it was declared, so the new key is registered rather than smuggled in.


Shahroz16 (approved):
The new records distinguish evaluator decisions from delivered events and carry the expected signed edge, accuracy, age and corroboration flag. The module builds and all 98 focused tests pass; the logging, sign and corroboration mutations fail as expected. This addresses the calibration logging follow-up from #882. Approving this diagnostic change; the stack still needs #882's confirmation-lifetime issue addressed.

PR #885, opened 2026-09-16
test(geofence): the recorded-scenario format and its matcher

mahmoud-elmorabea:
**Android replay harness — 1 of 4**

`base` → **#885** → #886 → #887 → #888

The file format the harness reads, and the matcher that compares what the SDK emitted against what was expected.

Contributes to MBL-2309
Contributes to MBL-2421


geofence/src/test/java/io/customer/geofence/replay/Scenario.kt
```diff
@@ -0,0 +1,178 @@
...
+        )
+    }
+
+    private fun parseRecord(obj: JsonObject, fileName: String, lineNumber: Int): ScenarioRecord? {
+        val kind = when (obj["k"]?.jsonPrimitive?.content) {
+            "given" -> Scenario.Kind.GIVEN
+            "when" -> Scenario.Kind.WHEN
+            "then" -> Scenario.Kind.THEN
+            "note" -> Scenario.Kind.NOTE
+            else -> return null
```
Shahroz16: Could we reject an unknown `k` here instead of returning null? `mapIndexedNotNull` currently drops that record while still treating the file as readable, so a malformed capture can disappear from the replay without making discovery fail.

Shahroz16 (approved):
Approving the format and matcher changes. The exact head is clean in hosted checks, and the parser and matcher contract is consistent with the recorded fields. One follow-up would be to reject an unknown `k` instead of dropping it in `mapIndexedNotNull`, since a malformed record would otherwise disappear from a replay without making discovery fail.

geofence/src/test/java/io/customer/geofence/replay/Scenario.kt
mahmoud-elmorabea: Done in e571bccf — `parseRecord` now throws naming the file, the line and the offending `k`, and the call site is `mapIndexed` rather than `mapIndexedNotNull`.

iOS already threw on the same input (`LoadError.malformedLine(_, "unknown kind ...")`), so this was a divergence between two readers of one shared format. Added `load_givenUnknownKind_expectRejected` to hold it. Safe for the corpus: the only `k` values present across all 39 files are given/when/then/note/scenario.

mahmoud-elmorabea:
Force-pushed with the unknown-`k` fix from your review — flagging rather than re-requesting, since this is approved.

Details on the thread. Short version: `parseRecord` throws now, matching what iOS already did, and `load_givenUnknownKind_expectRejected` holds it. No corpus file is affected.

geofence/src/test/java/io/customer/geofence/replay/Scenario.kt
```diff
@@ -0,0 +1,181 @@
...
+     */
+    fun geofenceIds(): List<String> {
+        string("ids")?.let { raw ->
+            return raw.split(",").map { it.trim() }.filter { it.isNotEmpty() }
+        }
+        return listOfNotNull(string("id"))
+    }
+}
+
+internal data class ScenarioFence(
```
Shahroz16, line 84: Could we preserve the polygon payload in the scenario fence before using this harness for polygon QA? parseBody currently requires top-level latitude, longitude, and radius and drops shape, geometry, and enclosing_circle. ReplayApiService then rebuilds a circle-only GeofenceApiRegion. A geometry-only polygon is silently omitted, while a flattened enclosing circle can replay green without ever exercising polygon decoding or containment. Keeping the raw region JSON or the typed polygon fields, plus one synthetic polygon replay, would cover the real boundary. At minimum, rejecting unsupported shape records would make this fail closed instead of quietly grading the wrong path.

Shahroz16 (commented):
I exercised the harness for the polygon use case. Its own tests pass, but the fixture contract cannot currently carry a polygon, so this needs an extension before it can validate polygon behavior.

PR #886, opened 2026-09-16
test(geofence): the seams the replay substitutes

mahmoud-elmorabea:
**Android replay harness — 2 of 4**

`base` → #885 → **#886** → #887 → #888

The four seams the harness substitutes: Play Services, the network, the wall clock, and event delivery. Everything else runs for real.

Contributes to MBL-2309
Contributes to MBL-2421


geofence/src/test/java/io/customer/geofence/replay/ReplayDoubles.kt
```diff
@@ -0,0 +1,312 @@
...
+ * populated by a fix that already arrived. The harness feeds it from the stimuli it has replayed
+ * so the anchor the SDK computes is the one the drive would have computed.
+ */
+internal class ReplayLocationServices : LocationServices {
+    /** Fixes the SDK asked for. A drive's foreground entries should each produce one. */
+    var silentRequestCount = 0
+        private set
+
+    var lastKnown: LocationCoordinates? = null
+
```
Shahroz16, line 299: Could we wire the replayed fixes through `setLastKnownLocation` or narrow this claim? The runner only updates its private `lastFix`; this property is reset and read but never populated, so `resolveAnchor` never exercises the fallback branch when registration has no center.

geofence/src/test/java/io/customer/geofence/replay/ReplayDoubles.kt
```diff
@@ -0,0 +1,312 @@
...
+        return Result.success(Unit)
+    }
+
+    override suspend fun replaceGeofencesForBootRestore(regions: List<GeofenceRegion>): Result<Unit> =
+        replaceGeofences(regions, emptySet())
+
+    override suspend fun removeGeofencesByIds(ids: List<String>): Result<Unit> {
+        roundTrip("remove(${ids.joinToString(",")})", removeAnswers)
+        registeredIds.removeAll(ids.toSet())
+        return Result.success(Unit)
```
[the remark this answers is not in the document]
Shahroz16: Confirmed from the production path: an empty replace or remove returns immediately without a Play Services call. The replay double still parks and consumes an answer, so it can shift later virtual timestamps. I am treating this as a required replay-fidelity fix.

Shahroz16 (changes requested):
Requesting two small replay-fidelity fixes before this stack moves on. `ReplayRegistrar` waits for every `replaceGeofences` call, but production returns immediately when the region list is empty, so a maxBusinessGeofences=0 replay can consume a fake OS answer and shift later timestamps. Also, `ReplayLocationServices.lastKnown` is never populated by the runner, so the fallback anchor path described in this PR is not exercised. Could we make empty calls no-ops and either feed lastKnown from delivered fixes or narrow that KDoc? Hosted checks are green.

geofence/src/test/java/io/customer/geofence/replay/ReplayDoubles.kt
mahmoud-elmorabea: Took the second option — narrowed the claim — because the evidence says wiring it would be the bug.

`setLastKnownLocation` is a host-app API: "sets the last known location from the host app's existing location system." Nothing inside the SDK calls it, so a `prov=bus` fix does not populate it, and the only caller anywhere in the repo is a manual screen in the sample app that no recorded drive used. `getLastKnownLocation()` therefore returned null throughout every capture, which is exactly what the double does — so `resolveAnchor` fell through to its `registrationCenter` branch on the phone as it does here.

Feeding it from replayed fixes would hand the replay an anchor production never had, which is the failure the runner's own `identity.changed` comment describes from the 2026-09-12 drive.

66926bfb rewrites the KDoc to say it stays empty and why. If you would rather exercise that fallback branch, I think it wants an authored conformance scenario with an explicit host-set position rather than the runner inferring one — happy to add that separately.

mahmoud-elmorabea:
Addressed in 66926bfb.

`replaceGeofences` and `removeGeofencesByIds` now return early on an empty list, matching `GeofenceManager` (`replaceGeofencesInternal` returns at the empty-regions guard, `removeGeofencesByIds` at the empty-ids guard, neither touching Play Services). The double no longer spends a recorded answer on a call the OS never saw, and no longer clears `registeredIds` where production leaves the registrations alone.

Latent for the current corpus — no recorded drive registers zero fences, the smallest is 9 — so no result changes.

`lastKnown` answered on its thread: narrowed the KDoc rather than wiring it, because the wiring would be the bug.

geofence/src/test/java/io/customer/geofence/replay/ReplayDoubles.kt
Shahroz16: Thanks for narrowing the comment, but following the production getter one level deeper still leaves a fidelity gap. `LocationServicesImpl.getLastKnownLocation()` reads `LocationTracker.lastKnownLocation`, and the silent location path updates that value before it publishes `LocationAcquired`. A replayed bus fix therefore does create SDK last-known state even though the host-facing setter is never called. Could we mirror that update when the runner delivers the fix, and add a case where registration center is absent so the fallback is exercised?

Shahroz16 (changes requested):
The empty registrar calls are now faithful and the replay tests pass locally. The remaining last-known-location path is still different from production, as noted in the existing thread. The hosted messaging-push failure is an unrelated Robolectric artifact download, not a failing assertion in this PR.

geofence/src/test/java/io/customer/geofence/replay/ReplayDoubles.kt
Shahroz16: You're right that this head does not deliver recorded fixes yet, so the mismatch is not reachable in this PR. One nuance for the runner follow-up: production does populate LocationTracker.lastKnownLocation internally before publishing LocationAcquired, even without the host setter. We should mirror that when fix delivery lands, but it should not hold this foundation PR. Clearing my block.

Shahroz16 (approved):
Rechecked the exact head and ran the replay tests. The last-known parity point is not reachable until the runner delivers fixes, so it does not block this foundation PR. The remaining concurrency markers are not reachable under the serialized replay contract. Looks good.

PR #888, opened 2026-09-16
test(geofence): replay recorded drives and grade the SDK against them

mahmoud-elmorabea:
**Android replay harness — 4 of 4**

`base` → #885 → #886 → #887 → **#888**

The suite that grades the SDK against its inputs.

**This suite reports as skipped in CI.** Its inputs are supplied only when running locally; two assumptions rather than an early return, because an early return reports as passed.

Contributes to MBL-2309
Contributes to MBL-2421


Shahroz16 (approved):
Approving the recorded-drive suite. Discovery now reports unreadable files and refuses a present corpus with no recorded drives, and hosted checks pass. The external corpus remains intentionally skipped when it is not mounted, as documented.

mahmoud-elmorabea:
Rebase only — no content change. Carries fixes made in #885, #886 and #887 from this review round. Corpus run green: 14 tests, 0 failures.

mahmoud-elmorabea:
Force-pushed a reporting change from device testing.

The replay looped every drive inside one test method, so a hundred drives with one bad reported as a single failure whose message carried the entire corpus. It now parameterises on the same discovery list the iOS suite uses, one case per drive, named by scenario:

```
ScenarioReplayTest: 4 tests, 1 failure
  ✓ [android-Google-Pixel-8-Pro-20260918-1619]
  ✓ [android-Google-Pixel-8-Pro-20260921-1850]
  ✓ [android-samsung-SM-S921B-20260921-1842]
  ✗ [android-samsung-SM-S921B-20260918-1617]
```

Discovery moves to its own `ScenarioDiscoveryTest`. It is a check about the corpus as a whole and would otherwise re-run, and re-report, once per drive.

An absent corpus yields one placeholder case that skips — JUnit refuses a parameterised class with no parameters, and that error would read as a broken suite rather than a missing checkout. Verified both ways: 15 tests with drives present, 14 with 2 skipped when absent. ktlint clean.

Shahroz16 (changes requested):
Could we fail when `CIO_GEOFENCE_SCENARIOS` is set but does not point to a directory? Right now `takeIf { it.isDirectory }` turns that into `root == null`, so the discovery check and the placeholder replay both skip. A typo or moved checkout therefore looks exactly like an intentionally absent corpus and the harness can go green without replaying anything. Treating an explicit but unusable override as an error would close that gap.

The per-drive parameterization itself looks good. I ran both focused classes locally; without the private corpus they correctly report two skipped cases, and CI is green at this head.

mahmoud-elmorabea:
Done in 72aba5ee.

An override that names something other than a directory now throws, naming the path and what it should point at, so a typo or a moved checkout fails the run instead of emptying it. Previously it resolved to `root == null` — the same answer as "no corpus on this machine" — and since every caller guards on `assumeTrue(Scenarios.isAvailable)`, discovery and every replay skipped together and the suite went green having replayed nothing.

Blank is treated as unset rather than as an error. It asserts no location, and an environment that exports the variable empty is not claiming to know where the corpus is — happy to make that throw too if you would rather it did.

The resolution moved into `Scenarios.resolve(override, from:)` so the rules are testable at all: `root` reads `getenv` once and caches it, which no test can arrange. Four cases cover it — a path that does not exist, a path naming a file rather than its directory (the likelier typo), a usable directory, and blank or absent falling through to the sibling walk.

48 replay tests green, and that includes all 31 recorded-drive cases against the private corpus rather than skipping them, so the new failure mode is the only behaviour that changed.

Shahroz16 (approved):
The explicit override now fails closed, and the new resolution cases pass locally. This addresses my earlier concern; the current head and hosted checks are green.

PR #889, opened 2026-09-16
chore(geofence): pin the polygon corroboration and coarse-enter paths

mrehan27:
Two regression tests for the polygon evaluation path. No behaviour change.

**A repeated fix cannot corroborate its own arrival.** iOS hit this: their corroboration required the second fix to postdate the last fix their resolver had *delivered*, not the fix being corroborated, so a cached fix echoed back cleared the guard and a marginal arrival confirmed itself from one observation.

Ours is already correct. The per-fence monotonic check in `PolygonRouteProcessor` runs ahead of every evaluation path, so an echo never reaches `confirmArrival`. But nothing pinned it on the corroboration path, and the defect is invisible in a capture because an echo and a real second opinion produce identical records.

**A delivered coarse ENTER reaches the store.** On the 2026-09-16 drive the device crossed fence 16's wake circle cleanly and the region store came back unchanged. The logs that would say whether GMS delivered were lost, so this pins the half that can be tested without a device: given a delivered callback for a registered, routable polygon, the store must change.

It passes, which is the useful result. It narrows that failure to delivery rather than to anything the controller does with a callback it receives.

Uses the real `GeofenceRegionStoreImpl`. Every precondition the controller checks reads through the store, and a relaxed mock answers all of them affirmatively, which is the shape that would hide exactly this.

**Verification**
- Mutation: removing the monotonic guard fails the first test; a narrow mutant that bypasses it only on the corroboration path fails that test and nothing else in 879
- Mutation: dropping the two persistence calls from `activate` fails the second test
- ktlint 0, Android Lint 0, 879 tests, 0 failures

Independent of 890, both off `feature/geofence-polygons`.


geofence/src/test/java/io/customer/geofence/polygon/PolygonCoarseEnterRealStoreTest.kt
```diff
@@ -0,0 +1,103 @@
...
+import org.robolectric.RobolectricTestRunner
+
+/**
+ * The coarse polygon ENTER path against the real [GeofenceRegionStoreImpl].
+ *
+ * Written after the 2026-09-16 drive, where the OS had a clean inbound crossing of fence 16's
+ * wake circle and the region store recorded nothing at all. The logs that would say whether GMS
+ * delivered were lost, so this pins the half we can test without a device: given a delivered
+ * callback for a registered, routable polygon, the store must come out changed. A green run here
+ * narrows that failure to delivery; a red one would have found it without another drive.
```
Shahroz16: Could we narrow this to proving controller persistence once activation is reached? The test bypasses receiver routing and the revision check, so a green result doesn't isolate the drive failure to GMS delivery. The real-store check is useful, but those other paths still need to stay open in the diagnosis. The relaxed-mock explanation below also needs a tweak: its empty routable set rejects activation unless explicitly stubbed.

Shahroz16 (approved):
The module builds and all 27 focused tests pass. Removing the duplicate-fix guard or the controller persistence calls makes the corresponding new test fail, and restoring them passes again. The completed Claude Code Opus pass found no runtime blocker; one non-blocking scope clarification below.

geofence/src/test/java/io/customer/geofence/polygon/PolygonCoarseEnterRealStoreTest.kt
mrehan27: Both correct, fixed in 0f2d68f1.

Scope narrowed: the KDoc now says the test calls the controller directly, so it covers neither receiver routing nor the revision check (passed as null), and that a green run does not isolate the failure to delivery. Routing and revision stay open suspects.

The mock rationale was backwards and I have corrected it. A relaxed mock relaxes getRoutableRegisteredIds() to an empty set, which rejects activation rather than waving it through. The real reason for the real store is that the persistence being asserted is the store's own.

Comment only, no assertion change.

PR #890, opened 2026-09-16
chore(geofence): register a polygon's wake circle at 400 m, not 1 km

mrehan27:
Lowers the polygon wake-circle floor from 1000 m to 400 m.

**Why 1000 m cannot work.** The approach session is capped at two minutes and starts when the wake circle is crossed. At 1 km and the 26 km/h measured on the 2026-09-16 drive, that is 139 s of approach against a 120 s budget: the session expires before arrival and the arrival is never evaluated, even when GMS delivers the callback perfectly. 400 m leaves roughly four samples inside the budget at that speed.

**The old floor's justification does not apply to polygons.** It cited a measured 889 m GMS coarse-containment error. Both containment paths filter `!isPolygon`, so that error never reaches a polygon's registered radius. It only ever justified the floor for circles, which do not read this value.

**What still holds.** Activation admits a polygon only when `distance + accuracy <= radius`, and indoor fixes measure around 100 m. At 400 m that admits from 300 m out on an indoor fix, where the backend's 101 m retail circle could not admit from anywhere. This is why the floor is not simply removed.

Calibration step, not a derived value. A fixed radius against a fixed session cap cannot serve walking and driving at once. That is the real defect and is not addressed here.

**Verification**
- `toDomainRegions_givenValidPolygon_...` parses a real API response with `base_radius_m: 100` and asserts the registered radius is 400
- On device (Pixel 6, real workspace): fences 16/17/18/21 re-registered at `rad=400`; fence 20 stays at `rad=551` because its backend circle already exceeds the floor
- ktlint 0, Android Lint 0, 879 tests, 0 failures

Independent of #889, both off `feature/geofence-polygons`.


geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonWakeCircle.kt
```diff
@@ -19,21 +19,28 @@ internal data class PolygonTriggerCircle(
...
- * 1000 m is one observed worst-case containment error, rounded. It is the weakest part of this and
- * a drive that records ENTER reliability against a smaller registration would replace it. Adding a
- * kilometre on top of a circle already kilometres wide answers neither constraint, so a large
- * polygon is registered as the backend describes it.
+ * 400 m is a calibration step down from the 1000 m this carried until 2026-09-17, not a derived
+ * value. Two things moved it.
+ *
+ * The old floor's headline justification was a measured GMS coarse-containment error of 889 m on
+ * a 500 m fence. It does not apply to a polygon: both containment paths filter `!isPolygon`, so
+ * that error never reaches a polygon's registered radius. It only ever justified the floor for
```
Shahroz16: Could we keep this framed as a calibration hypothesis rather than ruling out the old radius? The `!isPolygon` filters skip SDK-side containment synthesis, but polygon wake circles still depend on GMS callbacks. A successful movement-trigger handoff also stops sampling and can open another session later, so one 120-second window doesn't establish that 1 km cannot work. The 400 m experiment is reasonable to test; these distinctions will matter when interpreting the next drive.

Shahroz16 (approved):
Approving the 400 m floor as a calibration step. The module builds, all 109 focused tests pass, and an extra boundary check preserves backend radii between 400 m and the old floor. The completed Claude Code Opus pass found no introduced runtime blocker. Could we rerun the failing Compose instrumentation job before merging? Its failure is in the unchanged SDK teardown/in-app network path, not the wake-radius code. The rationale clarification below is non-blocking.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonWakeCircle.kt
mrehan27: Agreed on both, fixed in a3c08cae.

The comment now frames 400 m as a hypothesis the next drive tests, not a demonstration that 1 km cannot work. I have also stopped dismissing the 889 m figure: it does not justify the floor through the containment path, but it is evidence about GMS's own position error, and a wake circle depends entirely on GMS callbacks, so it argues against shrinking as much as the old text argued for it. The comment says that tension is unresolved.

Also added that a movement-trigger handoff stops sampling and a later crossing can open a fresh session, so one 120 s window does not mean the arrival is lost.

No behaviour change, 400 m stays as the experiment.

PR #891, opened 2026-09-17
chore(geofence): remove the polygon evaluation policy nothing reaches

mrehan27:
Removes the CONFIRMED polygon evaluation policy. Net **−363 lines**.

**It has never run.** `PolygonLocationEngine` is the only caller of `PolygonRouteProcessor.process` and passes `DECISIVE_SINGLE_FIX`. The CONFIRMED branch was reachable only through a parameter default no caller uses.

**Nobody chose to keep it.** `evidenceFor`, `minimumEvidenceIntervalNanos` and `requiredConfirmations` all arrived in `41efef3b`, the original spike, and rode through each re-split because they were already in the tree.

**It was actively misleading, twice.** A QA guide writer was about to publish its five constants as live tester-visible behaviour, reasonably, since a constant in the evaluator looks like the shipping rule. And two of our own test harnesses had drifted onto the dead path through that same default.

**Removed:** `evidenceFor` and its perimeter-sampling helpers; `PERIMETER_SAMPLE_COUNT`, `ENTER_CONFIDENCE_THRESHOLD`, `EXIT_ACCURACY_INFLATION_METERS`, `MAX_EXIT_ACCURACY_METERS`, `MAX_EVALUATED_FIX_ACCURACY_METERS`; the `PolygonEvidencePolicy` enum and parameter; `minimumEvidenceIntervalNanos`; the three-confirmation state machine whose result only the dead branch consumed; `lastEvidenceElapsedNanos`. `PolygonTransitionStateMachine`'s confirmation count is now required, since the one remaining caller passes 2.

**Tests: 16 removed, 4 rewritten.** The removals all asserted behaviour that no longer exists. The rewrites are the part worth your attention: those tests assert guarantees that still hold, but expressed them as confirmation counts.

**Two of my rewrites were vacuous on the first attempt, caught by mutation:**

- The replay-guard test passed with the monotonic guard **deleted**. A decisive fix commits immediately and committed state then absorbs every replay, so the guard was invisible. Now uses marginal fixes, where a replay would falsely corroborate.
- The re-arm test passed with `clear()` **no longer discarding pending evidence**. Correct and broken both emit one arrival across two fixes, differing only in which fix produces it. Now asserts per fix.

Mutation results: removing the monotonic guard fails 3 tests, breaking `clear()` fails 1.

ktlint 0, Android Lint 0, 863 tests, 0 failures.


Shahroz16 (approved):
Approving the policy cleanup. I verified the removed branch is not reached by the current production caller and the focused `:geofence:testDebugUnitTest` suite passes. Please clean up the remaining KDoc that still describes the removed exit-inflation and clamp constants, and consider restoring a narrow boundary test around the 50 m behavior in a follow-up.

PR #892, opened 2026-09-17
chore(geofence): record why a polygon callback or arrival goes unreported

mrehan27:

## Why

A drive on 2026-09-17 lost a real arrival. The device was inside the ring with an 18 m fix, GMS delivered the wake, and no event was reported. The capture gave no reason because there was none to give.

The fence was evaluated, decided ENTER, and then withheld pending a second agreeing fix. That is the one branch in the pipeline that consumes a fix, decides something, and logs nothing. In a trace it is indistinguishable from a callback that never arrived, which is why three reproduction attempts went looking in the delivery path.

An iOS device recorded the same venue at `edge=+10 acc=5.1`, so the arrival was real, not inferred.

## What

No behaviour change. Every path that can discard a callback, skip an evaluation, drop an approach sample or end a session now says why.

| record | covers |
|---|---|
| `os.callback.dropped` | controller refusals, netting against their receipt |
| `polygon.evaluation.skipped` | engine refusals, deliberately off the callback key |
| `polygon.arrival.pending` / `.expired` | a held arrival, and one that died waiting |
| `polygon.sampling.skipped` | discarded approach samples |
| `polygon.approach.stop_refused` / `.request_discarded` | a declined stop, and a request torn down while arming |
| `polygon.approach.stopped` | now carries `n`, the session's sample count |

`polygon.evaluation.skipped` is kept off `os.callback.dropped` because the approach session reaches that path too, so those records have no receipt to net against and would inflate the dropped-callback count.

Seven record defects on the same seam, found after the review rounds and all from one ambiguity. GMS compares `PendingIntent`s with `Intent.filterEquals`, which ignores extras, so every session in a user-state generation shares one registration. Two questions then need two keys, and conflating them made "no entry" mean both "not reported yet" and "already reported":

- the sample count belongs to a session, so it keys on the deadline and is dropped once reported, which is what keeps the age prune safe
- whether an ending has been reported belongs to the registration, so it keys on the generation and is never evicted, because nothing cancels the `PendingIntent` and no age makes a repeat impossible

The 2026-09-18 capture carried 14 teardown records against two real sessions, two per sync pass. Suppression is bounded by process lifetime rather than by that set, so a capture spanning restarts still shows one such record per restart.

`n=0` on a stopped session is the one that matters: two captures show sessions arming and never delivering a sample, which is what starved the held arrival.

## Testing

909 unit tests green, parsed from the result XMLs. ktlint, `:geofence:lintDebug` and `:geofence:apiCheck` clean.

Teardown interleavings are enumerated rather than hand-picked, because each review round found the next case on this one seam. `PolygonApproachTeardownInvariantTest` runs every sequence of ten lifecycle operations up to length three, 1000 runs, each drained, asserting that no ending is reported without a count, no count is reported twice, and endings never outnumber starts. It found two defects review had not, including the removal retry re-issuing without the keys, which 1000 sequences reduce to 139 failures against one named test.

Per-site mutation: each log call deleted individually against the full suite. 16 of 23 sites are killed by a named test. The other 7 are mid-pass re-checks of a condition an outer gate already held, documented in `PolygonDropVisibilityTest` as expected-unreachable so the field can falsify them.

Reviewed across seven rounds by a second session. Its findings in this work: the expiry record firing for arrivals that never existed and again for ones that succeeded, two log calls holding a lock while invoking host-supplied code, a sample counter that could only ever hide a zero, a removal that restarted the request and reported an ending anyway, a late sample resurrecting a reported session, and the removal retry dropping the keys its success path needed.

## Reading the capture: absent `n` is not zero

`polygon.approach.stopped` carries `n`, the session's sample count, and `n=0` means the session armed and received nothing. That is the finding the field is looking for.

A teardown that cannot name its session omits `n` entirely rather than reporting zero, because guessing zero would manufacture that finding. **Anything summing or defaulting this field must treat absent as unknown, not as zero.**

## Locking: closed

Every polygon record is now emitted outside both `stateLock` and `controllerLock`. `PolygonRouteProcessor` returns its records instead of logging them, and the five engine teardown methods return the arrival holds they discarded, so the controller reports them after releasing its lock.

`PolygonLockFreedomTest` is what keeps that true. It drives the real paths through a `Logger` that refuses to be called while either lock is held, so it fails for any record on any path, including ones added later.

One path escaped it, because a sweep only covers what its fixture can reach. The coarse EXIT tore its session down by calling the public `deactivate` from inside `controllerLock`, and `synchronized` is reentrant, so the discarded holds were reported when only the inner block had closed. It needs a callback with no usable fix and a polygon still registered; the sweep unregisters first, so it never got there. The state half is now private and returns the holds. That left the public entry point with no production caller, so it is gone and its test drives the coarse EXIT instead.

Review suggested carrying out only the records this PR adds and leaving the rest as a follow-up. This carries all six out instead. With three of six still logging under `stateLock`, the test above needs an allowlist, and an invariant with a documented exception is not an enforced one. That expansion is roughly 400 lines, mostly the mechanical conversion of expression bodies to block bodies; the one existing test file it touched shrank by 52 lines. If the diff size makes review harder, splitting the expansion into its own PR is preferable to dropping it.

## Not in scope

Corroboration itself is unchanged. Whether a 24 m fence should need two fixes is a calibration decision, and it needs the field data this makes readable. iOS has since changed theirs to block rather than gate, so the platforms have diverged on that rule.


geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonApproachMonitor.kt
```diff
@@ -94,6 +95,10 @@ internal class PolygonApproachMonitor(
             registrationRetryAttempt = 0
             registrationRetryJob?.cancel()
             registrationRetryJob = null
+            // Reset when the session arms, not when it ends: an ending that bypasses the normal
+            // stop would otherwise carry its samples into the next session's count, and this
+            // instrument exists to make "ended with n=0" decisive. It can only ever hide a zero.
+            samplesThisSession.set(0)
```
Shahroz16: Could we keep this count with the session or `PendingIntent` it belongs to? When generation 7 has delivered samples and generation 8 starts, this resets the shared counter before `removeUpdates(previous)` completes. The old session can then be logged as `n=0`, or with samples from the new session, which makes the exact diagnostic this field is meant to provide unreliable. A per-intent count, or a captured count passed through removal and its retries, would keep the attribution stable.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonRouteProcessor.kt
```diff
@@ -137,8 +148,24 @@ internal class PolygonRouteProcessor(
...
+            return
+        }
         val observedAt = arrivalConfirmationNanos[polygonId] ?: return
-        if (elapsedRealtimeNanos - observedAt > MAX_CORROBORATION_GAP_NANOS) {
+        val heldForNanos = elapsedRealtimeNanos - observedAt
+        if (heldForNanos > MAX_CORROBORATION_GAP_NANOS) {
+            // The counterpart to the pending record. Without it a capture cannot separate a hold
+            // that completed from one that died, which is the whole question the field has to
+            // answer about corroboration.
+            logger.logPolygonArrivalExpired(
```
Shahroz16: Could we also record what happened when a pending arrival is cleared because the sampling session ends? This expiry only runs when another fix arrives. In the field case where sampling goes quiet and the engine calls `clear()` or `clear(id)`, the pending hold is still removed without a counterpart record, so the capture cannot tell whether it completed or disappeared. A regression test that ends the session without another fix would cover that path.

geofence/src/test/java/io/customer/geofence/polygon/PolygonCorroborationOnRealFencesTest.kt
```diff
@@ -0,0 +1,186 @@
...
+
+/**
+ * What corroboration costs on the workspace's own fences, at the accuracies the 2026-09-17 drive
+ * actually produced.
+ *
+ * Characterization, not verification: every assertion here is an evaluator or route-processor
+ * return value, so all of it passes with or without the instrumentation shipped alongside it. It is
+ * here to give a calibration decision real numbers from real rings, and it must be read that way.
+ *
+ * Vertices and fix coordinates are copied from the device's own region store and capture. Fence 17
```
Shahroz16: Could we translate these shapes and fixes to synthetic coordinates and use neutral venue names before this lands in the public SDK repository? The file currently combines exact business locations with a device's precise position and wall-clock time from a real drive. Translating every point by the same offset preserves the geometry, clearance, and all of these assertions without publishing that location history.

geofence/src/main/kotlin/io/customer/geofence/GeofenceLogger.kt
```diff
@@ -1128,6 +1178,111 @@ internal class GeofenceLogger(private val logger: Logger) {
...
+        horizontalAccuracyMeters: Double?,
+        fixAgeSeconds: Double?
+    ) {
+        logger.debug(
+            "Polygon '$geofenceId' reads as an arrival but its accuracy circle reaches the boundary. Holding for a second agreeing fix." +
+                tail(
+                    "polygon.arrival.pending",
+                    GeofenceLogIo.OUTPUT,
+                    listOf(
+                        "id" to token(geofenceId),
```
Shahroz16: Could we keep the identifier encoding consistent with the decision and expiry records? `token()` lowercases and rewrites punctuation, while the other polygon records pass the id directly through the tail sanitizer. A workspace id such as `Store-17` therefore cannot be joined from `arrival.pending` to `polygon.decided`, which defeats the trace this pair is meant to provide.

geofence/src/main/kotlin/io/customer/geofence/GeofenceLogger.kt
```diff
@@ -1128,6 +1178,111 @@ internal class GeofenceLogger(private val logger: Logger) {
...
+    /**
+     * One record per discarded approach sample. Sampling is what supplies a corroborating fix, so
+     * a silent gap here is indistinguishable from the session never having run.
+     */
+    fun logPolygonSamplingSkipped(reason: PolygonSamplingSkip) {
+        logger.debug(
+            "Polygon approach sample skipped — ${reason.detail}." +
+                tail(
+                    "polygon.sampling.skipped",
+                    GeofenceLogIo.INPUT,
```
Shahroz16: Could we separate the delivered-input facts from the SDK's skip decision here, or classify this record as an output? `EMPTY_BATCH` and `NO_USABLE_FIX` describe the input, while `NOT_CURRENT_SESSION` and `NO_POLYGON_IN_RANGE` are SDK decisions. Putting all four under `io=in` means replay treats a decision produced by the SDK as something it should feed back.

Shahroz16 (changes requested):
The latest commit fixes the skip reason and lock-held logging issues. The focused geofence tests pass locally, but I am holding approval on the session attribution, silent pending-arrival teardown, replay-field consistency, and public location fixture concerns in the inline comments.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonApproachMonitor.kt
mrehan27: Fixed, and the shared counter turned out to be the smaller half of it. Keying per PendingIntent was not enough either: setData carries the generation but the deadline is an extra, and filterEquals ignores extras under a fixed request code, so two sessions in one generation build an equal PendingIntent. Since the sampler arms on every callback while a generation changes only on identify, same-generation overlap is the common case.

Now keyed on the session deadline, which the receiver already reads from the delivering intent, so the batch names its own session and nothing infers the owner from what is active. Where a teardown genuinely cannot name a session the count is omitted rather than reported as zero, since zero is the finding this field exists to make.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonRouteProcessor.kt
mrehan27: Fixed. The expiry record now carries a reason, window_elapsed or session_ended, and clear() returns what it discarded so the engine can report it after its lock closes rather than logging under one.

Your point that this is the field case is what made it worth doing properly: the window path only fires when another fix arrives, which is exactly what does not happen when sampling goes quiet. Covered by a test that ends the session with no further fix, and it fails if the record is removed.

geofence/src/test/java/io/customer/geofence/polygon/PolygonCorroborationOnRealFencesTest.kt
mrehan27: Fixed, and thank you for catching it. Every point is translated by a constant longitude offset and the venues are renamed. The latitude band is unchanged, so the metric scale factor is identical and every clearance and assertion holds unmodified.

geofence/src/main/kotlin/io/customer/geofence/GeofenceLogger.kt
mrehan27: Fixed. Both records now pass the id straight through as every other record does. You were right that token() defeated the pair: the whole value of arrival.pending is joining it to the decision that follows.

geofence/src/main/kotlin/io/customer/geofence/GeofenceLogger.kt
mrehan27: Fixed, reclassified as output. Agreed on the reasoning: all four are the SDK's own decision about a delivered sample, not a fact about the input, so replay should not feed them back.

geofence/src/test/java/io/customer/geofence/polygon/PolygonCorroborationOnRealFencesTest.kt
Shahroz16: Thanks for neutralizing the venue names. The coordinates are still reversible: this file keeps the exact `31.37` latitude band and adds 60 to longitude, while `PolygonCoarseEnterRealStoreTest` still contains the untranslated `74.18` capture. Subtracting 60 recovers the real rings and device fix. Could we also shift latitude, rescaling longitude if needed, or replace these with synthetic rings that preserve the clearances?

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonRouteProcessor.kt
Shahroz16: Thanks, the explicit stop and deactivate paths are covered now. One common hold-clear path is still unpaired: when the next fix is ambiguous or agrees with the committed state, `confirmArrival()` clears the pending state at lines 83–87, but only the undecided or unchanged record is emitted. The `arrival.pending` never gets an `arrival.expired`, so machine accounting cannot tell that hold was broken. Could we emit a narrow reason such as `evidence_broken` when an existing hold is cleared, with a regression test for a marginal ENTER followed by a coarse fix?

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonApproachMonitor.kt
Shahroz16: Thanks, the deadline key and retry propagation address the earlier cross-session counter. There is still an attribution race in `start()`: `previousDeadline` is stored as instance state under `lock`, then read after the lock is released. A concurrent `start()` can overwrite it before the first call invokes `removeUpdates`, so the old request consumes the wrong session counter and the real one can later report `n=0`. Could we return the outgoing deadline alongside the pending intents from the synchronized block and pin that interleaving in a test?

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonGeofenceServiceController.kt
```diff
@@ -238,19 +268,32 @@ internal class PolygonGeofenceServiceController(
...
+            val fix = location.toPolygonLocationFix()
+            if (fix == null) {
+                logger.logPolygonSamplingSkipped(PolygonSamplingSkip.NO_USABLE_FIX)
+                continue
+            }
             val shouldProcess = synchronized(controllerLock) {
                 if (!hasMatchingIdentifiedUserLocked() ||
                     store.userStateGeneration() != expectedUserStateGeneration
                 ) {
+                    logger.logPolygonSamplingSkipped(PolygonSamplingSkip.NOT_CURRENT_SESSION)
```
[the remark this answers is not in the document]
Shahroz16: Thanks for moving the direct skip logs. The new arrival records still invoke the host logger under locks: `engine.deactivate()`, `stop()`, and `resetEvidence()` are called while `controllerLock` is held, and `routeProcessor.process()` logs pending, expired, and decided records while `stateLock` is held. Could we carry those records out with the state result and emit them only after both locks are released? That preserves the customer-dispatcher boundary this PR is trying to enforce.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonApproachMonitor.kt
```diff
@@ -269,20 +346,34 @@ internal class PolygonApproachMonitor(
                             desired && activePendingIntent == pendingIntent
                         }
                     }
-                    logger.logPolygonApproachMonitoringStopped()
+                    // Absent rather than zero when the session cannot be named: an unattributable
+                    // teardown reporting n=0 would read as "armed and delivered nothing", which is
+                    // the exact finding this field exists to make, so it must never be guessed.
+                    logger.logPolygonApproachMonitoringStopped(
+                        samplesReceived = sessionDeadlineElapsedRealtimeMs
+                            ?.let { samplesByDeadline.remove(it)?.get() ?: 0 }
```
Shahroz16: Could we avoid treating a missing counter entry as zero here? `processApproachLocations()` returns `STALE` before `recordSampleDelivered()`, so an expired stale batch reaches `stop(deadline)` and reports `n=0` even though GMS delivered it. The bounded eviction path can produce the same false zero. Initializing the counter when this process arms a session, and omitting `n` when the entry is absent, would keep zero as observed evidence rather than a guess.

Shahroz16 (changes requested):
Thanks, the latest commits resolve the five original findings and the retry-attribution gap. The focused geofence and monitor tests pass locally. I am holding approval on the remaining lock-boundary, counter-attribution, hold-accounting, and reversible-fixture issues in the existing threads.

geofence/src/test/java/io/customer/geofence/polygon/PolygonCorroborationOnRealFencesTest.kt
mrehan27: You are right, and the second file was the bigger hole. Both are now moved to a different latitude band with longitude deltas rescaled by the ratio of the two cosines, so there is no constant offset to subtract. I compared every edge length before and after and they are identical, so the clearances and all assertions are unchanged.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonRouteProcessor.kt
mrehan27: Good catch, and that is the commonest of the three. Added evidence_broken, emitted when a fix that cannot be judged or disagrees clears an existing hold, with a regression test for a marginal ENTER followed by a coarse fix. Every arrival.pending now has a counterpart.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonApproachMonitor.kt
mrehan27: Fixed. The outgoing deadline is returned from the synchronized block instead of being stashed on the instance, so a concurrent start cannot repoint it. I have not added a test that pins the interleaving; it needs two threads racing inside start and I did not want to add a timing-sensitive test at this hour. Happy to add one if you would rather it were pinned than argued.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonApproachMonitor.kt
mrehan27: Fixed, and your framing is the right one. A session that arms now seeds its counter at zero, so n=0 is observed evidence and an absent n means the teardown could not be attributed. The eviction path could also produce a false zero by deleting a live entry while its removal was in flight; counts are now pruned only once a deadline is far enough past that no session can still report against it.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonGeofenceServiceController.kt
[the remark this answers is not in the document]
mrehan27: Not addressed, deliberately, and it is the one open item on this PR.

You are right about all of it, including routeProcessor.process logging under stateLock. It is also pre-existing rather than new here: at e41b90de, retryRegistration logs as its first statement and is reachable synchronously from start(), which the controller calls under controllerLock.

Carrying the records out through both locks is the correct fix and it touches around 17 call sites. I did not want to make that change unreviewed at 2am on a branch where four review rounds have each found a real defect, two of them mine from the last hour. The failure mode is latency inside the goAsync budget rather than a wrong record or a lost arrival, so it cannot corrupt what this PR measures.

It is written up in the PR description with the three options and the tradeoffs. If you would rather it were fixed here than tracked, say so and I will do it as the first thing rather than ship around it.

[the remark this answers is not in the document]
Shahroz16: Thanks for spelling out the tradeoff. I agree the broader locking pattern predates this PR, but these new arrival records still call the host dispatcher while `stateLock` or `controllerLock` is held, even though this diff documents that boundary as unsafe. A slow customer dispatcher can still stall geofence processing here. Could we carry the new records outside both locks before this lands, and keep the broader cleanup as a follow-up?

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonApproachMonitor.kt
Shahroz16: The captured outgoing deadline fixes the concurrent-start race, thanks. There is still one reachable rollover where its count is never consumed: `start_givenTheSameGenerationAfterItsDeadlinePassed_expectOneRegistrationAndNoFalseStop` creates a new deadline while the old timeout job is delayed, but the equal `PendingIntent` skips `removeUpdates`, so the old counter is neither reported nor removed. Could we close the old accounting record without removing the equal live request, and pin that assertion in this test?

Shahroz16 (changes requested):
The latest commit resolves the cross-thread counter attribution, broken-hold accounting, missing-count default, and fixture concerns. The focused tests and release lint pass locally, and hosted CI is green. I am holding approval on the two remaining diagnostic-integrity issues in the existing lock-boundary and same-generation rollover threads.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonGeofenceServiceController.kt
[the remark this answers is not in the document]
mrehan27: Done. All six route-processor records now come out through the result and are emitted after the lock, and the five engine teardown methods return the arrival holds they discarded so the controller reports them once `controllerLock` is released.

`PolygonLockFreedomTest` drives the real paths through a `Logger` that refuses to be called while either lock is held. It found `polygon.decided` still logging under `stateLock` straight away, which is why this covers all six rather than the two arrival records.

To be explicit about that choice: you offered to keep the broader cleanup as a follow-up and I have not taken the offer. Carrying out two of six leaves the invariant false, so the test above would need an allowlist, and an invariant with a documented exception is not an enforced one. The extra work is roughly 400 lines, mostly expression bodies becoming block bodies, and the one existing test file it touched got 52 lines shorter.

If the size makes this harder to review, I would rather split the expansion into its own PR than drop it.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonApproachMonitor.kt
mrehan27: Fixed, and the assertion is pinned in that test.

The outgoing deadline's count is now consumed after the lock when the re-arm builds an equal `PendingIntent`, so the old accounting record closes without touching the live request. `start_givenTheSameGenerationAfterItsDeadlinePassed_expectOneRegistrationAndNoFalseStop` keeps its zero-removal assertion and now also expects `logPolygonApproachMonitoringStopped(0)` exactly once, pinned to zero rather than `any()` because the count is the whole point of the record.

Your rollover led me to a second one on the same seam, which the 2026-09-18 drive had already been logging all morning. The removal succeeds whether or not a request is still attached, and nothing cancels the `PendingIntent`, so once any session had built one for the live generation, every later stop found it and logged a teardown that never happened. Fourteen of those against two real sessions, two per sync pass, in the record we use to tell whether the session ever sampled. Gated now on the last removed request, which arming clears, with a test for the repeat case and a control that a genuine re-arm still reports both teardowns.

Worth noting for a follow-up: the structural fix is to cancel the `PendingIntent` on the removal success path, which would make `existingPendingIntentOrNull` return null and remove the bug class rather than the symptom. I did not want to change intent lifecycle in this PR.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonGeofenceServiceController.kt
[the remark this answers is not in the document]
Shahroz16: Thanks, the broad extraction fixes the direct paths. There is still a nested path in `onCoarseExit`: the outer block holds `controllerLock`, calls the public `deactivate()`, and `deactivate()` reports the returned holds after only its inner reentrant block closes. A coarse EXIT with no usable triggering location leaves the arrival hold intact, so `session_ended` reaches the host dispatcher under the outer lock. Could we return the discarded ids from this branch and report them after the outer block, then add that null-location hold case to the lock-freedom test? The current sweep unregisters the polygon before `onCoarseExit`, so it never reaches this branch.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonApproachMonitor.kt
Shahroz16: Thanks, the rollover count is pinned now. The one-slot removal memo still has an ABA across a re-arm: stop session D1, arm same-generation D2 before D1’s removal listener runs, then D1’s listener writes the shared `PendingIntent` back into `lastRemovedPendingIntent`. D2 uses that equal intent, so its real stop is suppressed and its deadline count remains stranded. Clearing the memo on arm cannot protect against the older listener writing afterward. Could we track teardown consumption by session deadline, or otherwise make this ordering atomic, and add the delayed-listener sequence to the test?

Shahroz16 (changes requested):
The latest commits fix the original session-count rollover and most of the lock-boundary paths. Two reachable cases still make the new diagnostics unreliable: the re-arm ABA around the one-slot removal memo, and the nested coarse-exit arrival record under `controllerLock`. I left both in the existing discussions and am holding approval until they are addressed. The focused geofence tests pass locally and hosted CI is green.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonGeofenceServiceController.kt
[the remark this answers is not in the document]
mrehan27: Fixed, and you were right that the sweep was the reason I missed it.

The state half of `deactivate` is now private, returns the holds, and `onCoarseExit` reports them after its own block closes. That left the public entry point with no production caller, so I removed it rather than keep a path production does not take; its test now drives the coarse EXIT that replaced it.

`coarseExitWithNoFixWhileAnArrivalIsHeld_expectTheDiscardIsEmittedOutsideBothLocks` pins it. Reporting the holds inside the lock again kills that test and nothing else, including the sweep, which is the point you were making.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonApproachMonitor.kt
mrehan27: Fixed by changing the identity rather than the ordering.

The memo cannot name a session, so no write ordering saves it. Named teardowns now consume the session's own accounting entry, which already exists from arm until report, so removing it is both the permission to report and the count to report, atomically and for that session alone. Your ABA cannot arise: D2 has its own entry, so D1's late listener cannot speak for it.

The memo stays for the one case with no entry to consult, a request that outlived the process which armed it, and it is now read only there.

`stop_givenTheEarlierRemovalLandsAfterAReArm_expectTheNewSessionStillReportsItsOwnCount` runs your sequence with a delayed listener. D2 delivers a sample and D1 does not, so the two records are distinguishable. Against the previous code it fails and `stop_givenTheSessionReArmedBetweenTeardowns_expectBothReported` still passes, so it is specific to this defect.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonApproachMonitor.kt
```diff
@@ -253,36 +367,79 @@ internal class PolygonApproachMonitor(
...
+                    //
+                    // The accounting entry decides it. One exists from the moment a session arms
+                    // until its teardown is reported, so removing it is both the permission to
+                    // report and the count to report, atomically and for that session alone. See
+                    // [lastRemoval] for why the memo cannot do this job.
+                    //
+                    // n=0 is observed evidence that nothing was delivered, because a session that
+                    // armed in this process always has an entry seeded at zero. Absent means the
+                    // teardown could not name a session at all.
+                    val samples = sessionDeadlineElapsedRealtimeMs
```
Shahroz16: Could we preserve the stopped record when a named teardown has no local accounting entry? This is reachable after a cold-process callback: an expired batch with no identified user calls stop with the recorded deadline, finds and removes the prior process's PendingIntent, but this process never seeded samplesByDeadline. The removal succeeds, samples is null, and unnamedTeardown is false because the deadline is present, so nothing is logged. Treating a missing entry as an unattributed teardown would keep this real session ending visible without inventing n=0.

Shahroz16 (changes requested):
The earlier counter and lock issues are addressed. One cold-process teardown path still becomes silent, so I am holding approval on that narrow diagnostic gap.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonApproachMonitor.kt
mrehan27: Good catch, and it removed the special case rather than adding one.

You are right that it is reachable and that nothing logged: the entry belongs to the process that armed the session, and my gate only reported an uncountable ending when the deadline was absent entirely.

The two questions are now asked separately. Whether a teardown happened is answered by whether this process has removed that registration before, which is what the memo is for and is sound because equal intents are one registration. Whether it can be counted is answered by the accounting entry, independently. A first removal with no entry is a real ending reported with no count, so both cross-process shapes are covered: reached by generation alone, or by a deadline this process never seeded. The deadline-null branch is gone.

`stop_givenAColdProcessNamesADeadlineItNeverArmed_expectTheEndingIsStillReported` pins your sequence and asserts absent rather than zero. Restoring the gate you reviewed kills it, and reporting every first removal even when an entry exists kills `stop_givenRepeatedStopsAfterTheSessionEnded_expectOneRecordNotOnePerPass`, so the two directions are both held.

Pushed as `109a6e39`. 903 tests, ktlint and lintDebug clean.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonApproachMonitor.kt
```diff
@@ -253,36 +366,82 @@ internal class PolygonApproachMonitor(
...
+                    // never seeded. It is reported with no count rather than dropped.
+                    //
+                    // n=0 is therefore observed evidence that nothing was delivered, and absent
+                    // means the ending could not be attributed. Neither is a default.
+                    val samples = sessionDeadlineElapsedRealtimeMs
+                        ?.let { samplesByDeadline.remove(it) }
+                    when {
+                        samples != null ->
+                            logger.logPolygonApproachMonitoringStopped(samplesReceived = samples.get())
+                        firstRemovalInThisProcess ->
```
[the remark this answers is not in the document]
Shahroz16: I traced this against both teardown paths and do not think this duplicate is reachable. A stop clears the active request before its asynchronous removal starts, so a re-arm after that point has no prior active request to consume. In the opposite ordering, the same-generation re-arm cancels the prior timeout before it can start a later teardown. I am not holding approval on this.

Shahroz16 (approved):
The cold-process teardown now stays visible without inventing an n=0 sample count, and the per-deadline accounting still prevents repeated stops from creating duplicate records. I also traced the new automated duplicate-report concern and could not reach that ordering through the production state transitions. The focused PolygonApproachMonitor test class passes locally at 109a6e39.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonApproachMonitor.kt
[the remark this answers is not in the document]
mrehan27: Thanks for the approval, and your reading of the ordering is right. I am not asking you to revisit that, but the defect turns out not to need it, so I have pushed a fix rather than merge on the approval.

You could not reach that ordering through the production state transitions in one process. Agreed. The field shape is two steps across a process boundary and needs no ordering at all:

1. A session arms and the request fails with `SecurityException`. The `PendingIntent` was minted before the failure and the OS keeps it.
2. That failure is a location-permission change, and a runtime permission revocation kills the app process. Measured on API 37 rather than assumed: `ActivityManager: Killing 7426:...java_layout (adj 0): permissions revoked`.
3. The next process reaches `stop()` naming that generation, or `removeStaleGeneration` from the approach receiver on a stale batch. `existingPendingIntentOrNull` finds the minted intent, GMS answers success for a request that was never attached, and in a fresh process the memo is empty and there is no accounting entry, so it reports an ending for a session that never registered.

That is the same fixture shape as `stop_givenColdProcessGeneration_expectRemovesWhatTheEarlierProcessRegistered`, which you approved: two monitor instances, one registration outliving the first. There the first registration succeeded, so its record is correct. Replace that success with a `SecurityException` and the same two steps produce a false one.

The fix is that a registration which never took effect claims the removal memo and drops its accounting entry, so the ending is neither counted nor reported. The intent is deliberately not cancelled, because on a same-generation failure the equal intent belongs to a request that may still be live.

Two more, both found after your approval and both wrong records rather than wrong state:

A removal landing while the request is still wanted re-issues it, so the session continues, yet an ending was reported and the live session's count consumed. The guard is now on the session rather than the registration: a removal naming the live deadline reports nothing and restarts, one naming an older deadline reports that session and still restarts the live request.

A late batch could resurrect a session already reported, because `recordSampleDelivered` created entries. It now only increments existing ones, so a slightly-late batch still counts, a reported session cannot revive, and a batch for a session this process never armed stays uncounted rather than reported as a partial.

On validation, since this seam has now produced several rounds of exactly this: `PolygonApproachTeardownInvariantTest` runs every sequence of six lifecycle operations up to length three, 216 runs, drains each and asserts invariants rather than expected outputs. No ending without a count for a single instance, no count reported twice, endings never outnumbering starts. It found the `SecurityException` case, and it also caught two mistakes in my own fixes, including a first attempt that compared a parameter with the field it shadowed and silently suppressed every teardown record. The parameter is renamed.

`96f9f1ee`. 907 tests, ktlint and lintDebug clean, each fix killed by exactly one named test under mutation. Independently reviewed by the other Android session before this push.

mrehan27:
Two commits since your approval at 109a6e39, so this needs another look rather than merging on it.

The seam had one ambiguity behind all of it: equal `PendingIntent`s share one GMS registration, so "no accounting entry" meant both "not reported yet" and "already reported". The count now keys on the session deadline and is dropped once reported; whether an ending was reported keys on the registration generation and is never evicted. Two more defects were found after that: a removal that restarted the request reported an ending anyway, and the removal retry re-issued without either key, so a removal GMS rejected once and accepted on retry reported the ending twice.

Rather than another named test per round, the interleavings are now enumerated: every sequence of ten lifecycle operations up to length three, 1000 runs, asserted against invariants. It found the retry defect and reduces it to 139 failing sequences.

909 tests green from the XMLs, ktlint, lintDebug and apiCheck clean.

PR #893, opened 2026-09-18
chore(geofence): decide a departure before reading the fix's accuracy

mrehan27:
Part of [MBL-2482](https://linear.app/customerio/issue/MBL-2482/mobile-run-cross-platform-polygon-geofence-qa)

The accuracy ceiling was the first statement in `decisiveEvidenceFor`, before anything branched on committed state, so it threw away departures as well as arrivals.

## The defect

Across four field captures, 8 fixes clear of their ring by 201 m to 988 m were refused as `accuracy_too_low`. A fix 988 m outside a ring is not ambiguous at 122 m accuracy.

Seven of those were at fences already committed OUTSIDE, so they were information discarded rather than events lost. One (fence 26, `edge=-321.7 acc=52.4`) was committed INSIDE, and that visit closed 5 minutes and 2 seconds later on the next callback. So the measured cost is one delayed departure, not a lost one, and the case for fixing it is that refusing a fix 321 m outside its ring at 52 m accuracy is wrong under any rule, not that it cost an event.

## The change

Clearance is read first. Outside the ring by more than the fix's own accuracy plus the departure margin settles which side the device is on, whatever the accuracy. The ceiling then governs arrivals alone, at the same value, so **which arrivals get reported does not move in this PR**.

A decisive outside fix that agrees with an OUTSIDE commitment is now `polygon.unchanged` rather than `polygon.undecided why=accuracy_too_low`. That matters for calibration: decisive observations were landing in the refusal bucket and inflating it.

## Validation

Replayed against all four captures: 10 verdicts change, all on the departure side, no arrival affected. Fence 26 exits 5 minutes earlier than it did.

Four mutations run, all killed: restoring the old order, dropping the margin, claiming a transition when already outside, and removing the arrival ceiling.

iOS reads the same two questions in the same order, confirmed from their source rather than assumed: their outside-decision returns above their ceiling guard, and they have never seen an `accuracy_too_low` in either direction across 1640 verdicts.

## Stack

```
feature/geofence-polygons
  mbl-2480-fix-dropped-polygon-callback   #892
    mbl-2482-polygon-departure-clearance  <- this PR
```


Shahroz16 (approved):
The reordered decision is supported by the recorded drives and matches the existing iOS ordering. I challenged the unbounded-accuracy case separately, but the rule here explicitly requires clearance beyond the fix’s own uncertainty plus the margin, and the replay changes only departure-side verdicts. The focused evaluator and the current #892 polygon integration and lock suites pass together on a clean local stack.

PR #894, opened 2026-09-18
chore(geofence): wake on the backend's own circle instead of a padded one

mrehan27:
Part of [MBL-2482](https://linear.app/customerio/issue/MBL-2482/mobile-run-cross-platform-polygon-geofence-qa)

The wake circle is now registered exactly as the backend sends it. The 400 m floor is gone.

## Why

On the 2026-09-18 Rainbow trip the padded circle woke us 193 m outside a shop whose own circle is 101 m. No fix was ever taken inside the polygon during a 44 minute visit and nothing was reported. iOS registers that 101 m verbatim, woke 1 m inside the ring on the same trip, and recorded the visit.

The circle already encloses the ring, so padding it only moves the wake further from the venue.

Both arguments for the floor fail:

- The measured GMS coarse-containment error of 889 m on a 500 m fence reached the floor through SDK-side containment, and both containment paths filter `!isPolygon`. It never applied to a polygon by that route.
- The approach admission gate needed the whole accuracy circle inside the radius. That gate feeds interval sampling, which is not serviced for a background app: measured the same day, 3 batches in 120 s in the foreground against 1 batch in 44 minutes in the field, same build. The padding was protecting a mechanism that cannot run in the regime it was padded for.

## The change

`PolygonWakeCircleValidator` keeps the upper bound and drops the floor. `radius` and `baseRadiusMeters` now agree; both are still carried because they answer different questions.

The admission gate goes from `distance + accuracy <= radius` to point containment. Against a circle only tens of metres wider than the ring, no indoor fix could satisfy the old rule from anywhere, so the polygon would never be evaluated. Activation opens a sampling session, it does not report an arrival, so admitting a coarse fix costs sampling while refusing one loses a visit.

## Measured on the emulator, including the cost

Validated with the SDK build on a Play-services AVD, driving an interpolated approach into the real Rainbow ring (registered radius 101 m):

| fix accuracy | ENTER callback | n |
|---|---|---|
| 10, 50, 90, 95, 105 m | fires | 1-2 each |
| 110, 115, 120 m | does not fire | 2 each |

A 551 m fence still fires at 120 m accuracy, so the threshold tracks the radius rather than being an absolute accuracy ceiling. **GMS stops triggering a circle once the fix is about as coarse as the radius.**

That is the cost of this change, and it is quantifiable: 11 of 54 callback fixes in the captures are coarser than 100 m, so roughly one in five of the wake opportunities we have observed would not wake a ~100 m circle.

Against that, a coarse wake on a 400 m circle bought very little. It reports the device outside the ring, which is correct and uninformative, and the sampling session it opens does not deliver in the background. By area, a 400 m circle around the Rainbow ring is inside the polygon about 6% of the time. So the trade is a fifth of the coarse wakes, which mostly carried no containment information, against waking at the ring on every decent fix.

Registration verified in the SDK's own store: every polygon now registers its own radius (101, 100, 104, 109, 163, 276, 371, 551), where seven of those were previously floored to 400.

Still not field-checked below 250 m, and no floor is invented to cover that; the next capture reads which radius registered and whether it fired.

## Testing

907 unit tests green, parsed from the result XMLs. ktlint and `:geofence:lintDebug` clean. Reinstating the floor as a mutation kills four tests; restoring the accuracy term in the admission gate kills one.

## Stack

```
feature/geofence-polygons
  mbl-2480-fix-dropped-polygon-callback     #892
    mbl-2482-polygon-departure-clearance    #893
      mbl-2482-polygon-covering-circle      <- this PR
```


Shahroz16 (approved):
The backend radius now flows through without the old padding, and the admission check is consistent with the smaller wake circle. I also exercised the resulting stack on an API 35 emulator against a real 326 m polygon. This calibration increment looks good to me.

PR #895, opened 2026-09-18
chore(geofence): ask for a precise fix when the delivered one decided nothing

mrehan27:
Part of [MBL-2481](https://linear.app/customerio/issue/MBL-2481/mobile-measure-indoor-polygon-arrival-behavior)

A polygon verdict that comes back undecided now asks for one precise fix and re-evaluates against it.

## Why

This is the only fix in the pipeline whose accuracy class the SDK picks. Every other one arrives attached to a GMS callback, and on a parked device those measure 100 m and upward, above the arrival ceiling of every real fence we monitor.

Nothing better follows them. Across four field captures the approach sampling session never contributed a distinct fix to any evaluation, so a held arrival waiting for the next delivered fix was not waiting slowly, it was waiting for something that does not come.

## Undecided, not stale

iOS gates the same request on a cached fix older than 30 s, because their region callbacks leave them to resolve a fix themselves. Ours arrive with one. Measured across 54 callbacks in seven captures, its age has a median of 0.6 s and a maximum of 9.7 s, so that gate would never fire here. What goes wrong for us is accuracy, not freshness. The divergence is deliberate and the numbers are in the commit.

## Bounds, each measured rather than borrowed

- **3 s wait.** This spends the receiver's 8 s dispatch budget ahead of the movement-refresh join. Callback to `movement.registered` measures 0.1 s to 6.57 s over 22 observations, median 0.31 s, so at 3 s the join keeps its full window for 21 of the 22.
- **One request per 30 s**, and none at all when the delivered fix decided, so an ordinary drive-by crossing pays for no sensor time.
- **2 s batch reuse**, not 30 s. Within it, a fix already obtained answers the rest of the same GMS batch, so several undecided fences in one delivery cost one request. Consecutive callbacks inside a batch are 0.07 s to 0.36 s apart. A reused fix bypasses the rate limit and the engine's own fix-age ceiling is 120 s, so a half-minute-old one would pass every gate and decide an arrival from a position the device may have left: 180 m at the 26 km/h measured on 2026-09-18, against a wake circle of about 100 m.
- **The requested fix and its rate limit are dropped wherever the coarse-transition memo is.** Sign-out leads to a sync and a registration with `INITIAL_TRIGGER_ENTER`, so the next user's first callback lands within seconds and would otherwise be refused its own fix because the previous session asked for one.

A callback that arrives with no location asks too. That path is fail-open surface rather than an observed case: all 49 callbacks in the captures carried a fix.

`polygon.freshfix.requested` is paired with exactly one of `received` or `skipped`, so a capture can tell a request that went unanswered from one that was never made.

## Validated on device, on the real fence geometry

Parked inside Tim Hortons (inradius 23.6 m) with a 90 m fix, which is inside the ring and above the arrival ceiling, then made a precise fix available while the request was in flight:

```
os.callback.received  ids=17 t=enter acc=90.0
polygon.undecided     id=17 why=accuracy_too_low edge=23.9 acc=90.0
polygon.undecided     id=27 why=accuracy_too_low edge=-69.3 acc=90.0
polygon.freshfix.requested  ids=17,27 n=2
polygon.freshfix.received   waited=0.1 fixsrc=fresh_request acc=8.0
polygon.decided       id=17 t=enter edge=23.9 acc=8.0
polygon.decided       id=27 t=exit  edge=-69.3 acc=8.0
```

That is the Passport Office shape: a real arrival 23.9 m inside a ring, refused because the delivered fix was 90 m, reported because we asked for a better one. Two things fall out of the same run: one request served both undecided fences, and the coarse samples that followed produced no further requests, so the rate limit holds.

Still not validated in the field. The accuracy of a requested fix on real hardware is the number we do not have, and the weekend build is what produces it. On the emulator the answer comes from the injected mock, so `waited=0.1` and `acc=8.0` are properties of the instrument, not of GMS.

The movement-trigger EXIT path evaluates a fix the same way and does not ask. Deliberately out of scope here.

## Testing

926 unit tests green, parsed from the result XMLs. ktlint, `:geofence:lintDebug` and `:geofence:apiCheck` clean. Twelve new tests. Mutations: never asking kills seven of them, dropping the within-batch reuse kills exactly the one that covers it, and widening the reuse window back to 30 s kills the expiry test.

The rate limit is global by key and user-scoped by lifetime, so every entry point that ends a user's scope has to lift it. Review found that only one of the six was actually covered: removing the reset from `completeUserReset` and from the session-open path both survived the full suite. There is now one test that enumerates all six entry points and reports every one that carried the cooldown across, and it names both of those sites when they regress.

## Stack

```
feature/geofence-polygons
  mbl-2480-fix-dropped-polygon-callback       #892
    mbl-2482-polygon-departure-clearance      #893
      mbl-2482-polygon-covering-circle        #894
        mbl-2481-polygon-callback-fix         <- this PR
```


Shahroz16 (approved):
The callback asks for one bounded high-accuracy fix only after the delivered fix is genuinely undecided, with cooldown and teardown coverage. I exercised the coarse-to-precise recovery on the top stack and saw it obtain the fresh fix and reach the polygon evaluator. This increment looks good to me.

PR #896, opened 2026-09-18
chore(geofence): re-check polygons on a schedule, not only on a callback

mrehan27:

## Why

A polygon is evaluated only because its wake circle reported a crossing. There is no second wake, so a crossing that is never reported is a visit that cannot be recovered.

That case is not hypothetical. `GeofenceRepository.emitInitialEnters` exists because GMS unreliably drops `INITIAL_TRIGGER_ENTER` for a fence added around a stationary device — and it contains `if (region.isPolygon) continue`, because the sync fix carries no accuracy evidence to judge a ring with. `insideNow` in `registerNearestAndPersist` excludes polygons for the same reason.

So a device sitting inside a wake circle when that circle is registered can get no callback and no backstop. Nothing re-derives a polygon arrival, and the visit stays unreachable until the user leaves the circle and comes back.

## What

A periodic worker, scheduled while any polygon is registered. It asks where the device is, then hands each polygon whose wake circle contains the fix to the same `activate` a coarse ENTER calls.

No new evaluation path. Dedupe, the accuracy decision, the on-demand precise fix, the corroboration hold and emission are the existing paths, so this adds a trigger rather than a second set of rules.

Three decisions worth stating:

- **Keyed on what is registered, not what is active.** An active polygon is one already woken, so gating on the active set would disable this in exactly the case it exists for.
- **`KEEP`, not `REPLACE`.** Every sync reaches the scheduler, and `REPLACE` re-anchors the period, so an app syncing more often than the interval would never re-check.
- **The worker retires its own work** when nothing is registered, so a missed cancel site costs one wake rather than a permanent job.
- **Every teardown path cancels it.** The scheduler reached one call site while `approachMonitor.stop` reached eleven, so the work outlived every teardown but one. The cancel is unconditional rather than keyed on what is still registered, because `stopAll` and `clearUserSessionRetainingOsRegistrations` leave registrations in place on purpose.
- **A departure clears a fallback activation.** `activate` records the polygon coarse-inside and only a GMS coarse EXIT clears it, so a polygon recovered from an ENTER that GMS never issued could stay active for the life of the install. A fix that is decisively outside the circle now clears it through `onCoarseExit`, which keeps a committed arrival active. The accuracy is subtracted rather than gated on the 50 m decisive ceiling, which a balanced fix would rarely meet, and the fix must actually report accuracy since `Location.accuracy` is 0 when unset.
- **It asks for a balanced fix, not a GPS one.** It only has to answer whether the device is inside a circle hundreds of metres across, and registered polygons are merely the nearest set, so a user kilometres from all of them would otherwise pay a long GPS wake every interval forever. Precision is bought downstream by the undecided path, and only once a cheap fix has shown it could matter.

Two records: `polygon.recheck.ran` carries the candidate count, the admitted ids and the cleared ids, and `polygon.recheck.skipped` carries why. Both counts, because a wake that admits nothing is the expected case and has to be distinguishable from one that never ran. A revoked permission is its own skip reason, since the fix source reports it as the same null a silent GPS gives.

## Limitation, accepted

Latency is WorkManager's 15 minute floor, measured at 18-20 minutes on an API 37 emulator. A stop shorter than that inside a dead zone is still missed. The alternatives that would close it are a foreground service with continuous location, or inner fences tiling the polygon, and both were rejected on cost.

## Testing

946 unit tests green, parsed from the result XMLs with `--rerun-tasks`. ktlint, `:geofence:lintDebug` and `:geofence:apiCheck` clean.

Twelve mutations, all killed: gating the schedule on the active set, admitting every registered polygon regardless of its circle, `UPDATE` in place of `KEEP`, removing the self-cancel, requesting a GPS fix instead of a balanced one, and reading the user-state generation after the fix await instead of before.

The last one survived at first. The ordering matters because the await is up to 20 s and an identify inside it would attribute this wake to the new user, but the stubbed generation was constant so nothing could see the difference. The stub now moves during the await, and the mutant dies.

The review round added six more: dropping the cancel from any of the five teardown paths, issuing it under `controllerLock`, halving the admission radius, doubling the departure radius, dropping the accuracy term, and dropping the active-set restriction. Two needed better fixtures rather than new code. The admission test had put the fix at the circle's exact centre with its negative case 111 km away, and the departure test sat 300 m outside a 120 m circle, so neither pinned its threshold. Both now use a pair ten metres either side of the boundary.

**Verified executing on a device**, not only in unit tests. On an emulator with a real synced session the worker started, waited 16 s for its balanced fix inside the 20 s budget, and returned `SUCCESS`:

```
polygon.recheck.ran cand=8 n=1 ids=17 fixsrc=fresh_request acc=100.0
polygon.undecided   id=17 why=accuracy_too_low edge=23.9 acc=100.0
```

It screened 8 registered polygons, admitted the one whose circle contained the fix, and handed it to `activate`, which evaluated and came back undecided. Note that WorkManager refuses a forced early run of periodic work, so this had to be waited for rather than triggered.

What the emulator cannot show is the accuracy benefit: its mock provider reports 100 m for every fix, which is why the verdict above is undecided. That part needs the field.

## Stack

```
feature/geofence-polygons
  #892 #893 #894 #895                   MERGED into feature/geofence-polygons
  mbl-2480-polygon-periodic-recheck     <- this PR
    mbl-2480-polygon-passive-listener   #897
      mbl-2480-unbalanced-exit-guard    #898
      mbl-2480-escalation-backoff       #899
      mbl-2480-corroboration-window     #900
```

The three on top of #897 are siblings, not a chain.


geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonRecheckWorker.kt
```diff
@@ -0,0 +1,183 @@
...
+        // polygon whose circle does not contain this fix has nothing to re-check.
+        val admitted = polygons.filter { it.distanceTo(fix.latitude, fix.longitude) <= it.radius }
+        logger.logPolygonRecheckRan(
+            location = fix,
+            candidateCount = polygons.size,
+            admittedIds = admitted.map { it.id }.sorted()
+        )
+        val controller = SDKComponent.android().polygonGeofenceServiceController
+        admitted.forEach { region ->
+            controller.activate(
```
Shahroz16, line 185: Could we route the scheduled fix through an activation path that does not assert a GMS coarse ENTER, and let a later outside fix clear any fallback-only activation? This call persists the polygon in coarse_inside_polygon_ids, but only the GMS coarse-EXIT path clears that state. If this worker recovers an ENTER that GMS missed, the polygon can therefore remain active after the device leaves, and later rechecks can keep reopening the sampling session.

Shahroz16 (changes requested):
The scheduled fallback currently writes coarse-containment state that it cannot reliably clear. There is also one teardown gap: stopAll() does not cancel this periodic work, and the missing-Location-module path leaves routable IDs populated, so the worker can keep fetching fixes and reactivate polygon sampling after geofencing was stopped. I cannot approve this until those lifecycle paths are closed.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonRecheckWorker.kt
Shahroz16: I added a focused regression against this exact head using the real GeofenceRegionStoreImpl. After fallback activation, a confidently outside approach fix still leaves activePolygonIds as [16] instead of clearing it, so this is now reproduced deterministically rather than inferred from the control flow. A separate teardown check also records no scheduler cancellation after stopAll().

mrehan27:
Both closed in 3187ccad.

**Teardown.** Your two paths led me to check the rest. `approachMonitor.stop` reaches eleven sites and the scheduler reached one, so the work outlived every teardown except `reconcileRegisteredPolygons`. It now cancels from `invalidateOsRegistrationState`, `stopAll`, `clearUserScopedState`, `clearUserSessionRetainingOsRegistrations` and `completeUserReset`. Unconditional, not keyed on what is still registered, because the last two leave registrations in place on purpose.

**Containment.** Took your second suggestion. A fix that is decisively outside the wake circle clears the activation through `onCoarseExit`, restricted to polygons in the active set, so a committed arrival stays active and discarded holds are still reported. Predicate is `distanceTo - accuracy > radius` with no 50 m ceiling, which a balanced fix would rarely meet, plus a `hasAccuracy()` check since `Location.accuracy` is 0 when unset.

One correction on the cost, since it changes what this fixes. The coarse-inside flag never blocks a business EXIT: committed state is the entered set, and the engine emits EXIT from any decisive-outside fix reaching an active polygon. What a sticky flag does is keep the polygon in the active set after departure, so `engine.stop` never runs and every fix reaching the engine re-evaluates it. That is the part this clears. The repeated sampling sessions are a separate thing: `activate` calls `approachMonitor.start` on every admitted fix, so they occur whether or not the flag is sticky, they are bounded at two minutes, and they only re-arm while the device is inside the circle. That cost belongs to wake-circle admission.

946 tests, 0 failures. Twelve mutants killed. Two of those exposed fixtures that did not pin their thresholds at all: the admission test put the fix at the circle's exact centre with its negative case 111 km out, and the departure test sat 300 m outside a 120 m circle. Both now use a pair ten metres either side of the boundary.

The same two fixes for the passive listener are ready on #897 and will go up with its next push.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonRecheckWorker.kt
Shahroz16: Thanks, the outside re-check now clears through onCoarseExit, and the teardown paths cancel future work. There is still one in-flight window after the balanced fix returns: stopAll() keeps the same routable IDs and user generation, so a worker already past the wait can reach activate() and re-arm the polygon. I reproduced this with the real worker, controller, and store by running stopAll() between logPolygonRecheckRan and activate(); even immediate coroutine cancellation left active IDs as [venue] instead of []. Could we invalidate these scheduled activations with a teardown or run token checked when activate() takes the controller lock? Cancellation can still clean up the work, but it cannot enforce this boundary by itself.

mrehan27:
Fixed in `f03717f1`. You were right that cancellation cannot enforce this by itself: the coroutine may already be past the suspension point when the teardown lands.

The boundary now has its own read. A counter bumped by every teardown, captured by the worker before it waits, and checked when `activate()` takes the controller lock. Teardown deliberately leaves the routable ids and the user generation in place, so neither of the checks `activate()` already made could see it. Cancellation still cleans the work up; this is what makes the outcome correct when it does not win the race.

Two tests, and both halves die under mutation: dropping the check in `activate` fails the re-arm test, and moving the worker's read to after the wait fails the other, since it would then hand over the post-teardown value.

One thing worth flagging, because it changed my approach on #897. The same token does not work for the passive path: a passive fix is dispatched before the receiver runs, so a token captured when the receiver starts is already the post-teardown value and refuses nothing. I built it that way first and a surviving mutant caught it. That PR uses the registration's own liveness instead.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonGeofenceServiceController.kt
```diff
@@ -94,10 +117,14 @@ internal class PolygonGeofenceServiceController(
     fun activate(
         polygonId: String,
         expectedUserStateGeneration: Long = store.userStateGeneration(),
-        expectedRegionRevision: Int? = null
+        expectedRegionRevision: Int? = null,
```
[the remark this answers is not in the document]
Shahroz16: I reproduced this on the latest head with the real store. A teardown after the early check but before the inner locked activation still restores the polygon to the active set. Passing the same teardown token into the inner activation makes the focused regression pass.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonRecheckWorker.kt
```diff
@@ -0,0 +1,225 @@
...
+            admittedIds = admitted.map { it.id }.sorted(),
+            clearedIds = departed.map { it.id }.sorted()
+        )
+        val controller = SDKComponent.android().polygonGeofenceServiceController
+        // onCoarseExit, not the store flag and not a direct deactivate. It is the reviewed order:
+        // record outside, drop a duplicate delivery, evaluate the fix, keep the polygon active if
+        // the arrival was already committed, and otherwise deactivate and report the holds it
+        // discarded. Clearing the flag alone leaves an already-emitted business EXIT with nothing
+        // to re-run, so the polygon would stay active regardless.
+        departed.forEach { region ->
```
Shahroz16, line 222: Could we carry the teardown token through this scheduled departure path too? I reproduced the worker selecting an entered polygon, stopAll() running before dispatch, and onCoarseExit() reactivating the polygon from the retained entered state. Forwarding the token only through activate() fixes the arrival path but leaves this one able to undo teardown.

Shahroz16 (changes requested):
Thanks, the generation guard fixes the original wait-to-wake window. I reproduced two narrower dispatch windows that can still re-arm after teardown: the token is dropped before the final locked activation, and the scheduled departure path does not carry it at all. I added both cases inline.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonRecheckWorker.kt
mrehan27: Fixed. The token was reaching activate() and then being dropped before the write it guards: the suspend entry checked it, then called the locked activation without passing it, and the registration and dedupe reads in between take and release the lock.

It is now checked inside controllerLock, next to the write. Every teardown bumps the generation under that same lock, so nothing can land between the check and the activation. A refused activation also no longer goes on to evaluate its fix.

Mutations: dropping the token from the inner activation, and removing that gate, each fail their own test.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonRecheckWorker.kt
mrehan27: Done, and it needed the token more than activate did. Its tail re-arms: a polygon still in the entered set is kept active instead of deactivated, and teardown retains that set on purpose, so a departure dispatched before a teardown and run after it undid the teardown through the exit path.

The worker now hands the same token to both calls, and it is checked at both lock acquisitions rather than only at the tail, because evaluateCallbackFix suspends between them. That also covers the coarse-outside write and the fix request, which Bugbot flagged separately.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonRecheckWorker.kt
Shahroz16: Thanks, both of the earlier dispatch windows are fixed on this head. I found one earlier boundary in the same worker: the teardown token is captured only after the routable catalog and permission reads. If stopAll() runs during that initial work, the worker reads the new token and can still activate the polygon after shutdown. I reproduced that with the real worker, controller, engine, and store. Active IDs end at [16] instead of empty, while moving the token capture ahead of the initial reads makes the regression pass. Could we make the ownership snapshot cover the worker from the start?

mrehan27: Fixed, and both halves of the snapshot moved, not just the token.

You are right that the capture sat behind the catalog and permission reads, so a teardown during that work was captured as though it had always been current. It is now the first thing `doWork` does.

The user generation moved with it, because it was read late for the same reason and carries the mirror defect: the catalog is read for the outgoing user and the generation sampled after an identify lands, so the run is attributed to the incoming user with the previous user's catalog. Same shape, other direction.

One mutation run and killed: putting either capture back after the reads.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonRecheckWorker.kt
```diff
@@ -0,0 +1,228 @@
...
+        val expectedTeardownGeneration =
+            SDKComponent.android().polygonGeofenceServiceController.teardownGeneration()
+        val routableIds = store.getRoutableRegisteredIds()
+        val polygons = store.getCachedRegions().filter { it.id in routableIds && it.isPolygon }
+        if (polygons.isEmpty()) {
+            // Self-healing rather than relying on every cancel site being wired: if the set is
+            // empty this work has nothing to do again, so it retires itself. A missed cancel
+            // therefore costs one wake, not a permanent periodic job.
+            logger.logPolygonRecheckSkipped(PolygonRecheckSkip.NOTHING_REGISTERED)
+            SDKComponent.android().polygonRecheckScheduler.cancel()
```
Shahroz16: Could we make this self-retirement conditional on the catalog still being empty when the cancellation is issued? I reproduced a sync scheduling the unique periodic recheck after this worker read an empty catalog but before it reached cancel(). The worker then cancels that newer schedule, leaving no periodic recheck until another sync or movement. That is the stationary-device gap this worker is meant to cover. My focused regression expected [schedule] and got [schedule, cancel]. The new ownership snapshot itself passes the earlier teardown regressions; this cancel-after-stale-read window is the remaining blocker I found.

mrehan27: Fixed, and closed by ordering rather than made conditional, which is better than what you asked for.

Making the retirement conditional on a re-read before the cancel narrows the window but does not close it: a sync landing between the re-read and the cancel is still lost. I had that version written and threw it away, because a mechanism that still occasionally produces the expensive failure to avoid the cheap one is the wrong shape.

What it does now is cancel, re-read, and re-assert with the existing KEEP schedule if the catalog filled:

    scheduler.cancel()
    if (registeredPolygons(store).isNotEmpty()) scheduler.schedule()

That has no window. Both callers write the routable ids before they call reconcile, and reconcile's `schedule()` comes after that write, so there are only two interleavings. Either the sync's write lands before the re-read, and this run sees it and re-asserts; or it lands after, in which case the sync's own `schedule()` also lands after this cancel and survives it. There is no third. I checked the premise at both call sites rather than assuming it: `GeofenceRepository` writes routable at 585 and reconciles at 594, and writes at 787 and reconciles at 834.

Two assumptions worth stating because they are load-bearing. WorkManager runs a cancel and an enqueue from one process in call order, via its serial task executor. And the cancellation this cancel provokes cannot interleave between the two calls: neither suspends, and a CoroutineWorker's job is only observable at a suspension point.

You are right that the trade is asymmetric, and the comment now says which side is expensive. A wrongly kept schedule costs one wake that does two store reads and returns before `awaitFreshFix`. A wrongly cancelled one costs the coverage until the next sync.

Both reads go through one helper so the re-check cannot drift from the decision it re-tests.

Three mutations killed: dropping the re-assert, reversing it with the cancel, and making it unconditional, the last one by the pre-existing nothing-registered control.

Shahroz16: Thanks, the cancel, re-read, and KEEP re-assert closes the stale empty-catalog cancellation window. I reran the focused worker and controller regressions on this head, and the full CI matrix is green.

Shahroz16 (approved):
The worker now preserves a concurrent sync’s newer periodic schedule. The focused regressions and the current CI matrix are green.

PR #897, opened 2026-09-18
chore(geofence): listen for fixes other apps already paid for

mrehan27:

## Why

The periodic re-check in the parent PR is bounded below by WorkManager's 15 minute floor, measured at 18-20 minutes. A `PRIORITY_PASSIVE` request never turns a sensor on: it asks the OS to hand over location some other app requested anyway. So it costs nothing beyond the delivery, it can land at any point inside that window, and it guarantees nothing, because a device where nothing else asks for location delivers nothing.

That asymmetry is the whole argument. It cannot make anything worse, and it can only narrow the gap.

## What

Registered for as long as any polygon is. A delivered fix takes the same route as the re-check's: admit the polygons whose wake circle contains it, then hand each to the `activate` a GMS callback uses. No second evaluation path.

Deliberately much simpler than the approach monitor, which cost six defects on one seam. There is no session here — no deadline, no sample count, no per-session identity — because the registration's lifetime is "some polygon is registered" and nothing has to tell two generations apart. The receiver reads the live generation at delivery, and a late delivery is refused by the controller's own check.

Two details that are load-bearing:

- **It retires itself** when a delivery finds nothing registered, the way the worker cancels its own work. `reconcile` is the only other caller of `stop()` and it runs from the registration paths alone, so a sign-out or the kill switch would otherwise leave the request live with GMS across process death, with no later registration ever arriving to remove it, waking the process on every other app's fix indefinitely.
- **A departure clears the activation.** `activate` records the polygon coarse-inside and only a GMS coarse EXIT clears it, so a polygon this listener activated from an ENTER GMS never issued could stay active for the life of the install. A fix that is decisively outside the wake circle now clears it through `onCoarseExit`, which keeps a committed arrival active and still reports discarded holds. The accuracy is subtracted rather than gated on the 50 m decisive ceiling, which a passive fix would rarely meet, and the fix must report accuracy at all since `Location.accuracy` is 0 when unset.
- **Every teardown path stops it.** `stop()` reached one call site, so a sign-out or the kill switch left the GMS request live. The shared helper that cancels the periodic re-check now stops the listener too, from all five paths that disable geofencing or end the session.
- **Only the newest fix in a batch is used.** The older ones describe the way in, and each would buy another evaluation of the same arrival.

The minimum interval is **unmeasured** and recorded as such. It throttles process wakes rather than sensor use, since an undecided verdict can only buy a precise fix once per the existing cooldown.

A passive `PendingIntent` does not survive a reboot, unlike a geofence. Boot restore re-registers and reconcile starts it again, so `polygon.passive.started` after every boot is expected rather than a restart loop.

## Validation

963 unit tests green, parsed from the result XMLs. ktlint, `:geofence:lintDebug` and `:geofence:apiCheck` clean. `PolygonPassiveReceiver` is added to the API dump because a manifest-declared receiver has to be public, as `PolygonApproachReceiver` already is.

Twelve mutations, all killed: `PASSIVE` to `BALANCED`, dropping the throttle, admitting every registered polygon regardless of its circle, never stopping the listener on reconcile, oldest fix instead of newest, removing the self-retirement, doubling the departure radius, dropping the accuracy term, dropping the active-set restriction, reverting the early return to admitted-only, reading the user generation after the first store read instead of before, and removing the listener from the shared teardown helper.

Three of those needed better fixtures rather than new code, and they were only found by mutating tests that already passed. The admission test had put the fix at the circle's exact centre with its negative case 111 km away, so halving the wake circle passed it. The departure test sat 300 m outside a 120 m circle, so doubling the radius passed it. Both now use a pair ten metres either side of the boundary. The `FLAG_NO_CREATE` on `stop()` was also unasserted, which mattered because without it a stop mints the request it is removing and `polygon.passive.stopped` would appear for a device that never registered one.

**Verified end to end on a device.** On an emulator with a real synced session:

```
polygon.passive.started
polygon.passive.received   cand=8 n=1 ids=17 acc=100.0
polygon.undecided          id=17 why=accuracy_too_low edge=23.9 acc=100.0
polygon.freshfix.requested ids=17 n=1
```

A fix nobody paid for arrived, screened 8 polygons, admitted one, and the evaluation came back undecided at 100 m accuracy for a device reading 23.9 m inside the ring — which is the accuracy failure this stack exists for, and it triggered the on-demand precise request. The request returned nothing because an emulator has no GPS to answer it.

## Stack

```
feature/geofence-polygons
  #892 #893 #894 #895                   MERGED into feature/geofence-polygons
  mbl-2480-polygon-periodic-recheck     #896
    mbl-2480-polygon-passive-listener   <- this PR
      mbl-2480-unbalanced-exit-guard    #898
      mbl-2480-escalation-backoff       #899
      mbl-2480-corroboration-window     #900
```

The three on top of #897 are siblings, not a chain.


geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonPassiveReceiver.kt
```diff
@@ -0,0 +1,96 @@
...
+        val admitted = polygons.filter { it.distanceTo(fix.latitude, fix.longitude) <= it.radius }
+        logger.logPolygonPassiveReceived(
+            location = fix,
+            candidateCount = polygons.size,
+            admittedIds = admitted.map(GeofenceRegion::id).sorted()
+        )
+        if (admitted.isEmpty()) return
+        val controller = SDKComponent.android().polygonGeofenceServiceController
+        admitted.forEach { region ->
+            controller.activate(
```
Shahroz16, line 121: Could we use a fallback-specific activation path here instead of the GMS coarse-ENTER path? This call records the polygon in coarse_inside_polygon_ids, but an outside passive fix is filtered out and never clears it. If the listener recovers a missed GMS ENTER, the polygon can stay active after the user leaves and later passive fixes can keep reopening the sampling session.

Shahroz16 (changes requested):
The passive wake worked on the live route, but it has the same persistent coarse-state problem called out inline. It also is not stopped by stopAll(): that path leaves routable IDs populated, so the receiver does not self-retire and can reactivate sampling after the SDK has disabled geofencing. I cannot approve this until those lifecycle paths are closed.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonPassiveReceiver.kt
Shahroz16: I added a focused regression against this exact head using the real GeofenceRegionStoreImpl. After fallback activation, a confidently outside approach fix still leaves activePolygonIds as [16] instead of clearing it, so this is now reproduced deterministically rather than inferred from the control flow. A separate teardown check also records no passive-listener stop after stopAll().

mrehan27:
Both closed in 3c6ca9d8. Same two fixes as #896, applied to the listener.

**Teardown.** `stop()` reached one call site, inside `reconcileRegisteredPolygons`, so a sign-out or the kill switch left the GMS request live across process death with no later registration to remove it. The shared helper that cancels the periodic re-check now stops the listener too, from `invalidateOsRegistrationState`, `stopAll`, `clearUserScopedState`, `clearUserSessionRetainingOsRegistrations` and `completeUserReset`. Unconditional, because the last two leave registrations in place on purpose.

**Containment.** A fix that is decisively outside the wake circle clears the activation through `onCoarseExit`, restricted to polygons in the active set. Predicate is `distanceTo - accuracy > radius` plus a `hasAccuracy()` check, with no 50 m ceiling since a passive fix is whatever another app asked for and would almost never meet it.

Note the early return changed: an all-outside fix used to return before doing anything and now has work to do.

963 tests, 0 failures. Twelve mutants killed. Three of them exposed fixtures that passed while asserting nothing: the admission test put the fix at the circle's exact centre with its negative case 111 km out, the departure test sat 300 m outside a 120 m circle, and `FLAG_NO_CREATE` on `stop()` was unasserted. That last one matters for reading a capture, since without the flag a stop mints the request it is removing and `polygon.passive.stopped` would appear for a device that never registered one.

One correction on the cost, since it changes what these two PRs fix. The coarse-inside flag never blocked a business EXIT: committed state is the entered set, and the engine emits EXIT from any decisive-outside fix reaching an active polygon. A sticky flag keeps the polygon in the active set after departure, so `engine.stop` never runs and every fix reaching the engine re-evaluates it, which is what this clears. The repeated sampling sessions are separate: `activate` calls `approachMonitor.start` on every admitted fix, so they happen whether or not the flag is sticky, they are bounded at two minutes, and they only re-arm while the device is inside the circle. That cost belongs to wake-circle admission rather than here.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonPassiveReceiver.kt
Shahroz16: Thanks, the next confidently outside passive fix now clears the fallback activation, and teardown stops the listener. A fix already queued before that removal completes can still arrive after stopAll(): the routable IDs and user generation survive, so the receiver reaches activate() and re-arms [venue]. I reproduced this with the real receiver, controller, and store. Could we invalidate already-dispatched passive fixes at activation, using the same teardown or run token as the scheduled path? This is the remaining blocker. An outside approach fix still leaves the activation around until a later passive or scheduled fix, but because that recovery now exists I am treating the delay as non-blocking.

mrehan27:
Fixed in `32b398c3`, and your suggestion did not survive contact, which is worth writing down.

I built the same token as #896 first: capture it in the receiver, check it at activation. A surviving mutant showed it refuses nothing. A passive fix is dispatched by the OS *before* the receiver runs, so by the time there is anything to capture, the teardown has already happened and the token matches. It can only ever agree.

The registration is the thing that actually changed. `stop()` now cancels the intent as well as removing the request, and the receiver drops a delivery that no longer has a live registration behind it. Read at delivery rather than captured beforehand, which is the opposite of the scheduled path and correct for the same reason. The system owns the intent, so a cold-process delivery under a live registration is still admitted, which a process-local flag would have broken.

Both halves die under mutation: dropping the receiver's check fails the delivery test, and dropping the cancel fails the monitor's own arming test, which is what makes the check answerable at all.

Also in this push: the branch is rebased onto #896's fix, since the teardown counter it uses lives in the shared controller.

On your non-blocking note about an outside approach fix leaving the activation until a later passive or scheduled fix: agreed, and I have left it, since the recovery now exists.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonPassiveReceiver.kt
Shahroz16: Thanks, this closes the case where delivery begins after the registration has already stopped. I can still reproduce stopAll() landing after isArmed() succeeds but before activate(): the receiver restores the polygon to the active set. Capturing the teardown token before the registration check and enforcing it during activation made the focused regression pass, so I’m keeping this narrowly blocked on that remaining dispatch window.

Shahroz16 (changes requested):
The new registration check correctly rejects a delivery that starts after teardown. I can still reproduce teardown during the receiver’s dispatch window re-arming the polygon, because activate() does not receive a teardown token. I added the exact remaining case to the existing thread.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonPassiveReceiver.kt
mrehan27: Fixed. The token is captured on entry, before the isArmed() check, and passed to both the activation and the departure call.

The two checks cover different windows and neither replaces the other. isArmed() catches a teardown that finished before this delivery began, which no in-process token can see, because the OS dispatched the fix before the receiver ran. The token catches one that lands while the handler is still running: every teardown bumps the generation under the controller lock and only then cancels the registration, so a delivery isArmed() admitted is still refused at the write.

Mutations: dropping the token from either call, or re-reading it at dispatch instead of on entry, all fail.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonPassiveReceiver.kt
```diff
@@ -0,0 +1,128 @@
...
+                triggeringLocation = fix,
+                expectedUserStateGeneration = expectedUserStateGeneration
+            )
+        }
+        admitted.forEach { region ->
+            controller.activate(
+                polygonId = region.id,
+                triggeringLocation = fix,
+                expectedUserStateGeneration = expectedUserStateGeneration
+            )
```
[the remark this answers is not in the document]
mrehan27: Not fixing this one, and the reason is the clearing route rather than a disagreement about the mechanism.

Agreed on the mechanism. A passive or worker admission calls `activate()`, which records the polygon coarse-inside, and `processApproachLocations` computes `approachOnly = active - coarseInside - entered`, so approach fixes will never deactivate it. One correction on the fine EXIT: the engine emits the business EXIT regardless. It is only the deactivation after the EXIT that requires the polygon to be neither coarse-inside nor entered, so the event is not suppressed, only the deactivation.

Where "indefinitely" does not hold: both fallbacks have a departure half. The re-check worker and this receiver each call `onCoarseExit` for an active polygon whose fix is decisively outside its wake circle, and `onCoarseExit` never asks how the polygon became active, so a passive-only or worker-only activation is reachable. Its middle block records coarse-outside, which is the clear.

The bound is the worker interval, measured 18 to 20 minutes on device, conditional on a fix arriving that satisfies `distance - accuracy > radius`. A device that stays within its own accuracy of the circle keeps the polygon active, and that is the intended state, because it is still inside the wake circle.

On the case worth raising, which is a polygon that committed an arrival: `onCoarseExit`'s tail keeps it active when it is still in the entered set, and that is deliberate. A committed ENTER owes a business EXIT, the engine only evaluates active polygons, and `reconcileContainment` skips polygons, so deactivating there would orphan the visit with no EXIT ever emitted. In the decisively-outside case that branch does not even run: the middle block has already cleared coarse-inside, the evaluator's departure clause is clearance-first with no accuracy ceiling, so the same fix emits the EXIT, the entered set drops the id, and the polygon is deactivated in that same pass. The entered branch only fires when the fix was decisively outside the circle but not the ring by a further 20 m, or was refused as too old, and the next worker or passive fix closes it.

For completeness, the routes that clear coarse-inside are `deactivateLocked`, `onCoarseExit`'s coarse-outside write, the teardown paths via `retainCoarseInsidePolygonIds(emptySet())`, `reconcileRegisteredPolygons` via `retainCoarseInsidePolygonIds(ids)` when a polygon leaves registration, and the store's user-session open removing the key.

Verified against the source and independently reviewed rather than assumed. Recorded as a standing dependency: the worker's departure half is now load-bearing beyond its own PR, and removing or gating it on the active set would turn this into a real leak with no other clearer.

geofence/src/main/kotlin/io/customer/geofence/polygon/PolygonPassiveReceiver.kt
Shahroz16: Thanks, the previously reported window after isArmed() is fixed. There is an opposite ordering still open: stopAll() advances the teardown token before the passive PendingIntent is cancelled. A delivery during removeLocationUpdates() can therefore capture the new token, still see the listener as armed, and activate the polygon after shutdown. I reproduced this through the real receiver, controller, store, engine, and GMS monitor. stopAll() returned with active IDs [venue]; the live admission and completed-stop controls both passed. Could we make the listener invalidation and token handoff one coherent boundary so an old registration cannot adopt the post-stop token?

mrehan27: Fixed, and you found that my comment asserted the opposite of the truth. It claimed the bump-then-cancel order was what made the receiver safe. It is what made it unsafe, exactly as you describe.

`stopScheduledWakes()` now runs before the synchronized state wipe on all five teardown paths, so the cancel precedes the token bump. That makes the two checks one boundary rather than two independent ones:

- The cancel shuts the gate synchronously and permanently, so any delivery arriving afterwards is refused by `isArmed()` whatever its token says, including one the OS had already dispatched. That is the case no in-process token can see.
- Because the token is read before `isArmed()`, any delivery that passes `isArmed()` before the cancel necessarily holds a pre-bump token, so its activation is refused at the write. There is no longer a path to a post-bump token with an open gate, which was the defect.

A delivery admitted in the instant before the cancel can still arm, and that is now harmless rather than permanent: the wipe runs after it. Checked on all five paths rather than assumed. `stopAll` and `invalidateOsRegistrationState` clear the active ids directly; the other three route through the store's `completeUserReset`, which removes `KEY_ACTIVE_POLYGON_IDS` and `KEY_COARSE_INSIDE_POLYGON_IDS` unconditionally. The activation is also atomic against the wipe, since both sit in the same `controllerLock` block, so it either arms fully and is undone, or reaches the lock after the bump and is refused.

Two new tests: one drives the delivery from inside the monitor's `stop()`, which is the cancel itself, with a stateful store so the assertion is the state left behind; the other asserts stop-before-first-wipe across all five paths. Mutations killed: reverting the order in `stopAll` alone, reverting it on all five, and removing `pendingIntent.cancel()` from `stop()`, which the real-PendingIntent test catches.

The comment now says which check covers which window.

Shahroz16: Thanks, the passive teardown window is fixed on this head. I also checked the stop-before-wipe ordering against concurrent reconciliation. It can briefly restart the wakes, but the shipped reset paths clear routable IDs so they self-retire, and the stopAll callers either clear routing first or run before sync subscriptions. I am not holding this PR on that hardening. The stack is still waiting on the new periodic-worker cancellation finding on #896, which this branch inherits; once that base is fixed and restacked, I can finish this approval.

mrehan27: Base is fixed and this branch is restacked on it, tip `a9471701`.

Agreed on the reconciliation hardening, and thanks for checking it rather than just flagging it. Your reading matches mine: it can briefly restart the wakes, the reset paths clear routable so they self-retire, and for `stopAll` the registrations are retained so an armed listener is the correct state anyway. Not holding this PR on it either. If it turns out to matter I would rather fix it where the schedule lifetime is owned than add a second guard here.

The #896 finding is closed by ordering rather than by a conditional: cancel, re-read, re-assert. Details on that thread.

Shahroz16: Thanks, the base fix is now restacked here. The passive teardown regressions and the inherited worker regression pass on the combined head, and CI is green.

Shahroz16 (approved):
The passive teardown fix and the restacked periodic-worker fix both hold on this combined head. Focused tests and CI are green.

PR #900, opened 2026-09-21
chore(geofence): confirm a marginal arrival from a second measurement

mrehan27:
Part of [MBL-2480](https://linear.app/customerio/issue/MBL-2480/mobile-fix-silently-dropped-geofence-callbacks)

`cor=true` appears in no capture. Arrival corroboration has never once completed on a device, and because arrival **gates** on it, every arrival marginal enough to need a second fix was silently dropped. Two are on the record, one of them 6.7 m inside its ring at 10.9 m accuracy.

Corroboration is required when the fix's accuracy circle reaches the boundary. Our rings run 24-42 m inradius and accuracy is rarely better than 10 m, so that is most of a polygon's area, not an edge case.

## Why it never completed, and why the window is not the fix

A held arrival emits `ArrivalPending` and no `Undecided`, so it never reached the precise-fix request #895 built. Nothing else supplies a second measurement inside the 60 s window in the background: the passive monitor is 5 min nominal and the re-check worker 15 min, and only the approach sampler could ever meet 60 s.

The window stays at 60 s. Widening it to reach the field's real ~300 s gap would contradict the #882 reproduction pinning that two marginal fixes 5 min apart are not one visit. The stronger reason is measured: the ~5 min re-emissions arrive at `acc=100.0`, read as ambiguous above the ceiling, so a wider window is defeated by the traffic inside it whatever the cadence.

Instead, `PolygonEvaluationOutcome` carries the pending-hold ids and the callback path asks for one fix for them, under the same cooldown as an undecided verdict.

## Corroboration now needs a second measurement, compared by position

The only independence guard was a per-fence strictly-newer `elapsedRealtimeNanos` dedupe. The fused provider re-emits a carried-forward position under a fresh stamp, so an echo passed it.

Position alone, not the whole measurement: position is the part measured to carry forward, accuracy the part measured to be substituted. Requiring both would let the same coordinate at a different sub-ceiling accuracy confirm an arrival from one observation, which is the defect this closes.

The cost is accepted and documented in source: a device whose coordinate is snapped identical cannot corroborate itself, so its arrival depends on the requested fix producing a different position and is dropped otherwise. Claiming a second opinion we do not have is worse.

## An identical position cannot break a hold either

Asking for a precise fix made this common rather than theoretical. That request returns over the accuracy ceiling most of the time at the coordinate already held, which read as disagreement and broke the run, killing the hold seconds after it opened. Asking would have lost the visit faster than not asking.

## Tests

Nine pre-existing tests corroborated by feeding one sample twice while their comments described a second sample, including one that asserted an echo *breaks* a hold. Their fixtures now resample about 2 m; the replay and restart tests keep byte-identical repeats, since that is what they assert.

`process_givenTwoMarginalFixesMinutesApart` stopped constraining the window once the echo guard landed, because it fed one sample twice. With distinct samples it fails when the window is widened.

`polygon.arrival.echo` carries `since`, the gap to the counted fix, because `edge` and `acc` alone cannot separate a re-delivery milliseconds later from an identical coordinate a minute later.

## Known and deliberately not fixed here

The echo guard is per fence, so a re-emitted older position could corroborate a different fence's hold. Both positions must be real observations inside that fence and the window caps the age, so it inherits timing rather than inventing a visit. A delivered-sample set is the fix.

## Stack

```
feature/geofence-polygons
  #892 #893 #894 #895 #896 #897 #898   MERGED into feature/geofence-polygons
  mbl-2480-escalation-backoff          #899
  mbl-2480-corroboration-window        <- this PR
```

Both remaining branches now sit directly on the feature branch.

## Merge note

This renames the callback request set from `undecidedPolygonIds` to `needing`. #899 reads the old name in `recordEscalationOutcomes`. A merge of the two conflicts in `PolygonGeofenceServiceController.kt`, so it cannot land silently.

**Decided order: this PR lands first, #899 second.** This one is approved and #899 is still in review, so the reconciliation is done on #899's branch under review rather than in a merge commit. Nothing is owed here.


Shahroz16 (approved):
The corroboration window now requires a genuinely second measurement, preserves a held arrival across unusable echoes, and respects user changes while waiting. The changed suite and additional lifecycle checks pass. The remaining parent teardown issue is outside this child diff.

PR #901, opened 2026-09-21
chore(geofence): gate the precise-fix request on the fences it serves

mrehan27:
Part of [MBL-2480](https://linear.app/customerio/issue/MBL-2480/mobile-fix-silently-dropped-geofence-callbacks)

Follow-up to #899, which Shahroz approved with this deferred.

The suppression gate read the callback's own fence. That fence is routinely not one the request would serve: a fix that decides it clears its memo, so the callback arrives carrying none while the only fence still needing a fix is already suppressed. The request got funded anyway, once per wake, which is the loop the backoff exists to stop.

Recording always covered the whole requested set. Only the check was narrower, so the two disagreed about what a request is for.

## The change

`requestWouldRepeatItselfLocked(needing, ...)` asks whether every fence in the request's own set is suppressed.

`all`, not `any`. One fence that could still be answered is worth the fix, so a suppressed co-tenant must not withhold it. Suppressing on `any` would keep the fix from a fence that is demonstrably answerable because a neighbour is parked, which is how this turns into a delayed arrival.

The cost of `all` is bounded. Every memo in a batch is recorded from the same fresh fix, so movement relative to one is movement relative to all, and the recording resynchronises the set on each request. The worst case is one extra request per fence, once, for a batch whose memos have drifted apart or whose 30 minute expiries are staggered. `any` by contrast would withhold a fix indefinitely while a neighbour sat parked.

Worth knowing before reading a capture, and documented in source: a fence in the set because its arrival is held never carries a memo, since `ArrivalPending` puts it among the evaluated rather than the undecided. A held arrival therefore always defeats this check and keeps the batch asking. That is correct, a second measurement is what a hold needs and the window caps it at 60 s, but batch suppression is off while any co-tenant hold is open.

Neither of the two designs discussed on #899 is used. Widening the recording suppresses a fence that is demonstrably decidable, and keying on whether a fence's verdict changed still asks the wrong question, because the callback's fence need not be in the set at all.

## Tests

| Reverted | Result |
|---|---|
| Gate back to the callback's fence | dies, 2 requests where 1 is due |
| `all` to `any` | dies, 1 where 2 are due |
| The `isNotEmpty()` guard | survives, documented in source as defensive |

The fixture needs one fix to reach two different verdicts. The accuracy ceiling is a flat 50 m rather than per fence, so the fences diverge geometrically instead: the second is shifted about 60 m north and committed INSIDE, so a venue-centre fix at 8 m accuracy is decisive for the first and `WITHIN_ACCURACY` for the second.

1006 geofence tests pass.

## Stack

```
feature/geofence-polygons              #892-#900 all MERGED
  mbl-2480-suppression-request-set     <- this PR
```

## Not in scope

`FUTILE_ESCALATION_RETRY_MS` is still 30 min and still unmeasured. Shahroz asked for that to go with the calibration work.


Shahroz16 (changes requested):
Could we keep held arrivals out of this suppression path? On a `NONE_ARRIVED` request, `needingPolygonIds` is recorded as still undecided, including any fence that is only waiting for corroboration. The next co-tenant callback can then suppress the second measurement for up to 30 minutes, even though the hold expires after 60 seconds, so a valid enter can be dropped. Passing the pending-arrival IDs into the gate and always allowing those requests, or excluding them from the timeout memo, would preserve the intended behavior. A regression test with a held arrival, a timed-out fresh-fix request, and a later co-tenant callback would cover the sequence.

Also, there are two consecutive KDoc blocks above `requestWouldRepeatItselfLocked`; could we move the field-capture rationale back above `escalationWouldRepeatItselfLocked` while touching this?

mrehan27:
Both fixed on `20c33997`. You were right, and the first one was a regression I introduced.

**Held arrivals.** Took your second option rather than the first: they are no longer recorded on the `NONE_ARRIVED` path at all. A hold is not a futile escalation, the fix decided ENTER and is waiting for corroboration, so memoising it records a failure that did not happen. Fixing the cause also avoids leaving an untrue record in place, which passing the pending ids into the gate would have done.

Your regression test is in, and the peer review of it caught that my first version was weaker than it looked: the co-tenant was never activated, so the timed-out request covered the held fence alone, and with no undecided fence in the set the mutant was indistinguishable. It is now the mixed case, one held and one undecided fence in one timed-out request.

Two things that came out of chasing it, neither of which you asked for:

Nothing anywhere pinned the timeout recording itself. Deleting it passed all 1007 tests. Pinned now.

I had documented "a held arrival never carries a memo" as unconditional. It is true of memo creation and false of memo persistence: a fence whose earlier request genuinely failed to decide it from this position carries a truthful memo, and a later hold on that fence from the same position is refused, falling back to the next delivered fix. That is in the KDoc and the description now. **Whether an open hold should bypass the gate outright is a policy call I have deliberately not made** - it is one condition if you want arrivals to win over cost here.

**KDoc.** Moved the field-capture rationale back above `escalationWouldRepeatItselfLocked`. My insertion had anchored before the wrong declaration and left it orphaned above the new function.

Five mutants now, four dead, listed in the description. 1008 tests, lint and API check clean.

Shahroz16 (approved):
The held-arrival timeout concern is resolved at this head. I also verified the focused regression fails when only the exclusion is reverted, then passes with the fix, along with the full 28-test fresh-fix suite and the geofence module build. Looks good from my side.

PR #902, opened 2026-09-21
test(geofence): read the corpus from mobile-replay-harness

mahmoud-elmorabea:
**Android replay harness — 5 of 5**

#885 → #886 → #887 → #888 → **this**

The recorded drives now live in their own private repo, `customerio/mobile-replay-harness`, as one flat `scenarios/` directory. This points the reader at it. Test-only; no production source.

Pairs with **customerio/mobile-replay-harness#3**, which ships the corpus, and with **customerio-ios#1304**, the same change on the other side.

## Three changes

**The sibling path** becomes `mobile-replay-harness/scenarios`.

**`conformanceRoot` and `conformance()` go.** There was a `recorded/` directory and a `conformance/` sibling beside it, discovered separately and concatenated. There is now one directory, and a scenario says in its own header where it runs.

**`expect` is replaced by `sourceKind`**, read from `source.kind`. Routing no longer reads it: a scenario runs here when `platform` is `android` or `any`.

## Why the `any` branch is needed even though recorded drives stay per-platform

The authored scenarios have always been the cross-platform tier and have always declared `platform: "any"`. They used to be routed by *which directory they sat in*. In a flat directory `recorded()` meets them and rejects both.

`sourceKind` is kept for one job only — so the "did discovery find any *drives*?" guard can still distinguish a recorded drive from an authored scenario without a second directory standing in for it.

A recorded drive still names the OS that produced it, and that is a statement about its **inputs**, not about the SDK. CoreLocation delivers every geofence event twice and Play Services does not — on the 2026-09-21 drive all six iPhone callbacks arrived as three identical pairs 11–26 ms apart, while both Android arms recorded none.

## Verification — and why CI cannot do it

Local, against mobile-replay-harness#3: **470 tests, 0 failed, 4 replay cases, discovery clean** — read from the JUnit XML, since `ignoreFailures=true` makes gradle's verdict meaningless.

**A green CI check proves nothing here.** The corpus repo is private and is not checked out in CI, so `isAvailable` is false, every replay test skips, and the job is green with or without this change. That is true of the whole stack, not just this PR.

I reverted this change and ran against the new corpus to confirm both failure modes:

| | result |
| --- | --- |
| old reader, no override | `ScenarioReplayTest` 1 test, **1 skipped** — green having graded zero drives |
| old reader, override set | `ScenarioDiscoveryTest` **FAILED**: both authored files rejected as `header platform is "any", expected "android" or "ios"`; only 2 of 5 scenarios graded |
| this change | 4 replay cases, discovery clean, 0 failed |

Contributes to MBL-2421


geofence/src/test/java/io/customer/geofence/replay/Scenario.kt
```diff
@@ -114,7 +118,7 @@ internal object ScenarioLoader {
         return Scenario(
             name = header["name"]?.jsonPrimitive?.content ?: file.nameWithoutExtension,
             platform = header["platform"]?.jsonPrimitive?.content ?: "unknown",
-            expect = header["expect"]?.jsonPrimitive?.content ?: "regression",
+            sourceKind = header["source"]?.jsonObject?.get("kind")?.jsonPrimitive?.content ?: "recorded",
```
Shahroz16: Could we reject a missing or unknown `source.kind` instead of defaulting it to `recorded`? I removed only `source` from the authored `first-arrival-and-crossing` scenario, then ran discovery and replay against that one-file corpus. Both passed, and discovery counted it as a recorded drive even though there was no phone capture. Validating `recorded`/`authored` the same way we validate `platform`, with a regression case for missing provenance, would keep this guard fail-closed.

Shahroz16 (changes requested):
The four corpus-backed Android cases pass locally, but missing provenance currently bypasses the recorded-drive guard. This head also conflicts with the updated #888 base, so please retain #888’s explicit invalid-override failure while restacking.

geofence/src/test/java/io/customer/geofence/replay/Scenario.kt
Shahroz16: Thanks, this now fails closed for missing or unknown provenance. I reran the one-file missing-source case and confirmed discovery reports source.kind as unknown instead of counting it as a recorded drive.

geofence/src/test/java/io/customer/geofence/replay/ScenarioReplayTest.kt
```diff
@@ -0,0 +1,256 @@
...
+         * A corpus that is absent yields one placeholder case that skips: JUnit refuses to run a
+         * parameterised class with no parameters at all, and an error there would look like a
+         * broken suite rather than an absent checkout.
+         */
+        @JvmStatic
+        @ParameterizedRobolectricTestRunner.Parameters(name = "{0}")
+        fun drives(): Collection<Array<Any?>> {
+            val files = if (Scenarios.isAvailable) Scenarios.replayable() else emptyList()
+            if (files.isEmpty()) return listOf(arrayOf("no corpus", null))
+            return files.map { arrayOf(it.name.removeSuffix(".scenario.ndjson"), it) }
```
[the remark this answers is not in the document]
Shahroz16: I reproduced this distinction. The replay class alone can skip when every file is invalid, but the normal suite runs the separate discovery guard and fails on those files. That behavior predates this corpus move, so I do not think it should block this test-only PR.

Shahroz16 (approved):
Thanks, the missing-provenance case now fails as expected, and the restack retains the invalid-path check. The loader and discovery tests passed, along with all four corpus-backed replay cases. Looks good.

